PR #7, opened 2016-09-09
Fix parenthesis mismatch.

UndeadKernel:
Small but in master version: the file module-asm.el is missing a closing parenthesis.


hlissner:
Whoops, my bad! Good catch.

PR #12, opened 2016-10-07
Fix bindings being shadowed by the doom popup

UndeadKernel:
Bindings defined in a major mode do not work in popups. These changes fix this.
Instead of using 'use-local-map' (which overrides the keybindings of the major mode), add new keybindings to the doom-popup-mode-map keymap.

I am not totally aware of why you were using 'use-local-map'. This fix removes the idea of replacing the local map.

### DIFF
--- a/core/defuns/defuns-popups.el
+++ b/core/defuns/defuns-popups.el
@@ -19,14 +19,6 @@
     map)
    "Active keymap in popup windows.")
 
-(defvar doom-popup-mode-local-map
-  (let ((map (make-sparse-keymap)))
-    (define-key map [remap evil-force-normal-state] 'doom/popup-close)
-    (define-key map [escape] 'doom/popup-close)
-    (define-key map (kbd "ESC") 'doom/popup-close)
-    map)
-  "Active keymap in popup windows with ESC bindings.")
-
 (advice-add 'doom/evil-window-move :around 'doom*popup-window-move)
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


UndeadKernel:
You might be wondering: What is it that this pull request fixes:

Short example of what didn't work: 
Open a clean emacs and see the help of the "-any" function (f1 h -any RET). Type the letter "o". This should run "ace-link-help" as defined [here](https://github.com/hlissner/.emacs.d/blob/master/private/my-bindings.el#L329). It doesn't work because the *Help* buffer overrides the bindings of help-mode when it uses 'use-local-map'. This fixes it.

hlissner:
The goal was to **only** enable `doom-popup-mode-local-map` in buffers that did not have a `:noesc` rule in `doom-popup-rules`.

I'll have to refuse your PR for two reasons:

1. You define the ESC keybindings in the non-local `doom-popup-mode-map`. This causes all popups, regardless of the `:noesc` rule, to be closable with ESC, because they all share the same map.

2. You removed the `map` let binding, but still use the `map` variable.

But you're correct about major-mode keybindings getting shadowed. I have a good solution for it, which I'll push after some testing in ~20 minutes.

hlissner:
There we go. My rudimentary tests show this fixed the issue with shadowed bindings. Let me know if you have any problems, and thanks for reporting the issue!

UndeadKernel:
Your elisp foo is strong with you.
Thank you for pointing out the problems with my PR.
You mentioned that you were using the 'map' variable in the 'let' binding. I could not see how this was the case. Could you please clarify this for the sake of knowledge?

hlissner:
Certainly. It was a minor error, and actually, I also made a mistake: **in the old code** I define `map` and don't use it:

```elisp
;       v
(let ((map doom-popup-mode-map))
  (unless (memq :noesc rules)
     (use-local-map doom-popup-mode-local-map)))
```

And **in the new code** (of this PR), you removed the let form and used `map`:

```elisp
(unless (memq :noesc rules)
  ;            v
  (define-key map [remap evil-force-normal-state] 'doom/popup-close)
  (define-key map [escape] 'doom/popup-close)
  (define-key map (kbd "ESC") 'doom/popup-close))
```

Hopefully that clarifies things!

PR #14, opened 2016-10-08
Fix wrong smart forward and backward bindings

UndeadKernel:
The smart-forward package does not provide the functions smart-left or smart-right. Instead, the functions should be named smart-forward and smart-backward.


hlissner:
Good catch! Thank you.

PR #17, opened 2016-11-06
Enable company mode when asked to autocomplete

UndeadKernel:
I'm trying to fix two things with this PL:

1) I noticed that when using C-x C-l in insert mode, the line-autocomplete would not work because company was not enabled. I now load company mode when using that command.

2) If using the command C-SPC to autocomplete, if company mode is not enabled, enable it.

Feel free to add both hunks, only the first one, or none ;)


hlissner:
Good idea. Thanks!

PR #24, opened 2017-01-25
doom-make-xpm: no need to reverse the data

ivanbrennan:
The data is same front to back, since it's a list of identical lists of identical elements, so there's no need to reverse it before passing it to pl/make-xpm.

@hlissner Let me know if I'm wrong about this, or if there's a conceptual reason to keep the `reverse`, but I don't think you need it.

### DIFF
--- a/core/core-modeline.el
+++ b/core/core-modeline.el
@@ -133,7 +133,7 @@ cached the first time."
      (let ((data nil)
            (i 0))
        (setq data (make-list height (make-list width 1)))
-       (pl/make-xpm "percent" color color (reverse data))))))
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks! You're quite right. However, I'll have to reject the merge because I'm in the middle of a comprehensive rewrite of my config and your fix would be lost in the slew of upcoming changes (lots of deleted/renamed/moved files).

In any case, thank you anyway! The function is present in the new version and I've removed the reversal there.

ivanbrennan:
Sounds good, I'm excited to see the rewrite. I've learned a lot by reading through your config. Thanks!

PR #26, opened 2017-02-20
Fix persistent-soft make dependency

rjww:
Add `persistent-soft` to `doom-protected-packages` to avoid install failure.


rjww:
Refreshing the autoloads should probably be added to `make install` too, because `doom/packages-install` is a requirement.

hlissner:
Good catch, thanks!

PR #52, opened 2017-05-06
add `doom/large-file-check to prompt literal open of large files

gilbertw1:
Add a prompt that will ask to open large files literally to avoid performance issues. A large file is determined by the checking the file size against the variable ```doom-large-file-size``` (default: 1MB). Opening a file literally essentially means that:

* Major and minor modes are disabled
* Buffer is readonly
* Undo is disabled

This results in drastic performance increases when opening a large file with the intention simply viewing it. This does not interfere with the built in emacs prompt when opening very large files.

Also, let me know if this functionality should be located somewhere else. ```core/editor``` seemed like the most logical location for it.


hlissner:
Excellent, thanks! And yes, core-editor seems the best fit for it.

PR #56, opened 2017-05-09
disable copy on visual state change for BOTH mac and linux

gilbertw1:
Disable automatic copying to clipboard anytime visual state changes. This was previously only applied when running on macOS, but has now been modified to apply to linux as well.


hlissner:
Huh, I hadn't realized this was a problem on Linux too. Thanks!

PR #61, opened 2017-05-16
only disable show-paren-mode locally for org mode buffers

gilbertw1:
Currently loading ```org-mode``` will disable ```show-paren-mode``` globally. Previously this wasn't noticeable because it was toggled on and off frequently via hooks, however it is not being re-activated after loading an ```org-mode``` buffer. This commit will make ```org-mode``` only disable show-paren-mode locally in the ```org-mode``` buffer.

Also, holy cow you've been slinging some elisp. All of your recent changes have been awesome!

### DIFF
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -47,7 +47,8 @@
   (setq line-spacing 1)
 
   ;; show-paren-mode causes problems for org-indent-mode
-  (show-paren-mode -1)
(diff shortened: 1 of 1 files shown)


hlissner:
Doh! I completely forgot show-paren-mode was global, haha. Thanks! 

And thanks for the praise, I've got more good stuff coming. Not to mention, your fork has inspired a couple changes too!

PR #62, opened 2017-05-16
make doom/blink-cursor interactive

gilbertw1:
Make ```doom/blink-cursor``` an interactive function, so that it can easily be called on demand to flash the cursor line.


hlissner:
Woops! My mistake. Thanks!

PR #76, opened 2017-05-21
Fix *scratch* immortality

binarin:
`eq` doesn't work on strings


hlissner:
Whoops! That must've been a buffer comparison at some point. Good catch, thanks!

PR #84, opened 2017-05-26
Fix link to mu dotfiles.

orther:


hlissner:
Whoops! Good catch, thanks!

Issue #92, opened 2017-06-03
Additional Packages


gilbertw1:
Hi again!

I've added a couple of packages to my fork, and I was wondering if you'd be interested in pull requests for either:

#### 1) [Eyebrowse](https://github.com/wasamasa/eyebrowse)

Eyebrowse is a simple window configuration (workspace) manager, that functions similarly to workspaces in i3 or XMonad. I used some code from Spacemacs to integrate it with persp-mode allowing me to have multiple workspaces within a single perspective that I can quickly flip between. Semantically, I look at a perspective like a project specific environment (like a sublime text window), and eyebrowse window configurations as workspaces within that environment (like tabs in that sublime text window).

#### 2) [Evil Goggles](https://github.com/edkolev/evil-goggles)

Evil goggles gives nice visual hints when you perform an action in evil mode (yank, delete, paste) that can be quite nice. I added a toggle function so that it can be toggled on and off easily, and I think might suit your presentation module well, even if you don't prefer to use it during regular editing.

Thanks!

hlissner:
> eyebrowse

I'm tempted, but I'll pass. My use of perspectives is more task-oriented than project-oriented. i.e. I spawn one when I've got something to do unrelated to what I'm doing -- which may traverse multiple projects -- and close them once I'm done.

There _is_ something I'd like though: bookmarks, but for window configurations. `+workspace/load` and `:sload`/`:ssave` let me do this, but it needs a friendlier interface.

> evil-goggles

Oh, I like that. I'd use it for regular editing all around. I wonder if that should go into its own `ui` module, in `feature/evil` or part of `app/present` though. Perhaps I should add a new category for `:evil` modules. Let me sleep on that and get back to you.

gilbertw1:
Using evil-goggles has worked out to be very pleasant over that past few days, it gives a nice feel to the editing commands.

I was thinking about it a bit, and I personally feel like I'm leaning a bit more towards a new category for evil related packages. The evil module has a lot of packages under it and it could be nice to split those up a bit. Maybe something like:
```
evil -
     |-> core
     |-> editor
     |-> ui
```
Although it might be difficult to decide where each package should live under a evil category. Failing that, maybe a new module under ui for things that are more or less bells and whistles? (nav-flash might work in there as well) 

Naming and structure aren't easy :)

gilbertw1:
Went ahead and created a pull request for this with the module located under ```ui```. I can modify the location if we decide on a different location.

PR #93, opened 2017-06-04
added module ui/evil-goggles

gilbertw1:
Added evil-goggles per #92 following the structure and location of nav-flash. I can move it to a different location if we decide it should live somewhere else.

Thanks!


hlissner:
Thanks! I think this is fine where it is, for now.

PR #112, opened 2017-06-16
app/irc: def-setting don't eval letvars

bandresen:
if not done this way it will try to call letvars and :user doesn't exists


hlissner:
Sorry, I have to reject this. This problem runs deeper (originating in `def-setting!`) and I already have a large-scale fix in the works. It will be up within the next couple of hours.

PR #118, opened 2017-06-23
fix :text-obj map macro

vsr625:
aB now correctly does a-block,
iB now correctly points to inner-block. It was the opposite before.


vsr625:
Sorry, I didn't ask you before the pull request, I thought it was a mistake in the code, so I did it without your consent. Was that intentionally done to point the opposite?

hlissner:
Not at all, that is certainly a mistake! Thanks for catching it.

PR #122, opened 2017-06-24
addition of Hy (hylang.org)

bandresen:
if you're interested :-)


hlissner:
Huh, Lisp and Python. An odd combination, but color me interested!

PR #123, opened 2017-06-24
app/irc: two small fixes

bandresen:


hlissner:
Good catches, thanks!

PR #127, opened 2017-06-27
Adding org-publish config.

matthewgraybosch:
Here's a rough version of an org-publish config. I haven't figured out how to get the org-to-rss functionality working, though, and the configuration assumes a particular directory structure:

**Source Directories**

* ~/work/website/
* ~/work/website/assets/
* ~/work/website/posts/
* ~/work/website/feed/

**Target Directory**

* ~/public_html/

### DIFF
--- a/modules/private/demifiend/config.el
+++ b/modules/private/demifiend/config.el
@@ -1,0 +1,79 @@
+;;; lang/org/+publish.el -*- lexical-binding: t; -*-
+
+;; Default config for org-publish.
+;; Contributed by Matthew Graybosch (public@matthewgraybosch.com)
+(setq user-login-name "demifiend")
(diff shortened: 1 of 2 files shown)


hlissner:
I can't accept the PR itself (because it's a private module), but this is an excellent reference. I'll look into it next weekend and integrate it into lang/org. I'll see if I can get org-to-rss working too.

matthewgraybosch:
That's perfectly fine. I figured this would be more useful to you as a reference implementation than a direct patch. In any case, this config needs a little more work. The html-head, html-head-extra, etc. values need to point to functions if my config is going to read HTML from partials.

matthewgraybosch:
Gonna close this PR. Please expect another one tonight once I've finished testing/debugging a new iteration suited for public distribution.

PR #129, opened 2017-06-28
org-publish in modules/lang/org

matthewgraybosch:
I've moved the org-publish config to modules/lang/org/. I haven't been able to make the functions that read partials work, though. I keep getting the following.

> org-html-template: Wrong type argument: sequencep, org-mode-head-partial


matthewgraybosch:
The basic config works now. I just need to create a test site with some content to verify that RSS export works correctly. I didn't get any errors when exporting, so it *should* work, but while I might be an Emacs Lisp n00b I've been in the trade long enough to know that "should work" isn't the same as "proven to work". :cat:

PR #137, opened 2017-07-02
Add tools/prodigy

bandresen:
I implemented this in my config and prodigy seems neat in general:
https://martinralbrecht.wordpress.com/2016/05/30/handling-email-with-emacs/


hlissner:
I'm very interested in this! In fact, I'm considering adding a new `feature/services` module for it, with some deeper integration into Doom. Thanks for letting me know about it!

It's a busy week though, so it may be a week or two until I get around to it though.

Issue #126, opened 2017-06-26
thoughts about addition of feature/hydra?


bandresen:
If you're unaware of [hydra](https://github.com/abo-abo/hydra), it is by abo-abo of ivy fame and it's very neat in situations where you're infrequently using a major mode 

My github fork is your .emacs.d with two initial commits to show hydra a little bit:
https://github.com/bandresen/emacsd/commits/master

As you can see in [5d46cc3](https://github.com/bandresen/emacsd/commit/5d46cc391189cbd85cca1a6c48e81aca1ffaf356) I'm requiring hydra the when-clause and I would like to have a defhydra! macro that adds an autoloaded definition so hydra is only loaded when the command is invoked. I don't know how to best accomplish this in the spirit of doom.

Your thoughts?

best regards,
benny

hlissner:
I've spied hydra for a while. I'll think on it and have a closer look at it next weekend.

bandresen:
Added smerge-mode to vcs including a hydra to show the usefulness of hydra.

PR #150, opened 2017-07-16
Hydra feature + smerge + git-gutter hydra

bandresen:
This is a better show case than what I had up before.


bandresen:
Just a note so it gets referenced in #126

PR #159, opened 2017-07-20
Add: app/irc: scroll to bottom on any input

bandresen:
more generalized which allows this to work for Emacs mode or for pastes.

Code inspired and adapted by eshell.


hlissner:
I rebased the fixup into a single commit and merged it independently. Thanks!

PR #173, opened 2017-08-09
Check if the hydra feature is enabled before opening smerge hydra

gilbertw1:
Pretty self explanatory ... I was receiving void function errors when ```+hydra-smerge/body``` was being called when the hydra feature was not enabled.

On the plus side, this promped me to enable the hydra module which is pretty awesome! 👍


hlissner:
Ah! Good catch, thanks!

Issue #172, opened 2017-08-08
Interest in doom-modeline change? (truncated paths)


bandresen:
I changed the modeline a little bit. I use a lot of long paths and the relative to root paths aren't always enough to tell me immediately in which project I'm in so I built a fish shell directory truncate function.

This is what I came up with:
![](http://i.imgur.com/M4rUlSR.png)

I also use it in the eshell prompt:
![](https://gitlab.com/bennya/shrink-path.el/raw/screenshots/eshell-prompt.png)

It's not yet mature but I wanted to ask that if you're into it I will polish it up and do a PR.

hlissner:
Sure, color me interested! I'd be happy to add it to both eshell and the modeline.

Issue #170, opened 2017-08-07
Modeline left hand-side doesn't appear in terminal mode


amoriello:
Hi

the left hand-side of modeline doesn't appear in terminal mode.

From what I see,

***
**modules/ui/doom-modeline/config.el**
```lisp

;; The bar regulates the height of the mode-line in GUI Emacs.
(def-modeline-segment! bar
  (when (display-graphic-p)
    (+doom-modeline--make-xpm
     (face-background (if (active)
                          'doom-modeline-bar
                        'doom-modeline-inactive-bar)
                      nil t)
     +doom-modeline-height
     +doom-modeline-bar-width)))


...



(def-modeline! main
  (bar matches " " buffer-info "  %l:%c %p  " selection-info)
  (buffer-encoding major-mode vcs flycheck))
```

❗️   bar returns nil when in terminal mode

***

**core/core-ui.el**
```lisp
(defsubst doom--prepare-modeline-segments (segments)
  (cl-loop for seg in segments
           if (stringp seg)
            collect seg
           else
            collect (list (intern (format "doom-modeline-segment--%s" (symbol-name seg))))))

(defmacro def-modeline! (name lhs &optional rhs)
  "Defines a modeline format and byte-compiles it. NAME is a symbol to identify
it (used by `doom-modeline' for retrieval). LHS and RHS are lists of symbols of
modeline segments defined with `def-modeline-segment!'.

Example:
  (def-modeline! minimal
    (bar matches \" \" buffer-info)
    (media-info major-mode))
  (doom-set-modeline 'minimal t)"
  (let ((sym (intern (format "doom-modeline-format--%s" name)))
        (lhs-forms (doom--prepare-modeline-segments lhs))
        (rhs-forms (doom--prepare-modeline-segments rhs)))
...
```
***

❗️ when `def-modline!` process the `lhs` part of the modeline, `doom--prepare-modeline-segments` doesn't expect any of the segments `seg` to be `nil`.

***

Making `def-modeline-segment! bar` return an empty string `""` instead of `nil` when not in GUI mode fixes the issue, but I'm pretty sure this isn't the right fix 😜 (thus, not proposing any PR, since I'm not fluent enough in lisp - working on it - ).

You're doing great stuff here, this project actually pushes me to learn lisp.

Hope this helps.

Thanks!

bandresen:
[...]

PR #176, opened 2017-08-11
Add: ui/doom-modeline: show project-root but truncated

bandresen:
Will display the project root's parent in truncated form (fish style) and the
relative directories as before.

Non-truncated path is shown by mouse-over on modeline.

New requirement:
shrink-path v0.3[1]

[1] https://gitlab.com/bennya/shrink-path.el


bandresen:
Forgot to mention relevant issue #172 where a screenshot is shown.

bandresen:
Doh. I guess I need to rebase again, as I can't use the commits on my own fork's master.

Now it should be good.

bandresen:
Oh right. It completely slipped my mind. Can't be merged yet as it currently doesn't work at all using `emacs -nw`.

In general there shouldn't be an issue as it's just using text properties but I still need to make sure this works in `-nw`.

modules/ui/doom-modeline/config.el
```diff
@@ -211,38 +214,43 @@ active."
...
+  (propertize (+doom-modeline--buffer-file-name 'shrink)
+              'help-echo (+doom-modeline--buffer-file-name nil)))
 
+(defun +doom-modeline--buffer-file-name (truncate-project-root-parent)
+  "Propertized `buffer-file-name'.
+If TRUNCATE-PROJECT-ROOT-PARENT is t space will be saved by truncating it down
+fish-shell style.
 
+Example:
+~/Projects/FOSS/emacs/lisp/comint.el => ~/P/F/emacs/lisp/comint.el"
```
rphillips: it should be ~/P/F/e/lisp/comint.el, correct? (emacs gets truncated)

bandresen: Depends on the parameters to the function. When using it like in the modeline.
emacs = project root
lisp/comint.el = relative path to file

It's a combination of the modeline behavior right this moment where it's relative-path to file from project root.
Now it just includes the truncated version of the project root's parent for context, the project root and the relative filename is as before.

A bit longer but way better context at least for my use case. :-)

bandresen:
Now it also includes a fix for the bar segment issue from #170 but the work was already done by @amoriello.

Also it works under `--no-window` but of course lacks the mouse-over expanded version of the path.

aminb:
Nice work, but could you please make this optional, perhaps using `def-setting!`? I usually work with *really* deep paths and my laptop's screen isn't wide enough to fit all of that (even with truncations) in the mode line; so I'd like to be able to stick with just the file name on my laptop.

amoriello:
This is great stuff, I was about to ask for more customization + doc, but you got that covered already.

Also thanks for fixing #170, though, as I said in the comment, I'm not sure this is the correct fix.

hlissner:
Sorry for taking so long to get to this (I've been sick all week). This is fantastic work, thank you!

PR #198, opened 2017-09-09
inf-haskell: fix mode-map name

Ptival:
I'll need someone to double-check this for me, but on my computer, I don't have any `inf-haskell-mode-map`, so I was getting errors like `(void-variable inf-haskell-mode-map)`.

There is however an `inferior-haskell-mode-map` (seems like poor naming convention on their side...), which at least seems to remove the error when I replace it with.


hlissner:
Good catch, thanks!

PR #200, opened 2017-09-09
fix(wrkspcs): interactive rename, save + load dir

russmatney:
Started getting into doom's workspace usage today, and these seemed to fix a few issues – not sure if they're unique to my machine or not.

This fixes renaming workspaces to work interactively by prompting for a new name from the minibuffer.

This also fixes the saving and loading of workspaces, however, an error still shows up after saving saying that it failed, despite writing to the persp file (and successful re-loads following).

I'm curious if the 'main' workspace can be renamed dynamically - i tried some code to do so but hit a few issues so thought I'd ask if it was a true necessity to protect it.

### DIFF
--- a/modules/feature/workspaces/autoload/workspaces.el
+++ b/modules/feature/workspaces/autoload/workspaces.el
@@ -116,7 +116,7 @@ Returns t on success, nil otherwise."
     (error "%s is not an available workspace" name))
   (persp-save-to-file-by-names
    +workspace-workspace-file *persp-hash* (list name) t)
-  (memq name (persp-list-persp-names-in-file +workspace-workspace-file)))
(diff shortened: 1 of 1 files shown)


hlissner:
Good catch. I'll look into workspace I/O soon and see if I can resolve that problem.

As for the main workspace, renaming it is problematic because it is a special case. `persp-mode` offers no unique id of any sort to identify a perspective, besides its name. I'll have to implement my own, or extend its struct. Then allowing users to rename ought to become trivial. In any case, I'll look into it.

hlissner:
I've pushed a fix for the main workspace to the `develop` branch. It is no longer protected and can be dynamically renamed with `+workspace/rename` (or `:tabr[ename]` if you're an evil user and are using my private module).

PR #204, opened 2017-09-13
Fix a spelling mistake during variable assignment

PigeonF:
_savehist-save-minibuffer-history_  was misspelled as _savehist-save-minibuffer-his**s**tory_. I assume that this was not intentional, so I fixed it.

I could not find any information about contributing (the link at the bottom of the readme is broken), so I hope that this is fine. If I made some kind of mistake, just tell me.


hlissner:
Good catch, thanks!

I've removed the link from the readme for the time being. Thanks for letting me know!

Issue #185, opened 2017-09-02
Enhancing the JavaScript lang module.


orther:
I've been working heavily on a large JavaScript project and slowly changing the provided doom `lang/javascript` module. I've broken things a lot but am really starting to feel stable. if you are open to changes in the JS module I'd like to clean things up and create pull request? I have added a lot of functionality that I do think some people will want but not all. So it would make sense to make certain features option but live within the `lang/javascript` module. So my questions are:

1. Are you open to changes to the JS module?
2. Is there an idiomatic way for modules to have opt-in functionality? (e.g. enabling jest test runner support, enabling automatic code format on save)

Before I refactor these our of my own private module I want to make sure it is something you and/or other people would want. Below is a list of what I've added and have been using. 

## What I've Added
Most of these are fully working but some clean organizing is needed:
 - Auto Format Code: eslint w/ prettier
 - Running tests w/ Jest or Mocha for current buffer or current project using [mocha.el](https://github.com/scottaj/mocha.el/)
   - NOTE: If I can enable RealGUD (I saw you disabled it a few months back) interactive debugging works for Jest and Mocha.
 - Add `{project-root}/node_module/.bin/` to path so locally installed binaries are used.
 - Use `eslint_d` if available otherwise fallback to `eslint`
 - Flow type suppport w/ company mode integration.
 - Code coverage visual overlay to easily see what is being tested using [coverlay.el](https://github.com/twada/coverlay.el)

hlissner:
That'd be fantastic. I'd happily welcome all those changes.

As for opt-in functionality, take a look at `def-project-mode!`. It lets you define project-specific minor modes. Here's an example from the javascript module:

```emacs-lisp
(def-project-mode! +javascript-npm-mode
  :modes (html-mode css-mode web-mode js2-mode markdown-mode)
  :files "package.json")
```

This minor mode only activates in html/css/web/js2/markdown-mode buffers AND when `package.json` is present in the project root.

This gives you the ability to have (among other things) project-specific...

+ keybindings (attach to `+javascript-npm-mode-map`)
+ snippets or file templates (in a `+javascript-npm-mode` folder in your yasnippet snippets / the file templates module)
+ hooks (`+javascript-npm-mode-hook`)

These modes can also be activated through `.dir-locals.el`, wherein you set `doom-project` to the name of the project mode you want to activate.

Presumably, you could create jest/mocha/tape/anything project minor modes, and attach your custom functionality to it. Is that what you're looking for?

[...]

PR #214, opened 2017-09-28
Add project-mode +javascript-eslintd-fix-mode

orther:
This the following features mentioned in #185 have been implemented using the `def-project-mode!`:

 - Auto Format Code: eslint w/ prettier
 - Use eslint_d if available otherwise fallback to eslint

I didn't know how you prefer the documentation for this feature be added so I am leaving it here in pull request. 

# Documentation
This project mode enables automatic formatting of javascript source on save using `eslint --fix`.

## Install
This project mode requires `eslint_d` for auto formatting.
```sh
npm -g install eslint_d
```

## Setup
To enable this feature for a specific project you can use `.dir-locals.el` to set `doom-project` to include `+javascript-eslintd-fix-mode`. Here is an example of my `.dir-locals.el` file which is located next to `package.json` in the root of my JS project: 
```elisp
;;; Directory Local Variables
;;; For more information see (info "(emacs) Directory Variables")

((nil . ((doom-project . (+javascript-eslintd-fix-mode)))))
```

### Prettier code formatter
You can get [Prettier](https://github.com/prettier/prettier)'s great AST based code formatting using this project mode with this eslint pluging that runs Prettier as an ESLint rule:
 - https://github.com/prettier/eslint-plugin-prettier

### DIFF
--- a/modules/lang/javascript/config.el
+++ b/modules/lang/javascript/config.el
@@ -6,7 +6,8 @@
   :config
   (setq js2-skip-preprocessor-directives t
         js2-highlight-external-variables nil
-        js2-mode-show-parse-errors nil)
+        js2-mode-show-parse-errors nil
+        js2-strict-trailing-comma-warning nil)
 
   (add-hook! 'js2-mode-hook
     #'(flycheck-mode highlight-indentation-mode rainbow-delimiters-mode))
@@ -136,6 +137,17 @@
   :init
   (map! :map* (json-mode js2-mode-map) :n "gQ" #'web-beautify-js))
 
+(def-package! eslintd-fix
+  :commands (eslintd-fix-mode eslintd-fix)
+  :init
+  (defun +javascript|init-eslintd-fix ()
+    (when (bound-and-true-p +javascript-eslintd-fix-mode)
+      (eslintd-fix-mode)
+      ;; update flycheck to use eslintd for more consistent results
+      (when-let (eslintd-executable (executable-find "eslint_d"))
+        (setq flycheck-javascript-eslint-executable eslintd-executable))))
+  (add-hook! (js2-mode rjsx-mode) #'+javascript|init-eslintd-fix))
+
 
 ;;
 ;; Skewer-mode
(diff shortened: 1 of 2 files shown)


hlissner:
This is a great start. Two issues jumped out to me:

1. I think we can silently `(setq flycheck-javascript-eslint-executable "eslint_d")` if `(executable-find "eslint_d")` succeeds, and we should do this outside of the hook. After all, if it's installed, it's safe to assume we want to use it globally, and seems to be safe as a drop-in replacement.
2. The definition of `+javascript-eslintd-fix-mode` is incomplete. It shouldn't be tied to `+javascript-npm-mode`. We can also do away with the `+javascript|init-eslintd-fix` hook and simply add `eslintd-fix-mode` to `+javascript-eslintd-fix-mode-hook`.

In any case, I've merged this and will refactor as necessary. Your PR has exposed a couple issues in `def-project-mode!` that I'll need to address too.

I look forward to your other lang/javascript improvements!

modules/lang/javascript/config.el
```diff
@@ -20,14 +19,16 @@
...
     (when (derived-mode-p 'js-mode)
       (when-let ((eslint (expand-file-name "node_modules/eslint/bin/eslint.js"
                                            (doom-project-root)))
                  (exists-p (file-exists-p eslint))
                  (executable-p (file-executable-p eslint)))
-        (setq-local flycheck-javascript-eslint-executable eslint))))
-  (add-hook 'flycheck-mode-hook #'+javascript|init-flycheck-elint)
+        (setq-local flycheck-javascript-eslint-executable eslint)
+        (setq-local js2-strict-trailing-comma-warning nil)
+        (setq-local js2-strict-missing-semi-warning nil))))
```
hlissner: @orther Do these two options apply to eslint_d too?

orther: Yes, once eslint is setup for flycheck linting I don't think it makes sense for js2 warnings to show since they can conflict with what is explicitly set in the eslint config. So there are possibly more that should be disabled but these are just the ones that popped up for me.

orther: Yes, once eslint is setup for flycheck linting I don't think it makes sense for js2 warnings to show since they can conflict with what is explicitly set in the eslint config. So there are possibly more that should be disabled but these are just the ones that popped up for me.

orther:
I appreciate the feedback as I'm rather new to elisp and still learning doom-emacs so it's extremely informative.

Flow, Jest and a few more JS related changes are incoming. I started with this small specific pull request to start with on my JS module update because I wanted feedback. Moving forward I will likely include several JS module changes into a single pull request (unless you prefer otherwise).

PR #217, opened 2017-10-03
Fix `SPC x` (function was renamed to 'doom/open-scratch-buffer)

aminb:


hlissner:
Good catch, thanks!

aminb:
Cheers!

PR #224, opened 2017-10-05
Fix markdown specific keybindings being added to global map

gilbertw1:
This pull request is pretty straight forward and prevents the markdown localleader keybindings from being added to the global keymap.


hlissner:
Good catch, thanks!

PR #239, opened 2017-10-09
Fix commands that would depend on 'git-link' being loaded.

UndeadKernel:
Some commands, e.g. `+vcs-root`, assume that "git-link" is loaded.
I have added all commands needed from "git-link" in the :commands declaration of its `def-package!`.

### DIFF
--- a/modules/feature/version-control/+git.el
+++ b/modules/feature/version-control/+git.el
@@ -85,5 +85,7 @@
 
 
 (def-package! git-link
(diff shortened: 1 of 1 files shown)


hlissner:
I would prefer that `+vcs-root` do a `(require 'git-link)` rather than add more autoloads to git-link's config, because the implementation of those `+vcs-*` functions (and the private functions they use) are more volatile. 

Could you make that change? If not, close this and I will shortly.

PR #254, opened 2017-10-26
Exclude .dir-locals.el from file-templates

amosbird:


hlissner:
I have a more robust solution for this in my local `develop`, so I'll have to turn it down. I'll push it up soon.

amosbird:
Great!

PR #256, opened 2017-10-26
Fix dump-jump with local fadvice macro.

amosbird:


core/core-lib.el
```diff
@@ -29,6 +29,18 @@
 ;; Helpers
 ;;
 
+;; stealed from https://emacs.stackexchange.com/a/16495/16662
```
vsr625: *stole

hlissner:
I'll have to turn this PR down. I'm hesitant to extract `doom-with-advice` into its own macro just for one usage. I'm currently trying out a slightly different solution to this in `develop`, as of https://github.com/hlissner/doom-emacs/commit/4c3f5e8cc28831b2fe6041a7cfbe452366056b2e.

If it isn't much too trouble, please give it a try.

PR #264, opened 2017-11-08
Fix 'Wrong type argument: commandp' in +workspace/kill-session-and-quit.

sarg:
image

After pressing SPC q Q, I get this error. Seems, that `(interactive)` fixes the problem.


hlissner:
Ah! Good catch, thanks!

Issue #241, opened 2017-10-10
Understanding how to efficiently setup packages.


teesloane:
After reading through the docs I'm still a bit confused on how to properly load packages. I have noticed an increase in emacs loading time (1s --> 3s) after having added 5 packages. 

My private module for doom is version controller [here](https://github.com/teesloane/dotfiles/tree/master/tees)... but I'm mostly wondering - when to use the `package!` and `def-package!` macros. 

Currently I'm doing the following in `packages.el`

```
;; -*- no-byte-compile: t; -*-
;;; private/tees/packages.el

(package! rjsx-mode)
(package! doom-themes)
(package! prettier-js)
(package! browse-at-remote)
(package! yaml-mode)
(package! markdown-mode)


;; This might be slowing boot ... should make this happen on hooks?
(def-package! rjsx-mode        :demand t :config)
;; (def-package! doom-themes      :demand t :config) ;; I think this is already installed
(def-package! prettier-js      :demand t :config)
(def-package! browse-at-remote :demand t :config)
(def-package! yaml-mode        :demand t :config)
(def-package! markdown-mode    :demand t :config)
```

I noticed that if I don't use `:demand` a package is not available in `M-x`. 

I'm super new to configuring emacs, but having lots of fun with this. So if you see anything that could be improved let me know...

Thanks for your time and all your hard work!

teesloane:
@hlissner thank you for such a detailed answer. That definitely clears thing up a lot. I had no idea most of those packages were already in Doom ha! 

As for the example for rjsx-mode -- that illuminates much. I'm guessing that if I wanted to, say, have all js files in a folder called `app` I could similarly create a similar `(def-package)` call, borrowing from the build in lang/javascript where I could override when the mode is invoked.

This has been a lot of fun for me, coming from spacemacs and learning more elisp and actually (kind of) knowing what's going on; so the details are much appreciated.

Issue #171, opened 2017-08-07
How can one move a REPL window?


Ptival:
When I open a REPL using `C-c C-l`, the window it creates seems special (darker background, exits the process when closed, ...).

It also always seems to split my screen horizontally, which is sometimes useful, sometimes annoying. Not knowing enough about emacs, I can't seem to find a way to move this window to the right side.
I have managed to get it in another frame by using the switch-buffer commands, but I can't close the original one because it kills the process and all frames that have the buffer in it.

Could you give me pointers as how to deal with this special buffer and how it is displayed? :(

Thanks!

hlissner:
Hmm, there are a lot of moving parts here. I'm not sure what you've got `C-c C-l` bound to. If you aren't using the `+eval/repl` command to start a mode's repl, then Doom will treat the repl buffer like a common special buffer.

In that case, the two relevant variables are `shackle-default-alignment` and `shackle-rules`, which tell Emacs how to deal with windows that meet a certain criteria.

At the end of `shackle-rules` there are two catch-all rules:

```emacs-lisp
(set! :popup 
  '("^\\*"  :regexp t :noselect t :autokill t)
  '("^ \\*" :regexp t :size 12 :noselect t :autokill t :autoclose t))))
```

The two problems here are `:autokill t` (which ensures a buffer and its associated processes are killed together) and since there is no `:align` specified, it resorts to `shackle-default-alignment`, which defaults to `'below`.

So, how do we fix this?

Here are three ways:

1. Simply specify an overriding rule for your repl in particular:

   ```emacs-lisp
   (set! :popup "*my-repl-buffer*" :align 'left :noesc t)
   ```

   See `M-x doom/describe-setting RET :popup` for details.

2. Enable the `+eval-repl-mode` minor mode in your repl:

   `(add-hook 'inferior-python-mode-hook #'+eval-repl-mode)`

   You'll find this popup rule in [modules/feature/eval/config.el](https://github.com/hlissner/.emacs.d/blob/master/modules/feature/eval/config.el#L43):

   ```emacs-lisp
   (set! :popup
     '(:custom (lambda (b &rest _) (buffer-local-value '+eval-repl-mode b)))
     :size 16 :noesc t)
   ```

3. Bind `C-c C-l` to `+eval/repl` and use that instead to invoke your repl. Most modes have reasonable defaults. If it's not opening the repl you want, you can change the major-mode->repl associate with the `:repl` setting, e.g.

[...]

PR #269, opened 2017-11-19
Updates to clojure module

teesloane:
Hey hey!

I've been writing some more clojure in my life so I thought I'd try and add some stuff to the clojure module. This pr isn't really complete -- it works but I'm still grokking how these modules work, (although I did research into use-package as you recommended in an issue I opened ( #241  ))

Either way, feel free to review if it interests you; or to dismiss. Either way - I've got something kind of half working, and wanted to check if it would add any value to your project!

I've got the localleader keybindings working, cider is working, but I couldn't get the repl to not be a popup, as mentioned in #171 

I looked at putting a +repl solution in, but wasn't sure how to go about that. I looked at the python example:

```
;;;###autoload
(defun +python/repl ()
  "Open the Python REPL."
  (interactive)
  (process-buffer (run-python nil t t)))

```

But perhaps this would be more complicated between clojurescript and clojure with cider. I have had issue with getting repls to work with clojurescript... there seem to be some edgecases sometimes, in my experience, so I'm not sure if it's wise to bake cider into the doom-repl feature. 

One last note - I had to add `clojure` to the `init.el` to get it to work -- although that file is gitignored

Cheers!


modules/org/org/config.el
```diff
@@ -6,7 +6,7 @@
   (push (file-name-directory path) load-path))
 
 ;; Custom variables
-(defvar +org-dir (expand-file-name "~/work/org/")
+(defvar +org-dir (expand-file-name "~/Dropbox/notes/")
```
teesloane: woops, this should not have been commited

hlissner:
This looks fantastic, great work! I've been away from the project for a short while, but will be back in the next few days to look into merging this asap.

teesloane:
Sounds good! Happy to make nay changes that you need as well. Cheers!

hlissner:
Alright! Now that I'm back I had some time to review your PR. It looks great! There are a few minor changes I'd like to ask for, to make this module more idiomatic:

1. Unnest the `(def-package! clj-refactor ...)` block and use `:after` to chain it to clojure-mode:

   ```emacs-lisp
   (def-package! clojure-mode ...)
  
   (def-package clj-refactor
     :after clojure-mode
     :config
     ...)
  
   ...
   ```

2. The `:commands (clojure-mode clojurescript-mode)` line is unnecessary. The `:mode` declarations add autoloads for these already.

3. To have shackle ignore cider, try the `:ignore` property:

   ```emacs-lisp
   (set! :popup "^\\*cider-.*" :ignore t)
   ```

4. I know little about Clojure and Cider, so I can't give any specific advice about integrating its repl into doom's repl system. If, for some reason, clojure's repl is completely incompatible with Doom's popup system, then it's best you avoid its repl system too.

    If this is not the case, doom's repl system works like this:

   + There must be an interactive function, canonically named `+module/repl`, e.g. `+python/repl`. It must open (and return) a repl buffer.
   + This function must be registered to a major mode, like so: `(set! :repl 'python-mode #'+python/repl)`
   + The repl can then be invoked from python-mode buffers using `M-x +eval/open-repl` (bound to `SPC o r` in my private module).

Hope that helps!

teesloane:
@hlissner Thanks for the feedback!

I made the requested changes (let me know if I missed anything though!).

Regarding the repl; I think the real frustration for me was that I kept hitting `ESC` (mapped to caps on my keyboard) and that would kill the repl. After reading the shackle code I learned about noesc, so that solves my problem entirely! (before I was taking the repl, and then moving it to the right with `SPC w L` (which was a bit annoying as the cider repl boots _two_ instances of the repl when you boot a clojurescript repl.)

Should be good now, cheers!

hlissner:
Excellent! One more thing: could you try:

```diff
-(set! :popup '("^\\*cider" :regexp t :noselect t :autokill f :autoclose f :noesc t))
+(set! :popup "^\\*cider" :regexp t :noselect t :noesc t)
```

If you omit :autokill or :autoclose, it _should_ achieve the same effect. If all works well, then it will be ready to be merged!

teesloane:
Changes made! Looks like all still works ok!

hlissner:
Great! And merged. Thanks again!

PR #281, opened 2017-12-03
enable hl-line-mode only when it's not term mode

showgood:
Enable `hl-line' after exiting evil-visual mode but only when major mode is not term-mode.


hlissner:
I have to reject this, not because I disagree, but because I already had a better solution in my local repo. I've pushed this to `develop` now, and it will make its way into master soon.

PR #297, opened 2017-12-20
Remove the trailing space of doom dashboard

amosbird:
This is a bit weird when displaying trailing white spaces is on.


hlissner:
I've fixed this another way: by "disabled" whitespace-mode in the doom dashboard instead, so I'll close this PR, but thanks for bringing this to my attention!

amosbird:
Hmm, I still see the trailing whitespaces after applying the patch.

hlissner:
Oh! You're using `show-trailing-whitespace`, not `whitespace-mode`, right?

amosbird:
It works fine now. Thanks!

PR #299, opened 2017-12-20
Fix private module readme's links

ruimams:
Currently the links in [modules/private/README.org](https://github.com/hlissner/doom-emacs/blob/3d323e6ecfca571e1e4d66f0b466711e6a108d6f/modules/private/README.org) are broken. I propose using absolute links, since they work in forks and locally.

Feel free to use relative links if you want.


hlissner:
Thanks!

PR #321, opened 2018-01-04
fix ivy.

amosbird:
Due to this commit 

https://github.com/abo-abo/swiper/commit/4f57b5a9eb3f9c971d48fb33dca54af43bd497ca


hlissner:
I'm already aware of this and have fixed it locally. I have held off pushing it because it is tied to many other _big_ commits. It should be up within the next 24 hours, so I must turn down this PR. Thanks for your help in any case!

PR #332, opened 2018-01-08
Latex module revamped

UndeadKernel:
Completion with company mode.
Selection of bibliography using Ivy or Helm.
Later preview panel or okular as viewers.
LatexMk for compiling code.
Prettified indentation with adaptive-wrap along with good indentation of environments.
Additional fontification of common commands.

### DIFF
--- a/modules/lang/latex/+fontification.el
+++ b/modules/lang/latex/+fontification.el
@@ -1,0 +1,89 @@
+;;; lang/latex/+fontification.el -*- lexical-binding: t; -*-
+
+;; Fontification taken from https://tex.stackexchange.com/a/86119/81279
+(setq font-latex-match-reference-keywords
+      '(
+        ;; biblatex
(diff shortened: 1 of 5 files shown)


hlissner:
This looks great! I look forward to trying it out. 

Just two adjustments I'd like to ask for:

+ `def-setting!` settings needs to return a quoted form (like `defmacro`), otherwise those settings won't run in a byte-compiled config.
+ Use `string-empty-p` rather than `s-blank?`. I understand ivy-bibtex depends on s.el, but where possible (and simple), I prefer that Doom code use built-in libraries over external ones.

UndeadKernel:
I have addressed the issues you mentioned @hlissner.
Let me know if you find anything else that needs changes.

hlissner:
Fantastic! I'll give it a more thorough test run this weekend. Thanks for your help!

PR #333, opened 2018-01-09
Call flycheck-popup-tip-mode correctly

dsdshcym:
- Fix issue: Symbol’s function definition is void: flycheck-popup-tip


hlissner:
Ah! What a silly mistake. Thank you!

PR #336, opened 2018-01-09
Add Vue file to web mode

Dragnucs:
Support `.vue` Vue.js files in web-mode.


hlissner:
Thanks!

tosiek88:
How to assigne .vue extension file to vue-mode, maybe it will be good to removed it from web-mode and do it as defualt vue-mode, I'm using vue-mode with Doom Emacs, it's all right. 
There is couple lacks with colors , for example method are same color as object but i can live with that :)  
```
(add-to-list 'auto-mode-alist '("\\.vue\\'" . vue-mode))
``` 
I put that line in config.el but is not working

PR #339, opened 2018-01-10
Fix popup evil command window execute

dsdshcym:
See commit descriptions for details.

### DIFF
--- a/modules/feature/popup/+hacks.el
+++ b/modules/feature/popup/+hacks.el
@@ -80,7 +80,7 @@ the command buffer."
           (select-window execute-window)
         (user-error "Originating buffer is no longer active"))
       ;; (kill-buffer "*Command Line*")
-      (+popup/close popup)
+      (delete-window popup)
       (funcall execute-fn result)
       (setq evil-command-window-current-buffer nil)))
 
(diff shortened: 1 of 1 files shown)


modules/feature/popup/+hacks.el
```diff
@@ -74,12 +74,13 @@ the command buffer."
...
                                     (line-end-position)))
           (execute-fn evil-command-window-execute-fn)
+          (execute-window (get-buffer-window evil-command-window-current-buffer))
           (popup (selected-window)))
-      (select-window doom-popup-other-window)
-      (unless (equal evil-command-window-current-buffer (current-buffer))
+      (if execute-window
+          (select-window execute-window)
         (user-error "Originating buffer is no longer active"))
       ;; (kill-buffer "*Command Line*")
```
dsdshcym: Why was this line commented?

I tried to uncomment it but it seems to be not working correctly (`*Command Line*` buffer was not killed)

hlissner:
It was commented out because it would do nothing. There is a function in `kill-buffer-query-functions` that prevents killing buffers which are visible. However, when a window is deleted, the popup system will take care of lingering buffers (think of it as the reference counting garbage collector of buffer management).

That said, this is good. I hadn't realized the command window wasn't working until now. Quick request: could you change that `+popup/close` to `delete-window`? The former will pollute the popup history.

dsdshcym:
@hlissner updated

PR #342, opened 2018-01-11
feature/evil: add mapping for evil-numbers

dsdshcym:
1. `evil-numbers` is in the `packages.el` of feature/evil, but not configured with any `def-package!` statements
2. About the keybinding, I think it would be better to use the same keybindings as vim (since doom is "An Emacs configuration for the stubborn martian vimmer"). It might be uncomfortable for Emacs users (since it takes `C-x`), but I've been using this config for some time, and it works fine for me (a former vim user and spacemacs user)

### DIFF
--- a/modules/feature/evil/config.el
+++ b/modules/feature/evil/config.el
@@ -229,6 +229,14 @@
   (add-hook 'doom-escape-hook #'+evil|escape-exchange))
 
 
+(def-package! evil-numbers
+  :commands (evil-numbers/inc-at-pt evil-numbers/dec-at-pt)
+  :init
+  (map!
+   :nv "C-a" #'evil-numbers/inc-at-pt
+   :nv "C-x" #'evil-numbers/dec-at-pt))
+
+
 (def-package! evil-matchit
   :commands (evilmi-jump-items evilmi-text-object global-evil-matchit-mode)
   :config (global-evil-matchit-mode 1)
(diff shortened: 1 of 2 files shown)


hlissner:
>evil-numbers is in the packages.el of feature/evil, but not configured with any def-package! statements

Whoops, must've slipped through on one of my late night code binges!

> About the keybinding, I think it would be better to use the same keybindings as vim

I prefer those keybinds myself, but I hesitate to rebind `C-x`. It seems too essential a prefix key to Emacs for me to override by default. Well, I guess I'll try it out for now.

Although, could you move those keybinds to [private/default/+bindings.el, after L366](https://github.com/hlissner/doom-emacs/blob/develop/modules/private/default/+bindings.el#L366). Keybinding should be avoided in package configs for anything but fixing or remapping keybinds, not adding new ones. It is also easier for people to discover new keybinds in private/default.

Thanks a lot for the help, in any case!

cosminilie:
I've tried the config with this mapped to C-x for a couple of days and i can say it's not working out. I would remap it something else.

hlissner:
After a lot of consideation, I've decided to change this C-x keybind to C-S-a. Sorry to say, rebinding C-x has proven far too imposing. The prefix is too important to Emacs.

In any case, you can always put `(map! :nv "C-x" #'evil-numbers/dec-at-pt)` in your private module.

PR #348, opened 2018-01-11
Correctly report when rtags's client or server are not found.

UndeadKernel:
If rtags is not installed, a more suitable message is displayed when cc-mode is activated.

Previously, if rtags was not installed, the error message would state that "nil" was not found.


hlissner:
Ah, good catch. Thanks!

PR #350, opened 2018-01-12
Fix race between eldoc and flycheck by delaying flycheck

vikigenius:
Eldoc uses a default idle delay timer of 0.5.
In your cc module flycheck also is using a `flycheck-display-errors-delay` of 0.5.

This is leading to a race between flycheck and eldoc for the minibuffer, and in my case flycheck never displays the error message and eldoc is showing the docstring even though there is an error in the function parameters.

So i have delayed the flycheck display from 0.5 to 0.7 so that it overrides eldoc in case of errors.


hlissner:
Good catch, thanks!

PR #355, opened 2018-01-13
Add: =calendar app

fuxialexander:
Added calendar app with `calfw` and `org-gcal`


hlissner:
This looks pretty good, great work!

PR #359, opened 2018-01-15
feature/evil: fix +evil/matchit-or-toggle-fold in magit modes

dsdshcym:
- this function only worked in magit-status-mode but not other modes like
  magit-log-mode, magit-process-mode, etc.

### DIFF
--- a/modules/feature/evil/autoload/folds.el
+++ b/modules/feature/evil/autoload/folds.el
@@ -88,7 +88,7 @@
   (interactive)
   (ignore-errors
     (call-interactively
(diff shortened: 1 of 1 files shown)


hlissner:
Sorry for the delay; would this still work for you if you used `(derived-mode-p 'magit-mode)` instead?

dsdshcym:
Yes, it still works.

Good to know that emacs has `derived-mode-p` function. Thanks!

hlissner:
Excellent, good work!

PR #364, opened 2018-01-17
Custom LaTeX modeline indicating how many words are in a region

UndeadKernel:
When writing text, it is often useful to know how many words are being written.
This change adds a custom modeline that indicates, with the character "W", how
many words are in a region.

@hlissner, I didn't want to change the **main** modeline for all buffers, so I created a custom one. However, it could be of interest to display also the number of words in the main modeline, maybe?

### DIFF
--- a/modules/lang/latex/+modeline.el
+++ b/modules/lang/latex/+modeline.el
@@ -1,0 +1,29 @@
+;;; lang/latex/+modeline.el -*- lexical-binding: t; -*-
+
+
+(def-modeline-segment! +latex-selection-info
+  "Information about the current selection customized for LaTeX
+buffers, such as how many characters, words and lines are
+selected, or the NxM dimensions of a block selection."
+  (when (and (active) (or mark-active (eq evil-state 'visual)))
+    (let ((reg-beg (region-beginning))
+          (reg-end (region-end)))
+      (propertize
+       (let ((lines (count-lines reg-beg (min (1+ reg-end) (point-max))))
+             (words (count-words reg-beg (min (1+ reg-end) (point-max)))))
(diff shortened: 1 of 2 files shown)


Paramonov77:
Мл

hlissner:
@UndeadKernel That's a good idea. I think I'll modify the main modeline to include a word-count in text-mode derived buffers (or add a toggleable option for it, if folks want it everywhere -- or in certain buffers). What do you think?

UndeadKernel:
I think it would also be a good idea to integrate word counts idea into the main modeline (for specific modes).
Feel free to close this pull request whenever it suites you best.

hlissner:
The `selection-info` segment now appends a word count if buffer-local `+doom-modeline-enable-word-count` is non-nil (this is set to `t` in text-mode derived buffers already), so I'll consider this resolved, and leave it open until it hits `master`.

hliebert:
The reported count is wrong in `visual-line-mode` but correct in `visual-mode`.

UndeadKernel:
I cannot reproduce your observation. While using `visual-line-mode`, the reported number of words is correct.

hliebert:
This illustrates the behavior, I just noticed it while I was checking the words in an abstract for the allowed limit. The first line isn't counted, and the cursor position influences the count although it probably should not. I'm sorry im not of more help, I am just starting out with emacs after years of vim. If I can dig deeper let me know how and I'll happily do it. 

peek 2018-04-18 18-50

hlissner:
@hliebert Found and fixed the issue. The modeline segment wasn't taking evil's visual-line mode into account and wasn't reading words beyond the cursor.

PR #383, opened 2018-01-29
add +bindings feature for private/default

driftcrow:
private/default module is convinience, but if I want create bindings for
my own, it's seems that change prefix key is not handy. for example:

(!map
(:leader
(:prefix "t"
....
)))

change all keybind prefix with "t" to prefix with "f" ....

thou, +bindings feature can disable load default binding, then I can load
binding in my private module ,but still use default module's others part.


hlissner:
I thought about this for a while and decided to introduce `+snippets` and `+evil-commands` as well as `+bindings`. Thanks for bringing this to my attention!

PR #387, opened 2018-01-31
Check if buffer-file-name is nil before passing it as argument.

sarg:
When I call `(doom/describe-module)` from dashboard the function raises an error, because buffer-file-name is not defined for dashboard.


hlissner:
Nice, thanks!

PR #392, opened 2018-02-01
Sensible alias

amosbird:


hlissner:
Heh. I got so used to my `lam` snippet that I forgot that that macro was difficult to type! Thanks!

PR #415, opened 2018-02-10
Fix rss module.

sarg:
Hi there. I have turned on rss module and found two bugs.
1. After fresh start of emacs org is not loaded. This prevents elfeed-org to be invoked when I start elfeed.
2. When I press `q` in elfeed I get error message about `kill-buffer` arguments.


hlissner:
Ah, good catch! I used to have a `doom-kill-buffer` that took two arguments. I must've missed this when I was replacing references to it. Thanks for your help!

PR #418, opened 2018-02-13
Fix adding evil-set-jump to rtags-jump-hook when evil is not enabled

azy2:
When evil is disabled functions like `rtags-find-symbol-at-point` don't work because the function evil-set-jump isn't defined. I'm proposing the simple fix of only adding the hook when `(featurep 'evil)`


hlissner:
You're quite right. Thanks for this!

PR #419, opened 2018-02-13
Fix +plantuml/install.

sarg:


hlissner:
Ah! Good catch, thanks!

PR #421, opened 2018-02-14
Fix keybinding

benj02:
Both flycheck-mode and toggle-frame-fullscreen seem to be bound to "SPC t f", which means that only the second binding gets applied and there's no binding to toggle flycheck. This PR simply changes the fullscreen binding to capital F.


hlissner:
A silly mistake. Thanks for fixing it!

PR #425, opened 2018-02-17
Disable flycheck for emacs-lisp in +private-config-path

azy2:
With the addition of +private-config-path we've moved our elisp outside of doom-emacs-dir and flycheck gets turned on for these files. This change keeps flycheck behavior consistent. I'm not sure if this is the best way to do it, though. I'm still learning elisp.


hlissner:
I've thought up a slightly more flexible version in 87f47e6.  Thanks for bringing this to my attention.

PR #437, opened 2018-02-26
Added helm to the popup ignore list

azy2:
Helm manages it's own windows and doesn't interface with popup very well.

### DIFF
--- a/modules/completion/helm/config.el
+++ b/modules/completion/helm/config.el
@@ -71,7 +71,8 @@
         [remap projectile-switch-project] #'helm-projectile-switch-project
         [remap projectile-find-file]      #'helm-projectile-find-file
         [remap imenu-anywhere]            #'helm-imenu-anywhere
-        [remap execute-extended-command]  #'helm-M-x))
+        [remap execute-extended-command]  #'helm-M-x)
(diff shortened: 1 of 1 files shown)


hlissner:
I'm afraid I can't accept this into Doom. I consider this as a personal customization and would rather leave this to private modules than make this the default, for a couple reasons:

1. I've tried to model the helm and ivy modules to act visually similar to CtrlP (vim), Goto-Anywhere (sublime text/atom), and command-T (textmate), as a reasonable default. I agree that it doesn'tt suit widescreens well, but I'd like to implement child frame support to address this).
2. The completion/helm module's defaults work fine with feature/popup. If this is not the case, that is a bug.
3. It is easier to put `(set! :popup "^\\*helm" :ignore)` in a private config than it is for another user to either remove that rule from `display-buffer-alist` or define a new rule to shadow the old one simply to reinstate the fallback behavior of `+all`.

azy2:
That seems reasonable to me. I didn't think it would do anything to the default setup and I figured it would make customizing helm less confusing. It certainly took me awhile to figure out why my helm settings weren't applying. Perhaps there's a fix in popup that's less drastic than `:ignore` that maintains the original behavior.

PR #438, opened 2018-02-26
feature/evil: add org-agenda-mode to evil-snipe-disabled-modes

AloisJanicek:
Before this commit, pressing `t` in org-agenda buffer resulted in activation of evil-snipe instead of setting todo state as how it is intended by [evil-org](https://github.com/Somelauw/evil-org-mode/tree/491b0b302b95d44ceb73d291dedbb9d5517ccee2#agenda)

(If this was on purpose and Doom has some custom way for setting todo states from org-agenda buffers I am not aware, I apology for pointless pull request)


hlissner:
This was definitely an oversight and I'm glad to see it corrected. Thanks!

Issue #446, opened 2018-03-01
Error (doom-init-hook) since commit 138ec2bf


dariooddenino:
1) git checkout 138ec2bf
2) make and then launch emacs
3) `Error (doom-init-hook): void-function in 'helm-mode' -> Symbol's function definition is void: helm-mode`

That's the first checkout that gives me the error and I still get it with the last develop commit

hlissner:
Try to delete `~/.emacs.d/.local/packages.el` then restart Emacs and see if that fixes your issue.

This is a temporary measure, I'm in the process of addressing issues with Doom's package management system.

dariooddenino:
I get the same error. These are the steps in case I'm doing something wrong:
1 git checkout develop
2 delete `.emcas.d/.local/packages.el`
3 make
4 launch emacs

dariooddenino:
Ok it works if I delete the file before launching emacs. I have to delete it everytime I want to launch it though :D

hlissner:
Yikes! I'll look into this.

hlissner:
Alright, this should resolve your issue with the bad cache. You may need to delete the file one last time though. Let me know if that isn't the case!

azy2:
@hlissner I still have to delete `.emacs.d/local/packages.el` every time before launching emacs. For me `Helm-M-x` and `magit-status` are broken.

dariooddenino:
It's fixed for me now

hlissner:
@avy2 @dariooddenino I've just pushed a possible fix for your haskell, helm and magit issues. Let me know if it doesn't fix your issues.

dariooddenino:
Looks like it's fixed now :)

PR #456, opened 2018-03-07
feature/evil: load evil-vimish-fold autoloads #446

azy2:
This fixes my issue with magit as described in #446.

### DIFF
--- a/modules/feature/evil/config.el
+++ b/modules/feature/evil/config.el
@@ -338,10 +338,9 @@ the new algorithm is confusing, like in python or ruby."
 
 
 (def-package! evil-vimish-fold
-  :commands (evil-vimish-fold/next-fold evil-vimish-fold/previous-fold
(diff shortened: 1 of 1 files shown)


hlissner:
How does evil-vimish-fold affect magit?

azy2:
Not sure. I use the package `evil-magit` which may have something to do with it. But whenever I try to run `magit-status` without deleting `~/.emacs.d/.local/packages.el` I get an error saying regarding evil-vimish-fold not being loaded. This mitigates the error. If it's a bug with `evil-magit` I can keep the change local though.

hlissner:
Ah! I've found the issue [in your private module, here.](https://github.com/azy2/doom-emacs-private/blob/master/modules/private/azy2/config.el#L32).

You don't need this anymore. As of https://github.com/hlissner/doom-emacs/commit/0d27eda45a32ef46117a575c2c29d46426cc35f0 Doom doesn't turn on evil-vimish-fold at all.

PR #457, opened 2018-03-09
Fix org files loading with errors because of not using evil mode.

UndeadKernel:
When opening org files, errors would prevent org mode to load. This would happen if the "evil" module was not chosen by the user in their "init.el". As the protector of all that is **holy** (un-evil), I'm fixing this issue with this PR.


hlissner:
I've pushed a fix for this that incorporates pieces of your PR, and ties lang/org's evil config to the evil-org package, so I'll consider this resolved.

PR #461, opened 2018-03-13
 Feature: +style feature that provide better org HTML export default

fuxialexander:
This PR added a new flag `+style` to `lang/org` module. Once enable, the org HTML export will have better appearance and functionality. The stylesheet and included script and be easily replaced using two variable `+org-html-export-style-dir` and `+org-html-export-style-alist`.

Functionality including:
- DataTables integration (searchable and sortable tables)
- Bootstrap 4 based table of contents and collapsible code blocks / tables
- Optionally embed image as base64 (`+org-html-embed-image`), i.e. self-contained HTML; ox-pandoc is also supported

- Demo 1
image

- Demo 2
image

- Demo 3
image

### DIFF
--- a/modules/lang/org/+export.el
+++ b/modules/lang/org/+export.el
@@ -7,10 +7,24 @@
 ;; `default-directory'. This is because all my org files are usually in one
 ;; place, and I want to be able to refer back to old exports if needed.
 
+
+(defvar +org-html-embed-image t
+  "whether image is embeded as base64 in html")
+
+(def-package! htmlize
+  :commands (htmlize-buffer
+             htmlize-file
+             htmlize-many-files
+             htmlize-many-files-dired)
+  :config
+  (setq-default htmlize-pre-style t))
+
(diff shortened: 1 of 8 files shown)


fuxialexander:
I might need to add some extra documentation / snippets of how to use/enable the DataTable and ToC in org files. Please do not merge it yet.

hlissner:
Sorry for this tremendously late response. lang/org has diverged a lot since this was first submitted. I'm still interested in the idea, but I'd like to come up with a generalized solution; preferably controlled either by something like a `+org-html-theme` variable (and we could store the js/css files in lang/org/themes/NAME/{css,js}/) or prompt the user for a theme when exporting html.

In any case, this PR would have to be redesigned, so I'll close this. I'll eventually get around to doing this myself, but it will take me a long time to get there. If you'd like to take a stab at it I'd be interested in working on it with you (and I promise to be more proactive about responding to it sooner!)

PR #467, opened 2018-03-19
Fix evil-collection support

dbordak:
Just tried out `(evil +everywhere)`, it doesn't load because it depends upon `company-tng`. This fixes it, dunno if you have any more specific requirements for PRs tho, sorry if this isn't good enough.

### DIFF
--- a/modules/feature/evil/config.el
+++ b/modules/feature/evil/config.el
@@ -23,10 +23,11 @@
 
 (def-package! evil-collection
   :when (featurep! +everywhere)
-  :after evil
+  :after (company evil)
   :preface
   (setq evil-want-integration nil) ; must be set before evil is loaded
   :config
+  (require 'company-tng)
   (evil-collection-init)
   ;; don't interfere with leader key
   (map! :after compile :map compilation-mode-map doom-leader-key nil)


fuxialexander:
`company-tng` is a front-end using overlay. Since it's a UX thing it would better to be a setting. The variable is only used in the last line in `evil-collection-company-setup`:
```lisp
(defun evil-collection-company-setup ()
  "Set up `evil' bindings for `company'."
  (define-key company-active-map (kbd "C-n") 'company-select-next-or-abort)
  (define-key company-active-map (kbd "C-p") 'company-select-previous-or-abort)
  (define-key company-active-map (kbd "C-j") 'company-select-next-or-abort)
  (define-key company-active-map (kbd "C-k") 'company-select-previous-or-abort)
  (define-key company-active-map (kbd "M-j") 'company-select-next)
  (define-key company-active-map (kbd "M-k") 'company-select-previous)

  (when evil-want-C-u-scroll
    (define-key company-active-map (kbd "C-u") 'company-previous-page))

  (when evil-want-C-d-scroll
    (define-key company-active-map (kbd "C-d") 'company-next-page))

  (define-key company-search-map (kbd "C-j") 'company-select-next-or-abort)
  (define-key company-search-map (kbd "C-k") 'company-select-previous-or-abort)
  (define-key company-search-map (kbd "M-j") 'company-select-next)
  (define-key company-search-map (kbd "M-k") 'company-select-previous)

  ;; Sets up YCMD like behavior.
  (when evil-collection-company-use-tng (company-tng-configure-default)))
```
Thus probably it's better to set the variable to `nil` first, and add a `+tng` flag to `company` module, where we call
```lisp
(after! company 
    (let ((evil-collection-company-use-tng t)) 
          (evil-collection-company-setup))
```

hlissner:
There's nothing wrong with the PR, and this is a good start, but I want to avoid aggressively loading company with evil.

I've set `evil-collection-company-use-tng` to nil and am leaving it to private modules to enable tng; it only takes the following two lines:

```elisp
(require 'company-tng)
(company-tng-configure-default)
```

I don't think that warrants its own module flag, but I'll consider it (`+auto` is an exception because it is an essential characteristic of code completion -- for most users).

In any case, I'll consider this issue resolved. Thanks for your help!

PR #473, opened 2018-03-23
Fix magic require of clojure.string

jcf:
Quick change that I've been running locally as the namespace is [`clojure.string`](https://clojure.github.io/clojure/clojure.string-api.html) as opposed to `clojure.str`.

My full list right now looks like this:

``` emacs-lisp
(nconc cljr-magic-require-namespaces
       '(("async" . "clojure.core.async")
         ("byte-streams" . "byte-streams")
         ("cache" . "clojure.core.cache")
         ("cli" . "clojure.tools.cli")
         ("codec" . "ring.util.codec")
         ("component" . "com.stuartsierra.component")
         ("d" . "datomic.api")
         ("data" . "clojure.data")
         ("deferred" . "manifold.deferred")
         ("gen" . "clojure.test.check.generators")
         ("jmx" . "clojure.java.jmx")
         ("json" . "cheshire.core")
         ("medley" . "medley.core")
         ("prop" . "clojure.test.check.properties")
         ("r" . "clojure.core.reducers")
         ("random" . "crypto.random")
         ("s" . "clojure.spec.alpha")
         ("sgen" . "clojure.spec.gen.alpha")))
```

I think some of the above is entirely subjective however and wonder where you draw the line on good defaults for all/what belongs in a private configuration…

Thanks for Doom BTW. It's super quick and has been very nice to work/play with.

### DIFF
--- a/modules/lang/clojure/config.el
+++ b/modules/lang/clojure/config.el
@@ -15,7 +15,7 @@
   (nconc cljr-magic-require-namespaces
          '(("re-frame" . "re-frame.core")
            ("reagent"  . "reagent.core")
-           ("str"      . "clojure.str"))))
+           ("str"      . "clojure.string"))))
 
 
 (def-package! cider


jcf:
I also have various indent rules for commonly-used libraries like om.next, core.match, Riemann, clojure.spec, Arachne, core.logic… happy to share those if there's interest in sharing such a thing in Doom (maybe tooling like cljfmt and/or zprint are the place for this kind of data).

hlissner:
Unfortunately, because I'm not a clojure user, it's difficult for me to draw that line too! I'm unsure how `cljr-magic-require-namespaces` is expected to work, for that matter. 

I'm all for sensible defaults, especially if you tell me those namespaces are used frequently, but ultimately, I have to let you be the judge of that. Do you think these are sensible defaults for Clojure users? If so, I'll happily merge this. If not, then it belongs in your private module.

jcf:
The change in this PR is definitely correct. Even if some people choose to alias `clojure.string` as something other than `str`, the namespace that alias resolves too is currently wrong in Doom.

In terms of my list of aliases, some are subjective enough that some Clojure users would need to remove them. Removing all third-party libs so that the magic requires only include `clojure.*` namespaces would be a sensible filter, but even then…

Although the results of the recent [State of Clojure](http://blog.cognitect.com/blog/2017/1/31/clojure-2018-results) survey show a large uptake of Clojure 1.9 where clojure.spec was introduced, some people will still be using [Schema](https://github.com/plumatic/schema) and commonly you'll see that take the `s` alias over `clojure.spec.alpha` (which of course will become `clojure.spec` at some point requiring a change to these magic requires).

I'd argue that the following list is highly unlikely to need anything removed in the foreseeable future:

``` clj
(nconc cljr-magic-require-namespaces
       '(("cache" . "clojure.core.cache")
         ("cli" . "clojure.tools.cli")
         ("data" . "clojure.data")
         ("gen" . "clojure.test.check.generators")
         ("prop" . "clojure.test.check.properties")
         ("r" . "clojure.core.reducers")))
```

If you could merge this PR as is that'd be appreciated. I'm happy to create another with a more exhaustive list of magic requires if you want.

hlissner:
Thank you for the detailed reply! I agree with filtering 3rd party libs, I think we should go with that.

That said, if you'd like to introduce sizeable changes to the `:lang clojure` module (that may be specific to certain toolchains, libraries or some such), consider a module flag.

Take a look at [these lines in the :lang haskell module](https://github.com/hlissner/doom-emacs/blob/develop/modules/lang/haskell/config.el#L3-L4). There are two development environments available for Haskell in Emacs, Intero (for stack) and Dante (for cabal). One can enable one or the other by supplying the `+intero` and `+dante` flags to the module in your `doom!` block. e.g.

```elisp
(doom! :lang (haskell +dante))
```

You mentioned indentation rules for common libraries; those, for instance, could be guarded by a `+indent` flag. My only rule for module flags is they they introduce popular and/or large-scale configuration that you can't otherwise pull off with a couple lines in your private module.

In any case, thanks again for your help!

jcf:
Thank you for the tip regarding module flags! I'll check them out.

PR #474, opened 2018-03-23
Add edn to Clojure mode extensions

jcf:
When opening an EDN file in Emacs I noticed Clojure code wasn't automatically applied. I think this is because the mode is deferred and Doom doesn't tell Emacs that `.edn` files belong to Clojure.

Fingers crossed I've fixed this in an idiomatic way!

### DIFF
--- a/modules/lang/clojure/config.el
+++ b/modules/lang/clojure/config.el
@@ -3,6 +3,7 @@
 (def-package! clojure-mode
   :mode "\\.clj$"
   :mode "\\.edn$"
+  :mode "\\(?:build\\|profile\\)\\.boot$"
   :mode ("\\.cljs$" . clojurescript-mode)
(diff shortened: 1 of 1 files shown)


hlissner:
This looks fine. Could you add one for Boot files too? e.g. `:mode "\\(?:build\\|profile\\)\\.boot$"`

jcf:
@hlissner pattern added in https://github.com/hlissner/doom-emacs/pull/474/commits/025ef9a7f8f95ad77529cc09c59628cf3f6e72a5.

jcf:
Happy to rebase/reword commits as you see fit. Not sure what the commit style is for Doom.

hlissner:
Excellent, thanks for this!

As for commit style, the only rules are: a) much be clear and b) if it's a breaking change, prefix it with a :boom: emoji. That's all, so these fit the bill perfectly!

PR #483, opened 2018-03-26
Add REPL definition for elixir-mode.

adamkruszewski:
Just a minor setting for defining REPL for elixir-mode.


hlissner:
Thanks!

PR #488, opened 2018-03-28
Update password-store +auth to only use auth-source-pass

orther:
Now that auth-password-store has been renamed to auth-source-pass and it matches
the name and functionality of the feature included in emacs 26 I've removed the conditional loading and use of different packages. 

I was getting an error on emacs load because `auth-source-store` wasn't loaded (and couldn't because the name is incorrect) but was defined with `def-package!`. There were several incorrect variations of the correct package names (e.g. `auth-store-pass`, `auth-source-store`) causing problems. Now that `auth-source-pass` works for both emacs 25 & 26 life should be better.

I updated auth-source-pass to only be installed when the +auth flag is set. I also moved the condition into the :when of the !def-package. If there is any reason to leave it as it was I'd appreciate any education you're willing to provide.

NOTE: I accidentally used incorrect variations of auth-password-store and auth-source-pass more than once working on this so I'm damn happy the package was renamed and updated to match emacs 26. I have to guess a lot of people are 😄 (see:  https://github.com/DamienCassou/auth-password-store/commit/e8d8733b1af67e4ea088d1ed015c554171feecb9#r23381057)

### DIFF
--- a/modules/tools/password-store/autoload.el
+++ b/modules/tools/password-store/autoload.el
@@ -12,9 +12,7 @@
 
 ;;;###autoload
 (defalias '+pass--get-entry
-  (if (featurep 'auth-store-pass)
-      #'auth-source-pass-parse-entry
-    #'auth-pass-parse-entry))
+  #'auth-source-pass-parse-entry)
 
 ;;;###autoload
 (defun +pass-get-field (entry fields)
--- a/modules/tools/password-store/config.el
+++ b/modules/tools/password-store/config.el
@@ -36,8 +36,6 @@
 
 
 ;; Is built into Emacs 26+
-(when (featurep! +auth)
-  (if (require 'auth-store-pass nil t)
-      (auth-source-pass-enable)
-    (def-package! auth-source-store
-      :config (auth-source-pass-enable))))
+(def-package! auth-source-pass
+  :when (featurep! +auth)
+  :config (auth-source-pass-enable))
--- a/modules/tools/password-store/packages.el
+++ b/modules/tools/password-store/packages.el
@@ -3,7 +3,9 @@
 
 (package! pass)
 (package! password-store)
-(package! auth-source-pass)
+
+(when (featurep! +auth)
+  (package! auth-source-pass))
 
 (when (featurep! :completion helm)
   (package! helm-pass))


hlissner:
I agree with the name change too; I was a little confused by it and didn't have enough time to look into it. This PR was exactly what Doom needed. Thank you for this!

orther:
I really should have mentioned that I haven't been able to get magithub to work with auth-source-pass yet. Still trying to track that down. I can pull values from password-store through auth-sources but it seems there are some edge cases causing problems.

fuxialexander:
@orther The cause is magithub cannot read the entry from password-store. I don't remember the function name, but basically magithub/ghub use `:host` and search for `api.github.com` in password-store, which will result in no return value even if you do that manually. However if you search use `:user` or `:HOST` that gonna work. In the end I just get around with it by not adding the pass to auth-source.

fuxialexander:
@orther It's here (in ghub.el):
```lisp
(defun ghub--auth-source-get (keys &rest spec)
  (declare (indent 1))
  (let ((plist (car (apply #'auth-source-search :max 1 spec))))
    (mapcar (lambda (k)
              (plist-get plist k))
            keys)))

(defun ghub--token (host username package &optional nocreate forge)
  (let* ((user (ghub--ident username package))
         (token (car (ghub--auth-source-get (list :secret)
                       :host host :user user))))
    (or (if (functionp token) (funcall token) token)
        (progn
          ;; Auth-Source caches the information that there is no
          ;; value, but in our case that is a situation that needs
          ;; fixing so we want to keep trying by invalidating that
          ;; information.  The (:max 1) is needed for Emacs releases
          ;; before 26.1.
          (auth-source-forget (list :max 1 :host host :user
          user))
          (and (not nocreate)
               (if (eq forge 'gitlab)
                   (error
                    (concat "Required Gitlab token does not exist.  See "
                            "https://magit.vc/manual/ghub/Gitlab-Support.html "
                            "for instructions."))
                 (ghub--confirm-create-token host username package)))))))
```

fuxialexander:
@orther Please ignore my previous reply, I've got it work:
You need to rename your magithub gpg file in pass directory to the following format:
`username^magithub@api.github.com.gpg`
And put the API key into it. That's enough.

PR #491, opened 2018-03-28
completion/ivy: fix ag/pt/grep file search args

MaskRay:


hlissner:
Ah, good catch. Thanks!

PR #495, opened 2018-03-29
Feature: lang/web: add local leader bindings for web-mode

benj02:
Web mode has a bunch of helpful commands under `C-c`, this replicates the same
bindings, but in leader-key style. E.g.: `C-c C-a k` becomes `SPC m a k`


hlissner:
This looks great, thanks!

benj02:
Awesome!

I should add, this is also a soft fix for an issue I run into a lot, where if you paste in text using any of the evil commands in web-mode it shows up unhighlighted. But now there is an easy keybind to rehighlight the buffer `SPC m h`

PR #496, opened 2018-03-29
Add mention of Discord to README

jcf:
Hey @hlissner,

Not sure you want this here, so more of a discussion than request. What do you think about mentioning Discord in the README?


hlissner:
There is already a discord badge in the README (on develop), but I understand if it's not very visible. The new readme in 2.1 has a more visible secondary link however. I'll turn down this PR since it won't be long until it's up.

PR #497, opened 2018-03-29
Add language support for Erlang

bitwalker:
Since Elixir is supported out of the box, it seems unfair to leave out Erlang :). I'm still relatively new to working with elisp, so let me know if I could be doing things differently. I added some customizations which add some additional support for rebar3 and some common Erlang files (e.g. `sys.config`). I think there is probably room here to expand this to maybe provide EDTS/Wrangler or something along those lines, but I wanted to start simple - this module is basically what I built for myself so far.

### DIFF
--- a/modules/lang/erlang/config.el
+++ b/modules/lang/erlang/config.el
@@ -1,0 +1,29 @@
+;;; private/erlang/config.el -*- lexical-binding: t; -*-
+
+(def-package! erlang
+  ;; customizations
+  :mode "\\.erlang$"
+  ;; rebar files
+  :mode "/rebar\\.config\\(?:\\.script\\)?$"
+  ;; erlang configs
+  :mode "/\\(?:app\\|sys\\)\\.config$")
+
+(def-package! flycheck-rebar3
+  :when (featurep! :feature syntax-checker)
+  :after erlang
+  :config
+  (flycheck-rebar3-setup))
+
+;; Completion via Ivy
+(def-package! ivy-erlang-complete
+  :when (featurep! :completion ivy)
+  :hook (erlang-mode . ivy-erlang-complete-init)
+  :config
+  (add-hook! 'erlang-mode-hook
+    (add-hook 'after-save-hook #'ivy-erlang-complete-reparse nil t)))
+
+
+;; Completion via Company
+(def-package! company-erlang
+  :when (featurep! :completion company)
+  :hook (erlang-mode . company-erlang-init))
--- a/modules/lang/erlang/packages.el
+++ b/modules/lang/erlang/packages.el
@@ -1,0 +1,13 @@
+;; -*- no-byte-compile: t; -*-
+;;; private/erlang/packages.el
+
+(package! erlang)
+
+(when (featurep! :feature syntax-checker)
+  (package! flycheck-rebar3))
+
+(when (featurep! :completion ivy)
+  (package! ivy-erlang-complete))
+
+(when (featurep! :completion company)
+  (package! company-erlang))
--- a/init.example.el
+++ b/init.example.el
@@ -93,6 +93,7 @@
        clojure           ; java with a lisp
        csharp            ; unity, .NET, and mono shenanigans
        data              ; config/data formats
+      ;erlang            ; an elegant language for a more civilized age
        elixir            ; erlang done right
        elm               ; care for a cup of TEA?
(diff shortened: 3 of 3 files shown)


hlissner:
This is a good start, but requires some work:

1. The `def-package-hook!` call is unnecessary. erlang is defined no-where else such that it can be reconfigured. Convert it to `def-package!` and move it to config.el (replacing the `(after! erlang ...)` block.
2. Use the `:mode` property in `def-package!` to register those file extensions, e.g.

```elisp
(def-package! erlang
  :mode "\\.erlang$"
  :mode "/rebar\\.config\\(?:\\.script\\)?$"
  :mode "/\\(?:app\\|sys\\)\\.config$"
  :config
  ...)
```

3. The two rebar3 functions should be moved into an `autoload.el` file (and the `init.el` file removed), and renamed to follow Doom's naming convention:
   + `+erlang-rebar3-flymake-get-include-dirs`
   + `+erlang-rebar3-flymake-get-code-path-dirs`
4. The `ivy-erlang-complete` and `company-erlang` packages should be conditionally installed in `packages.el`, e.g.

  ```elisp
    (package! ivy-erlang-complete))
    (package! company-erlang))
  ```
5. In `config.el`, you've defined the company and ivy packages as `erlang-ivy` and `erlang-company`, but these packages do not exist. The real names of these packages are `ivy-erlang-complete` and `company-erlang`, please correct them.
6. Those `:after` blocks are a little excessive, when `:hook` will do. Also, refrain from adding `ivy-erlang-complete-reparse` to the global `after-save-hook` hook, to avoid executing it when saving _any_ buffer:

```elisp
(def-package! ivy-erlang-complete
  :hook (erlang-mode . ivy-erlang-complete-init)
  :config
  (add-hook! 'erlang-mode-hook
    (add-hook 'after-save-hook #'ivy-erlang-complete-reparse nil t)))

(def-package! company-erlang
  :hook (erlang-mode . company-erlang-init))
```

-----

Also, side question: is there any reason why you aren't using the [flycheck-rebar3](https://github.com/joedevivo/flycheck-rebar3) plugin? Doom mainly uses `flycheck` over flymake. Unless that plugin is broken or inferior, I'd like to be consistent. I am not an elixir user, so I'm unable to reliably test this myself.

bitwalker:
@hlissner Thanks for the detailed review! I'm definitely learning as I go here haha

I've made the changes you requested, and switched over to using flycheck rather than flymake (I was just using flymake because that is what erlang-mode has support for out of the box), and things look a lot better now. Thanks for taking the time to walk me through :)

modules/lang/erlang/config.el
```diff
@@ -0,0 +1,30 @@
...
+  ;; rebar files
+  :mode "/rebar\\.config\\(?:\\.script\\)?$"
+  ;; erlang configs
+  :mode "/\\(?:app\\|sys\\)\\.config$")
+
+(def-package! flycheck-rebar3
+  :after erlang
+  :config
+  (progn
+    (require 'flycheck-rebar3)
```
hlissner: No need for this line. `flycheck-rebar3` is already loaded by this point.

Also, please remove the surrounding `(progn ...)`, as it is unnecessary here.

modules/lang/erlang/config.el
```diff
@@ -0,0 +1,30 @@
...
+(def-package! erlang
+  ;; customizations
+  :mode "\\.erlang$"
+  ;; rebar files
+  :mode "/rebar\\.config\\(?:\\.script\\)?$"
+  ;; erlang configs
+  :mode "/\\(?:app\\|sys\\)\\.config$")
+
+(def-package! flycheck-rebar3
+  :after erlang
```
hlissner: Please add a `:when (featurep! :feature syntax-checker)` before this line.

modules/lang/erlang/packages.el
```diff
@@ -0,0 +1,12 @@
+;; -*- no-byte-compile: t; -*-
+;;; private/erlang/packages.el
+
+(package! erlang)
+
+(package! flycheck-rebar3)
```
hlissner: Surround this in a `(when (featurep! :feature syntax-checker) ...)` so it isn't installed without the syntax-checker module enabled.

bitwalker:
I've updated the PR based on your feedback and squashed commits, let me know if anything else comes up!

hlissner:
Nice work, thanks!

Issue #503, opened 2018-04-02
Helm is broken


azy2:
If you enable helm in init.el and then run find file (or if you don't want to disable ivy M-x helm-find-files) you get an error stating: `Invalid function: with-helm-buffer`

hlissner:
I am unable to reproduce this. 

`with-helm-buffer` is [defined in helm-lib.el](https://github.com/emacs-helm/helm/blob/master/helm-lib.el#L1190-L1194), which is loaded by [helm.el](https://github.com/emacs-helm/helm/blob/master/helm.el#L34).

Have you updated helm recently? Are you sure you ran `make` or `make install` after enabling helm?

hlissner:
With 0739e8cd, I've tried to fix this issue. Please update and let me know if it does.

azy2:
Pulling develop I get the following issues. When I open emacs a minibuffer pops up and says: `Error (use-package): rspec-mode/:init: Symbol's value as variable is void: doom-auto-minor-mode-alist` and when I try to find file I get: `Wrong type argument: bufferp, "*helm find files*". I've also commented out all my personal helm configuration to make sure it is not related to that.

PR #507, opened 2018-04-03
Fix helm bug caused by old evil-collection

azy2:
Issue #503 is caused by a bug in evil-collection regarding 3rd party macros (described here: https://github.com/emacs-evil/evil-collection/issues/101). It is fixed in upstream but not on elpa (yet?). This pull request gets evil-collection from github and reverts the change to the helm module which was unnecessary. I suggest this change is reverted as soon as the bug fix in evil-collection makes it's way to elpa. It would also probably be advantageous to only pull from github if helm is enabled and otherwise from elpa but I do not know how to accomplish that.

### DIFF
--- a/modules/completion/helm/config.el
+++ b/modules/completion/helm/config.el
@@ -62,7 +62,7 @@
 
   (defun +helm*hide-minibuffer-maybe ()
     "Hide minibuffer in Helm session if we use the header line as input field."
-    (when (buffer-local-value 'helm-echo-input-in-header-line (helm-buffer-get))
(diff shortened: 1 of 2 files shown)


hlissner:
Considering how volatile the evil-collection package is turning out to be, I'm not too keen on tracking its HEAD. I'm even considering forking it, for Doom's sake. 

In any case, it has been two weeks since the fix was pushed upstream, so it must be reflected in MELPA by now. If so, is this PR still necessary?

PR #511, opened 2018-04-05
fix workspace switch-left cycle.

amosbird:
`+workspace/switch-left` doesn't work when on the leftmost workspace. This patch fixes it.


modules/feature/workspaces/autoload/workspaces.el
```diff
@@ -361,7 +361,7 @@ end of the workspace list."
                  (index (cl-position current-name persps)))
             (when (= perspc 1)
               (user-error "No other workspaces"))
-            (+workspace/switch-to (% (+ index n) perspc))
+            (+workspace/switch-to (% (+ (+ index n) perspc) perspc))
```
bandresen: No need to nest the addition operation here.

PR #517, opened 2018-04-08
Fix "z o".

scturtle:
PS: "d %" never works. Disabling evil-matchit fixes that.


hlissner:
Thanks!

Could you open a separate issue for `d %` and describe how it doesn't work? It seems to be working fine for me.

Issue #519, opened 2018-04-09
`d %` leaves one character.


scturtle:
System: archlinux, emacs 25.3, doom-emacs latest develop branch, latest packages.

`d %` on the first `(` of:
```
(defun foo ()
  )
```
leaves the matching `)`:
```
)
```
Same problem on macos and emacs HEAD. Disabling evil-matchit fixes that.
Probably this is an upstream problem but makes `d %` unusable.

hlissner:
I don't want to disable evil-matchit, beacuse it allows `%` to jump between start/end keywords (like `if` and `end` in ruby, or `function` and `endfunction` in lua) as well as parentheses.

I have a couple ideas for a fix and will address this soon.

hlissner:
Just pushed a fix that should have addressed your issue. Let me know if that isn't the case!

scturtle:
Almost fixed. Only the first time behavior after started emacs is buggy.

PR #526, opened 2018-04-11
Fix issue with `d%` leaving one character behind.

fatihict:
Resolves: #519


hlissner:
`evil-jump-items` does not exist. You must be thinking of `evilmi-jump-items`, which is what we're remapping `evil-jump-item` to.

This accidentally fixes #519 by effectively disabling evil-matchit by disabling its keybind.

PR #527, opened 2018-04-11
Fix: tools/magit: +magit--kill-buffer function call corrected

bandresen:


hlissner:
Ah, a careless mistake. Thanks for fixing it!

PR #528, opened 2018-04-12
Fix: app/email: only add unique contexts

bandresen:
for some reasons the settings are being invoked several times which causes the context to be duplicated in the context switcher.

The bug did not affect functionality.

### DIFF
--- a/modules/app/email/config.el
+++ b/modules/app/email/config.el
@@ -38,7 +38,8 @@ default/fallback account."
                            (string-prefix-p (format "/%s" ,label)
                                             (mu4e-message-field msg :maildir))))
                        :vars ,letvars)))
-         (push context mu4e-contexts)
+         (unless (seq-contains (seq-map #'mu4e-context-name mu4e-contexts) ,label)
+           (push context mu4e-contexts))
          ,(when default-p
             `(setq-default mu4e-context-current context))))))
 


modules/app/email/config.el
```diff
@@ -38,7 +38,8 @@ default/fallback account."
                            (string-prefix-p (format "/%s" ,label)
                                             (mu4e-message-field msg :maildir))))
                        :vars ,letvars)))
-         (push context mu4e-contexts)
+         (unless (seq-contains (seq-map #'mu4e-context-name mu4e-contexts) ,label)
+           (push context mu4e-contexts))
```
jcf: This means you can't update an existing context too, right?

Wouldn't it be better to replace an existing context by name so that reevaluating your config still updates mu4e after Emacs has started up?

I set `mu4e-contexts` by hand like so in my config, and it works well:

``` emacs-lisp


(setq
 mu4e-contexts
 `(,(make-mu4e-context
     :name "personal"
     :enter-func (lambda () (mu4e-message "Switched to Personal."))
     :leave-func (lambda () (mu4e-clear-caches))
     :match-func
     (lambda (msg)
       (when msg
         (string-prefix-p "/personal"
                          (mu4e-message-field msg :maildir))))
     :vars
     '((mu4e-sent-folder       . "/personal/sent")
       (mu4e-drafts-folder     . "/personal/drafts")
       (mu4e-trash-folder      . "/personal/trash")
       (mu4e-refile-folder     . "/personal/archive")
       (smtpmail-smtp-user     . "jame@example.com")
       (user-mail-address      . "james@example.com")
       (mu4e-compose-signature . "James")))
   ,(make-mu4e-context
     :name "work"
     :enter-func (lambda () (mu4e-message "Switched to Work."))
     :leave-func (lambda () (mu4e-clear-caches))
     :match-func
     (lambda (msg)
       (when msg
         (string-prefix-p "/work"
                          (mu4e-message-field msg :maildir))))
     :vars
     '((mu4e-sent-folder       . "/work/sent")
       (mu4e-drafts-folder     . "/work/drafts")
       (mu4e-trash-folder      . "/work/trash")
       (mu4e-refile-folder     . "/work/archive")
       (smtpmail-smtp-user     . "jcf@work.com")
       (user-mail-address      . "james@work.com")
       (mu4e-compose-signature . "James Conroy-Finn")))))
```

It's a little repetitive, which is why I guess this macro was added in the first place, but it works both when I start Emacs and if/when I need to change my mu4e configuration.

jcf: > As far as I can see you couldn't do that before either, at least not in a way that you were guaranteed to get the updated context when you selected it. Because it just appended the contexts and duplicate contexts shared the same keybinding.

Given that `push` prepends an item to the list…

> This macro creates a new list whose car is element and whose cdr is the list specified by listname, and saves that list in listname.

…and given that mu4e picks the first matching context by name, I think the most recent context would be picked meaning changes would have taken effect before this change - essentially the old data gets pushed further down the stack.

``` emacs-lisp
(defun mu4e-context-switch (&optional force name)
  "Switch context to a context with NAME which is part of
`mu4e-contexts'; if NAME is nil, query user.

If the new context is the same and the current context, only
switch (run associated functions) when prefix argument FORCE is
non-nil."
  (interactive "P")
  (unless mu4e-contexts
    (mu4e-error "No contexts defined"))
  (let* ((names (map 'list (lambda (context)
			     (cons (mu4e-context-name context) context))
		  mu4e-contexts))
	  (context
	    (if name
	      (cdr-safe (assoc name names))
	      (mu4e~context-ask-user "Switch to context: "))))
    (unless context (mu4e-error "No such context"))
    ;; if new context is same as old one one switch with FORCE is set.
    (when (or force (not (eq context (mu4e-context-current))))
      (when (and (mu4e-context-current)
	      (mu4e-context-leave-func mu4e~context-current))
	(funcall (mu4e-context-leave-func mu4e~context-current)))
      ;; enter the new context
      (when (mu4e-context-enter-func context)
	(funcall (mu4e-context-enter-func context)))
      (when (mu4e-context-vars context)
	(mapc #'(lambda (cell)
		  (set (car cell) (cdr cell)))
	  (mu4e-context-vars context)))
      (setq mu4e~context-current context)
      (unless (eq mu4e-split-view 'single-window)
        (mu4e~main-view-real nil nil))
      (mu4e-message "Switched context to %s" (mu4e-context-name context)))
    context))
```

> Also that use case of updating the context while running Emacs is not one that I've thought about because I haven't touched my `mu4e-contexts` in years.

Anyone who hasn't already setup mu4e is more than likely going to need to tweak a setting or two, and right now they'd have to either restart Emacs or `(setq mu4e-contexts nil)` for the changes to take effect.

Maybe this macro should just be removed and the README can recommend the same approach mu4e's docs does?

hlissner:
I pushed a change to the `:email` setting to a) replace contexts with the same name and b) dedup `mu4e-contexts`.

This should resolve this issue. Let me know if that isn't the case!

PR #529, opened 2018-04-13
Henrik doesn't maintain my Docker images

jcf:


jcf:
Your name's in loads of the file templates. Can we replace all hardcoded Henrik's with `user-full-name`?

bandresen:
I noticed that myself before and tried my hand at fixing it.
I didn't further investigate, but the snippet that should work, sets the `point` wrong and messes up the snippet. If you have a moment you could try that with a minimal yasnippet configuration vs the full doom.

```
FROM ${1:phusion/baseimage:latest}
MAINTAINER ${2:`(concat user-full-name " <" user-mail-address ">")`}

$0
```

PR #530, opened 2018-04-15
Don't use HTTPS when INSECURE variable is set

CeleritasCelery:
When working with a proxy or other insecure environment, the environment variable `INSECURE` will prevent TLS certificate check, but Doom still tries to use HTTPS repo's for packages. This prevents users without HTTPS from installing Doom. We can fix this by checking for this variable and using the the HTTP versions for `package-archives`.


hlissner:
Good idea, thanks!

PR #531, opened 2018-04-16
Update config.el

edmundmiller:
Fix doom leader not working in pdf mode. 

Replaces SPC default that moves the document down a page.

### DIFF
--- a/core/core-packages.el
+++ b/core/core-packages.el
@@ -96,9 +96,13 @@ missing) and shouldn't be deleted.")
       package-user-dir (expand-file-name "elpa" doom-packages-dir)
       package-enable-at-startup nil
       package-archives
-      '(("gnu"   . "https://elpa.gnu.org/packages/")
-        ("melpa" . "https://melpa.org/packages/")
-        ("org"   . "https://orgmode.org/elpa/"))
+      (if (getenv "INSECURE")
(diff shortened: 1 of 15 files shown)


UndeadKernel:
By removing the "q" binding, wouldn't that disable quitting a PDF for non-evil users?

Emiller88:
It keeps the q binding(I think it's more important than recording a macro in a PDF)

UndeadKernel:
The "q" binding seems to be replaced rather than kept.

Emiller88:
@UndeadKernel my mistake I'll fix that!

UndeadKernel:
A small detail still remains :)
Because you added the ":n" before the "q", now the binding will only work for evil users. For non-evil users, we wish to keep the original rule without the ":n".

Emiller88:
So would it be this?
```
(map! (:map (pdf-view-mode-map)
         "q" #'kill-this-buffer
         doom-leader-key nil))
```

hlissner:
@Emiller88 That should do it. Though, no need for the parentheses in this case:

```elisp
(map! :map pdf-view-mode-map
      "q" #'kill-this-buffer
      doom-leader-key nil)
```

Emiller88:
@hlissner Fixed it. Thanks, @UndeadKernel!

PR #535, opened 2018-04-18
EIN module updates & fixes

fuxialexander:
- Added a hydra inspired by spacemacs.
- Fix popup rules, company config, and default jupyter server setting


UndeadKernel:
Changes look great. I approve.

hlissner:
Thanks again!

hlissner:
Looks like this hydra is now throwing a `wrong-type-argument sequencep t` error (reported in #2057). This seems to be an issue with its definition, but I don't know hydra's specification well enough to know what the problem is. Any suggestions?

PR #536, opened 2018-04-18
enable which-key's evil support

fread2281:
Note: which-key says `which-key-show-operator-state-maps` is hacky, but imo it should be enabled anyways

### DIFF
--- a/core/core-keybinds.el
+++ b/core/core-keybinds.el
@@ -52,7 +52,9 @@ If any hook returns non-nil, all hooks after it are ignored.")
         which-key-add-column-padding 1
         which-key-max-display-columns nil
         which-key-min-display-lines 5
-        which-key-side-window-slot -10)
+        which-key-side-window-slot -10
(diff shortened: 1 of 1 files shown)


UndeadKernel:
Would this change affect non-evil users negatively?
It seems that the proposed changes would disable `which-keys` for non-evil users (them being the good guys, of course).

hlissner:
I originally disabled this because I ran into issues, though I don't remember what they were! I'll accept the PR and keep an eye out.

hlissner:
Looks like #542 is caused by this. I'm also noticing a difficult to reproduce scenario where evil's operator state is disabled entirely in some buffers; an uncaught error (even a `user-error`) thrown from the operator state can reproduce this. These issues go away with which-key evil integration disabled.

Sorry, but I think I will be reverting this PR for the time being.

fread2281:
I noticed that `dd` with the second press immediately works, but waiting for which-key popup in between doesn't

PR #537, opened 2018-04-18
[*] Removes tern in favor of tide, adds local node modules to path

jamestthompson3:
Tide provides much better support for code intel and completion and it is actively maintained unlike tern. Tide also does not require users to install outside dependencies such as the tern javascript package in order for it to work. Adding the local node_modules to the emacs exec path allows for users to then use the project versions of flow, prettier, jest, eslint, etc.


hlissner:
I'd left this on the backburner for a while. With a few minor edits I've merged it into develop. Thanks for your help!

PR #540, opened 2018-04-19
[*] Extends Tide support to RJSX mode, cleans up unused functions

jamestthompson3:
This adds Tide support to RJSX mode, cleans up unused function in RJSX mode and instead uses recently added functions.


hlissner:
Ah, just saw you merge as I was pushing! The mentioned corrections have been made and then some, so I'll considered this resolved.

PR #545, opened 2018-04-22
Adds better documentation to JS mode

jamestthompson3:
This adds a line in the README letting users know that they must add a `jsconfig.json` file to their project root in order to take advantage of tide features


hlissner:
Sorry for the delay. I have to turn this down only because `:lang javascript`'s documentation in 2.1 has been rewritten. I've included your suggestions in the rewrite, however, so this PR has still been helpful, thank you!

PR #547, opened 2018-04-24
Enable rainbow mode in LaTeX buffers.

UndeadKernel:
When a LaTeX document is opened, after all styles are applied, enable rainbow mode.


hlissner:
Is there a reason you're not using `(add-hook 'TeX-update-style-hook #'rainbow-delimiters-mode)`? Using `add-hook!` as you are will unnecessarily add a closure to the hook.

UndeadKernel:
I was not aware of the closure thing.
I was trying to do it more DOOM-like by using `add-hook!` :grinning: 
Anyhow, switched to normal (boring, but closure-less) `add-hook`.

PR #550, opened 2018-04-24
Add missing completion flags.

jwintz:


init.example.el
```diff
@@ -20,7 +20,9 @@
        workspaces        ; tab emulation, persistence & separate workspaces
 
        :completion
-       company           ; the ultimate code completion backend
+       (company          ; the ultimate code completion backend
+        +auto            ; enabble completion by default
```
rphillips: typo here: enabble

init.example.el
```diff
@@ -20,7 +20,9 @@
...
+        +childframe)     ; use childframes whenever possible
```
rphillips: childframes only work in emacs 26, right?

init.example.el
jwintz: Fixed. Thanks.

init.example.el
jwintz: Yes. Anyways, should be in comment?

rphillips: wouldn't hurt to put it in a comment and might cut down on any issues

jwintz: Let's do that then.

jwintz: Done.

PR #561, opened 2018-05-04
fix(lang/javascript): add flycheck-mode to typescript-mode hooks

zenlor:
Add flycheck-mode to typescript. The hook has been probably lost in the last refactor of the javascript module.


hlissner:
Ah, good catch. This was certainly an oversight from the last refactor. Thanks for this!

PR #562, opened 2018-05-05
Add move-text

Emiller88:
Add module for move-text.

Does not support evil visual. Plans are to fix at the source for evil support rather than make it doom specific. 

Adds:
-  <kbd>Meta</kbd>-<kbd>up</kbd> `move-text-up` (line or active region)
-  <kbd>Meta</kbd>-<kbd>down</kbd> `move-text-down` (line or active region)
If `evil`:
-  <kbd>Meta</kbd>-<kbd>k</kbd> `move-text-up` (line)
-  <kbd>Meta</kbd>-<kbd>j</kbd> `move-text-down` (line)

For lack of evil support see this issue:
https://github.com/emacsfodder/move-text/issues/11


modules/config/default/+bindings.el
```diff
@@ -585,6 +585,12 @@
         "C-b"    #'backward-word
         "C-f"    #'forward-word)
 
+      ;; move-text
+      (:after move-text
+        :map move-text-map)
```
MarkRedeman: I think the closing parentheses should be after `:n "M-j" #'move-text-down`

jwintz:
+1 for merging.

Emiller88:
@bluesd  I didn't go for drag-stuff because it was another minor mode and I didn't know how to implement that properly with doom. 

Try adding your solution to modules/config/default/+bindings.el where I added the others after move-text.

nealot:
Can't this functionality be replicated using Evil-Exchange? Which is, in my opinion, a superior utility since it can even use text objects.

Emiller88:
@nealot  I think you're right that you can replace it with Evil-Exchange. I was going for the meta-j feature similar to org-mode with ctrl-shift-j to move lines up and down quickly and visually. But I'm definitely going to start using evil-exchange!

Dragnucs:
Great then, how can I use it?

On May 12, 2018 6:20:23 PM GMT+01:00, Edmund Miller <notifications@github.com> wrote:
>@Dragnucs it actually works better if you don't use evil because moving
>the region works.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/hlissner/doom-emacs/pull/562#issuecomment-388570005

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Emiller88:
@Dragnucs  I added a README https://github.com/Emiller88/doom-emacs/blob/develop/modules/tools/move-text/README.org

hlissner:
Sorry for taking so long to get to this. This PR has become a bit messy. In its current state I can't accept it. Please either rebase until only the pertinent commits remain, or recreate it. Alternatively, wait until the 2.1 release and I will incorporate a move-text module with evil support.

Emiller88:
@hlissner I agree I'll close it.

Emiller88:
@hlissner I agree I'll close it.

PR #577, opened 2018-05-14
Update imenu/config.el

ppwwyyxx:
Add an autoload command


hlissner:
Ah, good catch, thanks!

PR #579, opened 2018-05-15
Add [slim-mode] to :lang web +html

benreyn:
[slim-mode]: https://github.com/slim-template/emacs-slim

Adds support for [slim](https://github.com/slim-template/slim) syntax hightlighting


hlissner:
Thanks for this!

PR #582, opened 2018-05-15
Add: ace-link-mu4e autoload

liujoey:
I uses mu4e a lot, `ace-link-mu4e` is a must have for viewing emails.


hlissner:
That's understandable. Thanks!

PR #583, opened 2018-05-15
Fix typo causing startup error in helm module

benreyn:
It looks like an extra letter snuck into a `nil` in 67dab98. This is causing a
startup error on develop.


hlissner:
Ack! Sorry about that. Thanks!

PR #585, opened 2018-05-15
Add rspec-mode to ruby-mode hook

benreyn:
Turns out, rspec mode wasnt being loaded because there wasnt anything in the
def-package! call that would initialize it. Its pretty common to add rspec mode
to the ruby-mode hook, and use-package makes this a simple one-liner in the
definition.


benreyn:
Thank @hlissner looks like issue with `associate!` was fixed in develop. This wont be needed anymore.

PR #591, opened 2018-05-16
Allow for evil-org todo feature set

igor-kupczynski:
If enabled adds the following keys to the normal mode in org:

-  `t`  :: `org-todo`
-  `T`  :: `org-insert-todo-heading`
- `M-t` :: `org-insert-todo-subheading`

This is useful if you want to quickly toggle TODO --> DONE. It is just `t d`
instead of `C-x C-t d`.

See https://github.com/Somelauw/evil-org-mode/blob/master/doc/keythemes.org#todo

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -90,6 +90,7 @@
        (org              ; organize your plain life in plain text
         +attach          ; custom attachment system
         +babel           ; running code in org
+       ;+evil-org-todo   ; `t d' and done
         +capture         ; org-capture in and outside of Emacs
         +export          ; Exporting org to whatever you want
         +present         ; Emacs for presentations
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -29,7 +29,10 @@
   :hook (org-load . evil-org-set-key-theme)
   :init
   (setq evil-org-key-theme '(navigation insert textobjects))
-  (add-hook 'org-load-hook #'+org|setup-evil))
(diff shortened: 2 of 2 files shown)


AloisJanicek:
If it was up to me, I would say no to `M-t` since it is a global shortcut for creating new work space.   
I would not mind `t` and `T`, since I don't use `evil-snipe`, but I am pretty used to my `SPC m t` for setting todo's in org and I can live without inserting todo headings directly, which I assume is a purpose of those functions judging by their name.

igor-kupczynski:
This is guarded on an option though `(org +evil-org-todo)` and the three keys are bundled together by the evil-org https://github.com/Somelauw/evil-org-mode/blob/master/doc/keythemes.org#todo

We cloud of course just add one key to the map, but I think it is better to offer folks the way to opt-in to that particular key theme from evil-org.

What do you think @AloisJanicek ?

hlissner:
Ah, I almost forgot:

Doom binds `M-RET` to some special behavior on TODOs: they do the same as `T` in your suggestion by adding a todo _below_ the current one. `S-M-RET` will add a todo _above_.

Pressing `TAB` or `S-TAB` in any org header will promote/demote (indent/dedent) that header, which is a good substitute for `M-t`.

Hope that helps!

igor-kupczynski:
Thanks for the detailed description :) I totally buy the arguments. The option to customize `evil-org-key-theme` is handy.

PR #594, opened 2018-05-18
Move text

Emiller88:
Reopened from #562 and cleaned up because I opened it on a personal developement branch, rather than a feature specific. Lesson learned!

Add module for move-text.

Does not support evil visual. Plans are to fix at the source for evil support rather than make it doom specific. 

Adds:
-  <kbd>Meta</kbd>-<kbd>up</kbd> `move-text-up` (line or active region)
-  <kbd>Meta</kbd>-<kbd>down</kbd> `move-text-down` (line or active region)

If `evil`:
-  <kbd>Meta</kbd>-<kbd>k</kbd> `move-text-up` (line)
-  <kbd>Meta</kbd>-<kbd>j</kbd> `move-text-down` (line)

For lack of evil support see this issue:
https://github.com/emacsfodder/move-text/issues/11


hlissner:
Sorry for the tremendously late response. I was hoping that evil+visual mode support would be implemented upstream, but more than a year later it hasn't happened and it looks like the package hasn't been updated in some time, so I will turn down this PR. I do want this functionality in Doom, however, so I'll revisit this sometime later. Thanks for bringing it to my attention!

PR #597, opened 2018-05-18
delete unused find-in-snippets menu item

chengyi:
find in snippets seems to be replaced by 'browse snippets' or  it lack of corresponding function.


hlissner:
I've restored the command, so this PR isn't necessary anymore. Thanks for bringing this to my attention!

PR #599, opened 2018-05-18
Solidity mode

Emiller88:
Moved my personal `solidity` setup to a module in case anyone else wants to make a blockchain calculator.

In the future, I'll add the snippets that I set up for it.

See the README for info about the module.


hlissner:
Excellent, thank you for this PR! I made a few minor changes, the biggest one being: removed `solidity-solc-path` and `solidity-solium-path` because it's preferrable that we pick them up from the user's PATH (useful for users running node/npm installed with the system package manager).

So long as you have `~/.node_modules/.bin` in your PATH, this shouldn't affect you.

Emiller88:
Awesome thanks!

PR #611, opened 2018-05-25
Support handlebars templates

jeffjewiss:
Thanks for Doom, I'm enjoying using it so far.

I work on a lot of Ember apps, which use Handlebars templates as `.hbs` files. I think this is all that's needed to use the correct mode.

Thanks!


hlissner:
Yes, that should be all that's needed. Thanks!

PR #612, opened 2018-05-26
Small fixes for go-mode

seanpile:
- Fixed typo for gocode command in `doctor.el`
- Only enable flycheck mode if the syntax-checker feature is enabled


hlissner:
Thank for your help!

PR #619, opened 2018-05-28
Fix typo in OCaml module.

secondwtq:
Just a typo in config (although it seems working fine on my machine).


hlissner:

PR #620, opened 2018-05-28
lang/org: +ipython

fuxialexander:
This PR adds ob-ipython and related config to doom-emacs as `+ipython` in `lang/org` module.

Main features including:
1. Support for remote kernels
2. Integrate `python-shell-*` and (if using) `lpy-mode` so that you can send code from `org-edit-src-code` window to `REPL`
3. (Optionally through `+right-popup`) popup editing window / REPL at right hand side.


hlissner:
This looks great. I look forward to trying it out. Thanks again for your help and hard work!

PR #621, opened 2018-05-28
Disable lui-mode line wrapping regardless of :feature spellcheck

crshd:
I don't see how those are related. Or am I missing something?


hlissner:
I don't recall why they were grouped together, but you're right, it seems odd that they are.

Would you kindly rebase 5703435 onto `develop`? A few orphaned commits were caught in your PR during a push --force (sorry!).

crshd:
Re-submitted as #651.

PR #625, opened 2018-05-30
add :ui/vinegar support

rphillips:
Adds a vinegar mode to doom-emacs. Pressing `-` pops up a minimal dired in the current buffer.

### DIFF
--- a/modules/ui/vinegar/config.el
+++ b/modules/ui/vinegar/config.el
@@ -1,0 +1,14 @@
+;;; ui/vinegar/config.el -*- lexical-binding: t; -*-
+
+(defun +vinegar|dired-setup ()
+  "Setup custom dired settings for vinegar"
+  (setq dired-omit-verbose nil)
+  (setq dired-hide-details-hide-symlink-targets nil)
+  (make-local-variable 'dired-hide-symlink-targets)
+  (dired-hide-details-mode t))
+
+(after! dired
+  :config
+  (add-hook 'dired-mode-hook '+vinegar|dired-setup))
+
+(map! :n "-" 'dired-jump)
--- a/init.example.el
+++ b/init.example.el
@@ -41,6 +41,7 @@
       ;tabbar            ; FIXME an (incomplete) tab bar for Emacs
       ;unicode           ; extended unicode support for various languages
        vi-tilde-fringe   ; fringe tildes to mark beyond EOB
+      ;vinegar           ; the salad dressing of file browsers
        window-select     ; visually switch windows
 
        :emacs
--- a/modules/emacs/dired/config.el
+++ b/modules/emacs/dired/config.el
@@ -2,6 +2,7 @@
 
 (def-package! dired
   :defer t
+  :commands (dired-jump)
   :init
   (setq ;; Always copy/delete recursively
         dired-recursive-copies  'always


modules/ui/vinegar/config.el
```diff
@@ -0,0 +1,13 @@
...
+  "Setup custom dired settings for vinegar"
+  (setq dired-omit-verbose nil)
+  (setq dired-hide-details-hide-symlink-targets nil)
+  (make-local-variable 'dired-hide-symlink-targets)
+  (dired-hide-details-mode t))
+
+(after! dired
+  :config
+  (add-hook 'dired-mode-hook '+vinegar|dired-setup)
+  (map! :n "-" 'dired-jump))
```
ar1a: wouldn't this map it in every buffer?

rphillips: true... is there a better selector?

ar1a: you could use `:map dired-mode-map` maybe?

rphillips: I want to be in a buffer, and press '-' to open dired just like in [vinegar](https://github.com/tpope/vim-vinegar). `dired-mode-map` will only map within a dired buffer which is not the intent. I think normal mode everywhere mimics vinegar's behavior.

ar1a: - seems too easy to hit for something that doesn't seem like it'd get much use. maybe `SPC- -`?

rphillips: tweaked the PR. Thanks.

rphillips: I tried the functionality out with spc - and I like it better with just -. It also matches the upstream functionality for vinegar https://github.com/tpope/vim-vinegar/blob/master/plugin/vinegar.vim#L113.

ar1a:
Maybe something like this?
https://github.com/hlissner/doom-emacs/commit/ba18b37d5cb2f731b5aa7965fd0515e50b3f5e47

rphillips:
@ar1a thanks for the pointer! tweaked the PR.

hlissner:
I don't object to the concept, but the difference between this and the`:emacs dired` module amounts to:

```elisp
(setq dired-hide-details-hide-symlink-targets nil)
(add-hook 'dired-mode-hook #'dired-hide-details-mode)
(map! :n "-" 'dired-jump)
```

Which I'm unconvinced constitutes its own module.

Two other issues:

+ `dired-omit-verbose` and `dired-hide-symlink-targets` don't appear to exist anywhere in the dired package(s). Perhaps you meant `dired-hide-details-hide-symlink-targets` for the latter?
+ I don't understand the intent behind making `dired-hide-details-hide-symlink-targets` a local variable.

-----

I'll add `(setq dired-hide-details-hide-symlink-targets nil)` in `:tools dired` and add `SPC o -` for dired (a global normal-mode keybind is pretty imposing and overwrites a default evil keybind, so I'd rather leave it up to users to beind themselves). If you'd like different "looks" for dired, I would recommend module flags for `:emacs dired` instead.

However, I'd be totally on board for a ranger/deer-mode + dired-hide-details-mode module (extra points if those dired settings can be isolated to deer-mode).

jcf:
> binding <kbd>-</kbd> globally is pretty imposing and overwrites a default evil keybind universally, so I'd rather leave it up to users to decide this key for themselves

:+1:

rphillips:
sounds good to me. thanks!

willmcclellan:
I really like deer and it's bindings over dired. Just a tip for anyone else who comes across this, make sure to disable the doom dired-k package if you do use deer. Seems like they clash otherwise which causes performance issues

PR #626, opened 2018-05-30
completion/helm: cleanup and set some more sensible defaults

ar1a:
Some sensible defaults I feel like would be widely appreciated, disabling them I just found myself re-enabling them out of frustration.

### DIFF
--- a/modules/completion/helm/config.el
+++ b/modules/completion/helm/config.el
@@ -38,23 +38,21 @@
 (def-package! helm
   :after helm-mode
   :init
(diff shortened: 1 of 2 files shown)


hlissner:
I've gated the fuzzy search engine behind the `+fuzzy` module flag and included `helm-flx` to sort out inaccuracies. I will add the same flag to the ivy module as well, since fuzzy search appears to be a popular request.

Thanks for bringing this to my attention!

PR #627, opened 2018-05-30
Add posframes to helm in emacs 26+

ar1a:
Some room for improvement, but definitely better than windows!


hlissner:
Nice work, thanks!

PR #631, opened 2018-05-31
Fix magithub not loading after magit

ar1a:
I'd have to run a magithub command to get everything to pop up like pull
requests and what not. I think this was an oversight -- just removing
this will still lazyload it, but instead when magit runs!


rphillips:
this fixes it for me as well.

hlissner:
This was certainly an oversight. Thanks!

PR #632, opened 2018-05-31
Add `q` button to magithub issue view

ar1a:


hlissner:
`q` is already bound to `magit-mode-bury-buffer` in
`magithub-issue-view-mode-map` (because that keymap is derived from
magit-mode-map, which binds `q` appropriately). Is this not the case for you?

ar1a:
it wasn't at the time, i'm too tired to check now :/

hlissner:
It does appear to be, so I'll close this for now.

PR #633, opened 2018-05-31
Change SPC-g-C to magithub-clone

ar1a:


hlissner:
Will this work with full git urls?

Also, `magithub-clone` doesn't appear to be autoloaded. Could you add a `:commands (magithub-clone)` to magithub in `:tools magit`?

ar1a:
that's solved with #631

ar1a:
and no, it doesn't support full urls

hlissner:
> that's solved with #631

Not in the case where `magithub-clone` is called before magit has loaded, which the keybind makes a possibility.

hlissner:
> and no, it doesn't support full urls

Hmm, I may hold back on this then. I'll look into the two commands and write a proxy that will accept both full urls and the user/repo format.

ar1a:
pr amended with the autoloaded command

hlissner:
And done. I wrote `+magit/clone` and bound it to `SPC-g-C`.

PR #617, opened 2018-05-26
python +lpy +conda; org +ipython

fuxialexander:
This PR adds a workflow to work with ipython in org-mode, which is particularly useful for data science. `lpy` and `conda` support were added to help python editing experience. (Conda is used more often than pyenv in data analysis using python).

Main features including:
1. Right pane for edit / run python codes
2. Painless remote ipython kernel support
3. Enjoyable lispy-like python editing experience provided by `lpy` from @abo-abo .
4. Everything are integrated together (conda, lpy, python-send-*, ob-ipython)
5. TRAMP support for `conda.el` (which regulates `anaconda-mode` and `lpy` to use correct python interpreter)

Demo:
1. Remote file with `anaconda-mode` setup to display doc for remote module. Also `conda env` is shown in modeline
image

2. `org-edit-src` setup code editing panel and REPL automatically. Also work for remote kernel.
image

3. Run code in REPL use `+python/repl-send-dwim` or `e` in `lpy`.
image

4. Remote-kernel-based completion works (through lpy `C-x C-o`).
image

PR #652, opened 2018-06-04
Add +conda

Emiller88:
Cleaned up from #617. I finally needed it so it got done.

@fuxialexander Would appreciate a README maybe and add anything I missed, please.

### DIFF
--- a/core/core-packages.el
+++ b/core/core-packages.el
@@ -315,7 +315,9 @@ them."
 
         ;; `quelpa-cache'
         (when (or force-p (not (bound-and-true-p quelpa-cache)))
-          (require 'quelpa)
+          ;; ensure un-byte-compiled version of quelpa is loaded
+          (unless (featurep 'quelpa)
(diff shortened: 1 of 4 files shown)


Emiller88:
Unless @fuxialexander says otherwise this should be ready  @hlissner

hlissner:
Nice work @Emiller88 @fuxialexander, this looks great! I'll revisit this once I understand conda better and see if it can be improved.

liujoey:
many issues in this PR.

1. +conda is not loaded in python/config.el. Should add this:
`(if (featurep! +conda)  (load! "+conda"))`

2. `start-pythonic` function doesn't exist, looked around, i believe it belongs to package 'pythonic.el', and the function name has been changed to `pythonic-start-process`

Emiller88:
@liujoey  Feel free to make a PR to my branch or make a new PR to doom with your changes!

PR #658, opened 2018-06-06
tools/magit: Show word-granularity on chosen hunk

ar1a:
This really seems like it should be a default, it's so handy to get better
diffs, sort of like how github does it. if set to 'all then it'll show on all of
them, but I think t is good enough.


UndeadKernel:
I already have this in my private config and would argue that it should indeed be a default. Using the value `t` seems to be a good balance between information and clean hunks.

hlissner:
I like it! Wasn't aware magit had the capability. Thanks!

PR #660, opened 2018-06-06
Fix broken links in README.md to point to real docs

jdemaris:
The links on master's README.md (and develop as well) are broken with regard to the keybindings. This makes it difficult for new users to get their bearings and is a pretty minor patch, so hopefully this can make it into both!


hlissner:
I apologize for taking so long to respond! Sorry to say, I have to turn down this PR, because all documentation in the 2.1 release (coming out any day now) has been _completely_ rewritten from the ground up. Merging this into master would complicate this gigantic merge and would be erased by the changes anyway. Thank you for the PR in any case!

PR #663, opened 2018-06-07
Fix term functions to have optional arguments so they work again

jdemaris:
In current `develop` the keybinding for opening `multi-term` in both current window (`SPC o t`) and in a popup (`SPC o T`) are broken and throwing the error message:

```
Wrong type argument: commandp, +term/open
```

The patch is pretty simple, so I hope it can get in and save other people this trouble! Part of it is also that the keybindings are in default but by default `term` is not enabled in the `init.example.el`.

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -48,7 +48,7 @@
        electric-indent   ; smarter, keyword-based electric-indent
        eshell            ; a consistent, cross-platform shell (WIP)
        imenu             ; an imenu sidebar and searchable code index
-      ;term              ; terminals in Emacs
+       term              ; terminals in Emacs
 
        :tools
        editorconfig      ; let someone else argue about tabs vs spaces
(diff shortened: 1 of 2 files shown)


hlissner:
The root of the problem is that:

a) `:emacs term` is not enabled by default,
b) And many keybindings in `:config default` are not conditionally bound. Meaning there are many keybinds for modules that may or may not be enabled.

I am working on a solution for B. As for A, I'd like to keep term disabled by default, because there is no consensus on which of term or eshell is a more reasonable default. Some users use one or the other, both or neither, and Doom tries to adhere to an "best default or opt-in" philosophy. If no best default exists, then it should be disabled.

I may add conditionals to `:config default` in the meantime or just wait until I roll out my WIP solution for B, but in either case, I must refuse this PR.

Thanks for bringing it to my attention, however!

PR #664, opened 2018-06-07
Fix latex C-g quitting latex after opening the reftex table of contents

UndeadKernel:
After opening the reftex TOC in a latex document, every call to `C-g` would mess up the open windows.
Remove the binding that was causing this problem.


hlissner:
Thanks!

PR #667, opened 2018-06-08
Better looking default indentation for nested latex lists

UndeadKernel:
By default, nested enumerates are indented with too many spaces (relative to how things are indented in DOOM).
This small PR changes the defaults to match DOOM indentation standards.


hlissner:
Thanks again!

PR #669, opened 2018-06-09
Add treemacs module

ar1a:
Seems pretty handy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ network-security.data
 # transient files
 *.cache-*
 *.log
+/.cache/treemacs-persist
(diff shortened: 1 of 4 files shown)


hlissner:
Thanks for this!

hlissner:
Oh, forgot to mention: I removed the `:leader` keybinds because they belong in `:config default`, _however_ I haven't set any default keys for treemacs. I need to decide what they should be, but will do that after the big keybind refactor planned for 2.1.1.

ar1a:
Great, I'll make an issue so we don't forget.

hlissner:
Side note: treemacs still creates `~/.emacs.d/.cache`, which I've filed a PR to fix (Alexander-Miller/treemacs#212).

PR #671, opened 2018-06-09
Fix wakatime breaking emacs if no api-key set

ar1a:


hlissner:
Ah! A silly oversight. Thanks for fixing it!

PR #673, opened 2018-06-09
bindings: magit | add `file-delete` as `SPC g x`

zaiste:
Relatively often I find myself in need to remove some files from a Git repository. I submit for your consideration the following binding (`SPC g x`) to make it conveniently available right under the fingers. ;)


hlissner:
Sounds good to me. There's a _small_ chance I may change it to `SPC g X`, since `magit-file-delete` is a reasonably dangerous command, but we'll see after some testing.

zaiste:
Yes, absolutely. There could be a convention that destructive actions need a `SHIFT` prefix too.

PR #675, opened 2018-06-10
lang/cc: set rtags-install-path to ${doom-etc-dir}/rtags/

ethercflow:
To keep rtags protocol consistency, I think it's better to use `rtags-install` to install or update rtags instead of compile it by ourselves. So put rtags with others tools under doom-etc-dir maybe a good idea.


hlissner:
Fortunately, this still allows folks to install/compile it themselves if they prefer (by simply not using `rtags-install`), so I'm more than happy to accept this. Thanks for your help!

PR #679, opened 2018-06-12
Fix: restore the correct display of current and total pages in PDFs

UndeadKernel:
A refactoring spree introduced a bug by incorrectly making a hook symbol a function. This fixes this.
Let there be page numbers again!


hlissner:
A careless mistake. Thanks for catching it!

PR #682, opened 2018-06-14
Fix parinfer hooks

ar1a:


hlissner:
This was already fixed in 6426da35. Also, I removed `common-lisp-mode` because it's an alias for `lisp-mode`. In any case, thanks for the PR!

PR #683, opened 2018-06-14
haskell/+intero: Fix goto-definition

maikf:
`intero-goto-definition` doesn't get called with `:lookup` set to
`'haskell-mode`. I'm not entirely sure why, but it works when I set it to `'intero-mode`.


hlissner:
There were some infrastructural changes made in the past day or two, which this PR is incompatible with. However, your solution was correct. The lookup handlers were set too late for the first `haskell-mode` buffer to see.

Thanks for this PR and bringing this to my attention!

PR #691, opened 2018-06-17
literate: Allow literate packages [HACKY]

HaoZeke:
Basically I was really enjoying the literate configuration, however the doom-emacs way is to split packages and config, so this way I can keep a literate `packages.org` file with a short description of why I need the package, while describing the keymap and other features in `config.org`.

Please feel free to edit the way this has been implemented, as I have basically 0 lisp knowledge.

### DIFF
--- a/modules/config/literate/init.el
+++ b/modules/config/literate/init.el
@@ -2,6 +2,8 @@
 
 ;;; config/literate/config.el -*- lexical-binding: t; -*-
 
+;;; config/literate/packages.el -*- lexical-binding: t; -*-
+
 (defvar +literate-config-file
   (expand-file-name "config.org" doom-private-dir)
   "The file path of your literate config file.")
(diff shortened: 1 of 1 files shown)


HaoZeke:
More generally I was wondering if it might be possible to allow all literate things. (like literate module additions in private)

fuxialexander:
@HaoZeke Theoretically that possible as long as proper header args is used. A org file can be tangled to multiple files. https://orgmode.org/manual/tangle.html#tangle

modules/config/literate/init.el
```diff
@@ -11,14 +13,24 @@
...
 (defun +literate-compile (&optional load)
-  "Tangles & compiles `+literate-config-file' if it has changed. If LOAD is
-non-nil, load it too!"
-  (let ((org +literate-config-file)
-        (el  +literate-config-dest-file))
-    (when (file-newer-than-file-p org el)
+  "Tangles & compiles `+literate-config-file' and `+literate-packages-file' if
+ it has changed. If LOAD is non-nil, load it too!"
+  (let ((orgConf +literate-config-file)
+        (orgPack +literate-packages-file)
```
fuxialexander: A little comment on code style: elisp usually use dash rather than camel case for visual separation.

HaoZeke:
Actually I decided to go another route for this... Basically setting 

```lisp
#+PROPERTY: header-args :tangle yes
#+PROPERTY: header-args+ :tangle "~/.doom.d/packages.el"
```

On a `packages.org` which is version controlled.

hlissner:
I've made some big improvements to the `bin/org-tangle` script that may be helpful for your use-case. Running `bin/org-tangle ~/.doom.d` and it will compile all your literate org files. It will also ignore trees that have a `:notangle:` tag.

This lets you make your entire config literate (init.el, config.el and packages.el).

Sadly, this must be done outside of Doom, as the :config literate module requires you load it via `doom!` in your init.el before it can compile your literate config on the fly.

HaoZeke:
Thanks @hlissner, I will look into it... Actually I currently have it set to tangle on save with:

```emacs-lisp
# Local Variables:
# eval: (add-hook 'after-save-hook (lambda () (org-babel-tangle)) nil t)
# End:
```

PR #693, opened 2018-06-17
Various fixes

jwintz:
- `cae962d4` Use `:modes` instead of `:in` in ruby layer.
- `3993c25b` Setup `solaire-mode` for doom opera theme.
- `e6d4e446` Test feature for `rtags`: this one is prone to discussion but I like it better this way.


hlissner:
I've cherry-picked cae962d and 3993c25. The +rtags flag I do intend to add, but it will require a little more than just that one test. I'll add it soon.

Thanks for your help!

jwintz:
Thanks !

  - ju

Issue #702, opened 2018-06-20
Open recent project files doesn't work (snippets error)


zeronone:
### Observed behavior

1. Open `doom-emacs`
2. Open recent project
3. Select project
4. Select file (error)

<details>
<summary>Stack Trace</summary>

```
Debugger entered--Lisp error: (void-function +snippets|enable-project-modes)
  +snippets|enable-project-modes(+data-vagrant-mode t)
  run-hook-with-args(+snippets|enable-project-modes +data-vagrant-mode t)
  (if (not +data-vagrant-mode) nil (run-hook-with-args 'doom-project-hook '+data-vagrant-mode +data-vagrant-mode) nil nil)
  (let ((last-message (current-message))) (setq +data-vagrant-mode (if (eq arg 'toggle) (not +data-vagrant-mode) (> (prefix-numeric-value arg) 0))) (if (not +data-vagrant-mode) nil (run-hook-with-args 'doom-project-hook '+data-vagrant-mode +data-vagrant-mode) nil nil) (run-hooks '+data-vagrant-mode-hook (if +data-vagrant-mode '+data-vagrant-mode-on-hook '+data-vagrant-mode-off-hook)) (if (called-interactively-p 'any) (progn nil (if (and (current-message) (not (equal last-message (current-message)))) nil (let ((local " in current buffer")) (message "+Data-Vagrant mode %sabled%s" (if +data-vagrant-mode "en" "dis") local))))))
  +data-vagrant-mode(1)
  (and (fboundp '+data-vagrant-mode) (not (and (boundp '+data-vagrant-mode) +data-vagrant-mode)) (and buffer-file-name (not (file-remote-p buffer-file-name))) t (and (file-exists-p (expand-file-name "Vagrantfile" (doom-project-root)))) t (+data-vagrant-mode 1))
  doom--init-mode-+data-vagrant-mode()
  run-hooks(after-change-major-mode-hook)
  run-mode-hooks(haskell-mode-hook)
  haskell-mode()
  set-auto-mode-0(haskell-mode nil)
  set-auto-mode()
  normal-mode(t)
[...]

PR #705, opened 2018-06-20
feature/snippets: add missing autoload cookie

AloisJanicek:
Fixes https://github.com/hlissner/doom-emacs/issues/702


hlissner:
Good catch. Thanks!

PR #707, opened 2018-06-20
Fix init.example `electric` module

mattly-cambia:
I've been keeping my init up-to-date with the example, but the example
still has `electric-indent` whereas `make` warns me it's been renamed to
`electric`. This provides the correct module name in the example.


hlissner:
Thanks!

PR #708, opened 2018-06-21
Look for the correct feature name for org+rest

mattly-cambia:
the feature is in `init.el` as `rest` not `restclient`; the way this was `ob-restclient` will never get installed


hlissner:
Good catch, thanks!

PR #711, opened 2018-06-21
Added Zathura as a viewer option

patrl:
Note that zathura with synctex is already supported as a viewer option by AUCTeX, making this a simple one-liner.


hlissner:
Thanks for the help!

PR #713, opened 2018-06-22
Minor refactor of bibtex features in latex module

patrl:
- the variables `bibtex-completion-bibliography` and `bibtex-completion-library-path` are defined in `bibtex-completion.el`, which is part of the `ivy-bibtex` package, and shared by `ivy-bibtex` and `helm-bibtex`. For some reason they were being set locally to `bibtex`, thereby invisible to `ivy/helm-bibtex`. These variables have been removed from `bibtex` and set locally to `ivy/helm-bibtex`.

- The config for `bibtex` was opinionated about the method used by `ivy/helm-bibtex` for opening pdf files. Doom now uses the upstream `ivy-bibtex` default  - `pdf-tools` is used if installed, and `docview` otherwise.

- The config for `bibtex` was opinionated about the method used to find the location of pdf files (it set the `bibtex-completion-pdf-field` variable). Doom now defaults to upstream `ivy-bibtex` behaviour - pdf file names in `bibtex-completion-library-path` match against citation keys.


hlissner:
I understand the motivation of these changes and agree with them, but I'd rather not duplicate them across the two packages. Could you extract them into a `(after! bibtex-completion ...)` block instead?

patrl:
Done!

hlissner:
Great! Thanks again for your help!

Issue #460, opened 2018-03-12
Proper way to add support for LSP-mode


The-BigDaddy:
I'm learning Elisp and I find Doom to be a really interesting implementation of Emacs. I really like the idea of having a set of core functions/vars/etc. I've been looking through a lot of the code to understand it, and maybe at some point figure out how to refactor a few things to create my own sort of personal twist on Doom-Emacs.

The LSP-modes for Emacs are amazing, and I think it'd be interesting to figure out how to make use of them in Doom-Emacs. Although I'm not entirely sure how to implement the base LSP packages ([lsp-mode](https://github.com/emacs-lsp/lsp-mode), [lsp-ui](https://github.com/emacs-lsp/lsp-ui), [company-lsp](https://github.com/tigersoldier/company-lsp)). My questions are

1. Should a language-server module be under the `lang`, `tool`, or be it's own module?

2. How would one go about loading the configuration for each language-server for each available language, ex. how will lsp-python catch the default configurations for lsp-mode and such?

3. Should there be any autoloads? 

4. Does this Doom-Emacs `config.el` code seem sane?

```
;;; lang/language-server/config.el -*- lexical-binding: t; -*-

(def-package! lsp-mode
  :after (:any cc-mode
               c-mode
               c++-mode
               objc-mode
               python 
               haskell-mode
               rust-mode 
               caml-mode
               vue-mode
               css-mode
               scss-mode 
               sass-mode
               less-mode
               stylus-mode)
  :config
  (setq lsp-enable-eldoc t)
  (setq lsp-enable-completion-at-point t))

(def-package! lsp-ui
  :requires lsp-mode
  :config 
  (with-eval-after-load "lsp-mode"
    (setq lsp-ui-flycheck-enable t)
  (map! :map lsp-ui-mode-map 
        [remap xref-find-definitions] . lsp-ui-peek-find-definitions
        [remap xref-find-references] . lsp-ui-peek-find-references)
  :hook
  (lsp-mode . lsp-ui-mode))
  
(def-package! company-lsp
  :after (:all lsp-mode lsp-ui)
  :config
  (set! :company-backend lsp-mode 'company-lsp)
  (setq company-lsp-async t)
  (setq company-lsp-enable-snippet t))
;;; EOF
```

hlissner:
The only issue I see is `:requires lsp-mode` under `lsp-ui`. 

[...]

PR #716, opened 2018-06-23
Feature/lsp

Emiller88:
@ar1a Did most of the work in her private config, I just wanted to bring it to the greater Doom community.

I have not tested this yet, I would appreciate some help on that front. 

Feel free to add any other languages that have lsp support you want to see.


patrl:
`doom re` throws up the following error on your feature branch:

```
Loading /home/patrl/GitHub/doom-emacs/modules/tools/lsp/packages.el (source)...
Eager macro-expansion failure: (wrong-number-of-arguments ((t) (module submodule &optional flags) "Declares that this module depends on another.

Only use this macro in a module's packages.el file.
```

modules/lang/rust/packages.el
```diff
@@ -9,5 +9,12 @@
...
 
-(when (featurep! :completion company)
-  (package! company-racer))
+(cond ((featurep! +lsp)
+       (depends-on! :tools lsp
+                    (package! lsp-rust)
+                    :recipe (:fetcher
+                             github
+                             :repo "emacs-lsp/lsp-javascript"
+                             :files ("lsp-typescript.el"))))
```
bandresen: This looks wrong.

bandresen:
Maybe it would be prudent if some of these configurations also checked for the presence of the program in `exec-path` before enabling the lsp configuration. (`pyls` or `bash-language-server`)

I don't have experience with LSP myself (I am very interested) but company can be quite annoying when it doesn't get what it expects in some circumstances that can lead to bad behavior.

modules/lang/rust/packages.el
Emiller88: What about it? The javacript and typescript difference?

Emiller88:
@patrl I'll test it out. 

@bandresen I think that's a great idea. I also don't have any idea how to implement that.

modules/lang/rust/packages.el
bandresen: This is the rust/packages.el :-)

Emiller88: I missed that! As you can see I wanted to try to keep things "similar". Should be fixed by 9914002

Emiller88:
@ar1a  That could work. Give it a shot and then add it.

hlissner:
Thank you for making this effort. I want to integrate LSP into Doom soon and this is a great start. 

I will hold off on merging this PR until 2.1.1, because there are some infrastructural decisions i must make that will impact this PR directly. Don't let that stop you from working on it, however. I will make the necessary adjustments once I am done.

Emiller88:
@hlissner  Definitely! I just created the PR so that everyone interested could work on it/ try it out, rather than a bunch of people working on their own personal configs.

modules/lang/go/packages.el
```diff
@@ -8,3 +8,7 @@
 
 (when (featurep! :completion company)
   (package! company-go))
+
+(cond ((feature! +lsp)
```
paschdan:
```
(cond ((featurep! +lsp)
```

you are missing a `p`

Emiller88: Fixed! Thanks!

Emiller88:
I'm thinking about a lsp +toggle feature that will keep lsp disabled unless you enable it for a file type in that session. Otherwise it'll use the light weight emacs package. Currently lsp seems pretty powerful, making it a full on ide, but it's pretty heavy about 2gb of memory for the python lsp. In the event you've pulled up 3 or 4 different lsps it might go over what a lot of systems can handle. Might just be a python thing It's the only language I've tested enough.

modules/lang/cc/config.el
```diff
@@ -250,3 +250,19 @@ compilation database is present in the project.")
   (def-package! helm-rtags
     :when (featurep! :completion helm)
     :config (setq rtags-display-result-backend 'helm)))
+
+(def-package! cquery
```
MaskRay: When this PR gets merged, I'll add my own language server ccls https://github.com/MaskRay/ccls/wiki/Emacs ... it now has a lot of improvement on pipeline, indexing and other aspects.

ar1a: why not do it now?

modules/lang/rust/packages.el
```diff
@@ -9,5 +9,8 @@
 (when (featurep! :feature syntax-checker)
   (package! flycheck-rust))
 
-(when (featurep! :completion company)
-  (package! company-racer))
+(cond ((and (depends-on! :tools lsp)
+            (featurep! +lsp))
+       (package! lsp-rust))
+      ((when (featurep! :completion company)
+         (package! company-racer))))
```
PoiScript: `racer` don't need to be installed defaultly any more. `rls` already [depended on it](https://github.com/rust-lang-nursery/rls/blob/master/Cargo.toml#L27).

edkolev:
FYI you might consider using eglot instead of lsp https://github.com/joaotavora/eglot

See here for a comparison between the two https://github.com/joaotavora/eglot#differences-to-lsp-modeel

myrgy:
Hi everyone!
@hlissner, do you have any estimation when this MR could be merged?
Thanks!

Emiller88:
@myrgy It will be after 2.1

myrgy:
Hi guys,

does anybody have any issues with `xref-find-referencies` ?
c++ project using ccls. semantic highlight and completion works fine.
please find stack trace below:
```
  gethash("uri" nil)
  lsp--location-to-td-position(nil)
  lsp--symbol-info-to-identifier(#<hash-table equal 6/65 0x18675d1>)
  mapcar(lsp--symbol-info-to-identifier (#<hash-table equal 6/65 0x18675d1> #<hash-table equal 6/65 0x1c293d5> #<hash-table equal 6/65 0x192baf1>))
  #f(compiled-function (function sequence) #<bytecode 0x52b765>)(lsp--symbol-info-to-identifier (#<hash-table equal 6/65 0x18675d1> #<hash-table equal 6/65 0x1c293d5> #<hash-table equal 6/65 0x192baf1>))
  apply(#f(compiled-function (function sequence) #<bytecode 0x52b765>) lsp--symbol-info-to-identifier (#<hash-table equal 6/65 0x18675d1> #<hash-table equal 6/65 0x1c293d5> #<hash-table equal 6/65 0x192baf1>) nil)
  seq-map(lsp--symbol-info-to-identifier (#<hash-table equal 6/65 0x18675d1> #<hash-table equal 6/65 0x1c293d5> #<hash-table equal 6/65 0x192baf1>))
  #f(compiled-function (backend) #<bytecode 0x1c5da09>)(xref-lsp)
  apply(#f(compiled-function (backend) #<bytecode 0x1c5da09>) xref-lsp nil)
  xref-backend-identifier-completion-table(xref-lsp)
  xref--read-identifier("Find references of: ")
  byte-code("\300\301!C\207" [xref--read-identifier "Find references of: "] 2)
  call-interactively(xref-find-references nil nil)
  command-execute(xref-find-references)
```

modules/lang/cc/packages.el
```diff
@@ -27,3 +27,13 @@
     (package! ivy-rtags))
   (when (featurep! :completion helm)
     (package! helm-rtags)))
+
+(cond ((featurep! +lsp)
+       (depends-on! :tools lsp)
+       (package! lsp-typescript)))
```
Capt0r:
```suggestion
       (package! lsp-cc)))
```

Emiller88:
Rebased before doing some work on the python module

modules/lang/cc/packages.el
```diff
@@ -27,3 +27,13 @@
     (package! ivy-rtags))
   (when (featurep! :completion helm)
     (package! helm-rtags)))
+
+(cond ((featurep! +lsp)
+       (depends-on! :tools lsp)
+       (package! lsp-cc))
```
bennylb: @Emiller88 not entirely sure but what is `lsp-cc`? Is this supposed to be `cquery`?

ar1a: https://github.com/MaskRay/ccls maybe this?

bennylb: Ah yeah that got me thinking, I think it's a temporary place holder. Replacing that with `cquery` works as intended.

Emiller88: Not sure what it is @bennylb. Does `cquery` use @MaskRay 's language server?

sveitser:
i just tried this out with typescript but didn't seem to get it to work.

I did 
- Add `+lsp` to the javascript layer.
- Install the node pacakage `typescript-language-server`.
- Run `bin/doom re`.

Is that correct? 

After opening typescript files it seems that `tide-mode` was still enabled and working but turning it off as well as enabling `lsp-mode` and `lsp-ui-mode` didn't make any lsp related things show up.

Emiller88:
@sveitser did you enable `:tools lsp`?

sveitser:
I forgot that. Thanks!

I'm now getting when turning on `typescript-mode`
```
helm-M-x: Symbol’s function definition is void: lsp-make-traverser
```

I think this was removed here

https://github.com/emacs-lsp/lsp-mode/commit/ec0baa150571945cb7316ecb235ad99ed31c413a#diff-3c9f975ff959eb4890c1927b91ce8f66L82

sveitser:
If anybody is interested, regarding `typescript` and `javascript`. Things worked out of the box with the following.

- Install node package `javascript-typescript-langserver`.

Add the following to doom config

```diff
diff --git a/home-nixpkgs/dotfiles/emacs/doom/config.el b/home-nixpkgs/dotfiles/emacs/doom/config.el
index 6ca248f..67da841 100644
--- a/home-nixpkgs/dotfiles/emacs/doom/config.el
+++ b/home-nixpkgs/dotfiles/emacs/doom/config.el
@@ -49,3 +49,8 @@
                    :ttl t))
 
 (setq evil-escape-key-sequence "hh")
+
+
+(require 'lsp)
+(require 'lsp-clients)
+(add-hook 'js2-mode-hook #'lsp)
diff --git a/home-nixpkgs/dotfiles/emacs/doom/packages.el b/home-nixpkgs/dotfiles/emacs/doom/packages.el
index c5c48c7..2f72426 100644
--- a/home-nixpkgs/dotfiles/emacs/doom/packages.el
+++ b/home-nixpkgs/dotfiles/emacs/doom/packages.el
@@ -17,3 +17,5 @@
 (package! solidity-mode)
 (package! terraform-mode)
 (package! weechat)
+(package! lsp-mode)
+(package! lsp-ui)
```

Emiller88:
@sveitser So typescript doesn't work out of the box with the way the package is written or is the documentation wrong? The user ideally shouldn't need to add anything to their private config

sveitser:
@Emiller88 It didn't work for me, no. What I posted in my previous comment works against doom `develop` for javascript/typescript.

I'm somewhat confused by the number of `lsp` related packges.

I think https://github.com/emacs-lsp/lsp-javascript used at https://github.com/hlissner/doom-emacs/pull/716/files#diff-87af0d4013454a052d821569780a430bR23 is no longer required and does no longer work with the current version of https://github.com/emacs-lsp/lsp-mode.

yyoncho:
@sveitser 

> Edit: I also had to change the node package to javascript-typescript-langserver (instead of typescript-language-server).

It will work with both.

Emiller88:
Something have probably changed with how we implement this due to https://github.com/emacs-lsp/lsp-mode#whats-new-in-lspel

Some language servers are now included by default.

.gitignore
```diff
@@ -29,4 +29,6 @@ network-security.data
 # transient/os temp files
 *.cache-*
 *.log
-.directory
+
+org-gcal/
```
ar1a: probably dont wanna commit this

init.example.el
```diff
@@ -79,6 +80,7 @@
        ;;tmux              ; an API for interacting with tmux
        ;;upload            ; map local to remote projects via ssh/ftp
        ;;wakatime
+       ;;lsp
```
ar1a: this is added twice to init.example.el

Emiller88:
@MaskRay I agree I think I'll just start this over and just add the `tools/lsp` to begin with since the changes that were made on lsp.

init.example.el
Emiller88: Probably from merging other PRs and missed it.

.gitignore
Emiller88: You are correct.

modules/lang/python/packages.el
```diff
@@ -20,3 +20,10 @@
 (when (package! anaconda-mode)
   (when (featurep! :completion company)
     (package! company-anaconda)))
+;; lsp
+(cond ((and (featurep! :tools +lsp)
```
mowind: @Emiller88  This condition `(featurep! :tools +lsp)` not effective, modify to `(featurep! :tools lsp)` may be correct.

Emiller88: Thanks for catching this!

rudedogg:
Swift would be another one:

- https://github.com/emacs-lsp/lsp-sourcekit
- https://github.com/apple/sourcekit-lsp/

hlissner:
Just a heads up, `set-lookup-handler!` now has async handler support. Everything after `:async t` in a set-lookup-handler! call will be treated as an async handler (i.e. it's positional).

```lisp
(set-lookup-handlers! 'tide-mode :async t
    :definition #'tide-jump-to-definition
    :references #'tide-references
    :documentation #'tide-documentation-at-point)
```

The only caveat is, if a mode has an async handler, it won't fall back to other backends. It always assumes the async handler worked.

If that's important at all, you'd have to write a blocking wrapper for `lsp-ui-peek-find-*`.

modules/lang/haskell/doctor.el
```diff
@@ -13,4 +13,6 @@
   (unless (executable-find "hlint")
     (warn! "Couldn't find hlint. Flycheck may have issues in haskell-mode")))
 
-
+(when (featurep! +lsp)
+  (unless (executable-find "hie")
```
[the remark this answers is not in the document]
hlissner: Please do not use use-package's `:ensure-system-packages`. It would circumvent Doom's startup optimizations (with package checks and installation) and impose too many difficult-to-reverse opinions on the setup of your system. 

Doom opts for a "user-knows-best" hands-off approach when it comes to dependencies outside of Emacs, because we programmers are a persnickety bunch, myself included.

That does mean a little more work for us and, admittedly, Doom suffers on the documentation front, but there are tools I hope that can help. In particular: `bin/doom doctor`, which will report missing external dependencies that haven't been installed. Doom should also fail gracefully in the absence of these dependencies.

ar1a:
the haskell lsp hook is wrong, with the new system you can't use :hook anymore because you need the package loaded before calling M-x lsp

modules/lang/haskell/doctor.el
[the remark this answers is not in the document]
CestDiego: I see your point about the `:ensure-system-packages` and I agree with you. I didn't really had an opinion on it other than it would help me when I switch machines and I have to install things all over again.

hlissner:
With the recent changes to lsp-mode, this PR is now outdated. However, I've implemented rudimentary LSP support in cb923ea for a variety of languages (see the commit message). Adding the `+lsp` is sufficient for enabling it in the language modules that support it. It still needs to be documented however, and I haven't yet grokked the replacement for `lsp-define-stdio-client`, but let me know if it can do with any improvements.

To everyone who contributed to this discussion and PR, thank you! Especially @Emiller88 for getting the ball rolling. I'll close this PR and move the discussion to #460.

PR #717, opened 2018-06-23
Add keybindings for circe

MarkRedeman:
This PR adds `SPC o I` as a keybinding to open IRC. When in a `circe` buffer some additional keybindings are available to make it easier to switch channels and sent private messages.

I also added an example of a pass entry to the documentation.

### DIFF
--- a/modules/app/irc/README.org
+++ b/modules/app/irc/README.org
@@ -3,10 +3,12 @@
 This module turns adds an IRC client to Emacs ([[https://github.com/jorgenschaefer/circe][~circe~)]] with native notifications ([[https://github.com/eqyiel/circe-notifications][circe-notifications]]).
 
 * Table of Contents :TOC:
-- [[#dependencies][Dependencies]]
-- [[#configure][Configure]]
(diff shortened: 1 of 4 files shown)


bandresen:
Very nice!

hlissner:
This looks great. Although, I have issue with the `SPC o I` keybinding. All app keybinds were removed not so long ago. The expected way to open them is through `M-x`, because:

+ Activating an app module is a drastic context shift and was too close to other less imposing keys like opening eshell, term, dired or neotree.
+ Part of the reason why app "launcher" commands were given a unique prefix (`=twitter`) was so they would be easy to locate and isolte in the `M-x` menu.

I may add a `SPC o =` prefix for apps later, but for the time being, could you remove the keybind?

MarkRedeman:
That makes sense. Having a `SPC o =` prefix would be nice as it helps the discoverability of the apps (when switching from Spacemacs to Doom it wasn't immediately obvious how to use the IRC module).
For now I've removed the keybinding and made a note about starting IRC using `=irc` in the documentation.

hlissner:
Excellent work. Thanks again!

PR #719, opened 2018-06-23
ui/vc emacs/vc: unify hydra design

bandresen:
tiny design niggle I had


bandresen:
@hlissner I do use them. I sometimes get the smerge-menu not popping up automatically but I use git gutter all the time.

PR #720, opened 2018-06-23
Cut: lang/clojure: unneeded + subjective parts

bandresen:
str => clojure.string is already defined in `cljr-magic-require-namespaces'
re-frame + reagent is subjective (for personal configuration IMO)

fighweel aspect is now handled upstream by giving a prompt
see: `cider-cljs-repl-types'


hlissner:
Thanks again!

PR #723, opened 2018-06-25
Fix disable ivy fuzzy search in counsel-grep and swiper

fuxialexander:
As the title said. The previous code does not work.


hlissner:
Thanks!

PR #726, opened 2018-06-25
Fix handler calls to reference lisp-mode.

areinisc:
`set-repl-handler!` helps with opening a repl when a particular mode is
active in a buffer. We want to be able to open a `sly-mrepl` whenever we
are in a lisp buffer, so we should have the repl handler look for
`'lisp-mode` to define opening a sly repl.

I assume the lookup handlers should be making sure we are in a lisp
buffer (similar reasoning to the repl-handler).

### DIFF
--- a/modules/lang/common-lisp/config.el
+++ b/modules/lang/common-lisp/config.el
@@ -5,8 +5,8 @@
   (setq inferior-lisp-program "sbcl")
 
   (set-popup-rule! "^\\*sly" :quit nil :ttl nil)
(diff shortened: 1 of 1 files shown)


hlissner:
It is preferrable these commands are used with minor modes (sly-mode) instead of major modes (lisp-mode), in case sly-mode is purposely disabled by the user (for whatever reason). However, though these commands _should_ work with minor modes, they currently do not, so I will accept this PR until they do.

PR #731, opened 2018-06-26
Factor out hindent

patrl:
`hindent` was previously only loaded as part of the `+intero` feature. It's now factored out as part of `config.el` with a `doom doc` check for the `hindent` executable.


hlissner:
Excellent. Thanks again!

PR #732, opened 2018-06-26
pdf: Set evil state to normal on initialization

HaoZeke:
Prevents starting in emacs mode.


HaoZeke:
To be fixed in the evil module

PR #743, opened 2018-07-06
Add support for multiple ligature'd fonts

jseba:
Refactors the `prettify-symbols' module to support more than one font for ligatures since Fira Code's ligature implementation is different from Iosevka's.


modules/ui/pretty-code/autoload/settings.el
```diff
@@ -0,0 +1,67 @@
...
+    :not-in        "∉"
+    :and           "∧"
+    :or            "∨"
+    :for           "∀"
+    :some          "∃"
+    :return        "⟼"
+    :yield         "⟻"
+    ;; Other
+    :tuple         "⨂"
+    :pipe          "" ;; FIXME: find a non-private char
```
jseba: The `:pipe' symbol is located in the Private Use Area which is not consistent between fonts and so maps to the wrong glyph in Fira. The two options I see are to not allow that plist to use characters in the PUA or to come up with a more versatile mapping system to the glyph codes.

hlissner:
I'm not certain how to handle the :pipe symbol. `+pretty-code-symbols` is pretty easy to modify, so I won't lose sleep over it, but I'm open to suggestions.

In any case, thank you for your hard work! I've done a little refactoring to fix a few errors, but otherwise it appears to work. I am not a big user of this module however, so let me know if I've missed anything!

hlissner:
@jseba I just noticed that there are two entries for `"<="` [here](https://github.com/hlissner/doom-emacs/blob/develop/modules/ui/pretty-code/+fira.el#L72) and [here](https://github.com/hlissner/doom-emacs/blob/develop/modules/ui/pretty-code/+fira.el#L94). I don't know enough about ligatures to resolve this. Could you advise?

jseba:
It looks like a typo in the list of character codes that I used to make Fira's ligatures work. I think the first one should be "=<". The characters look the same after rendering, but an updated version of the font has the "=<" version slightly tweaked. I'm not sure what language uses =< instead of <=.

PR #750, opened 2018-07-11
org: Update +export to generate revealjs stuff

HaoZeke:
Necessary to not create plain HTML presentations.

### DIFF
--- a/modules/lang/org/+export.el
+++ b/modules/lang/org/+export.el
@@ -18,7 +18,8 @@ path too.")
   (add-to-list 'org-export-backends 'pandoc nil #'eq)
   (setq org-pandoc-options
         '((standalone . t)
-          (mathjax . t))))
+          (mathjax . t)
+          (variable . "revealjs-url=https://cdn.jsdelivr.net/npm/reveal.js@3/"))))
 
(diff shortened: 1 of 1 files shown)


HaoZeke:
With a bit more work this could actually completely replace `ox-reveal`, but that wasn't the point of this.

hlissner:
ox-reveal appears to expose a lot of custom properties and settings. Do these work through pandoc?

Also, what if we _want_ to export plain HTML, would this setting prevent that?

HaoZeke:
@hlissner, 'ox-reveal' is a wrapper around revealJS and 'pandoc' also exposes the exact same set of features, just with a different syntax..

I can duplicate the 'ox-reveal' defaults here too... But for new users 'ox-reveal' is a lot easier so there is no point. 

This does not affect generating 'html', it just ensures that choosing 'revealJS' as the 'pandoc' export target will link to a valid stylesheet. Currently exporting to 'revealJS' with 'pandoc' does not really create a presentation, because it does not link to the 'revealJS' stylesheets or scripts.

hlissner:
Understood. Thanks for the PR!

PR #752, opened 2018-07-12
Improved repl handling for haskell module

patrl:
The haskell repl is now handled by `interactive-haskell-mode`, which supports a distinct haskell process per cabal project. Repl-handling is also set up for `.cabal` and `.lhs` (literate haskell) files.

`.hs` files also now have a module declaration inserted automatically (just like, e.g., `.el` files).

Still to do:
- [ ] disable for intero users


hlissner:
I went about this a bit differently with a76c080, 80c4abc, and 34db3f0. It should resolve the REPL situation with Haskell. Give it a try and let me know!

patrl:
This works great, thanks!

PR #753, opened 2018-07-12
Load electric ident

patrl:
Since `haskell-indentation-mode` (the default in `haskell-mode`) supports `electric-indent-mode`, load it if `electric` is enabled

### DIFF
--- a/modules/lang/haskell/config.el
+++ b/modules/lang/haskell/config.el
@@ -3,6 +3,9 @@
 (cond ((featurep! +intero) (load! "+intero"))
       ((featurep! +dante)  (load! "+dante")))
 
+(when (featurep! :emacs electric)
+  (add-hook 'haskell-mode-hook 'electric-indent-mode))
+
 ;;
 ;; Common plugins
 ;;
--- a/modules/lang/haskell/config.el
+++ b/modules/lang/haskell/config.el
@@ -3,8 +3,6 @@
 (cond ((featurep! +intero) (load! "+intero"))
       ((featurep! +dante)  (load! "+dante")))
 
-(when (featurep! :emacs electric)
-  (add-hook 'haskell-mode-hook 'electric-indent-local-mode))
 
 ;;
 ;; Common plugins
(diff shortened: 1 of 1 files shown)


hlissner:
`electric-indent-mode` is a global minor mode. I do not what electric-indent-mode enabled globally. What you probably meant was `electrin-indent-local-mode`. 

However, for two reasons, I'd much rather not enable electric-indent by default, in any mode. First, because it is _very_ easy for users to enable themselves (just  `(electric-indent-mode +1)` or `(add-hook 'some-mode-hook #'electric-indent-local-mode)`), I'd much rather it be left up them.

And second, haskell's indentation logic is quite poor and imprecise (more to do with the complex nature of whitespace in haskell than poor code design -- although `hindent` helps a lot with that), so I'd much rather not put it in charge of arbitrarily adjusting indentation when it thinks it needs to.

patrl:
Thanks for the explanation! Apologies for the dumb mistake - I've changed `electric-indent-mode` to `electric-indent-local-mode` in the PR. I'm quite happy to close this PR, it was admittedly made way too hastily. 

I will say however, that due to `haskell-indentation-common-electric-command` defined in `haskell-mode`, I personally found electric indent to improve the situation quite a bit. It's completely trivial for users to set this in their own configs however.

hlissner:
I looked deeper into this and found this tidbit in `haskell-mode`'s definition:

```elisp
  ;; Haskell is not generally suitable for electric indentation, since
  ;; there is no unambiguously correct indent level for any given line.
  (when (boundp 'electric-indent-inhibit)
    (setq electric-indent-inhibit t))
```

So it seems electric indentation does nothing in haskell-mode.

patrl:
Should I use a `def-package!` block to make sure this gets set before `haskell-mode` is loaded? i.e.

```lisp
(def-package! haskell-mode
  :init (setq haskell-indentation-electric-flag t)
  :config
  (set-repl-handler! 'haskell-mode #'switch-to-haskell)
  (add-to-list 'completion-ignored-extensions ".hi"))
```

patrl:
This works great! thanks!! Closing the PR.

PR #754, opened 2018-07-12
DONE: large refactor of latex

patrl:
The latex module probably needs cleaning up. It's (a) too opinionated, and (b) has several redundancies. This PR makes a start on doing that. 

- [x] general spring-cleaning
- [x] Consolidate existing config using `def-package!`. Everything should be lazily loaded in the right place now.
- [x] sub-modules moved into distinct files for ease of maintenance
- [x] citation and label completion are now handled by `company-reftex`, which is more sophisticated than the `company-auctex` backends.
- [x] symbol completion in math mode is now handled by `company-math`, which has a much more complete symbol-list than `company auctex`. Note: because there are so many candidates, this causes a hang on first load.
- [x] The variable `+latex-bibtex-dir` is removed. It was only used by `bibtex-completion` anyway. Users can set `bibtex-completion-library-path` themselves. Note: this has the potential to break existing configs.
- [x] hook for `LaTeX-math-mode` removed. It's rendered superfluous by symbol completion.
- [x] PDF viewers are configured upstream. They don't need to be configured here.
- [x] remove redundant package declaration for flyspell
- [x] add a `+unicode-math` option, which enables math-mode completion to unicode symbols. Useful for people using the latex unicode-math package.

### DIFF
--- a/modules/lang/latex/+latexmk.el
+++ b/modules/lang/latex/+latexmk.el
@@ -2,7 +2,8 @@
 ;;;###if (featurep! +latexmk)
 
 (def-package! auctex-latexmk
-  :after-call (latex-mode-hook LaTeX-mode-hook)
+  :defer t
+  :after latex
   :init
   ;; Pass the -pdf flag when TeX-PDF-mode is active
   (setq auctex-latexmk-inherit-TeX-PDF-mode t)
--- a/modules/lang/latex/+latexmk.el
+++ b/modules/lang/latex/+latexmk.el
@@ -1,0 +1,13 @@
+;;; lang/latex/+latexmk.el -*- lexical-binding: t; -*-
+;;;###if (featurep! +latexmk)
+
+(def-package! auctex-latexmk
+  :after-call (latex-mode-hook LaTeX-mode-hook)
+  :init
+  ;; Pass the -pdf flag when TeX-PDF-mode is active
+  (setq auctex-latexmk-inherit-TeX-PDF-mode t)
+  ;; Set LatexMk as the default
+  (setq-hook! LaTeX-mode TeX-command-default "LatexMk")
+  :config
+  ;; Add latexmk as a TeX target
+  (auctex-latexmk-setup))
--- a/modules/lang/latex/+latexmk.el
+++ b/modules/lang/latex/+latexmk.el
@@ -2,7 +2,6 @@
 ;;;###if (featurep! +latexmk)
 
 (def-package! auctex-latexmk
-  :defer t
   :after latex
   :init
   ;; Pass the -pdf flag when TeX-PDF-mode is active
--- a/modules/lang/latex/+ref.el
+++ b/modules/lang/latex/+ref.el
@@ -1,13 +1,8 @@
 ;;; lang/latex/+ref.el -*- lexical-binding: t; -*-
 
-(defvar +latex-bibtex-file ""
-  "File AUCTeX (specifically RefTeX) uses to search for citations.")
-
 (def-package! reftex
-  :hook ((latex-mode LaTeX-mode) . turn-on-reftex)
-  :init
-  (setq reftex-plug-into-AUCTeX t
-        reftex-toc-split-windows-fraction 0.3)
+  :defer t
+  :hook (LaTeX-mode . reftex-mode)
   :config
   ;; Get ReTeX working with biblatex
   ;; http://tex.stackexchange.com/questions/31966/setting-up-reftex-with-biblatex-citation-commands/31992#31992
@@ -18,7 +13,9 @@
           (?s . "\\smartcite[]{%l}")
           (?f . "\\footcite[]{%l}")
           (?n . "\\nocite{%l}")
-          (?b . "\\blockcquote[]{%l}{}")))
+          (?b . "\\blockcquote[]{%l}{}"))
+        reftex-plug-into-AUCTeX t
+        reftex-toc-split-windows-fraction 0.3)
   (unless (string-empty-p +latex-bibtex-file)
     (setq reftex-default-bibliography (list (expand-file-name +latex-bibtex-file))))
   (map! :map reftex-mode-map
@@ -31,14 +28,15 @@
           :e "q"   #'kill-buffer-and-window
           :e "ESC" #'kill-buffer-and-window)))
 
+;; set up completion for citations and references
 (def-package! company-reftex
   :when (featurep! :completion company)
   :after reftex
   :config
   (set-company-backend! 'reftex-mode 'company-reftex-labels 'company-reftex-citations))
 
+;; set up mode for bib files
 (def-package! bibtex
-  :defer t
   :config
   (setq bibtex-dialect 'biblatex
         bibtex-align-at-equal-sign t
--- a/modules/lang/latex/+ref.el
+++ b/modules/lang/latex/+ref.el
@@ -1,0 +1,60 @@
+;;; lang/latex/+ref.el -*- lexical-binding: t; -*-
+
+(defvar +latex-bibtex-file ""
+  "File AUCTeX (specifically RefTeX) uses to search for citations.")
+
+(def-package! reftex
+  :hook ((latex-mode LaTeX-mode) . turn-on-reftex)
+  :init
+  (setq reftex-plug-into-AUCTeX t
+        reftex-toc-split-windows-fraction 0.3)
+  :config
+  ;; Get ReTeX working with biblatex
+  ;; http://tex.stackexchange.com/questions/31966/setting-up-reftex-with-biblatex-citation-commands/31992#31992
+  (setq reftex-cite-format
+        '((?t . "\\textcite[]{%l}")
+          (?a . "\\autocite[]{%l}")
+          (?c . "\\cite[]{%l}")
+          (?s . "\\smartcite[]{%l}")
+          (?f . "\\footcite[]{%l}")
+          (?n . "\\nocite{%l}")
+          (?b . "\\blockcquote[]{%l}{}")))
+  (unless (string-empty-p +latex-bibtex-file)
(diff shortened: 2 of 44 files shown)


modules/lang/latex/config.el
```diff
@@ -14,9 +14,6 @@
 ;; Plugins
 ;;
 
-;; sp's default rules are obnoxious, so disable them
-(provide 'smartparens-latex)
```
hlissner: This was here to _disable_ smartparens' config for latex, which is obnoxious and slow. By removing this, you are re-enabling them, which is undesirable. Was that intentional?

patrl: D'oh - no, this was a dumb oversight, I thought this was enabling sp's latex support. I'll change this back.

modules/lang/latex/+latexmk.el
```diff
@@ -0,0 +1,14 @@
+;;; lang/latex/+latexmk.el -*- lexical-binding: t; -*-
+;;;###if (featurep! +latexmk)
+
+(def-package! auctex-latexmk
+  :defer t
```
hlissner: I'd suggest removing this. `:defer t` is redundant with `:after latex`

modules/lang/latex/+ref.el
```diff
@@ -0,0 +1,58 @@
+;;; lang/latex/+ref.el -*- lexical-binding: t; -*-
+
+(def-package! reftex
+  :defer t
```
hlissner: This should also be removed (`:defer t` is redundant with `:hook`)

modules/lang/latex/+ref.el
```diff
@@ -0,0 +1,58 @@
...
+
+;; set up completion for citations and references
+(def-package! company-reftex
+  :when (featurep! :completion company)
+  :after reftex
+  :config
+  (set-company-backend! 'reftex-mode 'company-reftex-labels 'company-reftex-citations))
+
+;; set up mode for bib files
+(def-package! bibtex
```
hlissner: Use `(after! bibtex ...)` here instead. `bibtex` is being eagerly loaded here because there is no `:defer t` or other deferring property.

modules/lang/latex/config.el
```diff
@@ -17,204 +13,101 @@
...
+        TeX-source-correlate-start-server nil
+        ;; automatically insert braces after sub/superscript in math mode
+        TeX-electric-sub-and-superscript t)
+  ;; prompt for master
   (setq-default TeX-master nil)
-  ;; Display the output of the latex commands in a popup.
+  ;; set-up chktex
+  (setcar (cdr (assoc "Check" TeX-command-list)) "chktex -v6 -H %s")
+  ;; tell emacs how to parse tex files
+ (add-hook! 'tex-mode-hook
```
hlissner: `add-hook!` is being used incorrectly here. I'd suggest either:

```elisp
(add-hook! 'tex-mode-hook (setq ispell-parser 'tex))
;; or 
(add-hook 'tex-mode-hook (lambda () (setq ispell-parser 'tex)))
```

modules/lang/latex/config.el
```diff
@@ -17,204 +13,101 @@
...
+           (lambda () (setq ispell-parser 'tex)))
+  ;; display output of latex commands in popup
   (set-popup-rule! " output\\*$" :size 15)
+  ;; Do not prompt for Master files, this allows auto-insert to add templates to
+  ;; .tex files
+  (add-hook! 'TeX-mode-hook (remove-hook 'find-file-hook
+                                         (cl-find-if #'byte-code-function-p find-file-hook)
+                                         'local))
+  ;; Enable rainbow mode after applying styles to the buffer
+  (add-hook! 'TeX-update-style-hook #'rainbow-delimiters-mode)
```
hlissner: The use of `add-hook!` is unnecessary when adding one function to one hook. Change this to the regular `add-hook`.

modules/lang/latex/config.el
```diff
@@ -17,204 +13,101 @@
...
-
-  ;; TeX Folding
-  (add-hook 'TeX-mode-hook #'TeX-fold-mode))
-
+; Fold TeX macros
+(def-package! tex-fold
+  :hook (TeX-mode . TeX-fold-mode))
 
-(after! latex
+(def-package! latex
```
hlissner: `(after! latex ...)` is preferred over `(def-package! latex :defer t :config ...)` because they are functionally equivalent, but we suffer additional overhead when using `use-package`.

modules/lang/latex/config.el
```diff
@@ -17,204 +13,101 @@
...
-    (add-hook 'TeX-after-compilation-finished-functions #'TeX-revert-document-buffer)))
+    (add-to-list 'LaTeX-indent-environment-list `(,env +latex/LaTeX-indent-item))))
 
-
-;; The preview package is currently broken with the latest AUCTeX version
-;; ("11.90.2.2017-07-25) ... and Ghostscript 9.22. It's now fixed in AUCTeX
-;; master, so we just have to wait.
+;; set-up preview package
 (def-package! preview
+  :defer t
```
hlissner: `:defer t` is redundant with `:hook`

modules/lang/latex/config.el
```diff
@@ -17,204 +13,101 @@
...
+                             company-auctex-environments))
+                          company-backends)))))
 
 ;; Nicely indent lines that have wrapped when visual line mode is activated
 (def-package! adaptive-wrap
   :hook (LaTeX-mode . adaptive-wrap-prefix-mode)
   :init (setq-default adaptive-wrap-extra-indent 0))
+
+;; referencing + bibtex setup
+(load! "+ref")
```
hlissner: The code in +ref.el references variables that are (presumably) defined in tex. This will throw an error if the package hasn't already been loaded, so it should be moved into tex or latex def-package block above.

modules/lang/latex/config.el
```diff
@@ -17,204 +13,101 @@
...
-  ;; We can't use the `set-company-backend!' because Auctex reports its
-  ;; major-mode as `latex-mode', but uses LaTeX-mode-hook for its mode, which is
-  ;; not something `set-company-backend!' anticipates (and shouldn't have to!)
-  (add-hook! LaTeX-mode
-    (make-local-variable 'company-backends)
-    (company-auctex-init)))
-
+  :after latex
+  :config
+  (def-package! company-math
```
hlissner: company-auctex and company-math are too tightly coupled here. If company-math is disabled, we implicitly disable company-auctex, and vice versa. I'll follow up with a suggestion once I get home later.

modules/lang/latex/+latexmk.el
patrl: Thanks, fixed in the latest commit!

modules/lang/latex/+ref.el
patrl: Thanks, fixed in the latest commit!

modules/lang/latex/+ref.el
patrl: Thanks, fixed in the latest commit!

modules/lang/latex/config.el
patrl: Thanks, fixed in the latest commit!

modules/lang/latex/config.el
patrl: Thanks, fixed in the latest commit!

modules/lang/latex/config.el
patrl: Thanks, fixed in the latest commit!

modules/lang/latex/config.el
patrl: Thanks, fixed in the latest commit!

modules/lang/latex/config.el
```diff
@@ -17,204 +13,101 @@
 ;; sp's default rules are obnoxious, so disable them
 (provide 'smartparens-latex)
 
-(after! tex
-  ;; Set some varibles to fontify common LaTeX commands.
+(def-package! tex
+  :mode ("\\.tex\\'" . TeX-latex-mode)
+  :hook (TeX-mode . visual-line-mode)
```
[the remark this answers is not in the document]
patrl: Thanks, fixed in the latest commit!

modules/lang/latex/config.el
hlissner: Ah! My mistake. I mixed up +viewers.el and +ref.el!

modules/lang/latex/config.el
hlissner: First draft:

```elisp
(defvar +latex--company-backends nil)

(def-package! company-auctex
  :defer t
  :init
  (add-to-list '+latex--company-backends 'company-auctex-environment nil #'eq)
  (add-to-list '+latex--company-backends 'company-auctex-macros nil #'eq))

(def-package! company-math
  :defer t
  :init
  (add-to-list '+latex--company-backends 'company-math-symbols-latex nil #'eq))

(when +latex--company-backends
  ;; We can't use the `set-company-backend!' because Auctex reports its
  ;; major-mode as `latex-mode', but uses LaTeX-mode-hook for its mode, which is
  ;; not something `set-company-backend!' anticipates (and shouldn't have to!)
  (add-hook! 'LaTeX-mode-hook
    (add-to-list (make-local-variable 'company-backends)
                 +latex--company-backends)))
```

patrl: Yeah, I agree. This just seemed like the most straightforward way of grouping and ordering the backends in the way that I wanted. I'm sure there's a better way.

patrl: Thanks! I'll test this now.

hlissner:
Sorry for taking so long to get to this! I'll write a wrapper company backend to make up for the removed unicode-math functionality soon.

In any case, thanks again for all your hard work! This was a monster of a module to refactor!

PR #755, opened 2018-07-13
Improvements for doom.cmd script

jseba:
I was trying out the batch script to run the replacement management commands on my Windows dev machine and ran across a couple issues.

### DIFF
--- a/bin/doom.cmd
+++ b/bin/doom.cmd
@@ -5,7 +5,7 @@ PUSHD "%~dp0" >NUL
 
 IF "%1"=="run" (
    SHIFT
-   emacs -Q %* -l ..\init.el -f "doom|run-all-startup-hooks"
+   start runemacs -Q %* -l ..\init.el -f "doom|run-all-startup-hooks"
 ) ELSE (
    emacs --quick --script .\doom -- %*
 )
--- a/bin/doom.cmd
+++ b/bin/doom.cmd
@@ -3,11 +3,11 @@
 @ECHO OFF
 PUSHD "%~dp0" >NUL
 
-IF %1=="run" (
+IF "%1"=="run" (
    SHIFT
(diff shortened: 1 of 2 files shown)


hlissner:
Thanks for this! I hadn't gotten around to testing bin/doom.cmd on a windows machine.

Though, I'm not sure I understand the reasoning behind the `s/boundp/bound-and-true-p` change. Emacs 27 loads early-init-file automatically, that's why it isn't necessary to load it if the variable exists (which suggests they are running 27+).

jseba:
It only loads `early-init-file` if `init-file-user` is non-nil, which both -Q and -q set to nil. If the file was loaded, `early-init-file` should be set to the absolute path to the source file that was loaded. I don't have my Macbook in front of me to test running bin/doom on it, but my Linux container ignores early-init.el when running bin/doom (since it calls Emacs with "-q --no-splash") and `early-init-file` is set to nil. If I start Emacs by just calling it directly, then it is loaded and set to the path "/home/jseba/.emacs.d/early-init.el".

I was looking at the code in Emacs in lisp/startup.el trying to understand what was happening; early-init.el  gets loaded with following (line 1150):
```emacs-lisp
  ;; Load the early init file, if found.
  (load-user-init-file
   (lambda ()
     (expand-file-name
      "early-init"
      (file-name-as-directory
       (concat "~" init-file-user "/.emacs.d")))))
  (setq early-init-file user-init-file)
```
where the work done by `load-user-init-file` is wrapped in `(when init-file-user ...)`.

gjnoonan:
LGTM 👍

hlissner:
Understood. Another question: I know virtually nothing about windows batch scripts, but is there no way to pass "all but the first argument" to runemacs? I'd like people to be able to pass their own arguments to the Emacs binary when using `bin/doom.cmd run`, if possible.

PR #757, opened 2018-07-14
Double quote doom binary location

xeijin:
This allows 'make quickstart' to work on Windows when using GNU make (appreciate Windows isn't officially supported)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-DOOM = bin/doom
+DOOM = "bin/doom"
 MODULES = $(patsubst modules/%/, %, $(sort $(dir $(wildcard modules/*/ modules/*/*/))))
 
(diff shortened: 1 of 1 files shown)


hlissner:
I don't mind accepting this PR, but I use GNU make on Linux and MacOS, and it has never complained about `DOOM` not being quoted. Does GNU make on Windows behave differently?

xeijin:
Yes, this is what I get when not quoted and running `make quickstart` from within `.emacs.d`:

`'bin' is not recognized as an internal or external command,
operable program or batch file.
make: *** [quickstart] Error 1`

Which is making it difficult to script as part of an automated installation.

hlissner:
Understood. Thanks for this!

PR #758, opened 2018-07-16
Intero hlint

patrl:
Adds the `hlint` checker for `intero` users. Note that this used to be the default in `intero`: https://github.com/commercialhaskell/intero/commit/06527fc4c9a5239e7921845f9c7cd4d2783eb030

Note this adds some redundancy, since the `hlint` checker is declared in two places. 

Since `intero` is broken on NixOS, this needs testing!


hlissner:
Looks good! I'll remove the redundancy in a bit. Thanks again!

PR #759, opened 2018-07-16
DONE: improve nix module

patrl:
- [x] Add company completion of NixOS options for `nix-mode`.
- [x] Add support for @jwiegley's `nix-update`
- [x] Make `nix-repl` available
- [x] Add major mode bindings
- [x] Add doctor check for `nix` and `nixfmt` executables

For future investigation:
  - [nix-company.el](https://github.com/NixOS/nix-mode/blob/master/nix-company.el) seems to be broken.
 - [nix-buffer](https://github.com/shlevy/nix-buffer) looks interesting but probably shouldn't be lazily loaded, due to its nature.

### DIFF
--- a/modules/lang/nix/config.el
+++ b/modules/lang/nix/config.el
@@ -1,16 +1,8 @@
-;;; config.el --- description -*- lexical-binding: t; -*-
-
-(def-package! nix-update
-  :commands (nix-update-fetch))
-
-(def-package! nix-repl
-  :commands (nix-repl-show))
-
-(def-package! helm-nixos-options
-  :when (featurep! :completion helm)
-  :commands (helm-nixos-options))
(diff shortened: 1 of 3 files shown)


modules/lang/nix/config.el
```diff
@@ -5,3 +5,22 @@
   :after nix-mode
   :config
   (set-company-backend! 'nix-mode 'company-nixos-options))
+
+(def-package! nix-update
+  :after nix-mode
+  :commands (nix-update-fetch))
+
+(def-package! nix-repl
```
ocharles: Does this need a `(package!)` entry in `packages.el`?

hlissner:
Excellent work! Thanks again!

PR #761, opened 2018-07-18
Haskell quality-of-life improvements

patrl:
- [x] define basic keybindings.
- [x] add `attrap-attrap` for apply suggestion support in `dante-mode` (`intero` has this built in)
- [x] modify haskell build process for intero users so that it uses `stack`

Testing by `intero` users is welcome!

### DIFF
--- a/modules/lang/haskell/+dante.el
+++ b/modules/lang/haskell/+dante.el
@@ -2,4 +2,17 @@
 ;;;###if (featurep! +dante)
 
 (def-package! dante
+  :init
+  (setq dante-load-flags '(;; defaults:
+                           "+c"
+                           "-Wwarn=missing-home-modules"
+                           "-fno-diagnostics-show-caret"
+                           ;; neccessary to make attrap-attrap useful:
+                           "-Wall"
+                           ;; necessary to make company completion useful:
+                           "-fdefer-typed-holes"
+                           "-fdefer-type-errors"))
   :hook (haskell-mode . dante-mode))
+
+(def-package! attrap
+  :commands (attrap-attrap))
--- a/modules/lang/haskell/+dante.el
+++ b/modules/lang/haskell/+dante.el
@@ -18,4 +18,12 @@
   :hook (haskell-mode . dante-mode)
   :config
   (when (featurep! :feature syntax-checker)
-    (flycheck-add-next-checker 'haskell-dante '(warning . haskell-hlint))))
+    (flycheck-add-next-checker 'haskell-dante '(warning . haskell-hlint)))
+
+  (map! :map dante-mode-map
+        :localleader
+        :n "t" #'dante-type-at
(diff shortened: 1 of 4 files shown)


modules/lang/haskell/config.el
```diff
@@ -10,10 +10,39 @@
...
+(after! haskell-mode
+  (map! :map haskell-mode-map
+        :localleader
+        ;; this is set to use cabal for dante users and stack for intero users:
+        :n "b" #'haskell-process-cabal-build
+        :n "c" #'haskell-cabal-visit-file
+        :n "p" #'hindent-reformat-buffer
+        :v "p" #'hindent-reformat-region))
+
+(after! intero-mode
```
hlissner: Could you move the intero and dante keybinds into the +intero and +dante modules? (And into their respective def-package! blocks, to reduce the number of `after!`'s)

modules/lang/haskell/config.el
```diff
@@ -10,10 +10,39 @@
 (def-package! hindent
   :hook (haskell-mode . hindent-mode))
 
+(after! haskell-mode
```
hlissner: Could these haskell keybinds be moved onto the end of the `(after! haskell-mode ...)` block later in the same file? Each `after!` block adds an entry to `after-load-alist`, which I try to avoid if it is inexpensive to do so.

modules/lang/haskell/+dante.el
```diff
@@ -1,7 +1,20 @@
 ;;; lang/haskell/+dante.el -*- lexical-binding: t; -*-
 ;;;###if (featurep! +dante)
 
+(def-package! attrap
```
hlissner: This `def-package!` block isn't necessary, because attrap autoloads attrap-attrap for us.

modules/lang/haskell/+dante.el
```diff
@@ -1,7 +1,20 @@
...
+  :commands (attrap-attrap))
+
 (def-package! dante
+  :init
```
hlissner: Could you move the :init block after :hook? A convention Doom follows is to have the properties that control how the package is loaded further up (like `:when` `:hook`, `:commands`, `:defer`, etc), and properties responsible for configuring the package further down (`:preface`, `:init` and `:config`, in that order).

modules/lang/haskell/config.el
patrl: Thanks! fixed in the latest commits.

modules/lang/haskell/config.el
patrl: Thanks! fixed in the latest commits.

modules/lang/haskell/+dante.el
patrl: Thanks! fixed in the latest commits.

hlissner:
Excellent work! Thanks again!

PR #762, opened 2018-07-18
DONE: Added configuration for editing fish shell files https://fishshell.com/

Brettm12345:


hlissner:
Thanks for this PR! I'm happy to merge it, but due to the size of this module, would you mind moving this into `:lang sh` instead, guarded by a `+fish` flag?

Additionally, both of these are unnecessary:

```sh
:mode "\\.fish$"
:interpreter "fish"
```

As the fish-mode package already does this in its autoloads file. Then the def-package! block can be reduced to:

```elisp
(def-package! fish-mode
  :when (featurep! +fish)
  :defer t
  :config
  (add-hook! fish-mode
    (add-hook 'before-save-hook #'fish_indent-before-save)))
```

Brettm12345:
Thanks. I moved it into `:lang sh +fish`

hlissner:
Nice work (and quick)! Thanks again!

PR #763, opened 2018-07-19
Add support for the dhall configuration language

patrl:
This collapses support for the [dhall configuration language](https://github.com/dhall-lang/dhall-lang) under the `data` module.


ocharles:
One more local module I can delete now :) Tanks @patrl !

PR #764, opened 2018-07-19
minor haskell tweaks

patrl:
- [x] add a hook for `subword-mode` to make navigating with camelCase a breeze, since camelCase is a convention for functions and variables in haskell.
- [x] add support for code-block collapsing via `haskell-collapse-mode` (built into `haskell-mode`)
- [x] enable some benign and useful features [as recommended in the haskell-mode manual](http://haskell.github.io/haskell-mode/manual/latest/Interactive-Haskell.html#Interactive-Haskell).
- [x] allow flycheck to handle error popups if `:feature syntax-checker` is enabled.

### DIFF
--- a/modules/lang/haskell/config.el
+++ b/modules/lang/haskell/config.el
@@ -11,8 +11,11 @@
   :hook (haskell-mode . hindent-mode))
 
 (after! haskell-mode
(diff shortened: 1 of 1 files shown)


hlissner:
Good work and thanks again!

PR #770, opened 2018-07-29
Highlight-numbers on every prog-mode

ar1a:
Seems inconsistent only added to a few select packages, this will just add it to all progmode stuff


hlissner:
Many major modes will benefit from this, but it will need more testing, particularly in modes that may have their own number/literal fontification (like C++).

In any case, thanks again!

PR #771, opened 2018-07-29
Preliminary rainbow-identifiers integration

ar1a:
add +better to syntax-checker

https://github.com/Fanael/rainbow-identifiers


hlissner:
This does not belong in `:feature syntax-checker`, and `+better` is a terribly ambiguous name for a flag. A `:ui rainbow` module would be a better fit; we could keep all rainbow-* plugins there.

PR #772, opened 2018-07-29
Add Racket lang

mdssjc:


hlissner:
I made a few adjustments (refer to the commit message for 238e37a for details), but this looks great! Thank you for your hard work!

hlissner:
@mdssjc Oh, I forgot to ask: is there a reason why `racket-smart-open-bracket-enable` can't be set directly, rather than in a hook?

mdssjc:
@hlissner No, it can be done directly.

PR #784, opened 2018-08-06
fix helm execute persistent action

duog:
With (helm +childframe) in doom! section, without this patch
execute-persistent-action opens an extra non-child frame. This
doesn't play nicely with my tiling window manager.

This patch is taken from helm-posframe, and seems to fix it.


hlissner:
Good catch, thanks!

PR #785, opened 2018-08-06
Fix treemacs typo

duog:


hlissner:
This was intentional, not a typo. In the most recent version of treemacs, this function's name was changed from `treemacs--current-visibility` to `treemacs-current-visibility`. 

Please update treemacs with `bin/doom update`.

duog:
Oops. Sorry for the spam My alias is now `YES=1 doom update && doom refresh && emacsclient -e '(doom/reload)'`. What can go wrong.

PR #786, opened 2018-08-07
Fix advice for `undo-tree-make-history-save-file-name'

jseba:
`:filter-return' expects a function to be given. If neither "zstd" nor "gzip" are found on the executable path, then it gets given nil instead, which results in sadness when trying to save files.

This is just the quick-and-dirty fix I made locally to fix the problem of not having either xz or gzip in the path (Windows problem).

### DIFF
--- a/core/core-editor.el
+++ b/core/core-editor.el
@@ -221,10 +221,11 @@ savehist file."
         `(("." . ,(concat doom-cache-dir "undo-tree-hist/"))))
   (global-undo-tree-mode +1)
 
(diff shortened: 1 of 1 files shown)


hlissner:
You've certainly caught an issue, but this solution doesn't work. The `cond` statement will either return a string or nil, neither of which are valid. I also think this could be more DRY.

Thanks for bringing this to my attention, in any case. I'll have a fix for this up in a couple minutes.

PR #790, opened 2018-08-08
Configurable sync backends for NotMuch

aidangilmore:
This is a simple fix to allow users to be able to specify the sync backend they
wish to use for NotMuch.


hlissner:
Excellent work, thanks!

PR #794, opened 2018-08-09
Fixed ess main load point

vikigenius:
The ess loading point has been changed from ess-site to ess-mode as described here: https://github.com/emacs-ess/ESS/pull/404 and here https://github.com/emacs-ess/ESS/issues/589.

Additionally [emacs-smart-equals](https://github.com/genovese/ess-smart-equals) package is broken because ess has made ess-S-assign obsolete and an alternative called ess-assign-list introduced. I have removed the package and disabled the functionality of smart assignment.

I will probably try to create a pull request to fix the emacs-smart-equals package, but it doesn't seem to be active. so no guarantee that it will be accepted.


hlissner:
Great work. Thank you for the help!

PR #795, opened 2018-08-10
added refactor shortcut to clojure

hackeryarn:
`clj-refactor` provides a great `hydra` for working with the refactors. This is just a small change to integrate a shortcut for activating the `hydra`.


hackeryarn:
That was my first thought too but <kbd>r</kbd> is already bound to `cider-eval-region`. 

I am open to rebinding either command if you have a suggestion.

hackeryarn:
Maybe we should move all the eval commands under <kbd>e</kbd> leader.

mattly:
I've always thought it was a bit odd that the Cider eval commands were under the minor-mode leader instead of `SPC c e` like with elisp. I mean, it makes sense, CIder is a whole 'nother thing, but it's inconsistent.

hackeryarn:
@mattly It tripped me up a few times. I've ran `SPC c e` and get a huge stack trace. Then I realize that I evaluated a my Clojure and elisp :man_facepalming:.

I like the idea of overwriting the elisp eval command when you are in another lisp mode. Let's see what @hlissner thinks about this approach.

hlissner:
Sorry for the late reply. I'm sick and slowly recovering!

> It tripped me up a few times. I've ran SPC c e and got a huge stack trace. Then, after a minute of being puzzled, I realized I evaluated Clojure as elisp man_facepalming.

That's a bug. `SPC c e` runs `+eval/buffer` in normal mode and `+eval/region` in visual mode. Both these commands will use whatever "evaluator" best fits the current major mode. It should only evaluate code as elisp if you are in `emacs-lisp-mode`.

The default evaluator for clojure-mode is `cider-eval-region`, as set by this line in lang/clojure/config.el:

```elisp
(set-eval-handler! 'clojure-mode #'cider-eval-region)
```

Is that not the case? At the very least, you should get either a `No linked CIDER sessions` or `not found [clojure] language information` error message, rather than trying to run clojure code as elisp.


> I am open to rebinding either command if you have a suggestion.

I'd like to rebind them. I'd like to reserve `r` for refactoring commands and `h` for minor mode-specific documentation commands.

> Maybe we should move all the eval commands under e leader.

Agreed. However, because I'm unfamiliar with cider (and clojure, for that matter). I'd appreciate a PR.

Also, fyi, I've made a few adjustments to the lang/clojure module, so please update before you make your changes.

hackeryarn:
@hlissner I updated the key binding for refactor. 

The eval behavior must have been fixed in a previous PR. `SPC c e` works as you described in Clojure now. With this functionality, we don't need the <kdb>e</kbd> leader binding for Clojure eval commands.

On a side note. I am more than happy to take on maintenance of the Clojure mode. It must be tough for you to maintain a mode you don't regularly use.

hlissner:
@hackeryarn Sorry for the terrible late response!

> I am more than happy to take on maintenance of the Clojure mode. It must be tough for you to maintain a mode you don't regularly use.

That would be fantastic! It certainly is difficult and I really appreciate your help.

PR #799, opened 2018-08-12
Fixes Racket

mdssjc:


hlissner:
```elisp
(company-capf company-abbrev company-dabbrev-code company-dabbrev company-files company-yasnippet)
```

Having all these company backends seems unnecessary. Can we reduce them or remove them entirely? The doom default for prog-modes is `(company-capf company-yasnippet)`, so we may not need to set anything for racket-mode at all.

1. Having both `company-dabbrev-code` and `company-dabbrev` means a lot of duplicates. These two backends are also naive, and only add to the noise. They should be used as last-resort backends in the event that `company-capf` isn't sufficient.

2. This will be slow and memory intensive, especially for as-you-type code completion, and in larger projects.

3. All of these backends are available through vim omnicompletion keybinds, for example:

  + `C-x C-f` for `company-files`
  + `C-x C-s` for `company-yasnippet`
  + `C-x C-n` and `C-x C-p` for `company-dabbrev-code`

mdssjc:
I agree!
The company-capf is the main one for the Racket engine, others are optional extras.

hlissner:
Alright, I've made the necessary adjustments. Thanks again for your help!

PR #802, opened 2018-08-14
Fix running make from inside doom

benj02:
The `term.el` package defines an environment variable `EMACS` inside its shell process, containing the Emacs and term.el version, in a string that looks like this: `26.1 (term:0.96)`. This interferes with the `bin/doom` command, which expects that environment variable to be a path to an Emacs binary. Trying to run make inside a doom terminal thus gives you this error:

```
Emacs isn't installed
make: *** [Makefile:5: all] Error 1
```

This simple fix just checks if `$EMACS` looks like a term version string, and ignores it if so.


hlissner:
Ah, a complete oversight on my part. Thanks for addressing this!

PR #804, opened 2018-08-15
DONE: add coq module

patrl:
Adds basic support for the coq proof assistant via proof general (sadly, not packaged on elpa/melpa) and company-coq.


hlissner:
Thanks again for your help!

PR #821, opened 2018-08-21
:lang agda (done)

patrl:
This PR adds basic support for the agda language. Since `agda2-mode` isn't packaged on (m)elpa, but instead accompanies each agda release, this module automatically finds the location of the emacs support files via the built-in `agda-mode locate` command. If the user doesn't have agda installed, this naturally won't work.

A couple of modifications are made to upstream `agda-mode`:
- local leader keybindings
- colorscheme agnostic syntax highlighting, mostly based on haskell and idris

### DIFF
--- a/modules/lang/agda/config.el
+++ b/modules/lang/agda/config.el
@@ -1,7 +1,7 @@
 ;;; lang/agda/config.el -*- lexical-binding: t; -*-
 
 (defvar +agda-dir
-  (when (executable-find "adga-mode")
+  (when (executable-find "agda-mode")
     (file-name-directory (shell-command-to-string "agda-mode locate")))
   "TODO")
 
--- a/modules/lang/agda/config.el
+++ b/modules/lang/agda/config.el
@@ -2,10 +2,18 @@
 
 (defvar +agda-dir
   (when (executable-find "agda-mode")
-    (file-name-directory (shell-command-to-string "agda-mode locate")))
-  "TODO")
+    (file-name-directory (shell-command-to-string "agda-mode locate"))))
 
+(def-package! agda-input
+  :load-path +agda-dir)
 
-(def-package! agda2
-  :load-path +agda-dir
-  :defer t)
+(def-package! agda2-mode
+  :mode "\\.agda\\'"
+  :after agda-input
+  :config
+  (map! :map agda2-mode-map
+                :localleader
+                :n "l" #'agda2-load
+                :n "d" #'agda2-infer-type-maybe-toplevel
+                :n "o" #'agda2-module-contents-maybe-toplevel
+                :n "n" #'agda2-compute-normalised-maybe-toplevel))
--- a/modules/lang/agda/config.el
+++ b/modules/lang/agda/config.el
@@ -11,8 +11,7 @@
   :mode "\\.agda\\'"
   :after agda-input
   :init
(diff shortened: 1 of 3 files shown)


hlissner:
Could you link me to agda-mode's project? I'd like a closer look at its source before approving this PR.

hlissner:
I'm down for that. I generally prefer to merge modules where it makes sense, but I leave the final decision to you. I'm unfamiliar with these tools.

patrl:
I thought about this a little more, and decided that it makes sense to have a distinct agda module. It has its own [textbook](https://dl.acm.org/citation.cfm?id=2841316) after all, and fwiw spacemacs gives it a distinct layer.

That said, please hold off on merging this until I remove the *in-progress* flag. I noticed some issues with syntax highlighting that should be addressed before this gets merged.

patrl:
I'm happy with this for now. I've been using it in my own work for a week or so.

modules/lang/agda/config.el
```diff
@@ -0,0 +1,67 @@
...
+(defvar +agda-dir
+  (when (executable-find "agda-mode")
+    (file-name-directory (shell-command-to-string "agda-mode locate"))))
+
+(def-package! agda-input
+  :load-path +agda-dir)
+
+(def-package! agda2-mode
+  :mode "\\.agda\\'"
+  :after agda-input
```
hlissner: Because `agda-input` is eager-loaded, this will cause adga2-mode to be eager-loaded too. Is `:mode "\\.agda\\'"` not enough?

patrl: Yep you're absolutely right. This slipped in after I made some other changes, will change in the next commit.

I meant to ask you actually whether you think it's a good idea to eagerly load the input mode - it's very useful for unicode math input more generally, and an improvement on the tex input mode imo.

modules/lang/agda/config.el
```diff
@@ -0,0 +1,67 @@
...
+  :init
+  ;; make syntax-highlighting more consistent with other major modes
+   (progn
+        (mapc
+         (lambda (x) (add-to-list 'face-remapping-alist x))
```
hlissner: These are added to the global face-remapping-alist, which means they're active everywhere, even in non-agda2-mode buffers. I'd like to avoid this, either by moving it to a hook or leaving it to themes to handle.

The other problem is: if a theme _does_ provide better defaults for agda2-mode, then these will overwrite them. It's difficult to detect such a case. Frankly, it should be the responsibility of the package to set sane face defaults in the first place. Maybe it can be PR'ed upstream?

modules/lang/agda/config.el
hlissner: Another problem occurred to me, if the agda-mode executable doesn't exist, then `:load-path +agda-dir` will cause problems, so it should be:

```elisp
(def-package! agda2
  :when +agda-dir
  :load-path +agda-dir)
```

modules/lang/agda/config.el
patrl: You're absolutely right, this really should be fixed upstream. Will submit a PR there and remove this for now.

hlissner:
Looks good! Thanks again for your help!

Also, I've added your agda2-mode faces to doom-themes.

patrl:
Awesome, thanks for doing that!

PR #822, opened 2018-08-21
:lang idris

patrl:
Adds basic support for the [idris](https://www.idris-lang.org/) programming language.


hlissner:
Looks good! You've inspired me to check out idris, so I look forward to giving this module a spin.

Thanks again!

PR #830, opened 2018-08-24
helm-buffers-list -> helm-mini

FirstLoveLife:
helm-mini is a better version of helm-buffers-list: https://github.com/emacs-helm/helm/wiki/Buffers:
> The helm-mini command displays buffers and **recently visited files**. It may be more useful than helm-buffers-list.

> Tip
> Use helm-mini instead of helm-buffers-list.

### DIFF
--- a/modules/completion/helm/config.el
+++ b/modules/completion/helm/config.el
@@ -51,7 +51,7 @@ be negative.")
     [remap imenu]                     #'helm-semantic-or-imenu
     [remap noop-show-kill-ring]       #'helm-show-kill-ring
     [remap persp-switch-to-buffer]    #'+helm/workspace-mini
-    [remap switch-to-buffer]          #'helm-buffers-list
+    [remap switch-to-buffer]          #'helm-mini
     [remap projectile-find-file]      #'+helm/projectile-find-file
     [remap projectile-recentf]        #'helm-projectile-recentf
     [remap projectile-switch-project] #'helm-projectile-switch-project


hlissner:
Side question: do you use the `:feature workspaces` module?

`persp-switch-to-buffer` is remapped to `+helm/workspace-mini`, which uses `helm-mini` (whose buffer-list is restricted to the current workspace). This is meant to be the defacto method of switching buffers (it's bound to `SPC b b` and `SPC ,`).

In this case, `switch-to-buffer` is intentionally remapped to `helm-buffers-list`, rather than `helm-mini`, to serve a very explicit purpose. It is assumed that, if you're calling switch-to-buffer, then you have a specific goal in mind (because otherwise you'd be using the other command); to find a specific, open  and/or buried buffer, possibly in another workspace. In that case, recentf is just noise.

If you _don't_ use `:feature workspaces`, then this flow breaks down. That's a bug that should be addressed, but in that case, I'd rather it be remapped conditionally. e.g.

```elisp
(define-key! 'global [remap switch-to-buffer]
  (if (featurep! :feature workspaces)
      #'helm-buffers-list
    #'helm-mini))
```

FirstLoveLife:
> Side question: do you use the :feature workspaces module?

Yes. Nice to know <kbd>SPC b b</kbd> and <kbd>SPC </kbd> is bound to `+helm/workspace-mini`. In the last seveal(two or three) days, to learn some emacs's core concepts, I go back to vanilla emacs and download package manually, use require/provide manually, get familiar with emacs's edit keybinding(the first thing I do is install doom theme and doom modeline, haha). I bound <kbd>C-x b</kbd> with `helm-mini`. Then, I realized many emacs users, like me, also use the built-in keybinding like <kbd>C-x b</kbd> and bind to functions like helm-mini provide by helm/ivy. 

> It is assumed that, if you're calling switch-to-buffer, then you have a specific goal in mind (because otherwise you'd be using the other command); to find a specific, open and/or buried buffer, possibly in another workspace. In that case, recentf is just noise.

Good point. You persuade me. 

BTW, I give [eyebrowse](https://github.com/wasamasa/eyebrowse) a try, <kbd>M-n(n: 1~9)</kbd> can create workspace dynamically without invoke function like `+workspaces/new`. I think this feature is important and didn't found alternative in doom's built-in workspace. My current workround is:
```
(defun name-of-recursive-function (fun n)
  (when (> n 0)
    (funcall fun)
    (name-of-recursive-function
     fun (- n 1))))


(run-with-idle-timer
 0.1 nil
 #'(lambda ()
     (name-of-recursive-function '+workspace/new 8)))
```

it will be nice If doom can also have this feature.

hlissner:
> M-n(n: 1~9) can create workspace dynamically without invoke function like +workspaces/new

Yikes, I hadn't anticipated that use case. I typically create a workspace when I need them.

However, I am considering switching persp-mode for something else. Eyebrowse is a great candidate. It'll have to wait until 2.1.1 or 2.1.2 though.

Anyhow, your workaround has inspired me to look for a better solution, and I found this:

```elisp
(save-window-excursion
  (delete-other-windows)
  (switch-to-buffer (doom-fallback-buffer))
  (require 'persp-mode)
  (let ((wconf (funcall persp-window-state-get-function (selected-frame))))
    (dotimes (i 3)
      (let ((persp (persp-add-new (format "#%d" (1+ i)))))
        (setf (persp-window-conf persp) wconf)))))
```

This should be instant, compared to `+workspace/new`. If it works for you, I'll replace the internals of `+workspace/new` to use this technique.

FirstLoveLife:
Your function works fine in *scratch* + eval-buffer, but once I put it in config.el:
```
Error in private config: config.el, (wrong-type-argument hash-table-p nil)
```

Why cannot it be in config.el?

hlissner:
My bad, forgot that it needs to happen when persp-mode is active:

```lisp
(add-hook! 'persp-mode-hook
  (save-window-excursion
    (delete-other-windows)
    (switch-to-buffer (doom-fallback-buffer))
    (require 'persp-mode)
    (let ((wconf (funcall persp-window-state-get-function (selected-frame))))
      (dotimes (i 9)
        (let ((persp (persp-add-new (format "#%d" (1+ i)))))
          (setf (persp-window-conf persp) wconf))))))
```

FirstLoveLife:
```
There was an unexpected error:
  Error in a Doom startup hook (doom-hook-error)
  +workspaces|init
  (void-function (setf persp-window-conf))
```

I add `(require 'persp-mode)`, then error fixed

FirstLoveLife:
@hlissner Hi, there is another issue: the startup workspace will be the last one by default. To make the first workspace be the default, here is my try:
```
(require 'persp-mode)
(add-hook! 'persp-mode-hook
  (save-window-excursion
    (delete-other-windows)
    (switch-to-buffer (doom-fallback-buffer))
    (require 'persp-mode)
    (let ((wconf (funcall persp-window-state-get-function (selected-frame))))
      (dotimes (i 9)
        (let ((persp (persp-add-new (format "#%d" (1+ i)))))
          (setf (persp-window-conf persp) wconf))))))
(add-hook 'emacs-startup-hook #'(lambda ()(+workspace/switch-to '0)))
```

But the startup become really slow, do you have nicer solution?

PR #835, opened 2018-08-26
:ui deft

patrl:
This PR adds a module for [deft](https://github.com/jrblevin/deft) - an emacs major mode for browsing and filtering notes written in plain text formats.

`.org` is the default note format, and readable filenames are automatically generated based on the note title.

### DIFF
--- a/modules/tools/deft/README.org
+++ b/modules/tools/deft/README.org
@@ -1,4 +1,11 @@
 #+TITLE: :tools deft
 
-* TODO write deft readme
+[[https://jblevins.org/projects/deft/][Deft]] is a major mode for browsing and filtering notes written in plain text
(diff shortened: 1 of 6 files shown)


hlissner:
Looking good, but could you move it to `:ui deft` instead? `:tools` is for modules that interact with external programs and services.

Besides that, evil integration is a bit spotty, but I think we can improve it once I bring general on board.

patrl:
Oh sure! My mistake.

hlissner:
Perfect! Thanks again!

HaoZeke:
The keybindings should be in bindings.

hlissner:
@HaoZeke localleader keybinds (and keybind fixes) are the exception to that.

PR #838, opened 2018-08-27
Company-coq tweaks

patrl:
`company-coq` does more than just provide company completion.

- define lookup handlers based on `company-coq` functions.
- disable company features if company is disabled.


hlissner:
Nice work, and thanks again!

PR #842, opened 2018-08-27
lang/org: Add evil-friendly calendar navigation

myme:
A few bindings for navigating the calendar in the `org-read-date` prompt. The default bindings are using `Shift` and `Shift-Meta` + the arrows, which is having us leave the comfort of the home row.


hlissner:
Looks good! Thank you for your help!

PR #843, opened 2018-08-28
Fix typo in ivy/README.org

benj02:
The wrong commands are listed for `SPC b b`/`SPC b B`. Just a simple typo fix


hlissner:
Good catch! Thanks!

PR #854, opened 2018-08-31
add shortcut to "magit-gitflow-popup" with '%' in magit-mode

shuxiao9058:
as title, thks.


hlissner:
I'm not sure if % is the best keybind, but I'll give it a try for a while. Thanks for your help!

shuxiao9058:
The shortcut was referenced from [spacemacs](https://github.com/syl20bnr/spacemacs/blob/c57dd4870a18ced11694f130d75f3f6472d3eb94/layers/%2Bsource-control/git/packages.el#L211),Thanks.

PR #857, opened 2018-08-31
adds npm-mode for javascript module

qleguennec:


hlissner:
There is a little overlap between +javascript-npm-mode and npm-mode (understandable), but I'll iron those out later. Thank you for your help!

PR #861, opened 2018-09-03
Unicode fallback

patrl:
- old behaviour: all unicode characters are displayed as `doom-unicode-font`, when set.
 - new behaviour: unicode characters are displayed in `doom-font` if available, and `doom-unicode-font` if not.

Note that to get emacs to use `doom-unicode-font` *at all*, I needed the following in my `config.el`:

```elisp
doom-unicode-font (font-spec :name "DejaVu Sans Mono")
```

Rather than the default, which I think was:

```elisp
doom-unicode-font (font-spec :family "DejaVu Sans Mono")
```


hlissner:
This makes much more sense. Thank you for looking into this!

alexx-oo:
This doesn't work for my unicode settings. Before this commit, my unicode setting works. Now, even after the #864 commit, my undicode settings still doesn't work. Here is my font setting:
```(setq doom-font (font-spec :family "Source Code Pro" :size 30)
      doom-variable-pitch-font (font-spec :family "Source Code Pro")
      doom-unicode-font (font-spec :name "Source Han Sans") ;; or :family
      doom-big-font (font-spec :family "Source Code Pro" :size 32))
```

For some Chinese characters, the Emacs shows this font: "  x:-misc-droid sans fallback-medium-r-normal-xxxxxxx". I have used the first fontconfig in this [link](https://wiki.archlinux.org/index.php/Font_Configuration/Chinese_(%E7%AE%80%E4%BD%93%E4%B8%AD%E6%96%87)) on my Archlinux. Does it matter?

Based on my experience, the new behavior is not perfect for some Asian languages because of the character width when `doom-font` and `doom-unicode-font` are not well selected.  At the end we can have some "big" characters with "small" characters.

I am so sorry if I put my comments in wrong places.

sdvcrx:
Same problem here on my ArchLinux

Some related discussion: https://emacs-china.org/t/doom-emacs/6967

alexx-oo:
Sorry for the late replay. After "cleaning" my fonts, my config works (without enabling the unicode module).

PR #864, opened 2018-09-04
Fix unicode regression

patrl:
Setting the charset to `'unicode` caused emacs to segfault. Setting it to `nil` retains failback behaviour.


hlissner:
Good catch, thanks!

PR #867, opened 2018-09-05
proof-general now on melpa

patrl:
Proof-general no longer has to be loaded via quelpa's git fetcher.

https://melpa.org/#/proof-general


hlissner:
Great, thanks!

PR #868, opened 2018-09-06
Add docker module

rynffoll:
Add docker module

https://github.com/Silex/docker.el


Emiller88:
Perhaps add it to init.example.el also

hlissner:
Good work! Thanks for your help and this PR.

I will remove dockerfile-mode from lang/data shortly.

PR #872, opened 2018-09-08
Fix clj-refactor bindings

rynffoll:
Fix `SPC m r` for `clojure-mode`


hlissner:
Thanks for catching this!

PR #873, opened 2018-09-08
Syntax-checking for clojure via joker

mattly:
This is provided as a +feature on clojure because joker is not included
in the standard clojure distribution, and I wouldn't assume a given
clojurist has or wants to have it installed. But I think installing it
is the least-fuss option for syntax-checking clojure/script code in
emacs with flycheck.

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -80,7 +80,7 @@
        :lang
       ;assembly          ; assembly for fun or debugging
       ;(cc +irony +rtags); C/C++/Obj-C madness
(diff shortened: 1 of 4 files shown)


mattly:
Updated to remove the +joker flag

hackeryarn:
You should add instructions in the readme on how to install `joker`. A command to install it would be even better, but that could be a separate PR.

hlissner:
Thanks again!

PR #874, opened 2018-09-09
add anaconda3 home path '/usr/local/anaconda3'

shuxiao9058:


hlissner:
Thanks! By the way, is `/usr/bin/anaconda3` a standard location? It doesn't seem right for it to be in `/usr/bin`.

shuxiao9058:
I don't know it yet, but on macOS with brew the  anaconda3's home location is  ` /usr/local/anaconda3 `.

PR #879, opened 2018-09-10
Improve +fci|disable-when-company-activates

re-mx:
Hello, and sorry for my poor English...

I think this change avoids the error caused when the variable `fci-mode` is void.

Thanks.


hlissner:
Thanks for your help!

PR #891, opened 2018-09-16
Load .elc even for themes

edwintorok:
Even when an `.elc` is preset `load-theme` would attempt to load the `.el` first,
to give the opportunity to inspect the theme.
However we are loading themes automatically from (M)ELPA, and loading a
theme package is no more dangerous than loading a regular package.

So override the search order for themes and atttempt to load the `.elc`
first.
This looks a bit ugly, but improves startup time by ~25ms with the default theme.

### DIFF
--- a/core/core-ui.el
+++ b/core/core-ui.el
@@ -448,6 +448,16 @@ instead). Meant for `kill-buffer-query-functions'."
     (powerline-reset)))
 (advice-add #'load-theme :around #'doom*disable-old-themes-first)
 
+(defun doom*prefer-compiled-theme (orig-fn &rest args)
+  "Make `load-theme' prioritize the byte-compiled theme (if it exists) for a
+moderate boost in startup (or theme switch) time."
+  (cl-letf* ((old-locate-file (symbol-function 'locate-file))
+             ((symbol-function 'locate-file)
+              (lambda (filename path &optional _suffixes predicate)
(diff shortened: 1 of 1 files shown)


hlissner:
Good catch! However, I think your advice casts too wide a net. Rather than affecting all uses of `locate-library`, I'd rather only affect its specific use in `load-theme` with the following:

```elisp
(defun doom*prefer-compiled-theme (orig-fn &rest args)
  "Make `load-theme' prioritize the byte-compiled theme (if it exists) for a
moderate boost in startup (or theme switch) time."
  (cl-letf* ((old-locate-file (symbol-function 'locate-file))
             ((symbol-function 'locate-file)
              (lambda (filename path &optional _suffixes predicate)
                (funcall old-locate-file filename path '("c" "") predicate))))
    (apply orig-fn args)))
(advice-add #'load-theme :around #'doom*prefer-compiled-theme)
```

It's certainly less elegant, but it is more robust.

Also, would you kindly move (and rebase) the advice to core/core-ui.el, beneath [line 464](https://github.com/hlissner/doom-emacs/blob/develop/core/core-ui.el#L464), so as to group the two load-theme hacks together. Thanks!

edwintorok:
Agreed, the smaller the scope when a workaround like this is active the less problems it can cause.

hlissner:
Great work! Thanks again!

PR #896, opened 2018-09-17
Deactivate expansion of abbreviations (ProofGeneral/Coq)

wildsebastian:
This line deactivates the expansion of abbreviations since it is hard to work with Coq in evil mode when this is active. For example, if my cursor is next to the letter L and I hit escape, it gets expanded to: `
Lemma # : #.
  Proof.
     #
  Qed.
`
That makes it hard to work with ProofGeneral.
The solution I found is from a [ProofGeneral Github issue](https://github.com/ProofGeneral/PG/issues/174). I was not sure where to add the line. Apparently it has to be before loading evil-mode.

Let me know if you want me to add it in a different place or just close the PR if you don't want to have this at all.

Thank you for your work. I really enjoy using emacs using doom.

### DIFF
--- a/modules/lang/coq/README.org
+++ b/modules/lang/coq/README.org
@@ -1,3 +1,6 @@
 #+TITLE: :lang coq
 
-This module adds [[https://coq.inria.fr][coq]] support, powered by [[https://proofgeneral.github.io][Proof General]], with code completion via [[https://github.com/cpitclaudel/company-coq][company-coq]].
+This module adds [[https://coq.inria.fr][coq]] support, powered by [[https://proofgeneral.github.io][Proof General]].
+
++ Code completion ([[https://github.com/cpitclaudel/company-coq][company-coq]])
++ [[https://github.com/hlissner/emacs-snippets/tree/master/coq-mode][Snippets]]
--- a/modules/lang/coq/config.el
+++ b/modules/lang/coq/config.el
@@ -3,6 +3,8 @@
 ;; `coq'
 (setq proof-electric-terminator-enable t)
 
+(setq coq-mode-abbrev-table '())
+
 (after! company-coq
   (set-lookup-handlers! 'company-coq-mode
     :definition #'company-coq-jump-to-definition


hlissner:
I'm inclined to refuse this PR on the basis that it is default vim behavior (in fact, I'd rather coq used snippets rather than abbreviations), but this is obviously undesirable behavior. I want to avoid disabling `evil-want-abbrev-expand-on-insert-exit` globally, so instead, give this a try:

```elisp
;; add this to your private config.el
(setq-hook! 'coq-mode-hook
  evil-insert-state-exit-hook (remq 'expand-abbrev evil-insert-state-exit-hook))
```

(A warning: I'm only guessing that `coq-mode-hook` is the correct hook -- I don't use coq)

wildsebastian:
That makes sense.
I did not get your suggestion to work (I am a beginner when it comes to elisp and emacs as you might have guessed ;) ), but I found a different solution. Emptying the abbreviation table for coq works as well.
`(setq coq-mode-abbrev-table '())`.
If you want I can add this to the coq module and update the PR, otherwise you can just close it.
Thank you for your feedback. Sorry for the PR, next time I will open an issue first and discuss it there.

hlissner:
> Sorry for the PR, next time I will open an issue first and discuss it there.

No apology necessary. Here's as good a place as any to discuss the matter!

> `(setq coq-mode-abbrev-table '())`

I'm not sure I want to do that either (because coq users may be expecting them). Decisions decisions!

I think the ideal solution would be to extract [all of coq-mode's abbreviations](https://github.com/ProofGeneral/PG/blob/733cd24a7368ee186884da488da0f59bbedb627e/coq/coq-syntax.el#L131-L947) into yasnippet snippets. I can throw together a script that'll do that over the weekend.

wildsebastian:
Sounds like a plan.
I found on the [company-coq readme](https://github.com/cpitclaudel/company-coq) that they use yasnippet somehow.
> Selecting a completion often inserts a snippet with holes at the current point (company-coq uses yasnippet as the snippet backend). You can move between holes by using <tab> and S-<tab>.

I did not try it out so far.
If you could give me some guidance on how you would do it, let me know.
I would be more than happy to do the work and extract the abbreviations into yasnippet snippets.

hlissner:
>Selecting a completion often inserts a snippet with holes at the current point (company-coq uses yasnippet as the snippet backend). You can move between holes by using <tab> and S-<tab>.

company-coq uses yasnippet for a different purpose; to make editable regions in the code they expand completions (which you can tab/shifttab between).

What I intend to do is a little different. If you head over to [hlissner/emacs-snippets](https://github.com/hlissner/emacs-snippets), you'll find my Emacs snippets library (and the default one for Doom, if you have `:config (default +snippets)` enabled). Each subdirectory there is an example of a snippet library for a major mode.

I intend to add one for `coq-mode`, and convert [its abbreviations](https://github.com/ProofGeneral/PG/blob/733cd24a7368ee186884da488da0f59bbedb627e/coq/coq-syntax.el#L131-L947) into snippets. Snippets can be expanded voluntarily with TAB after typing the trigger. They also offer more flexibility, in case want better interactivity (e.g. tab/shifttabing between fields, running code inside, etc).

As explained [in coq-db.el](https://github.com/ProofGeneral/PG/blob/733cd24a7368ee186884da488da0f59bbedb627e/coq/coq-db.el#L33), abbreviations follow this format:

```
\(MENUNAME ABBREV INSERT STATECH KWREG INSERT-FUN HIDE)
```

We only need MENUNAME, ABBREV, and INSERT. If any of these are missing (and there are quite a few), we can skip them (they're used for something else). Here are three examples:

1. `("assert by" "assb" "assert ( # : # ) by #" t "assert")` -> `{snippets}/coq-mode/assert_by` ->

```
# -*- mode: snippet -*-
# name: assert by
# key: assb
# --
assert ( $1 : $2 ) by $0
```

2. `("Extraction (in a file)" "extrf" "Extraction \"@{file}\" @{id}." nil)` -> `{snippets}/coq-mode/extraction_in_a_file` ->

```
# -*- mode: snippet -*-
# name: Extraction (in a file)
# key: extrf
# --
Extraction "${1:file}" ${2:id}.
```

3. `("Scheme" "sc" "Scheme @{name} := #." t "Scheme")` -> `{snippets}/coq-mode/scheme` ->

```
# -*- mode: snippet -*-
# name: Scheme
# key: sc
# --
Scheme ${1:name} := $2.
```

wildsebastian:
Cool, sounds good.
I don't know if I can work on it this weekend, but I will start this week.

hlissner:
Fantastic! Thanks a lot for your help!

wildsebastian:
Cool! You are welcome! 🙂

Ptival:
@wildsebastian Is it supposed to work out of the box now? I just started using PG again with Doom, and I am encountering the unwanted (and extremely obnoxious) behavior! (on develop branch)

wildsebastian:
@Ptival It works for me. Let me know if you encounter the behavior again.

PR #899, opened 2018-09-19
Emacs27+: reduce number of GCs on startup from 4 to 1

edwintorok:
describe-variables `gcs-done`, it used to be 4, now it is 1.

Unfortunately `emacs-snapshot` segfaults when compiling doom twice, so I'm  back to Emacs26.


hlissner:
Nice work, and thanks for the PR!

PR #900, opened 2018-09-19
defconst compile time evaluation

edwintorok:
I was learning more about Emacs by looking at the bytecode it generates for simple constructs, and discovered that `defconst` was not optimized. Here's a macro that can optimize it.
Don't expect visible changes in startup time from this, but it is nice for documentation purposes (and maybe someone else *can* figure how to use this to optimize startup time).

For now I'm quite happy with startup time at 0.784s on Emacs26 (and Emacs27  would further reduce that, but I'll have to wait until it is released, it is not stable yet).

### DIFF
--- a/core/core-os.el
+++ b/core/core-os.el
@@ -20,8 +20,8 @@
   (when (featurep 'exec-path-from-shell)
     `(exec-path-from-shell-copy-envs ,@vars)))
 
-(cond (IS-MAC
-       (setq mac-command-modifier 'meta
+(macroexpand (eval-when-compile (cond (IS-MAC
+       `(progn (setq mac-command-modifier 'meta
              mac-option-modifier  'alt
              ;; sane trackpad/mouse scroll settings
              mac-redisplay-dont-reset-vscroll t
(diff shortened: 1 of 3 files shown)


hlissner:
Sorry to say, I cannot accept this PR. 

I am aware that Elisp doesn't enforce immutability (there is no low-cost way of policing constness in Elisp), however, my use of `defvar` and `defconst` are not accidental.

1. `defvar` does not mutate variables if they are already defined. `defconst` does. I want the ability to define these variables before core/core.el is loaded. The bin/doom script does exactly this, allowing users to operate on _other_ doom configs (or other private configs) with the `-e PATH` and `-p PATH` options. This may even break our compatability with [Chemacs](https://github.com/plexus/chemacs).

2. An `eval-when-compile` alone would do, and even then, it is premature for all but perhaps `doom-emacs-dir` and `doom-private-dir` (in fact, I already use it for `doom-private-dir`, and I intend to use it for `doom-emacs-dir` once I use `file-truename` on it).

3. If we inline `doom-private-dir`'s value, we break bin/doom's `-e` option and the `DOOMDIR` envvar for byte-compiling users. In any case, `const!` imposes unnecessary overhead, especially so that it tries to replace native forms. In fact, looking into this has made me realize that `DOOMDIR` support is already broken for byte-compiling users, which I will fix soon.

4. `cond` is one of the most efficient language constructs in elisp (especially Emacs 26+, now that it uses a jump tables under the hood). By introducing this change to core-os.el, we imposed the overhead of `macroexpand`, `eval-when-compile`, and interpolating values into lists, onto users that don't byte-compile (which is Doom's priority), all for an imperceptible benefit to the minority of users that do. The net-benefit for relying solely on cond is far greater.

edwintorok:
Thanks for the detailed explanation, I thought this might be the case which is why I sent this in a separate PR. I'm quite happy with my current startup-time, I think I'll look at missing functionality compared to Spacemacs next (e.g. the ocaml module should know about a few more packages/tools and keybindings).

PR #904, opened 2018-09-22
Normalize 1 -> t

ar1a:
Cleanup from old code? Annoyed me as I was scrolling through stuff.

### DIFF
--- a/core/core-editor.el
+++ b/core/core-editor.el
@@ -80,7 +80,7 @@ fundamental-mode) for performance sake."
 
 (def-package! server
   :when (display-graphic-p)
-  :defer 1
+  :defer t
   :after-call (pre-command-hook after-find-file)
   :config
   (when-let* ((name (getenv "EMACS_SERVER_NAME")))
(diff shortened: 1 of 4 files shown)


core/core-editor.el
```diff
@@ -129,7 +129,11 @@ savehist file."
 
 (def-package! recentf
   ;; Keep track of recently opened files
+<<<<<<< HEAD
```
ar1a: sigh

hlissner:
These were intentional. `:defer 1` instructs the package to autoload after 1 second of idle time, but perhaps the new `:defer-incrementally` would be better, especially for helm.

ar1a:
ah, will org get that done to it as well?

hlissner:
[It already has!](https://github.com/hlissner/doom-emacs/blob/develop/modules/lang/org/config.el#L28-L31) Same with [magit](https://github.com/hlissner/doom-emacs/blob/develop/modules/tools/magit/config.el#L19).

PR #905, opened 2018-09-22
OCaml module improvements

edwintorok:
See the included [README.org](https://github.com/edwintorok/doom-emacs/blob/ocaml/modules/lang/ocaml/README.org)

In OCaml, `opam` is a source-based package manager that also installs `.el` files along with tools that have editor support. I preferred to load these from opam's `$(opam config var share)/emacs/site-lisp` directory instead of MELPA because some of these are not present on MELPA (e.g. `ocamlformat`), and others are outdated on MELPA (e.g. merlin is fetched from a several month outdated `beta` branch, the release version is much newer).
The only exception is tuareg-mode (which needs no executable), and `flycheck-ocaml` (which only exists on MELPA).

The disadvantage is that these were not byte-compiled, so I added code in packages.el to do that, but that doesn't fit in nicely with the rest of doom-emacs. I tried to use `:recipe (:fetcher file :path +ocaml-elisp-dir)` instead, but that didn't work because the macro appears to expect a string literal there at macro expansion time. If you have suggestions on how to make this fit more nicely I'd be happy to change it.

The prettify-symbols-alist code also is a bit of a hack because `tuareg-mode` sets it itself, and I didn't want to duplicate it here.

The README is not complete yet, but thought I'd send out the code for some early feedback.

### DIFF
--- a/modules/lang/ocaml/config.el
+++ b/modules/lang/ocaml/config.el
@@ -1,13 +1,83 @@
 ;;; lang/ocaml/config.el -*- lexical-binding: t; -*-
 
-(def-package! tuareg
-  :mode ("\\.ml[4ilpy]?\\'" . tuareg-mode))
+;; def-project-mode!/associate! doesn't work when a
+;; package is lazy loaded, and everything is compiled
 
+(def-package! tuareg
+  :defer t ;; modes set by autoload
+  :config
+  ;; tuareg-mode has the prettify symbols itself
+  (set-pretty-symbols! 'tuareg-mode :alist
+    (append tuareg-prettify-symbols-basic-alist
+            tuareg-prettify-symbols-extra-alist))
+  (setq tuareg-prettify-symbols-full t)
+  ;; Use opam to set environment
+  (setq tuareg-opam-insinuate t)
+  (tuareg-opam-update-env (tuareg-opam-current-compiler))
+  ;; Spell-check comments
+  (when (featurep! :feature spellcheck)
+    (add-hook 'tuareg-mode-hook #'flyspell-prog-mode)))
 
 (def-package! merlin
   :after tuareg
-  :hook (tuareg-mode . merlin-mode)
+  :init
+  (set-lookup-handlers! 'tuareg-mode
+    :definition #'merlin-locate
+    :references #'merlin-occurrences
+    :documentation #'merlin-document)
+  (defun +ocaml|init-merlin ()
+    (when (and (projectile-locate-dominating-file default-directory ".merlin")
+               (executable-find "ocamlmerlin"))
+      (merlin-mode)))
+  (add-hook 'tuareg-mode-hook #'+ocaml|init-merlin)
+
   :config
+  (map! :map tuareg-mode-map
+        :localleader
+        :n "t" #'merlin-type-enclosing
+        :n "a" #'tuareg-find-alternate-file)
   (set-company-backend! 'tuareg-mode 'merlin-company-backend)
-  (after! company
-    (remove-hook 'company-backends 'merlin-company-backend)))
+  (setq merlin-completion-with-doc t))
+
+(def-package! flycheck-ocaml
+  :when (featurep! :feature syntax-checker)
+  :after merlin
+  :config
+  ;; Disable Merlin's own error checking
+  (setq merlin-error-after-save nil)
+  ;; Enable Flycheck checker
+  (flycheck-ocaml-setup))
+
+(def-package! ocp-indent
+  ;; must be careful to always defer this, it has autoloads that adds hooks
+  ;; which we do not want if the executable can't be found
+  :defer t
+  :init
+  (defun +ocaml|init-ocp-indent ()
+    (when (executable-find "ocp-indent")
+      (ocp-setup-indent)))
+  (add-hook 'tuareg-mode-hook #'+ocaml|init-ocp-indent))
+
+(def-package! utop
+  :defer t ;; loaded by hook below
+  :when (featurep! :feature eval)
+  :init
+  (set-repl-handler! 'tuareg-mode #'utop)
+  (set-eval-handler! 'tuareg-mode #'utop-eval-region)
+  (defun +ocaml|init-utop ()
+    (when (executable-find "utop")
+      (utop-minor-mode)))
+  (add-hook 'tuareg-mode-hook #'+ocaml|init-utop))
+
+(def-package! ocamlformat
+  :after tuareg
+  :commands (ocamlformat)
+  :init
+  (set-formatter! 'ocamlformat #'ocamlformat
+    :modes '(caml-mode tuareg-mode))
+  (defun +ocaml|init-ocamlformat ()
+    (setq +format-with 'ocp-indent)
+    (when (and (executable-find "ocamlformat")
+               (locate-dominating-file default-directory ".ocamlformat"))
+      (setq +format-with 'ocamlformat)))
+  (add-hook 'tuareg-mode-hook #'+ocaml|init-ocamlformat))
--- a/modules/lang/ocaml/packages.el
+++ b/modules/lang/ocaml/packages.el
@@ -1,5 +1,30 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; lang/ocaml/packages.el
 
+(when (featurep! +opam-site-lisp)
+  (defvar +ocaml-elisp-dir
+    (when (executable-find "opam")
+      (let ((opam-share (ignore-errors (car (process-lines "opam" "config" "var" "share" "--safe")))))
+        (when (and opam-share (file-directory-p opam-share))
+          (expand-file-name "emacs/site-lisp" opam-share)))))
+
+  (defmacro localpackage! (name)
+    `(package! ,name :recipe (:fetcher file :path ,+ocaml-elisp-dir)))
+
+  (localpackage! opam-site-lisp))
+
 (package! tuareg)
-(package! merlin)
+
+(when (featurep! :feature syntax-checker)
+  (package! flycheck-ocaml))
+
+(unless (featurep! +opam-site-lisp)
+  (package! merlin)
+  (package! ocp-indent)
+  (when (featurep! :feature eval)
+    (package! utop))
+  (when (featurep! :editor format)
+    ;; by default quelpa generated a version 0pre0.20180929.192844, which got parsed into (0 -1 0 ...), which when compared with version nil (0)
+    ;; in package-installed-p always yielded false
+    (package! ocamlformat :recipe (:fetcher github :repo "ocaml-ppx/ocamlformat" :files ("emacs/*.el"))))
+  (package! dune :recipe (:fetcher github :repo "ocaml/dune" :files ("editor-integration/emacs/*.el"))))
--- a/modules/lang/ocaml/README.org
+++ b/modules/lang/ocaml/README.org
@@ -1,0 +1,79 @@
+#+TITLE: :lang ocaml
+
+This module adds  [[https://ocaml.org/][OCaml]] support, powered by [[https://github.com/ocaml/tuareg][tuareg-mode]].
+
++ Code completion, look up documentation, and code navigation ([[https://github.com/ocaml/merlin/wiki/emacs-from-scratch][merlin]])
++ REPL ([[https://github.com/ocaml-community/utop][utop]])
++ Syntax-checking (~merlin~ with [[https://github.com/flycheck/flycheck-ocaml][flycheck-ocaml]])
++ Auto-indentation ([[https://github.com/OCamlPro/ocp-indent][ocp-indent]])
++ Code formatting ([[https://github.com/ocaml-ppx/ocamlformat][ocamlformat]])
++ Dune file format ([[http://dune.build/][dune]])
+
+* Table of Contents :TOC:
+- [[Module Flags][Module Flags]]
+- [[Prerequisites][Prerequisites]]
+- [[Features][Features]]
+- [[Configuration][Configuration]]
+- [[Appendix][Appendix]]
+  - [[Commands][Commands]]
+  - [[Hacks][Hacks]]
+
+* Module Flags
+This module provides the ~+opam-site-lisp~ flag to compile editor support ~.el~
+files from  opam's =site-lisp= directory.
+By default all editor plugins are installed from MELPA/GitHub even if the
+corresponding =opam= package isn't installed.
+If this flag is enabled then you must have all of the packages listed in
+=package.el= installed via =opam=.
+
+To enable this, use:
+#+BEGIN_SRC emacs-lisp
+(doom! :lang (ocaml +opam-site-lisp))
+#+END_SRC
+
+* Prerequisites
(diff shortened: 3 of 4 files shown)


modules/lang/ocaml/config.el
```diff
@@ -1,13 +1,82 @@
...
+
+(def-project-mode! +ocaml-merlin-mode
+  :modes (tuareg-mode)
+  :when (locate-dominating-file default-directory ".merlin")
+  :add-hooks (merlin-mode))
+
+(def-package! ocp-indent
+  :after tuareg
+  :if (executable-find "ocp-indent")
+  :load-path +ocaml-elisp-dir
```
hlissner: All these `:load-path`'s are unnecessary overhead. I suggest doing it separately, at the top, with:

```elisp
(add-to-list 'load-path +ocaml-merlin-mode)
```

modules/lang/ocaml/config.el
```diff
@@ -1,13 +1,82 @@
...
+  ;; Spell-check comments
+  (when (featurep! :feature spell-checking)
+    (add-hook! 'tuareg-mode #'flyspell-prog-mode)))
 
+;; use packages that require a binary from opam
+;; merlin is outdated on MELPA
 (def-package! merlin
   :after tuareg
-  :hook (tuareg-mode . merlin-mode)
+  :if (executable-find "ocamlmerlin")
```
hlissner: You have quite a few `:if (executable-find ...)`'s. I'd rather this check be done more proactively, on the mode's hook, rather than at startup.

For example, you'll see these two hooks in the lang/cc module:

```elisp
  (defun +cc|init-irony-mode ()
    (when (and (memq major-mode '(c-mode c++-mode objc-mode))
               (file-directory-p irony-server-install-prefix))
      (irony-mode +1)))
  (add-hook 'c-mode-common-hook #'+cc|init-irony-mode)

  (defun +cc|init-rtags ()
    "Start an rtags server in c-mode and c++-mode buffers."
    (when (and (memq major-mode '(c-mode c++-mode))
               (rtags-executable-find "rdm"))
      (rtags-start-process-unless-running)))
  (add-hook 'c-mode-common-hook #'+cc|init-rtags)
```

`irony-mode` and `rtags` are checked each time you open a C/C++ buffer, and simply won't activate if they aren't available. I prefer this approach because it doesn't require users to restart Emacs while they're setting up their dev tools (to test that Emacs sees them).

I'd prefer that this be done for merlin, ocp-indent, utop and dune.

modules/lang/ocaml/config.el
```diff
@@ -1,13 +1,82 @@
...
+  :if (executable-find "ocp-indent")
+  :load-path +ocaml-elisp-dir
+  :hook (tuareg-mode . ocp-setup-indent))
+
+(def-package! utop
+  :when (featurep! :feature eval)
+  :after tuareg
+  :if (executable-find "utop")
+  :load-path +ocaml-elisp-dir
+  :hook (tuareg-mode . utop-minor-mode)
```
hlissner: The `:after tuareg` is redundant with `:hook (tuareg-mode . utop-minor-mode)` (the latter is the better option for lazy-loading this package).

modules/lang/ocaml/config.el
```diff
@@ -1,13 +1,82 @@
...
+  :config
+  (set-repl-handler! 'tuareg-mode #'utop)
+  (set-eval-handler! 'tuareg-mode #'utop-eval-phrase))
+
+(def-package! ocamlformat
```
hlissner: The editor/format module [uses ocp-indent as the ocaml formatter](https://github.com/lassik/emacs-format-all-the-code) by default. Is ocamlformat better? Should we offer users a choice?

From my limited research, it also seems ocamlformat struggles with partial code formatting (while ocp-indent does not).

In any case, I'd like to better integrate this with the editor/format module, with `set-formatter!`, or a wrapper command that falls back to ocp-indent if ocamlformat isn't installed (or we're formatting partial code). e.g.

```elisp
;; in lang/ocaml/autoload.el
;;;###autoload
(defun +ocaml/format ()
  (interactive)
  (if (or +format-region-p
          (not (executable-find "ocamlformat")))
      (let ((+format-with 'ocp-indent))
        (call-interactively #'+format/region-or-buffer))
    (call-interactively #'ocamlformat)))

;; in lang/ocaml/config.el (in taureg's def-package! block)
(set-formatter! 'ocamlformat #'+ocaml/format
  :modes '(caml-mode tuareg-mode))
```

modules/lang/ocaml/+flycheck.el
```diff
@@ -0,0 +1,11 @@
+;;; lang/ocaml/+flycheck.el -*- lexical-binding: t; -*-
+;;;###if (featurep! +flycheck)
```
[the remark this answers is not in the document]
edwintorok: Running flycheck can be expensive on large files, so I wanted to provide a way to turn it off. Although that has gotten better lately with syntax checking being more asynchronous.

I've dropped the flag, and now the behaviour is:
* syntax checking on save is always on, provide by `merlin` (this is one of its main features)
* syntax checking with flycheck is enabled only when `syntax-checker` is enabled.

modules/lang/ocaml/config.el
```diff
@@ -1,13 +1,82 @@
...
+    (let ((opam-share (ignore-errors (car (process-lines "opam" "config" "var" "share" "--safe")))))
+      (when (and opam-share (file-directory-p opam-share))
+        (expand-file-name "emacs/site-lisp" opam-share)))))
 
+(def-package! tuareg
+  :defer t
+  ;; modes set by autoload
+  :config
+  ;; tuareg-mode sets the prettify symbols itself
+  (when (featurep! :ui pretty-code)
```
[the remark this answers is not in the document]
edwintorok: I found that `tuareg.el` exposes `tuareg-prettify-symbols-basic-alist` and `tuareg-prettify-symbols-extra-alist`, so I replaced my hack with just calling `set-pretty-symbols!` with those.

modules/lang/ocaml/config.el
edwintorok: If the executable is not available then quite likely the package itself is not available either and `def-package!` would fail. I need to think more about how to best solve this, see  below for discussion on package installation.

modules/lang/ocaml/config.el
```diff
@@ -1,13 +1,82 @@
...
 
+;; use packages that require a binary from opam
+;; merlin is outdated on MELPA
 (def-package! merlin
   :after tuareg
-  :hook (tuareg-mode . merlin-mode)
+  :if (executable-find "ocamlmerlin")
+  :load-path +ocaml-elisp-dir
+  :init
+  (when (featurep! :feature lookup)
```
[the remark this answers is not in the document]
edwintorok: Good to know, updated.

modules/lang/ocaml/config.el
edwintorok: I dropped them completely, see below for packages.el change.

modules/lang/ocaml/config.el
edwintorok: `ocp-indent` is widely used and preferred, it indents line-by-line and only changes whitespace.
`ocamlformat` is more opinionated, it parses it all into an AST, which it then pretty-prints. This can completely change lines (e.g. break up single line into multiple, or collapse back), so it should not be enabled by default globally. It also errors out on comments that are ambiguously associated with AST elements, one has to fix these before ocamlformat can be used.

It is up to the project maintainer's (or submodule maintainer's) choice whether to use ocamlformat or not (which can be signaled by the presence of an `.ocamlformat` configuration file, perhaps empty, in a subdirectory or project root). If multiple people are actively working on branches in a project a careful coordination needs to take place before ocamlformat is deployed to not create too many conflicts (conflicts from `ocp-indent` are easier to solve by just ignoring whitespace while rebasing/merging and formatting the final commit again).

Adding `(not (locate-dominating-file default-directory ".ocamlformat")` should address the above concerns, however I could not get the `+format/region-or-buffer` call to work, I get this stack trace when calling `format-all-buffer` and then `+ocaml/format` (the former just tells me formatting failed):
```
Debugger entered--Lisp error: (wrong-type-argument listp t)
  car(t)
  (let* ((val (format-all-probe)) (x16 (car val)) (x17 (cdr val)) (x18 (car x17)) (x19 (cdr x17))) (let ((mode-result x18) (formatter x16)) (let* ((val (+format-buffer formatter mode-result (or +format-preserve-indentation +format-region-p)))) (cond ((eq val 'no-formatter) (if (called-interactively-p 'any) (progn (message "No formatter specified for %s" major-mode))) nil) ((eq val 'error) (message "Failed to format buffer due to errors") nil) ((eq val 'noop) (message "Buffer was already formatted") nil) (t (message "Formatted (%s)" formatter) t)))))
  (let ((+format-with (or (if arg (+format-completing-read)) +format-with))) (let* ((val (format-all-probe)) (x16 (car val)) (x17 (cdr val)) (x18 (car x17)) (x19 (cdr x17))) (let ((mode-result x18) (formatter x16)) (let* ((val (+format-buffer formatter mode-result (or +format-preserve-indentation +format-region-p)))) (cond ((eq val 'no-formatter) (if (called-interactively-p 'any) (progn (message "No formatter specified for %s" major-mode))) nil) ((eq val 'error) (message "Failed to format buffer due to errors") nil) ((eq val 'noop) (message "Buffer was already formatted") nil) (t (message "Formatted (%s)" formatter) t))))))
  +format/buffer(nil)
  funcall-interactively(+format/buffer nil)
  call-interactively(+format/buffer)
  (if (use-region-p) (+format/region beg end arg) (call-interactively (function +format/buffer)))
  +format/region-or-buffer(14 14)
  funcall-interactively(+format/region-or-buffer 14 14)
  call-interactively(+format/region-or-buffer)
  (let ((+format-with 'ocp-indent)) (call-interactively (function +format/region-or-buffer)))
  (if (or +format-region-p (not (executable-find "ocamlformat"))) (let ((+format-with 'ocp-indent)) (call-interactively (function +format/region-or-buffer))) (call-interactively (function ocamlformat)))
  +ocaml/format()
  funcall-interactively(+ocaml/format)
  call-interactively(+ocaml/format record nil)
  command-execute(+ocaml/format record)
  helm-M-x(nil #("+ocaml/format" 0 13 (match-part "+ocaml/format")))
  funcall-interactively(helm-M-x nil #("+ocaml/format" 0 13 (match-part "+ocaml/format")))
  call-interactively(helm-M-x nil nil)
  command-execute(helm-M-x)
```

modules/lang/ocaml/packages.el
```diff
@@ -1,5 +1,19 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; lang/ocaml/packages.el
 
+(defvar +ocaml-elisp-dir
```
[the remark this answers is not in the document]
edwintorok: I dropped `byte-recompile-directory` in favour of defining a local package:
```
(defmacro localpackage! (name)
 `(package! ,name :recipe (:fetcher file :path ,+ocaml-elisp-dir)))

(localpackage! opam-site-lisp)
```

This integrates  a bit better with the rest of doom-emacs, although see the problems mentioned before, if certain `opam` packages are not installed then I need all those `:if`s in `def-package!` to prevent the packages themselves from loading, which is a bit ugly, I should be checking for the presence of the package, not the presence of the executable (and also doesn't work well when the executables are not in PATH on startup, but get added by `tuareg-opam-update-env`.

The upstream recommendation and install instructions always recommend just adding the `site-lisp` to emacs's load-path (which is fine in general, when people are not using a package manager), however I think it might simplify things if I just add regular `package!` definitions here (and for packages not yet on MELPA, a custom recipe with github fetcher from master branch).

What are your thoughts on this?

edwintorok:
Thanks a lot for the review, I'm happy to answer questions about the OCaml tooling ecosystem.
I've opened https://github.com/melpa/melpa/pull/5742 for the oudated merlin package issue.
I'm still undecided how to best handle opam's site-lisp packages (although my latest solution is less ugly than the previous one), and I've run into some trouble trying the ocamlformat formatter you suggested, see the individual replies on your comments.

modules/lang/ocaml/config.el
hlissner: This looks much better, but it's still tricky. There's still the issue that ocamlformat can't handle partial formatting. I know users can use `ocp-indent-region`, but I'd rather `+format/region` not fail if it's avoidable.

I'll look into this another time, however. I just need to sit down and rethink `set-formatter!`.

modules/lang/ocaml/packages.el
```diff
@@ -1,5 +1,30 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; lang/ocaml/packages.el
 
+(when (featurep! :feature +opam-site-lisp)
```
hlissner: These two uses of `featurep!` are incorrect. They should be `(featurep! +opam-site-lisp)`.

However, I'm unconvinced this flag is even necessary. I say just byte-compile it regardless.

modules/lang/ocaml/config.el
```diff
@@ -1,13 +1,92 @@
...
+    :definition #'merlin-locate
+    :references #'merlin-occurrences
+    :documentation #'merlin-document)
+  (defun +ocaml|init-merlin ()
+    (when (and (projectile-locate-dominating-file default-directory ".merlin")
+               (executable-find "ocamlmerlin"))
+      (merlin-mode)))
+  (add-hook 'tuareg-mode-hook #'+ocaml|init-merlin)
+
+  :defer t
```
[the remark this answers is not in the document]
edwintorok: I put `:defer t` between `:init` and `:config` because AFAICT `:init` is always executed, and adding `:defer t` prevents adding an `(require '<your-package>)` at that position.
I can see why you would want it grouped before `:init` though, so I made that change.

edwintorok:
Thanks, I tried simplifying packages.el, but having a package defined both in opam's site lisp and from MELPA confused Emacs too much: it seemed to work fine initially, but then I disabled and reenabled ocaml in config.el (while running make each time) and that caused merlin to not get loaded at all, and dune and ocamlformat failed to install.
Maybe its not worth supplying a flag for this and just have the opam-site-lisp there commented out as a hint for the user to use it if they so wish?

hlissner:
Forgive the later reply! 

This looks much better. If you could rebase those fixups, I'm ready to merge this (and will revisit the ocamlformat issue later).

hlissner:
Fantastic! Thank you for your all your help, and putting up with my nitpicking!

hlissner:
A quick question about merlin: before `merlin-mode` is enabled, you check for a `.merlin` file in the project. Is this file a requirement for merlin to work? And is it safe to assume all merlin projects will have this file?

edwintorok:
Merlin can work without a .merlin file, but you will likely see a lot of errors because merlin won't be able to find all the libraries you use. If you use a modern build system like [dune](http://dune.build/) then it will create the `.merlin` file for you automatically, otherwise you have to create it manually and check it into the repo.

Perhaps a better way would be to enable merlin mode always, and enable flycheck only when .merlin is present (otherwise it can be very annoying if it keeps showing everything as red).

PR #916, opened 2018-09-29
magit: only implement leader key logic when doom-leader-key is non-nil

seanfarley:
As someone that doesn't use evil, it seems this call needs to be guarded.

### DIFF
--- a/modules/tools/magit/config.el
+++ b/modules/tools/magit/config.el
@@ -47,8 +47,8 @@ available.")
 
   ;; Don't replace the leader key
   ;; FIXME remove me when general.el is integrated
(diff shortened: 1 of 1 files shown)


seanfarley:
I suspect this might not be needed and is due to my misunderstanding of totally disabling evil mode. Still, it'd be nice to get an explanation either way :-)

hlissner:
Good catch. Its purpose is to make sure magit keybinds don't override the leader key for evil users. In any case, it will be removed once I've brought general.el into Doom.

Thank you for the PR!

seanfarley:
\o/ yay!

PR #924, opened 2018-10-02
mu4e-confirm-quit should be nil instead of t

ragone:
I believe the `mu4e-confirm-quit` is meant to be nil according to the comment
above. Feel free to close if this was intended.


hlissner:
Good catch, thanks!

PR #925, opened 2018-10-03
Fix write mode fonts and format.el

edwintorok:
I set a custom font for variable-pitch and it was not taking effect (it was quite obvious, because I set a serif font, and I got a sans serif one instead).
I also checked with `SPC u C-x =` and looks like the font I set had no effect.

Two reasons:
* the font only got set for new frames, and not for the main emacs frame
* the +write-mode-hook was not actually run, because we assigned to the hook before +write-mode minor mode got defined (and the autoloads only generate autoloads for the mode, not the hooks)

I fixed these, perhaps not in the most elegant way. Feel free to suggest improvements.

Also got a bugfix for format.el, so included it here (wrong type, `cons` expected a list as 2nd argument, but it really wanted to just build a list of 2 elements here).

### DIFF
--- a/core/core-ui.el
+++ b/core/core-ui.el
@@ -351,9 +351,9 @@ frame's window-system, the theme will be reloaded.")
               ((display-graphic-p)
                (setq doom-font (face-attribute 'default :font))))
         (when doom-serif-font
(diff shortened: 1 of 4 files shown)


hlissner:
Nice work, thanks for the PR!

PR #927, opened 2018-10-03
Fix some error path and typos spotted by `make compile`

edwintorok:
Couple things I noticed when running `make compile`


hlissner:
Ah, more good catches! Thanks!

PR #928, opened 2018-10-04
fix `+macos/open-in-default-project` under `ranger-mode`

ztlevi:
A quick fix for invoking `+macos/open-in-default-project` under `ranger-mode`


hlissner:
Thanks!

PR #929, opened 2018-10-04
Make tide-mode optional for javascript

MannySchneck:
* Condition tide on an init flag
* Change example.init to include flag

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -98,7 +98,8 @@
        ;;hy                ; readability of scheme w/ speed of python
        ;;idris             ;
        ;;(java +meghanada) ; the poster child for carpal tunnel syndrome
-       ;;javascript        ; all(hope(abandon(ye(who(enter(here))))))
+       (javascript         ; ; all(hope(abandon(ye(who(enter(here))))))
+        +tide)             ; slow down your editor with a flood of useful info
        ;;julia             ; a better, faster MATLAB
        ;;latex             ; writing papers in Emacs has never been so fun
        ;;ledger            ; an accounting system in Emacs
--- a/modules/lang/javascript/config.el
+++ b/modules/lang/javascript/config.el
@@ -121,6 +121,7 @@
 ;; Tools
 
 (def-package! tide
+  :when (featurep! +tide)
   :defer t
   :init
   ;; Don't let hard errors stop the user from opening js files.


hlissner:
I am hesitant to accept this PR, for two reasons:

1. On Discord, it was said the motivation for this PR was slowness/stuttering while editing. I don't think tide-mode is to blame, as it doesn't do anything as you type, directly.

   In that respect, there are many other packages that are suspect. Flycheck, smartparens, and company are three likely culprits. Company's as-you-type completion (with the `+auto` flag or a non-nil `company-idle-delay`) is particularly slow. If you use it, you may have better luck disabling this specifically for tide with:

   ```elisp
   (setq-hook! 'tide-mode-hook company-idle-delay nil)
   ```

2. I want to avoid adding flags for individual packages, unless absolutely necessary. Flags are meant to represent features (groups of packages or non-trivial configuration), tool integration, or exceptionally popular defaults (like the `+auto` flag for completion/company). The idiomatic way to disable individual packages in Doom is `(package! tide :disable t)` in your `~/.doom.d/packages.el`.

Perhaps once we bring LSP on board, a `+tide` flag may be more appropriate. 

What do you think?

ztlevi:
My perspective of lsp is it bring performance problem and is worse than `tide`. I've using lsp for roughly half year and `lsp-javascript-typescript` is a terrible choice for React development. It couldn't handle `Synthetic Imports` which means `jump-to-definition` over `import React from 'react'` is not working. 

`tide` and `lsp-typescript` are both wrapper for the langserver comes with `vscode` and tend to better choice, at least `Synthetic Imports` are working well as long as we defined the proper `jsconfig.json` or `tsconfig.json`. And according to the performance issue, I would say `tide` is a much better choice. And it's definitely enough for all kinds of JS/TS development.

To sum up, it's a good idea to avoid additional flags. Especially for this case, as far as I know, there is no such better choice other than `tide` for JS.

BTW: I would recommend to add additional documentation for `javascript` module. For example, to enable proper `Synthetic Imports`, we need add some doc like [this](https://github.com/ztlevi/spacemacs/blob/tide/layers/+tools/tide/README.org#jsx) documented in `tide` as well.

MannySchneck:
I'm withdrawing this PR, as Henrik's comments make sense.

PR #933, opened 2018-10-06
Updated documentation to use better gocode fork (supports Go 1.11+)

theSuess:
Updated the documentation to install the recommended gocode package.

nsf/gocode is obsolete since it does not support Go 1.10+.

This fork is the currently recommended one as seen here: https://github.com/golang/go/issues/24661

### DIFF
--- a/modules/lang/go/README.org
+++ b/modules/lang/go/README.org
@@ -51,7 +51,7 @@ This module requires a valid ~GOPATH~, and the following Go packages:
 export GOPATH=~/work/go
 
 go get -u github.com/motemen/gore
-go get -u github.com/nsf/gocode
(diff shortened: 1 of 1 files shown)


bittopaz:
@hlissner according to https://github.com/melpa/melpa/pull/5753, probably `github.com/stamblerre/gocode` is not the right package we should list in the README.

It should be https://github.com/mdempsky/gocode.

theSuess:
@bittopaz you're right! Thr stamblerre fork has been merged with mdempsky. At the time at which I created the PR this has not yet been the case. So @hlissner should I create another PR or will you fix this yourself?

hlissner:
Please create a new one. I'm not at a computer, but can merge it from my phone.

PR #939, opened 2018-10-08
Allow `deadgreap` folding in normal mode

zaiste:
`TAB` doesn't work in normal mode for `deadgreap` to hide results on per-file basis. This PR fixes that by explicitly distinguishing the `deadgrep` mode.


zaiste:
@hlissner could you describe what the ideal solution would look like ? I can spend some time on it and improve it.

PR #945, opened 2018-10-10
Capture template tweaks

patrl:
- The built-in capture templates now add a link to the file from which capture was initiated - this integrates well with a capture-based workflow, and also means that the built-in templates play nicely with, e.g., `org-mu4e-store-and-capture`.
- Checkboxes have been changed to TODOs - other components of doom, e.g., `magit-todos` assume TODOs and don't recognize checkboxes. In any case, in org-mode checkboxes are standardly used for subtasks.


hlissner:
I now remember why I used `[ ]`. It was for personal TODOs (since they don't fall in the purview of any project). However, it is better that we change the project ones to `TODO`. Could you revert only the first one?

patrl:
Done!

hlissner:
Perfect, thanks!

PR #948, opened 2018-10-10
Open org links with `pdf-view-mode` (with `:tools pdf`)

patrl:
When `:tools pdf` is enabled, org links to pdf files should be opened in `pdf-view-mode`. Additional support is added for links to a specific *page* in the pdf.


hlissner:
Nice work. Thanks again!

PR #961, opened 2018-10-15
 lang/clojure: add binding to switch/set namespace 

mfiano:
Add a local leader keybinding to both switch to the CIDER REPL buffer and set the namespace according to the file buffer.

Additionally, this fixes the clobbered "r" binding, and some minor changes.


hlissner:
Looks good. Thanks!

PR #964, opened 2018-10-16
Add vala-mode

syvanpera:
Just added one package: 'vala-mode'. Provides syntax coloring for Vala language. Nothing more, nothing less.


hlissner:
Nice and simple! Thanks!

PR #966, opened 2018-10-16
merlin: eldoc, imenu and iedit support

edwintorok:
Thanks for mergin my PR, I see you've been busy refactoring it :)
Here are a few more  tweaks, `merlin-eldoc` is something I've recently come across, and merlin already had `imenu` and `iedit` support in the `merlin` package, just had to bind it to some keys/enable it.


hlissner:
Great work! No nits to pick here! Thanks again for your help.

Issue #941, opened 2018-10-10
treemacs showes nothing


laishulu:
### Observed behavior
treemacs showes nothing, see the following screenshot.
I am use emacs in terminal.
It works well when I use the GUI `emacs.app`

<img width="918" alt="snipaste_2018-10-10_17-29-38" src="https://user-images.githubusercontent.com/1224321/46726978-2a41cb00-ccb2-11e8-9cbb-812cbed5938d.png">

### Expected behavior

show directory of the current directory

### Steps to reproduce

1. Select these example steps,
2. Delete them,
3. And replace them with precise steps to reproduce your issue.

### System information

<details>
<summary>Click to expand</summary>

```
- OS: darwin (x86_64-apple-darwin14.5.0)
- Emacs: 26.1 (May 31, 2018)
- Doom: 2.0.9 (develop cdf745e72a3e8bf9bef1f0fa4d051b00aab351e3)
- Graphic display: nil (daemon: nil)
- System features: NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES THREADS
- Details:
  ```elisp
  elc count: 0
  uname -a:  Darwin machw 18.0.0 Darwin Kernel Version 18.0.0: Wed Aug 22 20:13:40 PDT 2018; root:xnu-4903.201.2~1/RELEASE_X86_64 x86_64 i386 MacBookPro10,1 Darwin
  modules:   (:feature (evil +everywhere) file-templates (lookup +docsets) snippets spellcheck (syntax-checker +childframe) workspaces :completion (helm +fuzzy) :ui doom-dashboard doom-modeline doom-quit evil-goggles fci hl-todo nav-flash treemacs (popup +all +defaults) unicode vc-gutter vi-tilde-fringe window-select :editor rotate-text :emacs ediff electric hideshow imenu vc :tools magit :lang emacs-lisp markdown (org +attach +babel +capture +export +present) python rust (sh +fish) :config)
  packages:  n/a
[...]

PR #967, opened 2018-10-16
fix treemacs compilation error

edwintorok:
In toplevel form:
../../../ui/treemacs/config.el:16:1:Error: Custom icon cannot be nil
✕ Failed to compile modules/ui/treemacs/config.el

Might not be the best solution, but it fixes the compile problem.


hlissner:
I'm afraid this is only a bandaid on the real problem. It is also the cause of #941. I'm currently entertaining this as a possible fix:

```lisp
(defun doom*disable-all-the-icons-in-tty (orig-fn &rest args)
  (if (display-graphic-p)
      (apply orig-fn args)
    ""))
```

hlissner:
It appears to work, so I'll close this.

PR #971, opened 2018-10-18
app: Add application to manage Arch's updates

UndeadKernel:
Add an application to manage the .pacnew and .pacsave files left behind after updating Arch-based distributions.


hlissner:
Sorry for the tremendously overdue reply (almost exactly a year!). I archived this PR because I wasn't quite sure if it'd fit into Doom's default module list -- and it slipped my mind. I've finally decided to turn it down because the config is too small to warrant its own module and is a tad too niche (I am considering an :os category, where it might fit, but I'll revisit this PR if/when that happens).

In any case, thanks for the PR (and for pacfiles-mode).

PR #977, opened 2018-10-21
turn off fci when window is narrow than fill column

laishulu:
without those hook handles:
snipaste_2018-10-22_00-14-25

with those hook handles:
snipaste_2018-10-22_00-17-51

### DIFF
--- a/modules/ui/fci/config.el
+++ b/modules/ui/fci/config.el
@@ -21,6 +21,23 @@ Changes to this variable do not take effect until `fci-mode' is restarted.")
   ;; more important to me, so...
   (add-hook 'org-mode-hook #'turn-off-fci-mode)
 
+  ;; turn off fci if the window is narrow than fill column
+  (defun fci-conditional (window)
+    (let ((fci-enabled (symbol-value 'fci-mode))
+          (fci-column (if fci-rule-column fci-rule-column fill-column)))
+      (with-selected-window window
+        (if (and (eq fci-enabled nil)
+                 (< fci-column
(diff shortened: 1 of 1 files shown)


hlissner:
I was hesitant to accept this PR. It is an expensive fix for a plugin that is already notoriously slow, but there may be no better solution. In any case, thanks for your help!

hlissner:
After some testing, I had to revert this PR as it doesn't behave correctly. This width-workaround would toggle fci-mode on and off, each time `window-size-change-functions` or `window-configuration-change-hook` was triggered.

laishulu:
Yes. It's the expected behavior.
Also, I find that when `fci-mode` is turned on. 
It would disable `visual-line-mode` automatically.
So long lines is truncated with a `$` marked at the end of each line.
It is a disaster, because can't read each full line without horizental scroll for each line.

Also fci is too lagging, so I am now turned to `column-marker`

```elisp
;; hightlight fill column
(def-package! column-marker
  :config
  (define-minor-mode toggle-fill-column-marker-mode
    "Toggle column marker mode."
    ;; The initial value.
    :init-value t
    (if toggle-fill-column-marker-mode
        (column-marker-1 80)
      ;; C-u     M-x display-prefix  -| (4)
      (column-marker-1 '(4))))

  (add-hook 'after-change-major-mode-hook 'toggle-fill-column-marker-mode)
  (map! :leader
        (:desc "toggle" :prefix "t"
          :desc "fill column marker" :nv "m" #'toggle-fill-column-marker-mode)))
```

PR #983, opened 2018-10-27
added :editor lispyville

The-BigDaddy:
I didn't include the `additional-movements` key theme because it has yet to work (the keys in that key theme are not being mapped to any corresponding `lispyville-` function), and `text-objects` key theme because that seems to be more useful with [targets.el](https://github.com/noctuid/targets.el) which is a whole different package to configure.


hlissner:
Looks good! I also have my eyes on targets.el, so I'll look into it more after the documentation update. Thanks again!

PR #987, opened 2018-10-31
lang/ocaml: drop bad imenu initialization

edwintorok:
A copy+paste error slipped  through in my previous PR, this would cause problems due to `merlin-imenu` not being available at the time `merlin-iedit` is initialized.


hlissner:
Good catch, thanks!

PR #996, opened 2018-11-01
(WIP) Feature/ui/frame-workflow

qleguennec:
Not sure where I'm going with this. There are a lot of issues:
- Should this be bound to `:feature` or to `:ui`?
- Should this replace `:feature workspace`? Overwrite the existing keybindings?
- If not, what prefix should I bind for frame keybindings? `f` is for files, `tab` is for workspaces.

Also my `map!` form in `config.el` doesn't get executed at startup. Can't figure out why.

### DIFF
--- a/modules/ui/frame-workflow/autoload.el
+++ b/modules/ui/frame-workflow/autoload.el
@@ -1,6 +1,13 @@
 ;;; ui/frame-workflow/autoload.el -*- lexical-binding: t; -*-
 
 ;;;###autoload
+(defun +frame-workflow/make-frame-current-major-mode ()
+  "Make a frame for current major mode and switch to it"
+  (interactive)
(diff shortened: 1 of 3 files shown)


hlissner:
Sorry for the tremendously late response. I've decided not to accept frame-workflow into Doom's default module list because there is overlap between it and the frame-oriented features I have planned for ui/workspaces. Additionally, both frame-workflow and persp-mode will be entirely replaced by native tabs in Emacs master, once it hits stable.

If you're still using the module and would like to maintain it, feel free to move it to its own repo. I will eventually have a mechanism for installing 3rd party modules.

rgrinberg:
Would it perhaps makes sense to still include frame selection using ivy? That alone would be quite useful.

hlissner:
@rgrinberg Try `M-x select-frame-by-name`

PR #997, opened 2018-11-03
Close the ending quote in lispy-mode

niklascarlsson:
With this setting lispy closes the ending quote which is in alignment
with how Doom treats quotes in other modes.

### DIFF
--- a/modules/editor/lispy/config.el
+++ b/modules/editor/lispy/config.el
@@ -9,6 +9,7 @@
          (lfe-mode . lispy-mode)
          (clojure-mode . lispy-mode))
   :config
+  (setq lispy-close-quotes-at-end-p t)
   (add-hook 'lispy-mode-hook #'turn-off-smartparens-mode))
 
 (def-package! lispyville


ar1a:
isn't this something smartparens should do? b0f18047994af67eabfb9420f9a321ce415c8bab makes me think so

qleguennec:
> isn't this something smartparens should do? [b0f1804](https://github.com/hlissner/doom-emacs/commit/b0f18047994af67eabfb9420f9a321ce415c8bab) makes me think so

smartparens is currently disabled in `lispy-mode`.

I personally think we should leave the default lispy behaviors and let the user change it in their private doom if they want.

Unless @hlissner wants to have this behavior homogenized whatever module is enabled.

ar1a:
id argue this should be a default then, to make it consistent with the rest of doom

niklascarlsson:
I think harmonizing it with the rest of Doom would make sense as default. Adding the setting would also make it easy for users to find it and disable it if they want the default lispy behavior.

hlissner:
Makes sense to me, thanks!

PR #998, opened 2018-11-03
Add lispy to init.example.el

ar1a:


hlissner:
I've made it opt-in, but thanks for the PR

PR #1000, opened 2018-11-05
lang/rust: Added clippy command and :desc descriptions

moshohayeb:
Added proper descriptions to cargo's commands instead of 'lambda'
in which-key

Thank you for contributing to Doom!

Before you submit this PR, please make sure your PR is targeted at develop, not
master (unless this is a fix for a critical error). Then replace this message
with a description of your changes.

### DIFF
--- a/modules/lang/rust/config.el
+++ b/modules/lang/rust/config.el
@@ -7,11 +7,12 @@
 
   (map! :map rust-mode-map
         :localleader
-        :prefix "b"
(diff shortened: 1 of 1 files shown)


hlissner:
`:desc` is broken for `map!` for mode or buffer-local keybinds. e.g. A which-key description for `SPC a` in `rust-mode-map` will affect `SPC a` globally, not just in rust-mode.

Until this is resolved (when general is brought in), I cannot accept this PR.

PR #1005, opened 2018-11-12
add required package 'with-editor'

shuxiao9058:


filalex77:
I think `with-editor` is bundled with Magit, at least of version 2.90.0.

PR #1025, opened 2018-11-23
Fix 'SPC m b r' in go-mode.

ngkz:
This PR fixes a bug which caused 'SPC m b r' to fail in go mode.


hlissner:
Good catch, thanks!

PR #1026, opened 2018-11-23
Add pytest

Emiller88:
Adds a nice pytest interface that uses magit popup to run the commands.

### DIFF
--- a/modules/lang/python/config.el
+++ b/modules/lang/python/config.el
@@ -111,17 +111,11 @@
         :n "V" #'nosetests-pdb-module))
 
 (def-package! python-pytest
-  :commands pytest-mode
-  :preface (defvar pytest-mode-map (make-sparse-keymap))
-  :init (associate! pytest-mode :match "/test_.+\\.py$" :modes (python-mode))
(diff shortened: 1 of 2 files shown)


hlissner:
Thanks for the PR! It doesn't seem that python-pytest is lazy loaded, so its keybinds will override nose-mode's. It seems pytest and nose tests both have the same file name scheme (`test-*.py`), so I'm not sure how to reconcile the two. I don't mind replace nose with pytest, as this module's default, supported testing framework, however. What do you think?

Emiller88:
I kept the bindings from conflicting with nose's. 
The end user would have nose bindings and pytest bindings if they open a `test-*.py` file and the option to run either suite. 
If they're in any `*.py` file they'll be able to call the popup and run pytest commands since it's designed to be used from where ever in the python project.
Basically the two shouldn't conflict, the user may slip up and hit a nose command in a test file on accident is the worse case.

hlissner:
Ah, you're right, my mistake. In that case the only issue is that it isn't lazy-loaded, which I've corrected. Thanks for your help!

PR #1029, opened 2018-11-27
Don't scan frames other than current in popup

amosbird:
This accidentally reuses windows in other frames, especially when there are multiple compliation buffers.


hlissner:
Good catch! Thanks!

PR #1031, opened 2018-11-29
Use fd for populating projectile cache

ar1a:
[fd](https://github.com/sharkdp/fd) is a rust program **significantly faster** than find and git ls-files that **respects .gitignore**.

This change is also recommended in the [projectile docs](https://projectile.readthedocs.io/en/latest/configuration/), or a screenshot: screenshot

### DIFF
--- a/modules/tools/magit/config.el
+++ b/modules/tools/magit/config.el
@@ -53,7 +53,7 @@ what features are available.")
   ;; fd is a rust program that is significantly faster. It also respects
   ;; .gitignore. This is recommended in the projectile docs
   (when (executable-find "fd")
-    (setq projectile-git-command "fd . -0"
+    (setq projectile-git-command "fd . --type f -0"
           projectile-generic-command projectile-git-command)))
 
 (def-package! magit-todos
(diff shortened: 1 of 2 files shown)


bennylb:
I only found out about `fd` recently. Not only is it faster but a lot nicer to use.

UndeadKernel:
Using `fd` is a good idea.
I wouldn't put the change, however, in the magit module. `fd` is much more general.
I think the change should be integrated into core.

hlissner:
Great idea! I'm totally on board, but is there a reason it's in tools/magit and not core/core-project.el?

ar1a:
err, because i wasn't paying attention lol. ill go move it there :woman_facepalming:

ar1a:
maybe its because i was changing the git-command? ask sleepy me :P

modules/tools/magit/config.el
```diff
@@ -49,6 +49,7 @@ what features are available.")
   (when doom-leader-key
     (define-key magit-diff-mode-map (kbd doom-leader-key) nil)))
 
+
```
ar1a: oh no!

hlissner: madness!

hlissner:
Nice work, thanks!

PR #1035, opened 2018-12-02
Fixed +popup--kill-buffer visibility check

djeis97:
Changed visibility check to consult all frames


hlissner:
Good catch! Thanks again!

PR #1038, opened 2018-12-06
Adding Ranger module to doom

ztlevi:
Super efficient module for file browser and it's designed for vimmers. Copy paste files between folders are hard for me in emacs. But with ranger, I could select files with `tab` and then use vim operator keys, e.g. d, y, to perform delete, copy actions.


ztlevi:
I tested the ranger module without dired module.

ar1a:
Looks like it could use a lot of cleaning up, but overall some nice additions here.

modules/emacs/ranger/config.el
```diff
@@ -0,0 +1,42 @@
...
+
+
+(def-package! all-the-icons-dired
+  :defer t
+  :init
+  (when (display-graphic-p)
+    (add-hook! 'dired-mode-hook #'all-the-icons-dired-mode)))
+
+
+(def-package! font-lock+)
```
hlissner: font-lock+ isn't being lazy-loaded. Remove this if there isn't any configuration needed for it.

modules/emacs/ranger/config.el
```diff
@@ -0,0 +1,42 @@
...
+  (setq ranger-omit-regexp "^\.DS_Store$"
+        ranger-excluded-extensions '("mkv" "iso" "mp4")
+        ranger-deer-show-details nil
+        ranger-max-preview-size 10))
+
+
+(def-package! all-the-icons-dired
+  :defer t
+  :init
+  (when (display-graphic-p)
```
hlissner: This check should be done whenever `all-the-icons-dired-mode` is enabled, rather that only once, at startup. GUI Emacs can be started from a daemon.

modules/emacs/ranger/config.el
```diff
@@ -0,0 +1,42 @@
...
+  ;; initialized to make sure that `gls' is in `exec-path'
+  (when IS-MAC
+    (let ((gls (executable-find "gls")))
+      (if gls
+          (setq insert-directory-program gls)
+        (message "Please install `gls` using `brew instal coreutils`..."))))
+  (setq dired-listing-switches "-aBhl --group-directories-first"))
+
+
+(def-package! ranger
```
hlissner: This isn't lazy loaded, please add `:defer t`, or use one of `:commands` or `:hooks` as necessary.

modules/emacs/ranger/config.el
```diff
@@ -0,0 +1,42 @@
...
+  :init
+  (setq ranger-override-dired t)
+  ;; set up image-dired to allow picture resize
+  (setq image-dired-dir (concat doom-cache-dir "image-dir"))
+  (unless (file-directory-p image-dired-dir)
```
hlissner: To allow users to customize image-dired-dir, perform this directory check in `:config` instead of `:init`.

modules/emacs/ranger/config.el
```diff
@@ -0,0 +1,42 @@
...
+  :defer t
+  :init
+  (when (display-graphic-p)
+    (add-hook! 'dired-mode-hook #'all-the-icons-dired-mode)))
+
+
+(def-package! font-lock+)
+
+
+(def-package! dired-x)
```
hlissner: dired-x isn't lazy loaded and there is no config. Please remove this.

modules/emacs/ranger/autoload.el
```diff
@@ -0,0 +1,12 @@
+;;; private/ranger/autoload.el -*- lexical-binding: t; -*-
```
hlissner: The path in the comment header is incorrect.

ztlevi:
## font-lock+
`font-lock+` package is necessary for `all-the-icons-dired-mode`. Otherwise, the icons looks weird. But I don't know the reason. It probably fix some features for `font-lock` over the icon support. Functions insides `font-locks+` are not autoloaded so I cannot defer it.

## display-graphic-p
As for the `display-graphic-p` problem, I am afraid I don't know the proper way doing it. Did you mean I need to add a `daemonp` check? You can go ahead fix this problem once this PR merged.

## Personal perferences
Besides, I have couple personal ranger preferences.
1. I override the ranger's quit function because the original `ranger-disable` will delete  same buffer window. For example, I have two ranger buffer opened vertically split and I quit one of them. The other one will quit too. And I don't like this way because sometimes I will use two split ranger buffer to copy files between.
2. ranger buffers in other workspaces sometimes cause problem when I switch between workspaces.  So I decide to kill these ranger buffers without active windows.

The configs resolve problems above is showing below, and I put it in the `ranger`'s `:config` to override the original functions.
```
  (defun ranger-close-and-kill-inactive-buffers ()
    "ranger close current buffer and kill inactive ranger buffers"
    (interactive)
    (ranger-close)
    (ranger-kill-buffers-without-window))
  ;; do not kill buffer if exists in windows
  (defun ranger-disable ()
    "Interactively disable ranger-mode."
    (interactive)
    (ranger-revert))
```

## Key bindings
There are some keys I prefer but might not be necessary to add them by default. You can take a look at them. Many of them can be called via `;` prefix in ranger mode.

`ranger-go` is mapped to `g` by default but in this case, `g g` go to top does not work. So I mapped `ranger-go` to `M-g`
```
  (map!
   (:map ranger-normal-mode-map
     "g" nil
     "q" #'ranger-close-and-kill-inactive-buffers
     "f" #'counsel-find-file
     "M-g" #'ranger-go
     "C-<tab>" #'ranger-next-tab
     "C-S-<tab>" #'ranger-prev-tab
     "U" #'dired-unmark-all-files
     "u" #'dired-unmark
     "(" #'dired-hide-details-mode
     "+" #'dired-create-directory
     ))
```

ar1a:
"Functions insides font-locks+ are not autoloaded so I cannot defer it."

you can with :commands

modules/emacs/dired/config.el
```diff
@@ -60,6 +72,47 @@
...
+    (make-local-variable 'dired-hide-details-hide-symlink-targets)
+    (setq dired-hide-details-hide-symlink-targets nil)
+    ;; hide details by default
+    (dired-hide-details-mode t))
+  (add-hook 'dired-mode-hook #'ranger/dired-setup)
+
+  (setq ranger-override-dired t)
+  :config
+  ;; set up image-dired to allow picture resize
+  (setq image-dired-dir (concat doom-cache-dir "image-dir"))
```
hlissner: Please set `image-dired-dir` in :init. The variable should be set as soon as possible, but the directory check and `make-directory` should be done as late as possible, to give the user a chance to customize it.

modules/emacs/dired/config.el
```diff
@@ -60,6 +72,47 @@
...
+        ranger-excluded-extensions '("mkv" "iso" "mp4")
+        ranger-deer-show-details nil
+        ranger-max-preview-size 10))
+
+
+(def-package! all-the-icons-dired
+  :when (featurep! +icons)
+  :defer t
+  :init
+  (when (display-graphic-p)
```
hlissner: It turns out that all-the-icons-dired-mode [already performs its own display-graphic-p check the way I wanted](https://github.com/jtbm37/all-the-icons-dired/blob/master/all-the-icons-dired.el#L88), so this check isn't necessary.

hlissner:
Alright, looks much better, thanks again for your help!

PR #1041, opened 2018-12-07
Add descriptions to rust-mode-map

Emiller88:
Otherwise they all just say `lambda`.

### DIFF
--- a/modules/lang/rust/config.el
+++ b/modules/lang/rust/config.el
@@ -8,10 +8,10 @@
   (map! :map rust-mode-map
         :localleader
         :prefix "b"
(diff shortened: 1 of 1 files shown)


hlissner:
This is a duplicate of #1000. To quote my response to it:

> :desc is broken for map! for mode or buffer-local keybinds. e.g. A which-key description for SPC a in > rust-mode-map will affect SPC a globally, not just in rust-mode.
>
> Until this is resolved (when general is brought in), I cannot accept this PR.

Once General is in, I'll see to adding `:desc` to all buffer-local keybinds, including these.

PR #1042, opened 2018-12-08
Clean up after swoop removal

ar1a:
removed in 7f5e778b0c3abe14b2597f9b93fb399ff2335e62
some functions were left over that error out, this removes them


ar1a:
There might be many that doesn't exist anymore - like `+helm/pt`

hlissner:
Good catch. Thanks.

On a side note, `+helm/pt` is still defined, though I'm considering dropping pt support.

PR #1052, opened 2018-12-18
Enable magit by default

ar1a:
makes sense as almost a staple of emacs it should be on by default. not
to mention the vcs module is enabled by default too!

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -70,7 +70,7 @@
        ;;gist              ; interacting with github gists
        ;;macos             ; MacOS-specific commands
        ;;make              ; run make tasks from Emacs
-       ;;magit             ; a git porcelain for Emacs
+       magit             ; a git porcelain for Emacs
        ;;password-store    ; password manager for nerds
        ;;pdf               ; pdf enhancements
        ;;prodigy           ; FIXME managing external services & code builders


hlissner:
I've thought about this for a while now and am conflicted. I don't think it should be enabled by default. 

Magit is certainly a staple of Emacs, but it's a very specialized (git-centric) tool. it's also a _big_ package with _big_ dependencies (and growing, considering the recent transient update). Magit also has acceptable alternatives: `git` on the command line, any git GUI client, or other version control systems. `:tools` is also meant to be a category for optional, external tool support.

`:emacs vc`, on the other hand, expands on functionality built into Emacs. It provides the "bare minimum" of version control integration for our favorite OS (and doesn't only support git). I think that distinguishes it from magit as a universally essential module.

ocharles:
May be worth considering the weight of the votes above. If you're installing doom, you're already into pretty power user territory. Also, it's not like you can't just turn it off if you don't want the footprint.

ar1a:
I think it also makes sense as a sensible default. IMO I think doom should be pretty good to go by default, instead of a minimalist approach and the user turns on things.

ar1a:
This was fixed in cedadb3!

PR #1055, opened 2018-12-19
add rust key bingding description

chengyi:
add rust mode key bindings description


hlissner:
I was unable to accept this earlier, as `:desc` didn't work for buffer-local keybinds in the old version of `map!`. Now that the general.el update is up, `:desc` works again.

Thanks for your help!

PR #1057, opened 2018-12-23
Bind SPC-/-p to +helm/project-search if using helm

ar1a:
seems to have been lost in the binding wars of 2018


hlissner:
Good catch, thanks again!

PR #1061, opened 2018-12-24
Change .gitignore to exclude "modules/private" when attached as symlink

NXTk:
Change .gitignore to exclude "modules/private" when attached as symlink.


hlissner:
Thanks for your help!

PR #1064, opened 2018-12-25
config/default: fix `Wrong number of arguments: define-key`

dywedir:


hlissner:
Ah, good catch! Thanks for your help!

PR #1066, opened 2018-12-26
Fix highlight-non-default-indentation

andresilva:
It was highlighting the expected white space style.


hlissner:

PR #1072, opened 2018-12-27
fix: disable verbose message for dired-omit

ragone:
To avoid endless spam in \*Messages\*, disable "Omitted 2 lines" every
time a new dired buffer is opened.

verbose


hlissner:
Thanks!

PR #1076, opened 2018-12-27
[tools/magit] Add magit-forge

filalex77:


ztlevi:
`map! :leader` probly should go to `+evil-binding.el` and how about add `:after magit` in `def-package!`

modules/tools/magit/config.el
```diff
@@ -100,3 +101,9 @@ what features are available.")
     (evil-define-key* evil-magit-state git-rebase-mode-map
       "gj" #'git-rebase-move-line-down
       "gk" #'git-rebase-move-line-up)))
+
+(def-package! forge
+  :when (featurep! :tools magit +forge)
+  :config
+  (map! :leader
+        :desc "Forge ""gF" #'forge-dispatch))
```
ztlevi: Space position located after Forge is wrong.

filalex77: @ztlevi Thanks!

ztlevi:
@filalex77 I mean, you can put `map!` over here~ https://github.com/hlissner/doom-emacs/blob/develop/modules/config/default/+evil-bindings.el#L605

filalex77:
@ztlevi Oh, I totally blanked out on that, sorry. Will fix tomorrow :(

hlissner:
I haven't had time to investigate forge properly yet; is there any reason to keep magithub, if we bring in forge?

filalex77:
From what I can see, Forge does both more and less than Magithub. Forge supports more services, Magithub has some functionality (like forking) that I didn't see in Forge yet.

So, I think defaulting for Forge and leaving Magithub as an option is an acceptable course of action. Though, if anybody knows better, feel free to chime in :)
-------- Original Message --------

filalex77:
I think it is. If I find any Forge-specific configuration, I'll make a second PR.

luigy:
Should probably also set the db path within `doom-etc-dir`?
```elisp
(setq forge-database-file (concat doom-etc-dir "forge/forge-database.sqlite")))
```

filalex77:
@luigy Good idea, I think. I'm not familiar with `doom-etc-dir` and similar internals (mostly), so input from @hlissner or someone involved with Magit/Forge here would be most appreciated.

richerve:
@hlissner anything else to do here? waiting for this to be merged to migrate to forge, for me the features provided are enough and it's much more stable in how to communicate with the git interfaces.

hlissner:
It slipped my mind! Thanks for reminding me. @filalex77 Thanks for your help!

PR #1088, opened 2019-01-02
latex: enable flyspell correctly in latex buffers

UndeadKernel:
The varible `TeX-mode-local-vars-hook` doesn't exist. Instead, we use
`TeX-mode-hook` to enable flyspell. We append to the hook, however, to
enable disabling flyspell with
`(setq-hook! 'TeX-mode-hook +spellcheck-immediately nil)`

### DIFF
--- a/modules/lang/latex/config.el
+++ b/modules/lang/latex/config.el
@@ -66,7 +66,7 @@ If no viewers are found, `latex-preview-pane' is used.")
   ;; Enable rainbow mode after applying styles to the buffer
   (add-hook 'TeX-update-style-hook #'rainbow-delimiters-mode)
   (when (featurep! :feature spellcheck)
-    (add-hook 'TeX-mode-local-vars-hook #'flyspell-mode))
+    (add-hook 'latex-mode-local-vars-hook #'flyspell-mode))
   ;; All these excess pairs dramatically slow down typing in latex buffers, so
   ;; we remove them. Let snippets do their job.
   (after! smartparens-latex


hlissner:
`TeX-mode-local-vars-hook` does not exist, but _is_ executed. [It's a Doom thing](https://github.com/hlissner/doom-emacs/blob/develop/core/core.el#L307-L314). Its purpose is to offer a hook that runs after file/directory-local variables are read. Flyspell is initialized this late so that users can customize its behavior through file-local variables or .dir-locals.el files. This change would break that.

UndeadKernel:
I see what the thought process is, yes.
There is the issue, however, that this does not work with such a hook:
```lisp
(setq-hook! 'TeX-mode-hook +spellcheck-immediately nil)
```
That is, we are not able to set `+spellcheck-immediately` to `nil` so that TeX buffers are not checked automatically at start.

Of course I tried with
```lisp
(setq-hook! 'TeX-mode-hook +spellcheck-immediately nil :append)
```
but it doesn't work.

Flyspell is not initialized late enough so that `+spellcheck-immediately` can be set to `nil`.

hlissner:
Very odd. `TeX-mode-local-vars-hook` should trigger much later than `TeX-mode-hook`. I'll look into this over the weekend.

UndeadKernel:
@hlissner, I found the issue.
You do not need, in theory, to look further into this PR:
The `*-local-vars-hook`s are running correctly. However, once again, AucTeX was not reporting the correct major mode. I changed the hook to be the one reported by AucTeX and now spellchecking is correctly enabled when opening LaTeX buffers.

hlissner:
Ah ha! Good catch. Thanks for looking into this!

PR #1089, opened 2019-01-02
pdf-tools: fixes for the new `modeline` module

UndeadKernel:
When using the new modeline module, pdf tools was not displaying the modeline correctly. This PR fixes it.

Also, remove some TODO comments that have already been addressed.


hlissner:
Nice work. Thanks again!

PR #1090, opened 2019-01-02
bindings: default set of bindings for emacs users

UndeadKernel:
With the `default` module and the `+bindings` flag enabled, non-evil users will have a default set of bindings suitable for Doom emacs.

### DIFF
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -26,8 +26,13 @@
  ;; Editor related bindings
  "C-a"            #'doom/backward-to-bol-or-indent
  [remap newline]  #'newline-and-indent
- "C-j"           #'+default/newline
- "C-S-s"          #'swiper
+ "C-j"            #'+default/newline
+ (:when (featurep! :completion ivy)
+   "C-S-s"        #'swiper
+   "C-S-r"        #'ivy-resume)
+ (:when (featurep! :completion helm)
+   "C-S-s"        #'swiper-helm
+   "C-S-r"        #'helm-resume)
  ;; Buffer related bindings
  "C-x b"       #'persp-switch-to-buffer
  "C-x C-b"     #'ibuffer-list-buffers
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -13,6 +13,7 @@
 
 ;; Prefix key to invoke doom related commands
 (setq doom-leader-alt-key "C-c")
+(setq doom-localleader-alt-key "C-c l")
 
 (map!
  ;; Text scaling
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -17,9 +17,12 @@
 
 (map!
  ;; Text scaling
- "C-+" (λ! (text-scale-set 0))
- "C-=" #'text-scale-increase
- "C--" #'text-scale-decrease
+ "<C-mouse-4>" #'text-scale-increase
+ "<C-mouse-5>" #'text-scale-decrease
+ "<C-down-mouse-2>" (λ! (text-scale-set 0))
+ "M-+" (λ! (text-scale-set 0))
+ "M-=" #'text-scale-increase
+ "M--" #'text-scale-decrease
  ;; Editor related bindings
  "C-a"            #'doom/backward-to-bol-or-indent
  [remap newline]  #'newline-and-indent
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -174,7 +174,7 @@
      (:prefix ("I" . "irc")
        :desc "Open irc app" "i"      #'=irc
        :desc "Quit irc" "q"          #'+irc/quit
-       :desc "Reconnect all" "r"     #'circle-reconnect-all
+       :desc "Reconnect all" "r"     #'circe-reconnect-all
        :desc "Send message" "s"      #'+irc/send-message
        (:when (featurep! :completion ivy)
          :desc "Jump to channel" "j" #'irc/ivy-jump-to-channel)))
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -147,7 +147,31 @@
        :desc "Edit line starts"   "a"         #'mc/edit-beginnings-of-lines
        :desc "Mark tag"           "s"         #'mc/mark-sgml-tag-pair
        :desc "Mark in defun"      "d"         #'mc/mark-all-like-this-in-defun
-       :desc "Add cursor w/mouse" "<mouse-1>" #'mc/add-cursor-on-click)))
+       :desc "Add cursor w/mouse" "<mouse-1>" #'mc/add-cursor-on-click))
+
+   ;; APPs
+
+   ;; Email
+   (:when (featurep! :app email)
+     (:prefix ("M" . "email")
+       :desc "Open email app" "m" #'=email
+       :desc "Compose email" "c"  #'+email/compose))
+   ;; IRC
+   (:when (featurep! :app irc)
+     (:prefix ("I" . "irc")
+       :desc "Open irc app" "i"      #'=irc
+       :desc "Quit irc" "q"          #'+irc/quit
+       :desc "Reconnect all" "r"     #'circle-reconnect-all
+       :desc "Send message" "s"      #'+irc/send-message
+       (when (featurep! :completion ivy)
+         :desc "Jump to channel" "j" #'irc/ivy-jump-to-channel)))
+   ;; Twitter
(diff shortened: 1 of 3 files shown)


seanfarley:
Wooooo!

modules/config/default/+emacs-bindings.el
```diff
@@ -1,3 +1,258 @@
...
+(which-key-add-key-based-replacements "C-c e"   "perspective")
+(which-key-add-key-based-replacements "C-c p"   "projectile")
+
+;; Prefix key to invoke doom related commands
+(setq doom-leader-alt-key "C-c")
+
+(map!
+ ;; Text scaling
+ "C-+" (λ! (text-scale-set 0))
+ "C-=" #'text-scale-increase
```
Oghma: You swapped the functions. Should be `"C-=“ (λ! (text-scale-set 0))` and in the previous line `"C-+" #'text-scale-increase`

modules/config/default/+emacs-bindings.el
```diff
@@ -1,3 +1,258 @@
...
+       :desc "Edit line starts"   "a"         #'mc/edit-beginnings-of-lines
+       :desc "Mark tag"           "s"         #'mc/mark-sgml-tag-pair
+       :desc "Mark in defun"      "d"         #'mc/mark-all-like-this-in-defun
+       :desc "Add cursor w/mouse" "<mouse-1>" #'mc/add-cursor-on-click)))
+
+ ;; Plugins
+
+ ;; misc plugins
+ "<f9>" #'+neotree/open
+ "C-="  #'er/expand-region
```
Oghma: Is “C-=“ binded to `#'text-scale-increase`?

modules/config/default/+emacs-bindings.el
```diff
@@ -1,3 +1,258 @@
...
+       :desc "Edit line endings"  "e"         #'mc/edit-ends-of-lines
+       :desc "Edit line starts"   "a"         #'mc/edit-beginnings-of-lines
+       :desc "Mark tag"           "s"         #'mc/mark-sgml-tag-pair
+       :desc "Mark in defun"      "d"         #'mc/mark-all-like-this-in-defun
+       :desc "Add cursor w/mouse" "<mouse-1>" #'mc/add-cursor-on-click)))
+
+ ;; Plugins
+
+ ;; misc plugins
+ "<f9>" #'+neotree/open
```
Oghma: Is better? 
```
(:when (featurep! :ui neotree)
    "<f9>" #'+neotree/open
(:when (featurep! :ui treemacs)
    "<f9>" #’+treemacs
```

modules/config/default/+emacs-bindings.el
UndeadKernel: You are correct. I have changed the bindings for changing the font size.

modules/config/default/+emacs-bindings.el
UndeadKernel: This is intentional so that there is no need to hold shift to change font size. I've changed CTRL for META.

modules/config/default/+evil-bindings.el
```diff
@@ -12,10 +12,47 @@
             doom-leader-key doom-localleader-key
             doom-leader-alt-key doom-localleader-alt-key))
 
+;; OS specific fixes
+(when IS-MAC
+  ;; Fix frame-switching key on MacOS
+  (global-set-key (kbd "M-`") #'other-frame))
```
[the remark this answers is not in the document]
UndeadKernel: This only gets executed if you are running `evil`. We are not evil... right?

modules/config/default/+emacs-bindings.el
```diff
@@ -1,3 +1,304 @@
...
+       :desc "Open email app" "m" #'=email
+       :desc "Compose email" "c"  #'+email/compose))
+   ;; IRC
+   (:when (featurep! :app irc)
+     (:prefix ("I" . "irc")
+       :desc "Open irc app" "i"      #'=irc
+       :desc "Quit irc" "q"          #'+irc/quit
+       :desc "Reconnect all" "r"     #'circle-reconnect-all
+       :desc "Send message" "s"      #'+irc/send-message
+       (when (featurep! :completion ivy)
```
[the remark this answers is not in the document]
UndeadKernel: A typo, indeed.

modules/config/default/+evil-bindings.el
```diff
@@ -12,10 +12,58 @@
...
+  ;; textmate-esque newline insertion
+  [M-return]    #'evil-open-below
+  [M-S-return]  #'evil-open-above
+  ;; textmate-esque deletion
+  [M-backspace] #'doom/backward-kill-to-bol-and-indent)
+
+;; Smarter C-a/C-e for both Emacs and Evil. C-a will jump to indentation.
+;; Pressing it again will send you to the true bol. Same goes for C-e, except
+;; it will ignore comments+trailing whitespace before jumping to eol.
+(map! :gi "C-a" #'doom/backward-to-bol-or-indent
```
hlissner: Could you move these two keybinds (C-a & C-e) back to config/default/config.el? These keys are present in both evil and non-evil sessions, and are meant to make them behave a little more intelligently.

modules/config/default/+evil-bindings.el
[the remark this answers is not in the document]
hlissner: The <kbd>M-\`</kbd> keybind is necessary to restore its OS behavior on MacOS, Evil or not. Unless there's another reason to restrict this to evil users, please move it back to config/default/config.el.

modules/config/default/+emacs-bindings.el
```diff
@@ -1,3 +1,304 @@
...
+     "C-M-y" #'counsel-yank-pop
+     "C-h F" #'counsel-faces
+     "C-h p" #'counsel-package
+     "C-h a" #'counsel-apropos
+     "C-h V" #'counsel-set-variable
+     "C-'"   #'counsel-imenu))
+ ;; repl toggle
+ "C-c C-z" #'+eval/open-repl
+ ;; company mode
+ (:after company
```
hlissner: Is there any alternative for set-mark-command (on `C-SPC`)? I'd like to have C-SPC for invoking Company manually, since `+auto` is not enabled by default for company and it's a conventional code-completion for a variety of editors.

seanfarley: `C-SPC` not being for marks ... 😦😦😦?????

modules/config/default/+evil-bindings.el
UndeadKernel: I like the idea of making these two keys the default also in emacs.

modules/config/default/+emacs-bindings.el
hlissner: Fair enough. I think TAB is a good target. I've got [something similar going on in evil-bindings](https://github.com/hlissner/doom-emacs/blob/develop/modules/config/default/+evil-bindings.el#L28-L34) I can adapt to it.

@seanfarley You'll have to forgive me. I don't know much about key conventions in vanilla Emacs.

hlissner:
And merged. There are a few things I think are missing, like Doom's [custom help/documentation commands](https://github.com/hlissner/doom-emacs/blob/develop/modules/config/default/+evil-bindings.el#L635-L661) or better helm support (both completion modules remap commands, so we should be using the base commands anyway), but I'll look into these things next weekend.

modules/config/default/+emacs-bindings.el
seanfarley: Hehe, that was just my head imploding as @UndeadKernel suggested :-)

4goodapp:
Hi,
Emacs have very very few keybindings that work the same as in other (non-evil :p) editors,
So I think those are precious keybindings that shouldn't be overwritten (by default).

It only creates more inconsistency specially for new user imo.
Spacemacs which apparently this is inspired from have MANY custom keybinding but those "GENERAL" keybindings are not touched.

So that's why I suggest that you either disable this by default OR, remove certain keybindings cause, even on browsers and no non-editors programs have shortcut like Ctrl+left , Ctrl+right, Ctrl+shift+left/right used for the same purpose. even on Spacemacs.

I really don't understand how this can be consistent tbh

UndeadKernel:
Hi @4goodapp,

Doom emacs is described as a stubborn and opinionated emacs configuration. These key bindings fall into the category of "opinionated": The standard behavior of many emac's key bindings are not always doing what's more useful or intuitive. We therefore take the liberty of changing "standard" bindings to obtain a more desirable behavior from the start. Take the `C-a` command as an example, it is remapped to `#'doom/backward-to-bol-or-indent`. This modified behavior, although different from the original, is generally more useful.

Experienced emacs users that want to return to the standard behavior are free to disable the "defaults" module or remap the functions they don't like. An experienced user will easily know how to make these changes. The story is different for new and unexperienced users. They are better off having the "nicer" and more powerful functionality from the start.

PR #1092, opened 2019-01-03
pdf: fix error when not using evil

seanfarley:
This fixes an error I have when opening pdfs. The fix is just to make sure
doom-leader key is non-nil.


hlissner:
Actually, since the general update, unbinding doom-leader-key is no longer necessary. Could you remove that keybind instead? (Keep the `q` bind though)

seanfarley:
Ok, updated the pull request

seanfarley:
@hlissner Any other blocks for this one?

seanfarley:
🕺

PR #1104, opened 2019-01-09
org: let ob-async ignore ob-ipython

UndeadKernel:
ipython implements its own `async` keyword. As ob-async is loaded now by
default, we need to explicitly tell ob-async not to touch ob-ipython blocks.


hlissner:
Good to know. Thanks again for your help!

PR #1107, opened 2019-01-10
fix flickering PDFs in pdf-view-mode

mnick:
Zooming into PDFs in PDFView conflicts with blinking cursors in evil-mode what causes the PDFs to flicker. 

This applies the fix from https://github.com/politza/pdf-tools/issues/201 what resolves the issue for me.


hlissner:
Sorry or the late response. Is there any reason we can't simply set `evil-normal-state-cursor` in this case? e.g.

```diff
-(set-evil-initial-state! 'pdf-view-mode 'emacs)
-(add-hook 'pdf-view-mode-hook
-          (lambda ()
-            (set (make-local-variable 'evil-emacs-state-cursor) (list nil))))
+(setq-hook! 'pdf-view-mode-hook evil-normal-state-cursor (list nil))
```

`evil-collection` provides some evil integration for the pdf major modes that I'd like to preserve, if possible.

PR #1110, opened 2019-01-10
bindings: correctly detect if syntax-checker is on

UndeadKernel:
A bug left from legacy code.

syntax-checker is now in "features" and not in "completion".


hlissner:
Ah, good catch.

PR #1112, opened 2019-01-11
emacs/dired: Ranger package enhancement

ztlevi:
I still want to use `font-lock+` in favor of a fork of all-the-icons fix. This time, I hook it after `all-the-icons-dired` since the major issue is `all-the-icons-dired-mode` not showing proper icons.

With `all-the-icons` fix fork, some icons are not showing correctly if we use `swiper` inside deer/ranger.

image

While with `font-lock+` enabled, icons are showing properly.
image

### DIFF
--- a/core/packages.el
+++ b/core/packages.el
@@ -8,9 +8,7 @@
   (package! ns-auto-titlebar))
 
 ;; core-ui.el
-(package! all-the-icons
-  :recipe (:fetcher github :repo "ubolonton/all-the-icons.el"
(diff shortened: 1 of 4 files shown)


hlissner:
Unfortunately, font-lock+ breaks major modes that use specialized, external parsers, like `enh-ruby-mode` and `js2-mode`, as well as any major mode that uses font-lock parsing for things like indentation. Unless there's a better alternative, I can't accept font-lock+ into Doom.

PR #1113, opened 2019-01-11
Add jest

Emiller88:
Added a my first emacs package. 

Not sure whether to bind it to tide mode map or js2.

### DIFF
--- a/modules/lang/javascript/config.el
+++ b/modules/lang/javascript/config.el
@@ -227,6 +227,18 @@
       "r" #'npm-mode-npm-run
       "v" #'npm-mode-visit-project-file)
 
+;; Jest
+(def-package! jest
+  :defer t
+  :init
+  (map! :after js2-mode-map
+        :localleader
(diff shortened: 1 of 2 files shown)


hlissner:
Sorry for the late response. I'm not yet certain what to do with this PR. Javascript has a number of popular testing frameworks. I wouldn't mind incorporating support for each of them, but I'm not keen on supporting only one.

Ideally, we'd have mocha, jasmine, tape, AVA and jest support, and have a dispatcher scan a project's package.json (the `+javascript-npm-dep-p` can do this) for installed frameworks and invoke the appropriate backend. Unfortunately, there is no package for AVA and tape, though there is [mocha.el](https://github.com/scottaj/mocha.el) with support for both mocha and jasmine.

It will be a while until I can look into this further, but this is a good start.

Emiller88:
@hlissner Sounds good feel free to close it in the mean time if you like.

PR #1125, opened 2019-01-18
Improve projectile-git-command

ar1a:
When indexing it ignores hidden files, which is especially annoying for dotfiles
where everything starts with a dot. If you just have -H it'll index .git too
though, so we exclude that


hlissner:
I was worried -H would cause `fd` not to respect vcs ignore files, but looks like it doesn't. Thanks for the help!

PR #1130, opened 2019-01-22
irc: automatically add circe buffers to irc workspace

seanfarley:
This also adds a better `next-tracking-buffer` that restricts itself to the irc
workspace.


seanfarley:
High fives all 'round!

PR #1143, opened 2019-01-26
Fix Clojure mode lookup handlers

gsnewmark:
Both `cider-browse-ns-find-at-point` and `cider-browse-ns-doc-at-point` are functions that meant to be used inside of the Cider's [namespace browser](https://docs.cider.mx/en/latest/miscellaneous_features/#browsing-namespaces) and not in the regular source files, so I switched lookup handlers to use functions which actually work in source files. 

I've also added references lookup handler based on `clj-refactor` as Cider itself [doesn't support it (yet)](https://github.com/clojure-emacs/cider/issues/1840).


hlissner:
Sorry for the late response. Nice work and thanks for your help!

PR #1144, opened 2019-01-26
Add vterm layer

SteamedFish:
vterm layer uses emacs-libvterm for terminal support.

I'm an emacs newbie. I learnt how to use emacs by using Doom Emacs, and I learnt how to write elisp program by reading Doom Emacs's source code.

This is my first time to write an elisp program, most of the code was copied from the `:emacs term` layer without truly understanding. I'm really sorry if my bad code makes you in exasperation.

There are still two things I'm in doubt:

1. I don't know what category this layer should be put. Because I copied it from `:emacs term` layer, I put it to `:emacs vterm`. I don't know if it's the right decision.

2. I don't know how to make emacs compile the module when install the vterm package by `make` command. I make it compile at the boot time, which losses some boot speed.

3. Everytime when I rum `make update` and the vterm package updates, it should be recompiled. I don't know how to do this.

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -60,6 +60,7 @@
        hideshow          ; basic code-folding support
        imenu             ; an imenu sidebar and searchable code index
        ;;term              ; terminals in Emacs
+       ;;vterm             ; another terminals in Emacs
        vc                ; version-control and Emacs, sitting in a tree
 
        :tools
--- a/init.example.el
+++ b/init.example.el
@@ -60,7 +60,6 @@
        hideshow          ; basic code-folding support
        imenu             ; an imenu sidebar and searchable code index
        ;;term              ; terminals in Emacs
-       ;;vterm             ; another terminals in Emacs
        vc                ; version-control and Emacs, sitting in a tree
 
        :tools
@@ -80,6 +79,7 @@
        ;;tmux              ; an API for interacting with tmux
        ;;upload            ; map local to remote projects via ssh/ftp
        ;;wakatime
+       ;;vterm             ; another terminals in Emacs
 
        :lang
        ;;assembly          ; assembly for fun or debugging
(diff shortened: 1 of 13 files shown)


jwintz:
👍🏻

QiuMP:
👍

shuxiao9058:
👍

hlissner:
Sorry for the late follow-up. I'm interested in this PR, but I have four reservations about it that I haven't had enough time to look into in yet:

1. This module doesn't lazy load vterm. I want to avoid compiling vterm at startup. It may be better to leave it to the user to do (as Doom's philosophy -- when it comes to the system environment -- is "the user knows best"). Perhaps provide an autoloaded `+vterm/install` command to make it easier. Considering they'll need to do this every time the package updates, it may be more convenient this way.
2. I'm skeptical about that long list of `vterm--self-insert` keybinds, Would it not be better to keep vterm in emacs state?
3. The vterm popup rule should be defined inside this module with `set-popup-rule`, rather than in ui/popup.
4. I think this module belongs in `:tools` rather than `:emacs`, as `:emacs` is for modules that modify and extend built-in behavior and plugins (as term and eshell do). `:tools` is for interacting with external processes, services or servers. However, if the rationale is that it should replace term, then it'd be better placed inside `:tools term`, perhaps gated behind a `+vterm` flag or something. I know little about vterm, so I'll rely on your judgment.

in any case, nice work and thanks for your help!

hlissner:
>    3\. Everytime when I rum `make update` and the vterm package updates, it should be recompiled. I don't know how to do this.

vterm-module.so is compiled inside the package which gets deleted when it is updated, so the end result is it will get recompiled automatically.

----

In any case, this looks great and vterm is a pleasure to use. Thanks for bringing this to Doom!

PR #1145, opened 2019-01-26
Fix description for increase/decrease font size

versinib:
Using which-key, I noticed "increase text size" had the opposite effect then noticed that the actual function that increased text size had a vague description.

I assumed this was a description error because the contrary would mean a universal hotkey change.


hlissner:
Good catch! Thanks for your help.

PR #1147, opened 2019-01-28
More MacOS bindings

seanfarley:
I don't think anything in here is *too* crazy

### DIFF
--- a/modules/config/default/config.el
+++ b/modules/config/default/config.el
@@ -139,9 +139,14 @@
         "s-n" #'+default/new-buffer
         "s-N" #'make-frame
         "s-q" (if (daemonp) #'delete-frame #'save-buffers-kill-terminal)
+        "C-s-f" #'toggle-frame-fullscreen
+        ;; Restore somewhat common navigation
+        "s-l" #'goto-line
         ;; Restore OS undo, save, copy, & paste keys (without cua-mode, because
         ;; it imposes some other functionality and overhead we don't need)
+        "s-f" #'swiper
         "s-z" #'undo
+        "s-Z" #'redo
         "s-c" (if (featurep 'evil) #'evil-yank #'copy-region-as-kill)
         "s-v" #'yank
         "s-s" #'save-buffer
--- a/modules/config/default/config.el
+++ b/modules/config/default/config.el
@@ -156,6 +156,7 @@
         "s--" #'text-scale-decrease
         ;; Conventional text-editing keys & motions
         "s-a" #'mark-whole-buffer
+        "s-/" #'doom/toggle-comment-region-or-line
         :gi [s-return]    #'+default/newline-below
         :gi [s-S-return]  #'+default/newline-above
         :gi [s-backspace] #'doom/backward-kill-to-bol-and-indent
--- a/core/autoload/editor.el
+++ b/core/autoload/editor.el
@@ -270,6 +270,16 @@ Respects `require-final-newline'."
   (when (looking-at "\n+")
     (replace-match "")))
 
+;;;###autoload
+(defun doom/toggle-comment-region-or-line ()
+  "Comments or uncomments the whole region or if no region is
+selected, then the current line."
+  (interactive)
+  (let (beg end)
+    (if (region-active-p)
+        (setq beg (region-beginning) end (region-end))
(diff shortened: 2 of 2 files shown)


modules/config/default/config.el
```diff
@@ -139,9 +139,16 @@
...
+        "C-s-f" #'toggle-frame-fullscreen
+        ;; Restore somewhat common navigation
+        "s-l" #'goto-line
+        "s-}" #'forward-paragraph
+        "s-{" #'backward-paragraph
         ;; Restore OS undo, save, copy, & paste keys (without cua-mode, because
         ;; it imposes some other functionality and overhead we don't need)
+        "s-f" #'swiper
         "s-z" #'undo
+        "s-Z" #'undo-tree-redo
```
hlissner: Please bind `s-Z` to `redo` instead. It will be remapped to undo-tree-redo if undo-tree-mode is enabled.

modules/config/default/config.el
```diff
@@ -139,9 +139,16 @@
         "s-n" #'+default/new-buffer
         "s-N" #'make-frame
         "s-q" (if (daemonp) #'delete-frame #'evil-quit-all)
+        "C-s-f" #'toggle-frame-fullscreen
+        ;; Restore somewhat common navigation
+        "s-l" #'goto-line
+        "s-}" #'forward-paragraph
+        "s-{" #'backward-paragraph
```
hlissner: These three keys are suspect. While I recognize `s-l` from textmate and textedit, it selects the current line in atom, sublime text and bbedit.

And I don't know where `s-}` and `s-{` come from.

modules/config/default/+emacs-bindings.el
```diff
@@ -298,6 +301,10 @@
    (:map ivy-minibuffer-map
      "TAB" #'ivy-alt-done
      "C-g" #'keyboard-escape-quit))
+ ;; helpful
+ (:after helpful
+   (:map helpful-mode-map
+     "C-g" #'quit-window))
```
hlissner: C-g should already be bound to `doom/escape`, which closes helpful popup windows. In the case you have intentionally disabled the popup manager, I'd rather C-g's behavior not be changed.

modules/config/default/+emacs-bindings.el
```diff
@@ -15,6 +15,9 @@
 (setq doom-leader-alt-key "C-c")
 (setq doom-localleader-alt-key "C-c l")
 
+(when IS-MAC
+  (map! "s-q" (if (daemonp) #'delete-frame #'save-buffers-kill-terminal)))
```
hlissner: I already have this in a local commit (and it will be in `config/default/config.el` instead). Please remove this.

core/autoload/buffers.el
```diff
@@ -334,3 +334,14 @@ processes killed."
           (delete-process p)
           (cl-incf n))))
     n))
+
+;;;###autoload
+(defun doom/toggle-comment-region-or-line ()
```
hlissner: Could this be moved to `core/autoload/editor.el`? buffers.el is for buffer/window management functions.

modules/config/default/config.el
seanfarley: Ah, yeah, makes sense.

modules/config/default/config.el
seanfarley: Up to you, really.

seanfarley: Just FYI, I got the `s-l` from Xcode (literally go to line) and Safari (go to location).

I forget where `s-{}` come from. Most likely TextMate.

modules/config/default/+emacs-bindings.el
seanfarley: Will do

core/autoload/buffers.el
seanfarley: Ah, sure

modules/config/default/+emacs-bindings.el
seanfarley: It would seem so! Yay, less code!

seanfarley:
I think I resolved everything but let me otherwise!

seanfarley:
Yayyyy!

PR #1153, opened 2019-02-05
Harmonize previous/next commands

niklascarlsson:
I noticed that previous/next for some of the bindings had changed. When looking into the +evil-bindings function I saw that the description of the commands didn't align with the actual functions being called.

This patch updates the named sections with harmonized description and function calls.


ocharles:
The next/previous error binding change annoyed me, but not enough to get round to fixing it. This looks good!

hlissner:
Ah, what a blunder. Nice work; thanks for catching that!

niklascarlsson:
No problem, glad to help! :)

PR #1154, opened 2019-02-05
irc: fix lui track bar not being enabled

seanfarley:
It seemed I mistakenly put this in an `evil` block and didn't notice.


hlissner:
Good catch, thanks!

seanfarley:
Woooo!

PR #1157, opened 2019-02-07
Solidity Fix

### LINKED ISSUE #1152: Solidity loading packages
When opening a solidity file
```
Error (use-package): Cannot load solidity-flycheck
Error (use-package): Cannot load company-solidity
```
Only occurs when first opening the second time it just opens it with no error.

Emiller88:
Fixes #1152. Though there's still an upstream issue of the `solc` command should
now be `solc-js`

### DIFF
--- a/modules/lang/solidity/packages.el
+++ b/modules/lang/solidity/packages.el
@@ -2,3 +2,5 @@
 ;;; lang/solidity/packages.el
 
 (package! solidity-mode)
+(package! company-solidity)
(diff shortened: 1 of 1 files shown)


Emiller88:
Scratch my comment about `solc` I just was installing the wrapper for whatever reason.

hlissner:
Good catch. Thanks for your help! (Also, roger that about the solc issue)

PR #1165, opened 2019-02-10
Add missing variable name

jcf:
This enables compilation of the Ruby module where previously the
following error would cause a breaking error.

    ../../../lang/ruby/autoload.el:48:45:Error: &rest without variable name
    ✕ Failed to compile modules/lang/ruby/autoload.el


hlissner:
Ah, good catch, thanks!

PR #1178, opened 2019-02-18
Add transient quit evil-bindings

luigy:
this adds back the dismiss behavior of "magit popups" that was lost from the
magit-popup -> transient update

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -445,11 +445,14 @@
 
 ;;; :tools
 (map! (:when (featurep! :tools magit)
-        :after evil-magit
-        ;; fix conflicts with private bindings
-        :map (magit-status-mode-map magit-revision-mode-map)
-        "C-j" nil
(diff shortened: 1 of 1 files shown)


modules/config/default/+evil-bindings.el
```diff
@@ -445,11 +445,14 @@
...
-        ;; fix conflicts with private bindings
-        :map (magit-status-mode-map magit-revision-mode-map)
-        "C-j" nil
-        "C-k" nil)
+        (:after evil-magit
+          ;; fix conflicts with private bindings
+          :map (magit-status-mode-map magit-revision-mode-map)
+          "C-j" nil
+          "C-k" nil)
+        (:map transient-map
```
luigy: Not sure wether `transient-edit-map` or `transient-sticky-map` should also go here

hlissner:
Good idea. It appears to work fine on `transient-map` for the time being. Thanks for the PR!

Issue #158, opened 2017-07-20
Haskell + stack + intero vs. Haskell + cabal + dante


Ptival:
Hey @hlissner , do you have any infrastructure in place to describe alternative setups for a given language, short of splitting it in two language modes?

For instance, in the Haskell ecosystem, there is a choice between using cabal-based tools (like `dante`) or using stack-based tools (like `intero`). It's easy to add either to `lang/haskell`, but one user might want to choose between a `haskell-dante` and `haskell-intero` setup, where, I believe, the two are somewhat mutually exclusive.

Any thoughts on this?

hlissner:
I've anticipated this and am experimenting. The solution I'm leaning toward is boolean feature flags for `doom!` in init.el. Something like:

```emacs-lisp
(doom! :feature
       ;; [...]
       syntax-checker
       (version-control +git +hg)   
       workspaces
       ;; [...]
       :lang
       go
       (haskell +intero)  ;; or +dante
       hy
       ;; [...]
       )
```

I leave it to modules to interpret these flags however they wish. They can be detected using `featurep!`, like so:

+ Longhand form (works anywhere): `(featurep! :feature version-control +git)`
+ Shorthand form (syntactic sugar when used in a module's config.el or packages.el file): `(featurep! +git)`

Therefore:

```emacs-lisp
(when (featurep! +intero)
  (load! +intero))

;; or the simpler equivalent
(def-feature! +intero)
```

It's slated for `develop` over the weekend, I'll then add intero and dante features to `lang/haskell`. That said, `intero` will be the new default, since I only chose dante+cabal because I was new to Haskell at the time, and hadn't tried stack yet. These days, I've come to prefer stack and believe it makes a more sensible default.

Ptival:
Cool! I don't care too much about the default, I've used both and liked both in different contexts. Looking forward to those changes! 👍

hlissner:
Alright, this is now on `develop` and will be merged into master over the weekend, so I'll consider this resolved. 

I have a small haskell project coming up next week, so I'll use that to iron out any kinks in the new module.

Ptival:
Woo! I will also test it out! 👍

hlissner:
@Ptival Just wanted to let you know, in the recent push to master I've made module flags for the haskell module required. Meaning there is no default anymore. Hope that saves you a few minutes of frustration!

PR #1185, opened 2019-02-21
lang/haskell: only require intero when flag is set

luigy:
Looks like a leftover from https://github.com/hlissner/doom-emacs/issues/158


hlissner:
This was intentional, actually (i.e. if you're not using dante, you're using intero). That said, I suppose there's no harm in a more explicit conditional. Thanks for the help!

PR #1188, opened 2019-02-22
Add haskell +lsp

ar1a:
A few things are left wanting in this mode which has been mentioned on discord.
One of the glaring things is that when company-lsp has snippets, the line is
incorrectly indented when chosing a completion, and you have to unindent the
line. 

The other glaring issue is eldoc only shows what the function will return,
which would be bearable if company worked with snippets!

Some more digging re: eldoc should get you looking at `lsp-eldoc-render-all` and `lsp-markup-display-all`

[The lsp server is hie](https://github.com/haskell/haskell-ide-engine/)

### DIFF
--- a/modules/lang/haskell/+lsp.el
+++ b/modules/lang/haskell/+lsp.el
@@ -1,0 +1,8 @@
+;;; lang/haskell/+lsp.el -*- lexical-binding: t; -*-
+
+(def-package! lsp-haskell
+  :after haskell-mode
+  :init (add-hook 'haskell-mode-hook #'lsp)
+  :config
+  ;; Does some strange indentation if it pastes in the snippet
+  (setq company-lsp-enable-snippet nil))
--- a/modules/lang/haskell/+lsp.el
+++ b/modules/lang/haskell/+lsp.el
@@ -2,7 +2,7 @@
 
 (def-package! lsp-haskell
   :after haskell-mode
-  :init (add-hook 'haskell-mode-hook #'lsp)
+  :init (add-hook 'haskell-mode-hook #'+lsp|init)
   :config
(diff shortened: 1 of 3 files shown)


ar1a:
for people with big fonts,
```lisp
(after! lsp-haskell
  ;; These take up a lot of space on my big font size
  (setq lsp-ui-sideline-show-code-actions nil))
```

may help. And for people running nix,
```lisp
(setq lsp-haskell-process-wrapper-function
  (lambda (argv)
    (append
      (append (list "nix-shell" "-I" "." "--command")
              (list (mapconcat 'identity argv " ")))

      (list (concat (lsp-haskell--get-root) "/shell.nix")))))
``` will almost certainly help

**Comment by ocharles** [comment]
> And for people running nix,

People running Nix should really get on the direnv bandwagon. Then Emacs will just source your `shell.nix` appropriately when you change buffers and update `PATH` accordingly.

**Comment by ar1a** [comment]
generally if you have one project using nix, all the projects are using nix

**Comment by hlissner** [comment]
Since it uses yasnippet to expand snippets, could it be the `yas-indent-line` setting has something to do with it? It's set to `'auto` by default, which means it'll use the mode's indent-line-function. You could try `(setq yas-indent-line 'fixed)` or `(setq yas-indent-line nil)` and see if that changes anything.

**Comment by ar1a** [comment]
'fixed seems to have done it

**Comment by hlissner** [comment]

**Comment by ar1a** [comment]
ah, didn't know if you could setq-hook on a non-local variable or not

**Comment by hlissner** [comment]
`setq-hook!` forcibly makes variables buffer-local, i.e. it expands to

```elisp
(add-hook 'haskell-mode-hook (lambda () (setq-local yas-indent-line 'fixed)))
```

PR #1193, opened 2019-02-23
Add LSP for Rust

iovxw:
In fact, Rust LSP already works :+1: :tada:  , this PR just disable racer when +lsp


hlissner:

PR #1194, opened 2019-02-23
Fix wrong function names in non-evil yas-keymap

flatwhatson:


hlissner:
Ah, good catch! Thanks for the help!

PR #1195, opened 2019-02-24
Add: sumatrapdf as pdf viewer on Windows

xhcoding:
Add [SumatraPDF](https://www.sumatrapdfreader.org/free-pdf-reader.html) as PDF viewer on Windows


hlissner:
Shouldn't that be `(executable-find "SumatraPDF.exe")` (with the `.exe` extension)? Also, when installing SumatraPDF, does it add itself to `%PATH%`?

xhcoding:
`.exe` extension is not required. Users need to manually add `SumatraPDF.exe` file directory to %PATH%, which I don't think is a problem.

hlissner:

PR #1196, opened 2019-02-24
Fix wrong function name for non-evil repl

flatwhatson:


hlissner:
Another good catch. Thanks again!

PR #1197, opened 2019-02-24
Bind C-l to ivy-done in counsel-ag-map

luigy:
to make it more consistent with `ivy-minibuffer-map` rather than it being bound to `ivy-call-and-recenter` which is already mapped by `C-SPC`


hlissner:

PR #1199, opened 2019-02-25
Improved switch-to-buffer-other-window for ivy

flatwhatson:
Added `+ivy/switch-workspace-buffer-other-window` so we have matching behaviour for `C-x b` and `C-x 4 b`.  Ideally this would be a replacement for `persp-switch-to-buffer-other-window`, but that doesn't exist.

Also added `switch-to-buffer-other-window` binding for `C-x 4 B` to mirror `switch-to-buffer` on `C-x B`.

Also added the other-window variants to the ivy-rich enhancement list.


hlissner:
Looks good to me. Thanks again for the help!

PR #1200, opened 2019-02-25
Replace cquery with ccls for C/C++ LSP support

flatwhatson:
The `cquery` language server is unmaintained and simply broken at this point.  Instead, we should use `ccls` which is well maintained and works very nicely.

https://github.com/MaskRay/ccls

@MaskRay


hlissner:
I wasn't aware ccls had succeeded it. Nice work and thanks for the help!

PR #1201, opened 2019-02-25
Added example of how to activate +ein-hydra

kot-behemoth:
Small change to README.org with an example of how to bind ein hydra. The code is based on hlissner's suggestion when I asked about this on Discord. Might be useful for other users setting up ein who are not too familiar with hydras (I didn't know you have to append `/body` to hydra definition to actually call it).


hlissner:
Sorry for the tremendously late response. Thank you for your help!

PR #1202, opened 2019-02-26
Restore C/C++ default-header-file-mode support

flatwhatson:
Removes the call to the builtin `c-or-c++-mode` from doom's `+cc-c-c++-objc-mode`.  The built-in mode can't be customized and always falls back to `c-mode`, which is annoying if that's not actually correct.


hlissner:
Good point! Thanks again for the help!

PR #1203, opened 2019-02-26
Remove C/C++ tab-always-indent override

flatwhatson:
The default config module sets `c-tab-always-indent` to `nil` when smartparens are enabled, but it's not clear why (it seems to come from the earliest days of this project).  This setting is not consistent with other modes, or with "standard" emacs behaviour.  Note that `tab-always-indent` is set `t` in `core-editor.el`.


hlissner:
I don't recall myself why it was set, but I think you're right. I can't argue with consistency! Thanks again.

Issue #510, opened 2018-04-05
open brace indentation in cc-mode


ustcpxy:
I am writing a few C programs and I noticed that the default brace indentation is as follows:
```
if ( a < b )
    {
        // code 
    }
```

How would I go about modifying this behaviour to achieve the following instead:
```
if ( a < b )
{
    // code
}
```

BTW:
I add below code in my private config but it takes no effect. 
```
(add-to-list 'c-offsets-alist '(substatement-open . 0))
```

scturtle:
Cause:
https://github.com/hlissner/doom-emacs/blob/5dacbb7cb1c6ac246a9ccd15e6c4290def67757c/modules/lang/cc/config.el#L87-L88
Hot fix:
```lisp
(after! cc-mode
  (map! :after cc-mode :mode c-mode-base "{" #'c-electric-brace))
```

hlissner:
@scturtle That should have no baring on line indentation in cc modes. These settings are present in the `:config default` module, which are meant to replace electric keys in the C modes:

```elisp
  (dolist (brace '("(" "{" "["))
    (sp-pair brace nil
             :post-handlers '(("||\n[i]" "RET") ("| " "SPC"))
             ;; I likely don't want a new pair if adjacent to a word or opening brace
             :unless '(sp-point-before-word-p sp-point-before-same-p)))
```

@ustcpxy The default style used is [an adjusted version of the `linux` style](https://github.com/hlissner/doom-emacs/blob/develop/modules/lang/cc/config.el#L72-L99).

Have you changed `c-default-style`?

hlissner:
Scratch that, I see the problem now. @scturtle's solution works:

```elisp
(map! :after cc-mode
      :map c-mode-base-map
      "{" #'c-electric-brace)
```

I'll look into this to see if I can find a better solution.

ustcpxy:
@hlissner I don't change the `c-default-style`.

ustcpxy:
yes, @scturtle's solution works for me. Thanks.

hlissner:
I've merged @scturtle's solution into `develop`, so I'll consider this fixed. Let me know if that isn't the case.

PR #1206, opened 2019-02-27
fix the issue: open brace indentation in cc-mode #510

ustcpxy:
the braces should be indented like below:
if ( a < b )
{
    // code
}

### DIFF
--- a/modules/config/default/config.el
+++ b/modules/config/default/config.el
@@ -71,7 +71,7 @@
       (c-toggle-auto-newline -1)
       (setq c-tab-always-indent nil
             c-electric-flag nil)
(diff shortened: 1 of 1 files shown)


hlissner:
Try this instead:

```elisp
(setq-hook! 'c-mode-common-hook electric-indent-inhibit nil)
```

Doom uses a combination of smartparens and electric-indent-mode to manage indentation (universally, for all braced languages), so I'd like to use that instead of cc-mode's.

hlissner:
I've incorporated electric-indentation across the board, for a number of languages included the C family, so I'll consider this issue resolved and close this. Thanks for bringing it to my attention.

PR #1207, opened 2019-02-27
Use configured variables for rtags binary names

shanavas786:
`rtags` supports custom binary names.


hlissner:
Looks good. Thanks for your help!

PR #1209, opened 2019-02-28
Fix flycheck-posframe feature detection

flatwhatson:
The auto-enabling of `flycheck-posframe` was not working due to broken feature check.


flatwhatson:
Thanks! :bowing_man:

PR #1221, opened 2019-03-06
Improve +ivy/switch-workspace-buffer

flatwhatson:
Allow current buffer to be selected when switching other window, so you
can quickly get a second window to the current buffer.

Preselect the most recently selected buffer, so repeated C-x b RET will
switch between your two most recent buffers.


hlissner:
Great idea! Thanks again for your continued help; it is much appreciated!

flatwhatson:
Glad to help, thanks for the awesome emacs config!

PR #1222, opened 2019-03-06
Fix cc fontification for constants with numbers

flatwhatson:
The regex wasn't matching numbers; `FOO_BAR` would be pretty, but `FOO_BAR_1` would not.


hlissner:
Good catch! Thanks!

PR #1228, opened 2019-03-08
Tidy up +ivy-buffer-preview functionality

flatwhatson:
Let-binding `ivy-use-virtual-buffers` is neater and more robust than the previous solution using some copy-pasta.  Calling the unwind routine will become important once counsel is patched to restore buffer-list.

See: https://github.com/abo-abo/swiper/pull/1954


hlissner:
Thanks again!

PR #1231, opened 2019-03-08
Add editor/expand-region module

flatwhatson:
An essential module for those who haven't fallen.


hlissner:
I hadn't thought of non-evil users (sorry!), so I brought it back, along with a number of improvements to the non-evil experience in Doom, so this PR won't be necessary. Thanks for the help though!

flatwhatson:
Awesome, greatly appreciated!

PR #1232, opened 2019-03-08
Fix org-pdfview initialising too early

flatwhatson:
Fixes startup error when org and pdf modules enabled:

Error in a Doom module: modules/lang/org/config.el, (void-variable org-file-apps)

### DIFF
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -83,6 +83,7 @@
     :config (evil-org-agenda-set-keys)))
 
 (def-package! org-pdfview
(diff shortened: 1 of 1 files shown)


hlissner:
Sorry for the sudden onslaught of bugs. Shouldn't have pushed just before going to bed! This required a different fix; loading org-pdfview `:after org` loads it too soon. We use those modifications to `org-file-apps` to lazy load it.

In any case, thanks for bringing this to my attention (and for your vigilance).

PR #1233, opened 2019-03-08
Fix doom-leader-alt-key binding for +emacs-bindings

flatwhatson:
Since a4c0bc27 it's no longer possible to just `(setq doom-leader-alt-key ...)` from a module or even `$DOOMDIR/init.el`.  This is a workaround so non-vimmers aren't stuck with `M-SPC` in the interim.


hlissner:
Ah, you're right. This affected both evil and non-evil users, and this fix only addresses it for the latter. Those keybinds need to be set later in the startup process. This should be fixed as of c9e7d27.

Thanks again for your help!

PR #1243, opened 2019-03-12
add expand region key bindings

chengyi:
the expand region key bindings is missing, just add it


hlissner:
expand-region was removed. See #1256 for details.

PR #1247, opened 2019-03-12
Maximize window vertically/horizontally

jdnavarro:
The default behavior for the windows key bindings is to *enlargen* instead of replacing all windows in the given direction. For me it sounds more sensible to do replacement but if you don't agree it's fine with me. I can keep this functionality in my private separate configuration.

I copied the maximize functions from Spacemacs. Apparently, they were copied from a blog post that no longer exists.

The second commit just adds the missing register key mapping for helm.

### DIFF
--- a/core/autoload/ui.el
+++ b/core/autoload/ui.el
@@ -103,7 +103,7 @@ See `display-line-numbers' for what these values mean."
     (save-buffers-kill-emacs)))
 
 ;;;###autoload
-(defun doom/window-zoom ()
+(defun doom/window-maximize-buffer ()
   "Close other windows to focus on this one. Activate again to undo this. If the
 window changes before then, the undo expires.
 
@@ -119,7 +119,7 @@ Alternatively, use `doom/window-enlargen'."
 ;;;###autoload
 (defun doom/window-enlargen ()
   "Enlargen the current window to focus on this one. Does not close other
-windows (unlike `doom/window-zoom') Activate again to undo."
+windows (unlike `doom/window-maximize-buffer') Activate again to undo."
   (interactive)
   (setq doom--window-enlargened
         (if (and doom--window-enlargened
@@ -142,6 +142,24 @@ windows (unlike `doom/window-zoom') Activate again to undo."
                 (maximize-window))
             (maximize-window))
           t)))
+
+;;;###autoload
+(defun doom/window-maximize-horizontally ()
+  "Delete all windows to the left and right of the current window."
+  (interactive)
+  (require 'windmove)
+  (save-excursion
+    (while (ignore-errors (windmove-left)) (delete-window))
+    (while (ignore-errors (windmove-right)) (delete-window))))
+
+;;;###autoload
+(defun doom/window-maximize-vertically ()
+  "Delete all windows above and below the current window."
+  (interactive)
+  (require 'windmove)
+  (save-excursion
+    (while (ignore-errors (windmove-up)) (delete-window))
+    (while (ignore-errors (windmove-down)) (delete-window))))
 
 ;;;###autoload
 (defun doom/reload-font ()
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -95,11 +95,14 @@
         "L"       #'+evil/window-move-right
         "C-S-w"   #'ace-swap-window
         ;; Window undo/redo
+        (:prefix "m"
+            "m"   #'doom/window-maximize-buffer
+            "v"   #'doom/window-maximize-vertically
+            "s"   #'doom/window-maximize-horizontally)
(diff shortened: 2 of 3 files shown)


yuhan0:
Nice :) I have the same functions in my private config also copied from Spacemacs, and they're one of my more frequently used window commands. 

I think you have the bindings swapped around though, `"|"` should be maximize-vertically and `"_"` horizontally.

yuhan0:
On a similar note, `"SPC w m"` for toggling full maximization might also be a useful command to borrow from Spacemacs:
```
(defun doom/toggle-maximize-buffer ()
  "Toggle between maximized buffer and previous window configuration"
  (interactive)
  (save-excursion
    (if (and (= 1 (length (window-list)))
             (assoc ?_ register-alist))
        (jump-to-register ?_)
      (window-configuration-to-register ?_)
      (delete-other-windows))))
```

jhacksworth:
This looks great, I'm going to use it :)

The only question I have is with `doom/toggle-maximize-buffer`, which looks to be the same as [doom/window-zoom](https://github.com/hlissner/doom-emacs/blob/f7a92ac7d50120838e7ec59620576c07332a7138/core/autoload/ui.el#L106).

core/autoload/ui.el
```diff
@@ -143,6 +143,39 @@ windows (unlike `doom/window-zoom') Activate again to undo."
             (maximize-window))
           t)))
 
+;;;###autoload
+(defun doom/toggle-maximize-buffer ()
```
hlissner: `doom/toggle-maximize-buffer` is identical to `doom/window-zoom`. Please remove one or the other.

modules/completion/helm/config.el
```diff
@@ -52,6 +52,7 @@ be negative.")
         [remap projectile-switch-project] #'helm-projectile-switch-project
         [remap projectile-switch-to-buffer] #'helm-projectile-switch-to-buffer
         [remap recentf-open-files]        #'helm-recentf
+        [remap evil-ex-registers]         #'helm-register
```
hlissner: `evil-ex-registers` doesn't appear to exist. Did you mean `evil-show-registers`? In any case, this isn't relevant to this PR, please remove it.

core/autoload/ui.el
```diff
@@ -143,6 +143,39 @@ windows (unlike `doom/window-zoom') Activate again to undo."
...
+      (window-configuration-to-register ?_)
+      (delete-other-windows))))
+
+;;;###autoload
+(defun doom/window-maximize-horizontally ()
+  "Delete all windows to the left and right of the current window."
+  (interactive)
+  (require 'windmove)
+  (save-excursion
+    (while (condition-case nil (windmove-left) (error nil))
```
hlissner: I suggest using `(ignore-errors ...)` rather than `(condition-case nil ... (error nil))`.

modules/config/default/+evil-bindings.el
```diff
@@ -101,6 +101,9 @@
         "L"       #'+evil/window-move-right
         "C-S-w"   #'ace-swap-window
         ;; Window undo/redo
+        "m"       #'doom/toggle-maximize-buffer
+        "_"       #'doom/window-maximize-horizontally
```
hlissner: Both `|` and `_` already exist on evil-window-map (bound to `evil-window-set-width` and `evil-window-set-height` respectively) and will need different keybinds. How about making `m` into a prefix?

```elisp
(:prefix "m"
  "m" #'doom/toggle-maximize-buffer
  "v" #'doom/window-maximize-vertically
  "h" #'doom/window-maximize-horizontally)
```

core/autoload/ui.el
yuhan0: Ah, I didn't realise that `doom/window-zoom` was a thing already - since it's a toggle, would it make sense to have "toggle" in the command name and move it under the `SPC t` prefix?
`SPC t m` -> toggle maximized window
`SPC t M` -> toggle maximized frame
`SPC t F`(existing) -> toggle fullscreen frame

jdnavarro:
I removed the helm-register commit, I'll submit in a different PR.

modules/config/default/+evil-bindings.el
```diff
@@ -95,11 +95,15 @@
         "L"       #'+evil/window-move-right
         "C-S-w"   #'ace-swap-window
         ;; Window undo/redo
+        (:prefix "m"
+            "m"   #'doom/window-maximize-buffer
+            "v"   #'doom/window-maximize-vertically
+            "s"   #'doom/window-maximize-horizontally
```
jdnavarro: I used here `s` instead of `h` to keep it consistent with `evil-window-split` (`SPC w s`) which splits the window horizontally.

hlissner:
Looks good to me! Thanks for your help.

PR #1249, opened 2019-03-12
Fix autoloading helpful for doom/describe-symbol

flatwhatson:
We use `helpful--read-symbol` which isn't already autoloaded.


hlissner:
Thanks again!

PR #1250, opened 2019-03-13
Don't install robe if lsp is enabled

qhuyduong:


hlissner:
Good catch! Thanks!

PR #1259, opened 2019-03-15
Re-add missing bindings for rspec-mode

qhuyduong:
Last PR has wiped out basic bindings in rspec-mode.
This PR re-adds theme.


hlissner:
Good catch. Thanks!

PR #1266, opened 2019-03-16
Updates to evil-mc config

yuhan0:
`evil-mc-undo-last-added-cursor` was introduced upstsream and bound to "gru" by
default, with the `evil-mc-undo-all-cursors` binding shifted to "grq"

Bar cursors don't display well by default on Mac and Windows, see
https://github.com/gabesoft/evil-mc/issues/49


hlissner:

PR #1267, opened 2019-03-16
Fixes for doom/kill-this-buffer-in-all-windows

flatwhatson:
This was erroring when run inside a popup (eg. `M-x ielm` `C-x k`) due
to a missing check for `window-live-p`.  We don't need to do anything if
the window has already gone away.

The real vs unreal check was backwards; we should only try
`previous-buffer` if the current buffer is *not* real.


hlissner:
Good catch, thanks!

PR #1270, opened 2019-03-17
[WIP] Organize local leader maps

yuhan0:
This PR is an attempt to standardize local leader keymaps and prefixes across different language modules, which currently have some inconsistencies.

Since these changes might lead to breaking of workflows for some users, I'll be marking it as a WIP draft and pushing commits incrementally for review and discussion :)

Summary of changes:
- `"m"` prefix for macroexpansion across various lisps 
  - `"me"` to expand current / last sexp
  - Sole command in elisp given its own binding instead of being prefixed
- `"g"` prefix as "goto" across language major modes
  - This is a breaking change for csharp and python which used the "f" prefix
- `"b"` prefix as "build" in programming language modes
  - Breaking change for common-lisp which used "c"
- `"e"` prefix as "eval", with the following conventions:
  `eb` ~ eval entire buffer
  `ed` ~ eval top-level definition / defun
  `ee` ~ eval last expression
  `er`~ eval region
  `eu` ~ un-define definition
  - breaking change for the Ruby and Racket modules which used "s" as the prefix.
- new: `"c"` prefix as REPL "connection"
  - only applied in `clojure-mode` so far, but could be extended to other languages
  - Leaves the "r" prefix for "refactor" instead of sometimes meaning "repl"
  - Conceptually separates common REPL related commands into ones involving evaluating and sending things: "e" and those involving managing sessions and connections with language servers: "c".

### DIFF
--- a/modules/lang/agda/config.el
+++ b/modules/lang/agda/config.el
@@ -13,6 +13,7 @@
   :config
   (map! :map agda2-mode-map
         :localleader
+        ;; FIXME organize according to conventions?
         "?"   #'agda2-show-goals
         "."   #'agda2-goal-and-context-and-inferred
         ","   #'agda2-goal-and-context
--- a/modules/lang/clojure/config.el
+++ b/modules/lang/clojure/config.el
@@ -75,15 +75,27 @@
               "E" #'cider-macroexpand-all)
             (:prefix ("n" . "namespace")
               "n" #'cider-browse-ns
-              "N" #'cider-browse-ns-all)
-            (:prefix ("r" . "repl") ; FIXME r ~ refactor
+              "N" #'cider-browse-ns-all
+              "r" #'cider-ns-refresh)
+            (:prefix ("c" . "connection")
+              (:prefix ("l" . "link")
+                "b" 'sesman-link-with-buffer
+                "d" 'sesman-link-with-directory
+                "p" 'sesman-link-with-project
+                "u" 'sesman-unlink)
               "n" #'cider-repl-set-ns
               "q" #'cider-quit
-              "r" #'cider-refresh
               "R" #'cider-restart
               "b" #'cider-switch-to-repl-buffer
               "B" #'+clojure/cider-switch-to-repl-buffer-and-switch-ns
-              "c" #'cider-repl-clear-buffer)))
+              "c" #'cider-repl-clear-buffer
+              "i" #'sesman-info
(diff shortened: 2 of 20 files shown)


yuhan0:
I just marked this PR ready for review although it shouldn't be merged in its current state- seems like the recently merged #1288 conflicts with changes I made in the `common-lisp` module. 
Before I start rebasing this branch, I'd like to know whether it makes sense to have these standardizations throughout Doom? 

Pinging @mfiano since this would directly revert some prefix label changes from #1288 as well as reorganize some bindings, and I'm not a Common Lisp user myself. Thanks!

mfiano:
As maintainer of the Spacemacs `common-lisp-sly` layer as well, I just want to say that I'm against standardizing leader prefixes. It was a nightmare for the users of that layer. For example, you don't "build" a file or function, you expect it to be mnemonic. Common Lisp *is* a bit different than a lot of other languages. Also, "build" sort of implies the build infrastructure, like what CL calls loading a file, and retrieving sources from Quicklisp, etc. "b" is but one of my complaints though.

yuhan0:
Thanks for the input! I'm coming from Spacemacs too and these changes were definitely inspired by the prefix conventions there.

There's definitely a tension between maintaining mnemonics within a particular mode and having the same bindings map to roughly the same actions across different modes. 

I think this makes sense for certain things like evaluating functions or running tests - it's nice to just rely on muscle memory and type `SPC m e d` without having to context-switch depending on whether the major mode calls it "sending" a "function" or "evaluating" a "definition", even if the mnemonics no longer match up perfectly.

For other areas where the language mode has a more unusual or opinionated stance, I can see why this would be an overreach - E.g. I think the Coq / Idris / Agda configs should probably be left alone since their conceptual model is so different from other languages.

Is Common Lisp's "compile" really that different from the idea of "build"? In my mind they're nearly synonyms so I felt it was okay to clump it into the "b" prefix for consistency, as long as the language didn't have a distinct concept of "build" which would conflict.

hlissner:
When terminology and conventions can vary so much across languages, perhaps it makes sense that each module follow their own, rather than a common one. At most, we could impose conventions across _classes_ of languages, say, C-style languages -- or lisps, or functional languages.

A global convention probably makes sense only to us maintainers of an Emacs config, looking down at the forest while the end user deals with only a few, familiar trees, so to speak.

PR #1273, opened 2019-03-18
Choose which spell checker flyspell will use

UndeadKernel:
* Previously, a user could not choose which spell checker to use.
  Instead, the first one found was used.
* This PR adds the flags `+aspell` and `+hunspell` which allow a user to
  choose which spellchcker to use (even if both are installed).
* Report an error if spell checking is enabled but no spell checker
  is found.


hlissner:
And done. Thanks for bringing this to my attention!

PR #1277, opened 2019-03-21
Fix +ivy-buffer-preview no longer ignoring virtuals

flatwhatson:
Upstream updated `counsel--switch-buffer-update-fn` to check
`ivy--virtual-buffers` instead of `ivy-use-virtual-buffers`.


hlissner:
Thanks again!

PR #1283, opened 2019-03-25
[WIP]Add Module for terralang

aiverson:
A very simple conversion of the lua module to support terra.

Adds syntax highlighting, code execution, and basic autocompletion.

If someone has a better humorous language comment, I'd like to hear it.


hlissner:
I was a little unsure about this module (whether or not it should be part of `:lang lua`), but it looks good to me otherwise. Thanks for your help!

PR #1285, opened 2019-03-26
Pass 'f' through to `evil-surround`

gilbertw1:
Currently `evil-embrace` surrounds with the 'f' character instead of passing it through to `evil-surround` which specifically handles it to create a surrounding function.

An example is in scala mode. Currently if I were to highlight `test` and press `s f` then evil embrace would simply surround it with 'f': `ftestf`. After this update it would instead forward to `evil-surround` which would then prompt for a function name and surround it with a function call like `function(test)`.

I've added this as a general `prog-mode` hook because it does the correct thing in most C-based languages (scala, c, c++, java, javascript, etc) and I think people rarely actually want to surround something with 'f'. I've added an exception for lisp modes which you've already properly handled via `evil-surround`.

### DIFF
--- a/modules/feature/evil/config.el
+++ b/modules/feature/evil/config.el
@@ -184,6 +184,7 @@ line with a linewise comment.")
 
 (def-package! evil-embrace
   :commands (embrace-add-pair embrace-add-pair-regexp)
+  :hook (prog-mode . +evil|embrace-prog-mode-hook)
   :hook (LaTeX-mode . embrace-LaTeX-mode-hook)
   :hook (org-mode . embrace-org-mode-hook)
   :hook ((ruby-mode enh-ruby-mode) . embrace-ruby-mode-hook)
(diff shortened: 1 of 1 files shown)


hlissner:
Odd. Embrace [defines an global `f` pair](https://github.com/cute-jumper/embrace.el/blob/master/embrace.el#L499-L518) at the end of `embrace--setup-defaults`:

```elisp
(defun embrace--setup-defaults ()
  (dolist (pair '((?\( . ("(" . ")"))
                  (?\) . ("( " . " )"))
                  (?\{ . ("{" . "}"))
                  (?\} . ("{ " . " }"))
                  (?\[ . ("[" . "]"))
                  (?\] . ("[ " . " ]"))
                  (?< . ("<" . ">"))
                  (?> . ("< " . " >"))
                  (?\" . ("\"" . "\""))
                  (?\' . ("\'" . "\'"))
                  (?` . ("`" . "`"))))
    (embrace-add-pair (car pair) (cadr pair) (cddr pair)))
  (embrace-add-pair-regexp ?t "<[^>]*?>" "</[^>]*?>" 'embrace-with-tag
                           (embrace-build-help "<tag attr>" "</tag>"))
  (embrace-add-pair-regexp ?f "\\(\\w\\|\\s_\\)+?(" ")" 'embrace-with-function
                           (embrace-build-help "function(" ")")))

(embrace--setup-defaults)
(make-variable-buffer-local 'embrace--pairs-list)
```

So it _should_ be safe to leave `f` to embrace, rather than passing it on to evil-surround, but I can reproduce the behavior you're seeing. Perhaps this is a bug upstream?

gilbertw1:
Ah, makes sense! I'll close this PR then.

Thanks!

PR #1286, opened 2019-03-28
Don't open `doom-fallback-buffer` on del workspace

gilbertw1:
The `current-buffer` wasn't being updated after switching workspaces and
was therefore failing the `doom-buffer-frame-predicate` check. As a result
deleting a workspace resulted in the previous window's buffer being 
replaced with the doom dashboard.

Replacing it with `window-buffer` seems to target the intended buffer and 
prevents this from occuring.


hlissner:
Thanks for your help!

PR #1287, opened 2019-03-28
Changed some wrong common-lisp bindings

mfiano:
The Sly documentation which these were originally based off of is out of
date, causing some of the bindings to have no longer defined functions.


hlissner:
Nice work! Thanks again for your help.

PR #1288, opened 2019-03-28
Fix common-lisp local leader prefix labels

mfiano:


hlissner:
It is annoying that this is necessary to get which-key to behave, but I understand that it's a necessary evil. Thanks again for your help!

PR #1289, opened 2019-03-29
Add Kotlin module

filalex77:
Simple Kotlin module, modelled after the `lang/rust` module.

### DIFF
--- a/modules/lang/kotlin/autoload.el
+++ b/modules/lang/kotlin/autoload.el
@@ -1,7 +1,7 @@
 ;;; lang/kotlin/autoload.el -*- lexical-binding: t; -*-
 
 ;;;autoload
-(defun +kotlin/locate-gradlew-file ()
+(defun +kotlin-locate-gradlew-file ()
   "Gradlew file location for this project."
   (locate-dominating-file buffer-file-name "gradlew"))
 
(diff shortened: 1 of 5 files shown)


filalex77:
@hlissner I think I also found out how to do `compile` from project directory. You probably needed that in `lang/rust` :)

hlissner:
This looks ready to merge, except for one small problem: `+kotlin/locate-gradlew-file` uses the naming convention for interactive commands, but is not interactive. Please rename it `+kotlin-locate-gradlew-file`.

modules/lang/kotlin/autoload.el
```diff
@@ -0,0 +1,15 @@
+;;; lang/kotlin/autoload.el -*- lexical-binding: t; -*-
+
+;;;autoload
+(defun +kotlin/locate-gradlew-file ()
```
hlissner: `+kotlin/locate-gradlew-file` uses the naming convention for interactive commands, but is not interactive. Please rename it to `+kotlin-locate-gradlew-file`.

modules/lang/kotlin/doctor.el
```diff
@@ -0,0 +1,4 @@
+;;; lang/kotlin/doctor.el -*- lexical-binding: t; -*-
+
+(unless (executable-find "ktlint")
+  (warn "ktlint not found. flycheck-kotlin won't work."))
```
hlissner: Please use the `warn!` macro instead of `warn`.

filalex77:
You're welcome, and thanks for the review!

hlissner:
And done! Thanks again!

PR #1290, opened 2019-03-29
view gists in browser after posting

mfiano:

### DIFF
--- a/modules/tools/gist/config.el
+++ b/modules/tools/gist/config.el
@@ -6,7 +6,8 @@
 
 (after! gist
   (set-evil-initial-state! 'gist-list-mode 'normal)
-  (setq gist-view-gist t)
+  (when (featurep! +view)
+    (setq gist-view-gist t))
 
   (defun +gist*list-render (orig-fn &rest args)
     (funcall orig-fn (car args) t)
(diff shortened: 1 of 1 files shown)


hlissner:
I don't agree with this PR for two reasons:

1. Flags shouldn't be used for trivial customizations,
2. I don't believe this represents a reasonable default. I would rather any mechanism that opens an external program be opt-in, unless doing so were its purpose (e.g. `+lookup/online`). Especially so when it goes against the package's default.

mfiano:
I use this feature to preview for two different reasons:

1) To preview GitHub's markup language rendering, like Org and Markdown. Another gist updates my browser.

2) I have it open my browser, so that I can link to a line number when communicating with someone online about code we are writing.

I agree about the flags. As for #2, the flag is supposed to make it non-default. If not then this is a case where myself and others should set variables themselves in `after!` in their own config.

hlissner:
> If not then this is a case where myself and others should set variables themselves in after! in their own config.

That's what I was thinking. Not to mention, a simple, literal assignment doesn't require `after!`:

```elisp
(setq gist-view-gist t)
```

> To preview GitHub's markup language rendering, like Org and Markdown. Another gist updates my browser.

That does make sense. Perhaps it would be better UX for `gist-list` to open to place the cursor on the newly created gist? Then the user would have the full range of control over what to do with the new gist (`go` to open it in the browser, `d` to delete it, `y` to yank its url, `e` to edit its description, etc).

Issue #1190, opened 2019-02-22
calfw grid does not extend to the whole buffer


Pentaquark1:
I tried the =Calendar App. This is what it looks like for me:
![screen](https://user-images.githubusercontent.com/42049801/53277593-e6051380-3704-11e9-87b7-9921c1591158.png)

Unfortunately the grid does not dynamically extend to the whole buffer, as would be expected from the vanilla version.
This also occurs if one does not use the calendar app, but instead only installs the calfw package. 
It seems like some other package is messing with the layout here, any idea how to fix this?

hlissner:
Are you able to reproduce this in vanilla Emacs? Here's how you can easily test it (from GUI Emacs):

1. Open the vanilla sandbox (`SPC h E` or `M-x doom/open-vanilla-sandbox`)
2. Type in:

   ```elisp
   (require 'calfw)
   ```
3. Press `C-c C-c` to open the vanilla instance of Emacs
4. Open calfw

haji-ali:
I have the opposite problem where the calendar is so big that part of the month is hidden. I tried with vanilla emacs and didn't have an issue there.

Pentaquark1:
This is really confusing. If I follow the above steps, and the problem with the static grid persists.
I tried it yesterday without any doom, just a fresh new .emacs.d and getting calfw from melpa.
There the grid will adept to buffer size..

hlissner:
As of #1294, this has been resolved. Let me know if that isn't the case!

Thanks to @jscott22 for looking into this.

PR #1294, opened 2019-03-30
Create calfw calendar with correct grid width

jscott22:
## Description
This PR addresses issue #1190 regarding calfw grid size when opening the calendar with `=calendar` from the calendar module.

## Problem
* Calfw calendar displays with an incorrect grid size
* Caused by [the new calendar workspace](https://github.com/hlissner/doom-emacs/blob/be6efd8dbff18c29953c4af8bfef88b8091added/modules/app/calendar/autoload.el#L18) displaying the doom splash screen with window margins
* Calfw [uses the selected window](https://github.com/kiwanami/emacs-calfw/blob/03abce97620a4a7f7ec5f911e669da9031ab9088/calfw.el#L784) to determine the size of the calendar grid

## Solution
* Selects the scratch buffer before creating the calendar
* Calfw generates the calendar grid based off of the resulting correct window margins

## Verification
* Calendar displays correctly with both workspace feature enabled and disabled
* Calendar displays correctly  while having the distraction free `writeroom-mode` active

### DIFF
--- a/modules/app/calendar/autoload.el
+++ b/modules/app/calendar/autoload.el
@@ -16,6 +16,7 @@
   (if (featurep! :feature workspaces)
       (progn
         (+workspace-switch "Calendar" t)
+        (doom/switch-to-scratch-buffer)
         (+calendar--init)
         (+workspace/display))
     (setq +calendar--wconf (current-window-configuration))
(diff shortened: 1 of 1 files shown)


hlissner:
Good work! Thanks for the help and for figuring out this elusive bug!

PR #1296, opened 2019-03-31
Make Sly REPL a left popup

mfiano:


hlissner:
I'm hesitant to accept this PR, for consistency's sake. Everywhere else, popup REPLs are opened in pop-ups at the bottom of the frame.

Folks that want to switch to the REPL buffer (rather than open it in a popup) have `+eval/open-repl-same-window` (bound to `SPC o R` for evil users -- though I notice there is no similar keybind for non-evil users in +emacs-bindings.el. Perhaps it should be `C-c C-S-z`?).

There's also `+popup/raise` (bound to `C-~` for all users) for "promoting" a popup to a real window.

mfiano:
That is a respectable opinion. It just makes it really difficult to code with less than 10 lines, constantly raising it, growing it, etc, when all you want is a static popup distinct from all of your other code that has sufficient vertical space.

mfiano:
I guess this is another snippet I will have to give to Sly users for their own config.

hlissner:
How about making it taller?

```diff
   (set-popup-rules!
-    '(("^\\*sly-mrepl"       :vslot 2 :quit nil :ttl nil)
+    '(("^\\*sly-mrepl"       :vslot 2 :size 15 :quit nil :ttl nil)
       ("^\\*sly-compilation" :vslot 3 :ttl nil)
       ("^\\*sly-traces"      :vslot 4 :ttl nil)))

mfiano:
That is still far too small, unless you want to look at multiple pages every time you compile a function and there is compiler output. The REPL includes code, so it should have sufficient vertical space, and not sacrifice source buffers' vertical space in doing so.

hlissner:
Ultimately, I increased the size of popups to .3 -- I realize this is still not large enough, but I've decided that it is `SPC o R`'s responsibility to produce a permanent REPL, not `SPC o r` (which is the one that is affected by the popup rule). These are bound to `C-c C-S-z` and `C-c C-z`, respectively for non-evil users.

In any case, I'll consider this resolved.

rgrinberg:
A configuration option that I think would be useful is to have all repl's be ignored by popup. I also tend to find myself constantly raising existing repl's. I think it would be less of a hassle this way.

hlissner:
@rgrinberg That is what `SPC o R` is for.

PR #1299, opened 2019-03-31
Escape key regexps when building which-key leader prefix

jochu:
Fixes a bug where the regexp for which-key leader prefixes aren't escaped. This it makes it possible to use regex special characters leader keys, such as `C-\`.


hlissner:

PR #1300, opened 2019-04-01
Add vterm-mode to evil-escape-excluded-major-modes

zeronone:
`jk` doesn't work as expected in `vterm`. 

Do I need to also check for `vterm` feature flag?


hlissner:
Looks good to me. No need for the feature check. Thanks for the help!

PR #1302, opened 2019-04-01
Use variable for `fd-find` binary name

shanavas786:
Binary name is not `fd` in some distribution eg, it is `fdfind` in debian


hlissner:

PR #1321, opened 2019-04-11
Enable go auto fmt before save

ethercflow:


hlissner:
`gofmt-before-save` is superceded by the `:editor format` module (and its `+onsave` flag), which provides better gofmt (and/or goimports) support, so I cannot accept this PR. I will, however, mention this in the :lang go readme. Thanks for bringing this to my attention!

ethercflow:
That's great.

PR #1326, opened 2019-04-13
feat: org-habit

ragone:
Added a simple way to make the org-habit consistency graphs align to the
right margin as well as being responsive to the width of the window.

This is added as a feature to the org module.

### DIFF
--- a/modules/lang/org/+habit.el
+++ b/modules/lang/org/+habit.el
@@ -1,7 +1,5 @@
 ;;; lang/org/+habit.el -*- lexical-binding: t; -*-
 
-(require 'org-habit)
-
 (defvar +org-habit-graph-padding 2
   "The padding added to the end of the consistency graph")
 
@@ -11,8 +9,9 @@
 (defvar +org-habit-graph-window-ratio 0.3
   "The ratio of the consistency graphs relative to the window width")
 
-(defun +org-habit-resize-graph()
+(defun +org-habit|resize-graph()
   "Right align and resize the consistency graphs based on `+org-habit-graph-window-ratio'"
+  (require 'org-habit)
   (let* ((total-days (float (+ org-habit-preceding-days org-habit-following-days)))
          (preceding-days-ratio (/ org-habit-preceding-days total-days))
          (graph-width (floor (* (window-width) +org-habit-graph-window-ratio)))
@@ -26,4 +25,4 @@
     (setq-local org-habit-following-days following-days)
     (setq-local org-habit-graph-column graph-column-adjusted)))
 
-(add-hook! 'org-agenda-mode-hook #'+org-habit-resize-graph)
+(add-hook 'org-agenda-mode-hook #'+org-habit|resize-graph)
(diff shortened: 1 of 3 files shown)


modules/lang/org/+habit.el
```diff
@@ -0,0 +1,29 @@
+;;; lang/org/+habit.el -*- lexical-binding: t; -*-
+
+(require 'org-habit)
```
hlissner: Could this be moved into `+org-habit-resize-graph`? I don't think there's any reason to eagerly load org-habit this soon.

modules/lang/org/+habit.el
```diff
@@ -0,0 +1,29 @@
...
+(defvar +org-habit-graph-padding 2
+  "The padding added to the end of the consistency graph")
+
+(defvar +org-habit-min-width 30
+  "Hides the consistency graph if the `org-habit-graph-column' is less than this value")
+
+(defvar +org-habit-graph-window-ratio 0.3
+  "The ratio of the consistency graphs relative to the window width")
+
+(defun +org-habit-resize-graph()
```
hlissner: Please rename this `+org-habit|resize-graph`. This is the convention for functions in Doom meant to be used as a hook.

modules/lang/org/+habit.el
```diff
@@ -0,0 +1,29 @@
...
+         (following-days (- graph-width preceding-days))
+         (graph-column (- (window-width) (+ preceding-days following-days)))
+         (graph-column-adjusted (if (> graph-column +org-habit-min-width)
+                                    (- graph-column +org-habit-graph-padding)
+                                  nil)))
+    (setq-local org-habit-preceding-days preceding-days)
+    (setq-local org-habit-following-days following-days)
+    (setq-local org-habit-graph-column graph-column-adjusted)))
+
+(add-hook! 'org-agenda-mode-hook #'+org-habit-resize-graph)
```
hlissner: Please change `add-hook!` to `add-hook`. None of `add-hook!`'s benefits are being used here.

modules/lang/org/+habit.el
ragone: Moved to `+org-habit|resize-graph`

modules/lang/org/+habit.el
ragone: Renamed to `+org-habit|resize-graph`

modules/lang/org/+habit.el
ragone: Using `add-hook- instead of `add-hook`.

ragone:
The changes you suggested have been implemented. Thanks! :)

1. As you can see, I have rebased my branch - Do you prefer merge commits?
2. Also, do you prefer to squash commits before merging?

hlissner:
Looks good. Thanks for the help!

> 1. As you can see, I have rebased my branch - Do you prefer merge commits?

I prefer rebasing, unless multiple people are working on the PR.

> 2. Also, do you prefer to squash commits before merging?

I generally don't. I do want to avoid convoluting development history with trivial or fragmented commits, but I'll bring it up if I think it's a problem. I don't think it is with this PR.

PR #1327, opened 2019-04-14
[WIP] Implement fsharp :lang module

TOTBWF:
(WIP) This is a pretty simple F# language module

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -107,7 +107,7 @@
        ;;elm               ; care for a cup of TEA?
        emacs-lisp        ; drown in parentheses
        ;;ess               ; emacs speaks statistics
-       ;; fsharp           ; ML stands for Microsoft's Language
+       ;;fsharp           ; ML stands for Microsoft's Language
        ;;go                ; the hipster dialect
        ;;(haskell +intero) ; a language that's lazier than I am
        ;;hy                ; readability of scheme w/ speed of python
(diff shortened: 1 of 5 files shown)


TOTBWF:
As a side question, should I be using `after!` here, or `def-package!`? I'm a bit of an emacs/doom newcomer so I wasn't quite sure.

modules/lang/fsharp/doctor.el
```diff
@@ -0,0 +1,4 @@
+;;; lang/fsharp/doctor.el -*- lexical-binding: t; -*-
+
+(unless (-any #'fsharp-mode--executable-find '("fsharpc" "fsc"))
```
hlissner: 1. Please use `cl-some` instead of `-any`. Dash.el is not a core dependency in Doom.
2. Wrap this in `(when (require 'fsharp-mode nil t) ...)`, to ensure the package is loaded so that `fsharp-mode--executable-find` is available.

init.example.el
```diff
@@ -96,6 +96,7 @@
        ;;elm               ; care for a cup of TEA?
        emacs-lisp        ; drown in parentheses
        ;;ess               ; emacs speaks statistics
+       ;; fsharp           ; ML stands for Microsoft's Language
```
hlissner: Please remove the leading space before `fsharp`

TOTBWF:
I still need to figure out some issues with the REPL, but once that is done this should be good to go

TOTBWF:
Finally got around to fixing things. This should be good to go

hlissner:
Looks good to me. The readme could do some work, but I'll look into improving it another time. Thanks again for your help!

PR #1333, opened 2019-04-16
Fix wrong-number-of-args in project-search-from-cwd

flatwhatson:


hlissner:
Good catch, but could you do the same for `+helm/project-search-from-cwd`, as well as remove the `directory` argument for both commands?

flatwhatson:
Done! o7

hlissner:
Perfect! Thanks again for your help!

PR #1336, opened 2019-04-18
Document company +tng flag

thiagokokada:
Just documenting `+tng` flag that I saw looking at `company/config.el` code and it wasn't documented before.


hlissner:
I think this should be renamed to `+tab`. `+tng` isn't obvious, and was named that to reflect the package that provides the feature (`company-tng`). What do you think?

thiagokokada:
I am fine either way. `+tab` may be more newbie friendly, however people that knows more about `company` maybe more familiar with `+tng` (since this is an `company` official extension).

BTW, `tng` means "tab and go" if you look in the `company-tng` documentation.

hlissner:
I'll stick to `+tng` for now. Thanks for the help!

PR #1337, opened 2019-04-18
julia file association

isentropic:
Add `julia-mode` association with `*.jl` files


hlissner:
This isn't necessary. The `julia-mode` package [already does this for us](https://github.com/JuliaEditorSupport/julia-emacs/blob/master/julia-mode.el#L66-L67). Thanks for the PR however!

PR #1341, opened 2019-04-19
Add :editor objed module

flatwhatson:
Integrates `objed` with Doom Emacs to provide text-object manipulation for non-evil users.

See https://github.com/clemera/objed

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -49,6 +49,7 @@
        ;;(format +onsave)  ; automated prettiness
        ;;lispy             ; vim for lisp, for people who dont like vim
        multiple-cursors  ; editing in many places at once
+       ;;objed             ; text object editing for the innocent
        ;;parinfer          ; turn lisp into python, sort of
        rotate-text       ; cycle region at point between text candidates
        snippets          ; my elves. They type so I don't have to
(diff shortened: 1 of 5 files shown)


richerve:
Shouldn't be clear in the README or somewhere that this must not be use alongside evil? I understand the idea of being in the `editor` section as it is a sort of add-on for editing purposes but at the end is an alternative to evil at least in the modal editing aspect so not sure if it should be in `feature` instead.

flatwhatson:
@richerve Thanks, I've updated the README with a note regarding objed vs evil.  Let me know if you think this is clear enough, I'm happy to revise it.

Regarding whether this belongs in `:editor` or `:feature`, I'll leave that decision with @hlissner.

flatwhatson:
This is now ready for final review.

richerve:
Agree with the category decision to be up to @hlissner, in any case I think `:editor` works well.

hlissner:
The `:feature` category is due for removal in the next batch of commits. `:feature evil` will be in `:editor` from that point on. In any case, I'll review this over the weekend.

flatwhatson:
Rebased to fix conflict with modeline segment.

flatwhatson:
Updated to remove the dubious activation hacks as discussed.  Ready for review!

hlissner:
Looks good! Thanks again, and for your continued work around here. You've been a huge help!

flatwhatson:
Thanks :)

PR #1344, opened 2019-04-19
Add documentation for JS module keymaps

pedrovanzella:
I tried documenting all the keys available under the local leader for all the js modes here.


hlissner:

PR #1345, opened 2019-04-19
Update Go module README

ericsolomon:
Updated the Go README to correct the install link for gore.

See: https://github.com/motemen/gore#installation


hlissner:
Thanks for the help!

PR #1353, opened 2019-04-23
Correct Javascript module tide format mapping

ericsolomon:
Updated the format mapping for the Javascript module to use the correctly named tide-format.

See: https://github.com/ananthakumaran/tide#commands


hlissner:
Thanks!

PR #1355, opened 2019-04-23
add doctor.el for shell and update readme

chrunchyjesus:
* Added doctor.el check for shellcheck when flycheck is enabled.
* Updated dependency documentation in readme.org for sh lang.
* Removed commands section because none of the commands exists (and do not work either).


hlissner:
Sorry for the late response. Nice work and thank you for the help!

chrunchyjesus:
No problem.

Issue #1357, opened 2019-04-24
S-k does not work in clojure-mode


thiagokokada:
### Observed behavior

`S-k` (`K`) returns `Couldn't find documentation for 'thing'` instead of showing documentation for the thing at point. Calling `cider-doc` directly works, as also `SPC m h d` works too.

### Expected behavior

`S-k` would call `cider-doc` in `clojure-mode` and show documentation.

### Steps to reproduce

1. Open CIDER REPL
2. Eval any Clojure buffer
3. Use `S-k` to try to get documentation from something (like a  function)

### System information

<details>
<summary>Click to expand</summary>

- OS: gnu/linux (x86_64-pc-linux-gnu)
- Emacs: 26.1 (abr 08, 2019)
- Doom: 2.0.9 (HEAD -> develop, origin/develop 768d6e9c 2019-04-23 23:05:53 -0400)
- Graphic display: t (daemon: nil)
- System features: XPM JPEG TIFF GIF PNG RSVG SOUND DBUS GSETTINGS NOTIFY LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 MODULES THREADS LIBSYSTEMD
- Details:
  ```elisp
  env bootstrapper: nil
  elc count: 0
  uname -a:  Linux 4.19.35 #1-NixOS SMP Wed Apr 17 06:38:55 UTC 2019 x86_64
  modules:   (:feature eval (evil +everywhere) file-templates (lookup +docsets) snippets workspaces :completion company (ivy +fuzzy) :ui doom hl-todo indent-guides modeline nav-flash neotree ophints (popup +all +defaults) vc-gutter vi-tilde-fringe window-select :editor fold (format +onsave) multiple-cursors rotate-text :emacs dired electric eshell imenu term vc :tools ansible docker editorconfig flycheck magit make rgb terraform :lang cc clojure data elixir emacs-lisp hy javascript markdown nix (org +attach +babel +capture +export +present) python ruby (sh +zsh) web :config default)
[...]

PR #1359, opened 2019-04-24
Set Clojure lookup handlers for specific minor modes to avoid overwriting

gsnewmark:
Fixes issue #1357

Previously references lookup handler defined in `clj-refactor` package definition overwrote definition and documentation lookup handlers defined in `cider` package definition. To avoid this I've assigned handlers to their backing minor modes instead of the major Clojure mode.


hlissner:
Looks good. Thanks for the help!

PR #1362, opened 2019-04-25
Fix +ivy-buffer-icons counsel-projectile transformers

flatwhatson:
Our custom transformers might be added after `ivy-rich-mode` has registered its transformers with ivy.  We need to call `ivy-rich-reload` to ensure the latest transformers are installed.

This fixes the bug where icons are not present in `counsel-projectile-find-file` until toggling `ivy-rich-mode` off and on again.


hlissner:
1911f44 implements the +icons flag, using a combination of ivy-rich and all-the-icons-ivy, so this PR is no longer necessary. Thanks for the help and bringing it to my attention though!

flatwhatson:
Awesome, thanks!

PR #1363, opened 2019-04-25
fix rustc detection

chrunchyjesus:
There is no `rust` binary in Rust, but a `rustc` one for its compiler.


hlissner:
Good catch, thanks!

PR #1376, opened 2019-04-30
Update treemacs open command

Dragnucs:
Update treemacs open command toggle `+treemacs/toggle`

`+treemacs/open` is no longer defined.


hlissner:
Good catch, thanks!

PR #1379, opened 2019-05-03
One dot in wrong place

xvrdm:
Not sure what I am doing, but I cannot run the example with the `.` and the example on the previous line also use `:size` without a `.` before the size value.


hlissner:
Ah, good catch, thanks!

PR #1381, opened 2019-05-03
Add protobuf package to data module

pcasaretto:


modules/lang/data/packages.el
```diff
@@ -7,3 +7,4 @@
 (package! yaml-mode)
 (package! csv-mode)
 (package! dhall-mode)
+(package! protobuf-mode)
```
pcasaretto: worth it ordering this alphabetically?

hlissner: For now we can leave it as is.

hlissner:
Thanks!

PR #1382, opened 2019-05-04
readme: Fix broken URL references

tilpner:


hlissner:
Good catches. Thanks for your help!

PR #1387, opened 2019-05-05
[javascript] add prettier-js support

danielpa9708:
support formatting with prettier https://github.com/prettier/prettier, https://github.com/prettier/prettier-emacs

### DIFF
--- a/modules/lang/javascript/config.el
+++ b/modules/lang/javascript/config.el
@@ -62,6 +62,18 @@
         :localleader
         "S" #'+javascript/skewer-this-buffer))
 
+(def-package! prettier-js
+  :when (featurep! +prettier)
+  :init
+  (map!
+   :localleader
+   :map rjsx-mode-map
+   "p" #'prettier-js
+   :map js2-mode-map
+   "p" #'prettier-js
+   :map typescript-mode-map
+   "p" #'prettier-js))
+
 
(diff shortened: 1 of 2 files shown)


hlissner:
I'm afraid I cannot accept this PR, because it is redundant with the `:editor format` module (and its +onsave flag), which uses prettierjs by default and is the canonical way to enable code formatters in Doom. That said, I'm aware that this module lacks documentation and it is overdue for a rewrite; I'm working on it. In any case, thanks for the PR and for bringing it to my attention.

danielpa9708:
Ok, wasn't aware of it. I'll see what I can do to integrate `:editor format` with prettier in my config

AlexXiong97:
@danielpza do you have any idea how to have the prettier read from `.prettierc` config for each project with `:editor format` enabled? thanks

danielpza:
@AlexXiong97 I think it already does. I don't get what you are trying to do? Do you want to format with prettier only if the project has a .prettierrc ?

AlexXiong97:
@danielpza sorry, you are right. It already does.

alrayyes:
Is there any special extra setup for this? When i run `+format/buffer` using `org-edit-special` I get the following error:

`Don't know how to format js-mode code`

As far as i can tell formatting works properly with full TypeScript files.

PR #1389, opened 2019-05-06
modules/ui: Add zoom-frm module

thiagokokada:
This module adds `zoom-frm` support.

The hydra is based on the `doom-text-zoom-hydra`, just using `zoom-frm` functions instead.

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -37,6 +37,7 @@
        vi-tilde-fringe   ; fringe tildes to mark beyond EOB
        window-select     ; visually switch windows
        workspaces        ; tab emulation, persistence & separate workspaces
+       zoom-frm          ; allow to zoom a frame
 
        :editor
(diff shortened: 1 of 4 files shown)


hlissner:
After some thought, I decided to selectively extract the zoom functionality from `zoom-frm` into the new `doom/{increase,decrease,reset}-font-size` commands introduced in 0f0fdbc. I've also updated `doom-text-zoom-hydra` to use them. The decision was motivated by:

+ Avoiding pulling in more dependencies than we need (zoom-frm uses very little of frame-cmds and frame-fns)
+ Better integration with Doom (particularly the modeline and restoring `doom-font` when reverting to the base font size)
+ Reduce redundancy with `doom-big-font-mode` (it is now powered by these new commands)

In any case, these changes were inspired by this PR, so thank you for bringing it to my attention!

PR #1395, opened 2019-05-09
Use treemacs-magit package

jeffjewiss:
I thought this might be useful to setup and install when using both treemacs and magit. Cheers.

### DIFF
--- a/modules/ui/treemacs/config.el
+++ b/modules/ui/treemacs/config.el
@@ -39,3 +39,7 @@
 
 (def-package! treemacs-projectile
   :after treemacs)
+
+(def-package! treemacs-magit
+  :when (featurep! :tools magit)
+  :after treemacs magit)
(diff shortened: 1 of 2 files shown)


modules/ui/treemacs/config.el
```diff
@@ -39,3 +39,7 @@
 
 (def-package! treemacs-projectile
   :after treemacs)
+
+(def-package! treemacs-magit
+  :when (and (featurep! :ui treemacs) (featurep! :tools magit))
```
hlissner: Please remove this check for `:ui treemacs`. This is in the ui/treemacs module, none of which is evaluated if the containing module isn't enabled.

modules/ui/treemacs/packages.el
```diff
@@ -5,3 +5,5 @@
 (when (featurep! :editor evil +everywhere)
   (package! treemacs-evil))
 (package! treemacs-projectile)
+(when (and (featurep! :ui treemacs) (featurep! :tools magit))
```
hlissner: Also remove this check for `:ui treemacs`. Same reason as above.

jeffjewiss:
Whoops, sorry about the extra treemacs checks. Still new to emacs and lisp. Cheers, thanks!

hlissner:
Not a problem! Thanks for the quick fixes (and thanks again for your help).

PR #1398, opened 2019-05-10
Fix snarky ruby in example init.example.el

socksy:
No need for a `do` when you have `{}`.


hlissner:
Haha, nice. Thanks for your help!

PR #1400, opened 2019-05-10
Enable `better-jumper-mode` when loaded

gilbertw1:
As per request when submitting `better-jumper` to MELPA, I've added a minor mode. This pull request enables `better-jumper-mode` when the package is loaded.


hlissner:
Looks good to me! Let me know when it gets into MELPA and I will merge this.

PR #1405, opened 2019-05-13
fix command to set tags

brabalan:
I used to wrong function for the localleader binding to set tags in org files. Sorry.


hlissner:
Good catch, thanks!

PR #1408, opened 2019-05-13
Fix ivy-switch-buffer virtual buffer support

### LINKED ISSUE #1406: ivy recent files
For some reason, a new update in ivy seems to restrict virtual buffer to `caller`'s being `ivy-switch-buffer` or `counsel-switch-buffer`. This means that doom's buffer switcher cannot have virtual buffers (like recent files) enabled since its caller is `+ivy--switch-buffer`.

flatwhatson:
Fixes #1406

Yes this is a dubious hack, but it works, and ivy-resume still works.  Not sure that there's anything else which could be affected.


hlissner:
Lookgs good to me! Thanks for the help.

haji-ali:
With this, `+ivy--switch-buffer` in 
https://github.com/hlissner/doom-emacs/blob/13a20866aa955c323746ee5410e49c6168f83c36/modules/completion/ivy/config.el#L108
is redundant.

PR #1410, opened 2019-05-13
Add +prescient option for :completion ivy

flatwhatson:
This adds the `+prescient` flag to enable the `ivy-prescient` package for filtering and sorting of ivy completions.  When used with the `+fuzzy` flag, uses prescient's fuzzy completion instead of `flx`.

Also increased the value of `ivy-flx-limit` from the default 200 to 10000.  This is a performance vs utility tradeoff, but the impact is minimal in my testing.

### DIFF
--- a/modules/completion/company/config.el
+++ b/modules/completion/company/config.el
@@ -47,6 +47,7 @@
 (def-package! company-prescient
   :hook (company-mode . company-prescient-mode)
   :config
+  ;; NOTE prescient config duplicated with `ivy'
   (setq prescient-save-file (concat doom-cache-dir "prescient-save.el"))
   (prescient-persist-mode +1))
 
--- a/modules/completion/ivy/README.org
+++ b/modules/completion/ivy/README.org
@@ -37,7 +37,8 @@ lighter, simpler and faster in many cases.
 #+end_quote
 
 ** Module Flags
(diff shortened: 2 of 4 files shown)


flatwhatson:
Updated so it only applies to `(ivy +fuzzy)`.  This is ready for review!

flatwhatson:
Using this for a while, it works well.  That said, I'm not sure that prescient should *replace* flx for our fuzzy feature, particularly once `ivy-flx-limit` is set to a less conservative value.

Prescient is great for "learning" your workflows, but is otherwise less clever about sorting the results of fuzzy matches.  It may even be best used *without* fuzzy matching, as an accelerator for non-fuzzy workflows.

I'll rework this PR to make a `+prescient` option, so we can choose between `+fuzzy` using flx, `+prescient` for non-fuzzy prescient, and `+prescient +fuzzy` for the original behaviour of this PR.

hlissner:
Looks good to me. Nice work! And thanks again for your help!

flatwhatson:
Thanks!

osener:
I've been enjoying this addition, thank you. It seems to be interacting badly with `+ivy/rg` though. When  `+prescient` is enabled, it prevents `rg` from matching words separated by spaces individually. Is it possible to disable this completely for backends that handle sorting & filtering themselves?

PR #1412, opened 2019-05-13
Fix typo in modeline README

raugturi:


hlissner:
This PR reminded me that the ui/modeline module's README was overdue for a rewrite (now that we've adopted doom-modeline). I've gone ahead and done that in 0a0def6, in which the affected line is no longer present. That said, thanks for bringing it to my attention!

PR #1416, opened 2019-05-15
Added pyimport support

Mambu38:
Added pyimport support.
I figured that it just uses modules imported in currently opened python files, and don't looks for installed modules.

Keybindings :
- Import missing module via `<LocalLeader> i m`
- Remove unused imports via `<LocalLeader> r u`


hlissner:
Thanks for your help!

PR #1430, opened 2019-05-24
docs: fix FAQ link

zaiste:
The `docs/` is now empty. Dead links may create confusion. Here's a proposition to link it to the FAQ in Doom's Wiki.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,9 @@
   <a href="docs/index.org">documentation</a>&nbsp; |&nbsp;
   <a href="/../../tree/screenshots">screenshots</a>&nbsp; |&nbsp;
   <a href="docs/contributing.org">contribute</a>
-  <a href="docs/faq.org">faq</a>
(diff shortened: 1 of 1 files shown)


hlissner:
Sorry for the tremendously late reply. This is a response to #1435 and #1536, as well:

I understand the confusion and frustration this causes. Doom is in a weird place right now; I made the `develop` branch our default branch so fewer people would follow the outdated (and afaik non-functional) master branch installation instructions, but it means we now live on a develop branch with a README.md with links to files in `docs/` that don't exist yet.

In any case, I won't accept PRs to fix the links in the project README, because I'm not far from publishing the docs folder (which includes bigger changes to this README), and I fear the old wiki is so outdated that linking to it may cause as much confusion as it alleviates.

Sorry for the trouble, but thanks you for trying to help!

shonfeder:
No trouble! Thanks very much for the explanations and for your work :)

PR #1433, opened 2019-05-25
Fixed tutorial--saved-dir destination

sergeyklay:
`tutorial--saved-dir` is a function, not a variable.


hlissner:
I've removed the offending line and decided I'd rather not have that advice in Doom core -- I don't think it's important enough. I'll implement an `:app tutorial` module later, instead, for better integration into evil-tutor and help-with-tutorial (perhaps with a doom-centric addon). In any case, thanks for bringing it to my attention!

sergeyklay:
You're welcome

PR #1434, opened 2019-05-26
add option to run only one function

uelei:
also rename shortcut to pytest repeat to "r"


hlissner:
Thanks for your help!

PR #1439, opened 2019-05-28
Use doom-private-dir for default banner path

flatwhatson:
This is a bit more convenient than hiding them away in `~/.emacs.d/modules/ui/doom-dashboard/banners`.


hlissner:
The purpose of `+doom-dashboard-banner-dir` was to make it easier for the user to refer to one of the [not yet] included official banner images with a relative filename. I'd rather leave it to the user to change that themselves. `+doom-dashboard-banner-file` also accepts an absolute path, so no need to touch `+doom-dashboard-banner-dir` at all.

I'd welcome a PR to include mention of that in the dashboard module's README, however.

flatwhatson:
Makes sense, thanks, I'll update the README.

PR #1440, opened 2019-05-28
install auth-source-pass from melpa

### LINKED ISSUE #1438: auth-source-pass should always be installed from MELPA
`auth-source-pass` is kept at v2 when using emacs >= 26, while v5 as already been released. This is the only package in the repo that follow this behavior.

bricewge:
Fix #1438

### DIFF
--- a/modules/tools/pass/packages.el
+++ b/modules/tools/pass/packages.el
@@ -4,10 +4,7 @@
 (package! pass)
 (package! password-store)
 (package! password-store-otp)
-
-;; `auto-source-pass' is built into Emacs 26+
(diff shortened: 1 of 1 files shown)


bricewge:
BTW when doing `doom install` with this PR, the wrong message is displayed:
```
Looking for packages to install...
1 packages will be installed:

+ auth-source-pass (QUELPA->ELPA)
```
It should have been `BUILT-IN->ELPA` or just `ELPA`.

bricewge:
Disregard this PR, the package don't stick in doom; `doom refresh` always reinstall the package.

PR #1441, opened 2019-05-28
feat(format +onsave): allow disabling formatting

danielpza:
enable/disable formatting in current buffer through `format-on-save-mode`


hlissner:
I decided on a simpler solution: to use format-all-mode directly, which should make this PR unnecessary. Disabling this mode will disable format-on-save.

In any case, thanks for the help and bringing this to my attention.

PR #1446, opened 2019-05-31
Replace kill-this-buffer with kill-current-buffer

siddharthshekar:
As per the documentation for kill-this-buffer, it should only be invoked
from the menu and behaves unpredictably when invoked programmatically.


hlissner:
I wasn't aware of `kill-current-buffer` until now. Thanks for bringing it to my attention!

PR #1449, opened 2019-06-01
projectile: fd never use colors

bricewge:
Prevent `projectile-find-file` to be littered with escape codes when the
environment variable LS_COLORS is set.


hlissner:
Thanks for your help!

Issue #1444, opened 2019-05-30
kill-emacs hangs when recentf contains TRAMP paths


bricewge:
When exiting emacs if a TRAMP path, in my case a file opened with sudo, is contained in the `recentf` file then emacs will hang until the credentials are provided or it timeout.

Commits b33b5ed624f8c1c6f1313acca182e4f0c03bf713 and 39db0a38c5faf8b6b6d9fdb9c154ee63fd86672b probably introduced the issue. I have found a [thread](https://lists.gnu.org/archive/html/bug-gnu-emacs/2007-07/msg00006.html) in the emacs mailing list discussing this issue. I suggest that `recentf-cleanup` should skip remote files during the exit cleanup but try to clean them only after you explicitly connect to the TRAMP host.

My current workaround:
``` elisp
(after! recentf
  (add-to-list 'recentf-keep `remote-file-p))
```

bricewge:
Even with the above workaround it also hangs when:
1. Opening a file as sudo and entering correct password
2. `sudo --remove-timestamp`
3. `tramp-cleanup-all-connections`, just afterward it already asks the sudo password (seems to come from `direnv`)
4. Exiting emacs hangs for sudo password

Those are the minimal set of steps I have found to reproduce the issue.

hlissner:
I wish recentf wouldn't try to connect to the remote to determine whether it should be kept or not, but I don't have time to look into this, so as of b68fb2a remote files are excluded from recentf. Let me know if that isn't the case and I will reopen this. Thanks for bringing it to my attention!

bricewge:
With b68fb2a you have to manually remove the offending entries from `~/.emacs.d/.local/cache/recentf` to get stop the hanging of emacs. I 'll try to write a PR to implement what i suggest in the initial bug report, would you be willing to merge it?

About my second comment, I dug some more, and it seems an issue specific to `direnv`. Removing it from `~/.doom/init.el` prevent the issue, I'll report it upstream.

UndeadKernel:
I was hit by this issue too.
Another way to clean your `recentf` list: call `recentf-edit-list` and remove unwanted entries.

bricewge:
> I wish recentf wouldn't try to connect to the remote to determine whether it should be kept or not,

By default it doesn't, `recentf-keep`:
> The default is to keep readable files.  Remote files are checked for readability only in case a connection
is established to that remote system, otherwise they are kept in the recent list without
checking their readability.

[...]

PR #1451, opened 2019-06-01
keep remote files in recentf

bricewge:
Verify that there is a connection to the file before trying to find it's
truename when recentf-cleanup is called. Complete explanation at https://github.com/hlissner/doom-emacs/issues/1444#issuecomment-497962273.
Fix hlissner#1444 for good (hopefully).

### DIFF
--- a/core/autoload/files.el
+++ b/core/autoload/files.el
@@ -129,6 +129,10 @@ MATCH is a string regexp. Only entries that match it will be included."
            short-new-name)
           (short-new-name))))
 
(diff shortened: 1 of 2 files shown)


hlissner:
The solution needed a small adjustment made in d329f21, to ensure that non-remote paths are resolved (and to prevent a race condition by moving doom--file-truename into core-editor). Let me know if this doesn't address your issue. Thanks for bringing it to my attention!

PR #1457, opened 2019-06-02
editor/fold: feature vimish-fold is :editor evil

bricewge:
I think the feature test was meant to be on `:editor evil` and not `vimish-fold` since I couldn't find such feature and that the package `vimish-fold` comes from: https://github.com/hlissner/doom-emacs/blob/32e266cfb4be41951c8bbffd516ccf9456beeeb8/modules/editor/fold/packages.el#L4-L5

### DIFF
--- a/modules/editor/fold/autoload/fold.el
+++ b/modules/editor/fold/autoload/fold.el
@@ -13,7 +13,7 @@
     (hs-minor-mode +1)))
 
 (defun +fold--vimish-fold-p ()
-  (and (featurep 'vimish-fold)
+  (and (featurep! :editor evil)
        (cl-some #'vimish-fold--vimish-overlay-p
                 (overlays-at (point)))))
 
@@ -98,7 +98,7 @@ Targets `vimmish-fold', `hideshow' and `outline' folds."
   "Open folds at LEVEL (or all folds if LEVEL is nil)."
   (interactive
    (list (if current-prefix-arg (prefix-numeric-value current-prefix-arg))))
-  (when (featurep 'vimish-fold)
+  (when (featurep! :editor evil)
     (vimish-fold-unfold-all))
   (save-excursion
     (+fold--ensure-hideshow-mode)
@@ -117,7 +117,7 @@ Targets `vimmish-fold', `hideshow' and `outline' folds."
   (interactive
    (list (if current-prefix-arg (prefix-numeric-value current-prefix-arg))))
   (save-excursion
-    (when (featurep 'vimish-fold)
+    (when (featurep! :editor evil)
       (vimish-fold-refold-all))
     (+fold--ensure-hideshow-mode)
     (hs-life-goes-on
@@ -135,7 +135,7 @@ Targets `vimmish-fold', `hideshow' and `outline' folds."
                       (when hs-block-start-regexp
                         (car (+fold--invisible-points count))))
                     (lambda ()
-                      (when (featurep 'vimish-fold)
+                      (when (featurep! :editor evil)
                         (if (> count 0)
                             (evil-vimish-fold/next-fold count)
                           (evil-vimish-fold/previous-fold (- count))))
--- a/modules/editor/fold/autoload/hideshow.el
+++ b/modules/editor/fold/autoload/hideshow.el
@@ -22,7 +22,7 @@
 ;;;###autoload
 (defun +fold-hideshow-set-up-overlay (ov)
   (when (eq 'code (overlay-get ov 'hs))
-    (when (featurep 'vimish-fold)
+    (when (featurep! :editor evil)
       (overlay-put
        ov 'before-string
        (propertize "…" 'display


hlissner:
The confusion is understandable, but this was intended. `vimish-fold` is a dependency of `evil-vimish-fold`. Testing for it, rather than `:editor evil` is slightly more portable (less coupled with the `:editor evil` module) and more explicit, because we only need vimish-fold's API in these sections, it also doesn't require evil to be installed (or :editor evil to be enabled).

bricewge:
I see. The problem is even when using `:editor evil` and `:editor fold` those tests return `nil`. How can I use the feature `vimish-fold` then? And how can `vimish-fold` package can be used without `evil` since it's never installed in that case?

hlissner:
Sorry for the late response:

>  The problem is even when using :editor evil and :editor fold those tests return nil

I cannot reproduce that. Could you provide your `M-x doom/info`? If `:editor evil` and `:editor fold` are truly enabled, then those featurep! checks should work. How are you testing them?

> How can I use the feature vimish-fold then?

You can create vimish folds with these keys (like it is in vim):

```
+ `zf` (operator) Creates a vimish fold
+ `zd` Delete the vimish fold at point
+ `zE` Delete all vimish folds in buffer
```

And doom integrates vimish-fold (and hideshow+outline-minor-mode) into evil's folding keys:

+ `za` (toggle fold at point)
+ `zc` (close fold at point)
+ `zo` (open fold at point)
+ `zR` (open all folds)
+ `zM` (close all folds)

> And shouldn't vimish-fold package be installed for feature vimish-fold to work when evil is not used?

Very possible, but I've only focused on evil integration because vim workflows are what I know. A PR is welcome if you'd like to bring it to non-evil users, but from what I can tell, hideshow doesn't have any default keybinds so I wouldn't know where best to bind these keys.

bricewge:
https://github.com/hlissner/doom-emacs/blob/2835314022f1cf290f28c48e41e62feab20aa208/modules/editor/fold/autoload/fold.el#L101
`(featurep 'vimish-fold)` return `nil` whereas `(featurep! :editor fold)` return `t`.

Should I rework this PR by modify my changes from `(featurep! :editor evil)` to `(featurep! :editor fold)`?
<details>
  <summary>doom/info</summary>

- OS: gnu/linux (x86_64-pc-linux-gnu)
- Emacs: 26.1 (mai 22, 2019)
- Doom: 2.0.9 (HEAD -> develop, _upgrade/develop a1c1d881 2019-06-11 16:21:37 +0200)
- Graphic display: t (daemon: t)
- System features: XPM JPEG TIFF GIF PNG RSVG SOUND DBUS GSETTINGS NOTIFY LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 MODULES THREADS LIBSYSTEMD
- Details:
  ```elisp
  env bootstrapper: nil
  elc count: 0
  uname -a:  Linux 4.19.47 #1-NixOS SMP Fri May 31 13:46:35 UTC 2019 x86_64
  modules:   (:completion company (ivy +prescient +childframe) :ui deft doom doom-dashboard doom-quit fill-column hl-todo modeline nav-flash neotree ophints (popup +all +defaults) pretty-code vc-gutter vi-tilde-fringe window-select workspaces :editor evil file-templates fold lispy multiple-cursors rotate-text snippets :emacs dired electric vc :term eshell vterm :tools direnv docker editorconfig eval flycheck flyspell (lookup +docsets) lsp make magit (pass +auth) pdf rgb :lang assembly (cc +lsp) data emacs-lisp (go +lsp) (javascript +lsp) markdown nix (org +attach +babel +capture +export +present) (python +lsp) (rust +lsp) (sh +lsp) (web +lsp) :app regex :config default)
  packages:  (vue-mode systemd outshine poporg evil-lion solarized-theme yasnippet-snippets webpaste esh-autosuggest alert (org-caldav :recipe (org-caldav :fetcher github :repo grauschnabel/org-caldav)) org-noter nov guix geiser (doom-themes :recipe (doom-themes :fetcher github :repo bricewge/emacs-doom-themes :branch hl-fill-column :files (*.el themes/*.el))))
  exec-path: (~/bin /run/wrappers/bin ~/.nix-profile/bin /nix/var/nix/profiles/default/bin /run/current-system/sw/bin /etc/profiles/per-user/bricewge/bin /nix/store/jk935pwdp9m5cs8bdi29p3ivbvvbnf00-emacs-26.1/libexec/emacs/26.1/x86_64-pc-linux-gnu)
  ```

hlissner:
> (featurep 'vimish-fold) return nil whereas (featurep! :editor fold) return t.

Have you tested `(featurep 'vimish-fold)` _after_ using one of the vimish-fold keybinds? (zf, zd, or zE)

The package is lazy-loaded,. `(featurep 'vimish-fold)` will return `nil` if the package isn't loaded, and it won't be loaded until you try to use it.

> Should I rework this PR by modify my changes from (featurep! :editor evil) to (featurep! :editor fold)?

No, `(featurep 'vimish-fold)` is correct. It is also redundant to check if `:editor fold` is enabled from within the `:editor fold` module.

bricewge:
> Have you tested `(featurep 'vimish-fold)` _after_ using one of the vimish-fold keybinds? (zf, zd, or zE)

It returns `t`...

I feel stupid now, I'm sorry I wasted your time on this.

hlissner:
Haha, no problem at all. Emacs can be obtuse. I'm more than happy to help!

PR #1459, opened 2019-06-03
Fix magit-clone error

qhuyduong:
magit-clone function has changed signature to work with transient.
So using the new magit-clone-regular instead.


hlissner:
Thanks for your help!

PR #1460, opened 2019-06-03
Posix compliance

Provessor:
Modifies the scripts
- bin/doom
- bin/doom-doctor
- bin/org-capture
- bin/org-tangle
so they are compatible with any POSIX compliant shell.


bricewge:
`bin/doom` and `bin/doctor` seems to work on my hand, for what that's worth. Note that I'm on an exotic linux OS that don't follow FHS, NixOS.

hlissner:
Thanks for your help!

PR #1461, opened 2019-06-03
completion/ivy: ivy-posframe upstream update

bricewge:
`ivy-postframe` go updated and became a minor mode. It broke doom's setup in it's way.


hlissner:

PR #1476, opened 2019-06-08
fix shebang

chrunchyjesus:
See https://en.wikipedia.org/wiki/Shebang_(Unix)#Portability


chrunchyjesus:
Yes.

chrunchyjesus:
Done. Also fixed the shebang for the macos patch.

hlissner:

Issue #1470, opened 2019-06-06
Newest version of format-all doesn't work with doom (private function names changed)


Strayer:
Commit https://github.com/lassik/emacs-format-all-the-code/commit/f3ed09c430c80d17fcc34f130f32072177f4ef2b changed some function names in format-all and this apparently causes doom-emacs to break with the following error message:

`let*: Autoloading file /Users/strayer/.emacs.d/.local/packages/elpa/format-all-20190606.853/format-all.elc failed to define function format-all-probe`

See https://github.com/lassik/emacs-format-all-the-code/issues/36#issuecomment-499544541 for reference.

nate:
pinning the package to the commit before the changes upstream fixed it for me: 

```elisp
(package! format-all :recipe (:fetcher github :repo "lassik/emacs-format-all-the-code" :commit "7ab5345d83dad3df1f162c24e6f28cb7307975fd"))
```

Once #1477 is rolled out this could be removed.

PR #1477, opened 2019-06-08
fix: format-all upstream definitions

4rg0:
This should fix [1470](https://github.com/hlissner/doom-emacs/issues/1470)

I'm completely new to emacs and have no clue what I'm doing, so be warned and bear with me if this is is all bonkers ;)


4rg0:
Happy I could help. Thank you very much for doom-emacs!

PR #1484, opened 2019-06-12
Add pomodoro in org

iceiceiceJack:
add Pomodoro Technique to org

### DIFF
--- a/modules/lang/org/+pomodoro.el
+++ b/modules/lang/org/+pomodoro.el
@@ -1,0 +1,24 @@
+;;; lang/org/+pomodoro.el -*- lexical-binding: t; -*-
+;;;###if (featurep! +pomodoro)
+
+(defun +org-pomodoro/notification(title message)
+  "Send a notification"
+  (when IS-MAC
+    (do-applescript
+     (format "display notification \"%s\" with title \"%s\" sound name \"Ping\"" message title))))
+
+(defun +org-pomodoro/pomodoro-notification()
+  "show notifications when pomodoro end"
+  (progn
+    (add-hook 'org-pomodoro-finished-hook '(lambda ()
+                                             (+org-pomodoro/notification "Pomodoro Finished" "Have a break!")))
+    (add-hook 'org-pomodoro-short-break-finished-hook '(lambda ()
+                                                         (+org-pomodoro/notification "Short Break" "Ready to Go?")))
+    (add-hook 'org-pomodoro-long-break-finished-hook '(lambda ()
+                                                        (+org-pomodoro/notification "Long Break" "Ready to Go?")))))
+
+(def-package! org-pomodoro
+  :when (featurep! +pomodoro)
+  :after org
+  :config
+  (+org-pomodoro/pomodoro-notification))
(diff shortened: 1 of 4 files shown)


modules/lang/org/+pomodoro.el
```diff
@@ -0,0 +1,24 @@
+;;; lang/org/+pomodoro.el -*- lexical-binding: t; -*-
+;;;###if (featurep! +pomodoro)
+
+(defun +org-pomodoro/notification(title message)
```
hlissner: Please rename this to `+org-pomodoro-notification`. By Doom's naming conventions, the `/` signifies an interactive command, which this is not.

modules/lang/org/+pomodoro.el
```diff
@@ -0,0 +1,24 @@
...
+  "Send a notification"
+  (when IS-MAC
+    (do-applescript
```
hlissner: I'd prefer that this weren't exclusive to macos users. Could you use [alert](https://github.com/jwiegley/alert) instead? It can put out notifications to a variety of notification backends, including MacOS native notifications and growl, without excluding Linux users.

modules/lang/org/+pomodoro.el
```diff
@@ -0,0 +1,24 @@
...
+     (format "display notification \"%s\" with title \"%s\" sound name \"Ping\"" message title))))
+
+(defun +org-pomodoro/pomodoro-notification()
```
hlissner: Could you move the contents of +org-pomodoro/pomodoro-notification into the :config body of org-pomodoro? It doesn't need to be in its own function.

modules/lang/org/+pomodoro.el
```diff
@@ -0,0 +1,24 @@
...
+  "show notifications when pomodoro end"
+  (progn
+    (add-hook 'org-pomodoro-finished-hook '(lambda ()
+                                             (+org-pomodoro/notification "Pomodoro Finished" "Have a break!")))
+    (add-hook 'org-pomodoro-short-break-finished-hook '(lambda ()
+                                                         (+org-pomodoro/notification "Short Break" "Ready to Go?")))
+    (add-hook 'org-pomodoro-long-break-finished-hook '(lambda ()
```
hlissner: Please replace the three hook functions with named functions, rather than using anonymous (and quoted) lambdas. It's easier to add/remove named functions, in case users want to customize these hooks, and the hook functions can be advised if so desired.

modules/lang/org/+pomodoro.el
```diff
@@ -0,0 +1,24 @@
...
+                                                        (+org-pomodoro/notification "Long Break" "Ready to Go?")))))
+
+(def-package! org-pomodoro
+  :when (featurep! +pomodoro)
```
hlissner: This `featurep!` check is unnecessary. This file won't be loaded in the first place if the +pomodoro flag isn't active.

modules/lang/org/+pomodoro.el
iceiceiceJack: I'll fix these this weekends, thanks for your advice.

hlissner:
Has there been any progress on this PR?

PR #1488, opened 2019-06-13
Correct documentation of env_var file

gagbo:
Because of 3ed54e191bc7db1b8adf46915601f322a0d364e5 The `load-env-vars` advice is false now. Instead we can propose setting "doom-env-file" correctly, it seems to work for me.


gagbo:
I see

hlissner:
Thanks for your help!

PR #1496, opened 2019-06-15
use default `nix-indent-function`

dywedir:
See https://github.com/NixOS/nix-mode/pull/79

cc @ar1a


ar1a:
LGTM, hope to see it indenting better than what we currently have.

hlissner:
Thanks for your help!

PR #1505, opened 2019-06-21
fix: file not found in autoload

birdgg:
fix error: Cannot open load file: No such file or directory, emacs/vc/autoload


hlissner:
Thanks for the help!

PR #1506, opened 2019-06-23
Uniquify buffer names

emil-r-petersen:
This improves buffer switching, as in this screenshot, you can see that it's not obvious which buffer is which: https://imgur.com/vC8Jaio

But with this change, this is what we get: https://imgur.com/Iki1EVD

This makes it easier to switch buffers, and to be sure what you are changing to. If there is some shortcut to expand the buffer name in minibuffer, I'm not aware of it. Either way, I believe this is a better alternative, because it gives you more information right as you want to switch buffers.

### DIFF
--- a/core/core-ui.el
+++ b/core/core-ui.el
@@ -191,7 +191,7 @@ read-only or not file-visiting."
  mouse-yank-at-point t           ; middle-click paste at point, not at click
  resize-mini-windows 'grow-only  ; Minibuffer resizing
  show-help-function nil          ; hide :help-echo text
- uniquify-buffer-name-style nil  ; custom modeline will show file paths anyway
+ uniquify-buffer-name-style 'post-forward-angle-brackets ; to show hints when switching buffers
  use-dialog-box nil              ; always avoid GUI
  visible-cursor nil
  x-stretch-cursor nil
--- a/modules/ui/workspaces/config.el
+++ b/modules/ui/workspaces/config.el
@@ -100,11 +100,15 @@ Uses `+workspaces-main' to determine the name of the main workspace."
   (defun +workspaces|init-persp-mode ()
     (cond (persp-mode
            ;; `persp-kill-buffer-query-function' must be last
+           (setq my-uniquify-buffer-name-style uniquify-buffer-name-style)
+           (setq uniquify-buffer-name-style nil)
            (remove-hook 'kill-buffer-query-functions 'persp-kill-buffer-query-function)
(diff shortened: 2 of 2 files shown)


seanfarley:
Oh, ha, good timing. I was about to submit a similar pull request it differs from yours in that I prefer `'forward` instead of `'post-forward-angle-brackets`. To be honest, I don't have a strong opinion over which setting is used since it's super easy to set in my own config. For reference, this is what `'forward` looks like (basically, use the next higher up directory with no angle brackets):

Screen Shot 2019-06-23 at 12 58 57 AM

emil-r-petersen:
@seanfarley Honestly, forward looks fine to me, too. Strange, I don't always see the full path on the far right, only in some cases. Seems that my non-unique buffer names don't include that part.

emil-r-petersen:
And I would just set it in my own config, except I think anything but nil is a clear improvement.

emil-r-petersen:
@hlissner I believe I made it work as intended, but I have found that after enabling and disabling persp-mode, I have no valid perspectives  (and need to run persp-new before things like counsel-projectile-switch-project works).

I'm not sure if this is a bug I've introduced, as unfortunately I'm still early in the progress of becoming adept at Elisp. Resetting to before my two commits, though, this still seems to happen.

Should we have it run +workspace/new when persp-mode is enabled again, maybe?

hlissner:
@emil-r-petersen Ah, I see the issue. In that case could you rebase out the changes to ui/workspaces? Sorry for the trouble, but I have a few refactors for ui/workspaces incoming and I'll include a more robust solution to the problem, which includes moving the creation of the main workspace to persp-mode-hook.

hlissner:
(Oh, forgot to mention: could you change `uniquify-buffer-name-style` to `forward` instead? I much prefer a familiar file path format to the angle brackets.)

hlissner:
Whoops, didn't mean to close this. As of 86ebe08, ivy's switch-to-buffer will display `{dirname}/{filename}` if uniquify-buffer-name-style is nil, to reduce this sort of confusion. And 9823881 will unset `uniquify-buffer-name-style` when persp-mode is active, as well as resolve your perspective issue when reactivating persp-mode.

hlissner:
As of 92676bf I've added your initial change to core-ui and I believe this issue has been fully addressed, so I'll close this PR. Thanks for the help and for bringing it to my attention!

emil-r-petersen:
@hlissner Thank you.

Issue #1521, opened 2019-06-28
lang:go use lcp-find-defintion if lcp package installed


idanko:
Hi, since Language Server Protocol for Go namely `gopls` is quite mature (despite the fact it is under active development) and I also found it faster than `guru` or `godef`, I would suggest using `lcp-find-definition` as a definition tool on set-lookup-handlers when lsp package is provided.

hlissner:
The lookup module [already uses lsp-find-{definitions,references} as its lookup handlers](https://github.com/hlissner/doom-emacs/blob/develop/modules/tools/lsp/config.el#L27-L29) when lsp-mode is enabled. [As well as `lsp-describe-thing-at-point`](https://github.com/hlissner/doom-emacs/blob/develop/modules/tools/lsp/config.el#L9-L10).

Perhaps you haven't enabled the `+lsp` flag for the `:lang go` module?

idanko:
@hlissner Ok, I found that I disabled the lsp-ui package for personal reasons (mostly because it is quite verbose for me), and just use lsp-mode for completion and, so far, for goto definition property (by overriding guru's one). I noticed this setup pretty functional.

hlissner:
As of 2f268e8, the module now falls back to `lsp-find-{definition,references}` when lsp-ui-mode isn't present/enabled, so I'll consider this issue resolved. Let me know if that isn't the case and I will reopen this. Thanks for bringing it to my attention!

idanko:
@hlissner hi, looks like the behavior is not changed. What I've done:

## Setup
doom.d/init.el has lsp and go modules
doom.d/package.el has `(package! lsp-ui :disable t)`

## Flow to reproduce
* cd ~/.emacs.d
* git pull (develop branch)
* doom upgrade && doom re
* open some go file
* press `gd` on `os.Exit()` to fire up goto definition of os stdlib package and Exit() function

## Expected behavior
`lsp-find-definition` run by starting the lsp server

## Actual
guru run

hlissner:
@idanko Are there any warnings/errors in \*Messages\*? (See `SPC h e`). For instance, did the LSP server successfully start?

idanko:
@hlissner aren't, the *Messages* buffer hasn't any lsp messages, even if manually turn on lsp mode, the behavior doesn't change.

hlissner:
[...]

PR #1529, opened 2019-07-01
lang/go: prioritize lsp over guru

idanko:
## Summary
This is a potential fix for https://github.com/hlissner/doom-emacs/issues/1521

## What is done
- [x] switched lsp! and set-lookup-handlers!

### DIFF
--- a/modules/lang/go/config.el
+++ b/modules/lang/go/config.el
@@ -4,10 +4,6 @@
 ;; Packages
 
 (after! go-mode
-  (if (featurep! +lsp)
-      (add-hook 'go-mode-hook #'lsp!)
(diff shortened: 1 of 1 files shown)


modules/lang/go/config.el
```diff
@@ -4,6 +4,10 @@
 ;; Packages
 
 (after! go-mode
+  (if (featurep! +lsp)
```
hlissner: Could you move this back to its original location, and instead do the following:

```diff
-(add-hook 'go-mode-hook #'lsp!) 
+(add-hook 'go-mode-local-vars-hook #'lsp!) 
```

This should fix your issue, and is something I was going to do later anyway.

idanko: @hlissner tested, done.

PR #1537, opened 2019-07-04
use keybindings from cargo mode

chrunchyjesus:


modules/lang/rust/config.el
```diff
@@ -10,10 +10,19 @@
...
+        :desc "cargo add"    "a" #'cargo-process-add
+        :desc "cargo build"  "b" #'cargo-process-build
+        :desc "cargo build"  "B" #'cargo-process-bench
+        :desc "cargo check"  "c" #'cargo-process-check
+        :desc "cargo clippy" "C" #'cargo-process-clippy
+        :desc "cargo doc"    "d" #'cargo-process-doc
+        :desc "cargo run"    "r" #'cargo-process-run
+        :desc "cargo search" "s" #'cargo-process-search
+        :desc "cargo update" "u" #'cargo-process-update
+        (:prefix-map ("t" . "cargo test")
```
hlissner: Please use `:prefix` instead of `:prefix-map`. The latter isn't designed mostly for use on leader keys, and may cause conflicts if used prolifically.

modules/lang/rust/config.el
```diff
@@ -10,10 +10,19 @@
   (map! :map rust-mode-map
         :localleader
         :prefix "b"
```
hlissner: Was it intended for the cargo test prefix to be under `b`? e.g. `SPC m b t a`? Or did you want it to be a top level localleader key, e.g. `SPC m t a`? I'd prefer the latter myself, so please wrap the `b` prefixed keybinds in parentheses.

```lisp
(:prefix "b"
  :desc "cargo add"    "a" #'cargo-process-add
  ...)
(:prefix ("t" . "cargo test")
  ...)
```

modules/lang/rust/config.el
```diff
@@ -10,10 +10,19 @@
   (map! :map rust-mode-map
```
hlissner: Could you move all these new keybinds into a `def-package!` block for `cargo`? This way users will be able to disable it if want to. It should be nested within the rust-mode `def-package!` block, e.g.

```elisp
(def-package! rust
  ...
  :config
  ...
  (def-package! cargo
    :defer t
    :init
    (map! :map rust-mode-map
          :localleader
          ...)))
```

modules/lang/rust/config.el
chrunchyjesus: Now that I think about it, do we need the `b` prefix at all?

modules/lang/rust/config.el
chrunchyjesus: Fixed.

modules/lang/rust/config.el
chrunchyjesus: Fixed.

modules/lang/rust/config.el
chrunchyjesus: @hlissner I have removed the `b` prefix everywhere wit the latest commit. If that's not desired please tell me.

hlissner: I'd prefer we kept `b`, it is the canonical prefix for build commands. It's consistent with most other modules and although rust doesn't have any other localleader keys, that may change.

chrunchyjesus: Fixed.

hlissner:
Nice work! Thanks again for your help!

PR #1570, opened 2019-07-17
ansible: fix upstream changing function names

seanfarley:
A simple fix for upstream renaming all their functions.


hlissner:

PR #1571, opened 2019-07-17
bindings: fix emacs bindings and add more

UndeadKernel:
Hi @hlissner. I finally have a PR for you to fix some of the outstanding issues that the default emacs bindings had. There were some bindings that would shadow many emacs things.

I've also added missing emacs bindings for `eshell`. Could you take a look at these? I'm not sure if I have to repeat the binding or if there is a way to tell `map!` to use something for vi and emacs at the same time (i.e., using `:e`?).

### DIFF
--- a/modules/term/eshell/config.el
+++ b/modules/term/eshell/config.el
@@ -132,10 +132,12 @@ You should use `set-eshell-alias!' to change this.")
             :n "C"       #'+eshell/evil-change-line
             :n "d"       #'+eshell/evil-delete
             :n "D"       #'+eshell/evil-delete-line
-            :i "TAB"     #'+eshell/pcomplete
-            :i [tab]     #'+eshell/pcomplete
-            :i "C-d"     #'+eshell/quit-or-delete-char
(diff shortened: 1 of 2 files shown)


modules/term/eshell/config.el
```diff
@@ -136,6 +136,11 @@ You should use `set-eshell-alias!' to change this.")
             :i [tab]     #'+eshell/pcomplete
             :i "C-d"     #'+eshell/quit-or-delete-char
             "C-s"   #'+eshell/search-history
+            ;; Emacs bindings
+            "C-e"   #'end-of-line
+            "C-d"   #'+eshell/quit-or-delete-char
+            "TAB"   #'+eshell/pcomplete
```
hlissner: Could you combine these tab keybinds with the insert mode ones just above? (By changing `:i` to `:ig`)

UndeadKernel: I've also changed "C-d" as it's the same in normal emacs bindings.

hlissner:
Great! Thanks again for the help!

PR #1578, opened 2019-07-21
Remove flycheck-rust since it is no longer used.

jdemilledt:
Removing the dependency for flycheck-rust since it is no longer used and causes problems when required without being set up.


hlissner:
This makes sense to me. If you'd kindly retarget this PR to the straight branch, I'd be happy to merge it.

jdemilledt:
@hlissner Done.

hlissner:
Thanks for the help!

PR #1587, opened 2019-07-23
add lsp-cache to gitignore

tam5:
Adds `lsp-cache` folder to `.gitignore`.


hlissner:
Rather than adding it to .gitignore, I'd rather we set `lsp-intelephense-storage-path` to a place in `doom-cache-dir`. I will add this soon.

PR #1588, opened 2019-07-23
Add scheme module

elais:
Adds a simple scheme module with geiser backend, supports all major scheme implementations except racket, which I have left out considering racket has its own module that will probably serve users better.


hlissner:
Looks good to me, thanks!

PR #1591, opened 2019-07-24
Add more cider functions to the clojure keymap

nate:
There are quite a few useful cider functions that are not mapped in doom's clojure module. The keys chosen for running tests match the keys used in the default cider keybindings.


hlissner:

PR #1594, opened 2019-07-25
Fix quoting of command in λ!!

filalex77:
``` emacs-lisp
(lambda!! #'avy-goto-char-timer t)
```

Before:

``` emacs-lisp
(lambda nil
  (interactive)
  (let ((current-prefix-arg t))
    (call-interactively avy-goto-char-timer)))
```

After:

``` emacs-lisp
(lambda nil
  (interactive)
  (let ((current-prefix-arg t))
    (call-interactively
     (function avy-goto-char-timer))))
```

### DIFF
--- a/core/core-lib.el
+++ b/core/core-lib.el
@@ -180,9 +180,9 @@ at the values with which this function was called."
 (defun λ!! (command &optional arg)
   "Expands to a command that interactively calls COMMAND with prefix ARG."
   (declare (doc-string 1))
-  `(lambda () (interactive)
-     (let ((current-prefix-arg ,arg))
-       (call-interactively ,command))))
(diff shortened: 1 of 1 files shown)


hlissner:
Oh, my bad. Actually, the returned closure shouldn't be quoted at all. Would you mind correcting that for me? i.e.

```elisp
(defun λ!! (command &optional arg)
  "Expands to a command that interactively calls COMMAND with prefix ARG."
  (declare (doc-string 1))
  (lambda () (interactive)
     (let ((current-prefix-arg arg))
       (call-interactively command))))
```

filalex77:
@hlissner Sure!

hlissner:
Thanks for the help!

PR #1597, opened 2019-07-26
tools/lsp: cache company candidates automagically

ar1a:
When opening a buffer, company-mode is not activated and therefore there's no completion. this was introduced in 149b2617 and now fixes the mistake


ar1a:
I also snuck de7af71df8c971b6990f34c57c9554d7d2852529 in because it's pretty good for performance. lmk if you want me to revert it

PR #1599, opened 2019-07-26
lang/racket: set lookup handler to find definition

tchajed:
Add a lookup handler for racket-mode to enable `SPC c d` to jump to definition (`, g d` is already bound to `racket-visit-definition`, but not the generic Jump to definition).

This is a bit buggy if you don't already have a Racket REPL open, which `racket-visit-definition` seems to rely on. That said, racket-mode will open a basic REPL even if you have errors in the buffer and many things will still work.

### DIFF
--- a/modules/lang/racket/config.el
+++ b/modules/lang/racket/config.el
@@ -21,6 +21,7 @@
   (add-hook! 'racket-mode-hook
              #'rainbow-delimiters-mode
              #'highlight-quoted-mode)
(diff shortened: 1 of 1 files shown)


hlissner:
This looks good to me. However, would you mind changing this PR to target the straight branch? A fair bit has changed over there, and it will be merged into develop sometime late next week.

tchajed:
Sure, will do so straight away.

hlissner:
Excellent. Thanks for the help!

PR #1600, opened 2019-07-26
:lang/python - Add lsp-python-ms

jdemilledt:
Many people prefer Microsoft's Python language server over pyls. This allows them to use it.


modules/lang/python/packages.el
```diff
@@ -4,6 +4,10 @@
 ;; Major modes
 (package! pip-requirements)
 
+;; Microsoft LSP
+(when (featurep! +lsp-ms)
```
filalex77: If you add a `+lsp-ms` flag, you probably should document it in [`init.example.el`](https://github.com/hlissner/doom-emacs/blob/develop/init.example.el#L137).

hlissner:
Hi! Thanks for the PR. I agree that lsp-python-ms is the better choice. 

I have two requests:

1. Could you target the `straight` branch instead, and
2. Could you gate it behind `+lsp` instead of `+lsp-ms`? I'd rather it be installed by default if lsp is enabled.

jdemilledt:
@hlissner There are probably some people out there still using `pyls`, and `lsp-python-ms` sets itself as the default. Would you like me to do that anyway?  
Also, targeting the `straight` branch has some adverse effects.

hlissner:
>Also, targeting the straight branch has some adverse effects.

It appears you've merged onto #1588. Would you mind rebasing onto straight? Without the shceme module there shouldn't be any conflict.

> There are probably some people out there still using pyls, and lsp-python-ms sets itself as the default

Having inspected lsp-python-ms's source, I find it disturbing that it would forcibly install itself if it isn't present, rather than gracefully yielding to the next available lsp client.

I don't like that behavior at all considering a) `M-x lsp-python-ms-setup` exists and b) LSP has a client fallback mechanism (that would allow it to use pyls in lsp-python-ms's absence). I think the better solution would be:

```elisp
(defun +python--dont-auto-install-server-a (orig-fn)
  lsp-python-ms-executable)
(advice-add #'lsp-python-ms--command-string
            :override #'+python--dont-auto-install-server-a)
```

So lsp-mode do it's job and fall back to pyls.

jdemilledt:
Aaand I managed to do a weird thing with the branches. Give me a moment and I'll reopen this.

PR #1510, opened 2019-06-24
Add README for lang/elixir

filalex77:
- [X] basic info (packages, etc.)
- [X] Elixir installation tips
- [x] features roundup
- [x] setup for LSP with [`elixir-ls`](https://github.com/JakeBecker/elixir-ls)

PR #1601, opened 2019-07-26
Add support for Elixir LSP via elixir-ls

filalex77:
- Document Elixir with LSP via elixir-ls
- Add `+lsp` flag to `:lang elixir`

Goes with #1510.

### DIFF
--- a/modules/lang/elixir/config.el
+++ b/modules/lang/elixir/config.el
@@ -29,6 +29,9 @@
     (sp-local-pair "do " " end" :unless '(sp-in-comment-p sp-in-string-p))
     (sp-local-pair "fn " " end" :unless '(sp-in-comment-p sp-in-string-p)))
 
+  (when (featurep! +lsp)
+    (add-hook 'elixir-mode-local-vars-hook #'lsp!))
+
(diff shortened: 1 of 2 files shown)


modules/lang/elixir/config.el
```diff
@@ -29,6 +29,9 @@
     (sp-local-pair "do " " end" :unless '(sp-in-comment-p sp-in-string-p))
     (sp-local-pair "fn " " end" :unless '(sp-in-comment-p sp-in-string-p)))
 
+  (when (featurep! +lsp)
+    (add-hook 'elixir-mode-hook #'lsp!))
```
hlissner: Could you change `elixir-mode-hook` to `elixir-mode-local-vars-hook`?

These are doom-specific hooks that run on `hack-local-variables-hook` very late in the major mode's activation process (giving the user time to customize settings with file/dir-local variables, among other things).

filalex77: Ah, so this is what the undocumented hook in `:lang rust` was. Thanks for pointing it out!

hlissner:
Thanks for your help!

PR #1602, opened 2019-07-27
Adding lsp-python-ms with patch so it doesn't force itself to be installed

jdemilledt:
Building on #1600, my branch juggling broke that PR.


hlissner:
Excellent. Thanks for your help!

PR #1604, opened 2019-07-27
Uniquify completion candidates of whole lines

rgrinberg:
If the same line is present more than once in the buffer, it will be
offered more than once as a candidate. This commit deletes duplicate
lines from the completion list.

### DIFF
--- a/modules/completion/company/autoload.el
+++ b/modules/completion/company/autoload.el
@@ -129,12 +129,13 @@ C-x C-l."
     (`candidates
      (all-completions
       arg
-      (split-string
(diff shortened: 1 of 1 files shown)


hlissner:
Ack, sorry. Switching this to the straight branch messed up this PR a bit. Would you mind rebasing it and targeting to at the `straight` branch?

rgrinberg:
Done. Btw, why is this marked as a refactor? As far as I can s ee this is fixing a bug.

hlissner:
You're right. I was too hasty picking it! Anyhow, thanks for the help!

PR #1608, opened 2019-07-28
lang/org: Package orgit if magit is enabled

filalex77:
https://github.com/magit/orgit


hlissner:

PR #1609, opened 2019-07-28
Fix :fetcher warning to include package name

tchajed:
There was a format command but it was in a literal string.


hlissner:
Ah, good catch. Thanks!

PR #1611, opened 2019-07-28
Use +pretty-code-iosevka-font-name on complete set

gagbo:
The ligatures are defined for the fontset `#Xe100 -> #Xe1cc` but the hook function did not use that complete space for the ligatures


hlissner:
Good catch, thanks!

PR #1580, opened 2019-07-21
lang/racket: Fix smart open bracket insertion

tchajed:
`racket-smart-open-bracket-enable` has been replaced by a function that inserts the right type of bracket (see https://github.com/greghendershott/racket-mode/commit/669275b3a620a940ce1c52610909577e94abf55f), which we need to bind to [ in insert mode.

This feature used to work, then Issue #1452 was filed to request it be configurable, then it broke, and now I'm fixing it. This indicates that maybe instead of just fixing it we should make it configurable, or users should do this customization themselves and this PR should just stop setting the obsolete `racket-smart-open-bracket-enable` flag. Note that DrRacket does not enable this setting by default.

PR #1613, opened 2019-07-28
lang/racket: Fix bug in configuration

tchajed:
Looks like my last refactor in #1580 broke racket-mode (apparently I didn't test it after making that change). Sorry about that!


hlissner:
No worries. I missed it too. Thanks for the quick fix!

PR #1620, opened 2019-07-30
tools/docker: Update README.org

filalex77:
- Proofread a bit
- Add #+SINCE: v2.0.9
- Change some formatting to proper Org syntax


hlissner:
Looks good to me. Thanks again for your help! I especially appreciate any help I can get with documentation.

PR #1621, opened 2019-07-30
fix small spelling mistake in fonts.el

ashtonbaker:
There is a small spelling mistake in `fonts.el`. This is just a small PR to fix it.


hlissner:
This looks good to me, but would you mind retargeting/rebasing this PR onto the straight branch? It will be merged into develop soon and it'd be a big help if PRs between now and then targeted it instead.

ashtonbaker:
Sounds good. I think I have changed it accordingly.

hlissner:
Great. Thanks for the help!

PR #1622, opened 2019-07-31
Fix +go/test-rerun

jsravn:
This command wasn't implemented yet it seems. I've implemented it.

### DIFF
--- a/modules/lang/go/autoload.el
+++ b/modules/lang/go/autoload.el
@@ -6,17 +6,22 @@
 (defvar +go-test-last nil
   "The last test run.")
 
-(defun +go--run-tests (args)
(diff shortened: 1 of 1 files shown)


jsravn:
@hlissner do I need to rebase this on `straight`?

jsravn:
I tested it out. `compile` is a nice improvement. However I found without `save-selected-window` it will change focus to the compile buffer the second time you run tests, which is sort of annoying. I updated the code to use compile but kept the save-selected-window. Let me know what you think.

hlissner:
Sorry for the late response. Keeping `save-selected-window` makes sense to me. Thanks for the PR!

PR #1628, opened 2019-08-01
Replace (if|when)-let with '*' variants

filalex77:
This fixes the deprecation warnings when running on Emacs 27.

I'm not hoping for this to be merged now, as Doom is still in the process of 25 ⇒ 26 migration, but I thought this may be useful later.

Made using https://github.com/facebookincubator/fastmod:

```
fastmod '((?:when|if)-let) \((.*)\)$' '$1* (($2))'
```

Corner cases with multiline elements fixed by hand. YMMV, did not test all the modules.


filalex77:
As the deprecation seems to be [reconsidered](https://gitlab.com/emacs-ci/emacs/commit/441fe201ea129709ac9807b9b6b30caa45bbd293), this PR is no longer needed.

PR #1638, opened 2019-08-05
Correct argument name in `defadvice!' docstring

danielkoning:
Self-explanatory: make the docstring agree with the macro definition.


hlissner:
Thanks for the help!

Issue #1640, opened 2019-08-06
[BUG] +evil `SPC w c` hardcoded to #'+workspace/close-window-or-workspace


alexandergraul:
**Describe the issue**
Disabling `:ui workspaces` breaks `SPC w c` and `C-w c` (using evil keybindings).

The problem is in https://github.com/hlissner/doom-emacs/blob/develop/modules/config/default/%2Bevil-bindings.el#L143:
```lisp
(:map evil-window-map
  "c" #'+workspace/close-window-or-workspace)
```

I expect this binding to use `evil-window-delete` by default.I guess the current binding should be in the  `featurep! :ui workspaces` further down in the file.

If no one minds I'll send a PR with theses changes.

hlissner:
Now that #1641 is merged, I'll consider this resolved. Thanks for bringing it to my attention and again for fixing it!

PR #1641, opened 2019-08-06
Fix evil-window "c" binding

alexandergraul:
Instead of relying on the optional ":ui workspaces" feature, the
built-in `evil-window-delete` is used and
`+workspace/close-window-or-workspace` is only used if ":ui workspaces" is
enabled.

Fixes https://github.com/hlissner/doom-emacs/issues/1640.

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -140,7 +140,6 @@
         "C-r"     #'winner-redo
         "o"       #'doom/window-enlargen
         ;; Delete window
-        "c"       #'+workspace/close-window-or-workspace
         "C-C"     #'ace-delete-window)
 
       ;; Plugins


alexandergraul:
Someone on discord said that I should probably use straight as a base instead -> switching back.

modules/config/default/+evil-bindings.el
```diff
@@ -140,7 +140,7 @@
         "C-r"     #'winner-redo
         "o"       #'doom/window-enlargen
         ;; Delete window
-        "c"       #'+workspace/close-window-or-workspace
+        "c"       #'evil-window-delete
```
hlissner: This keybind already exists in evil-window-map and can be removed altogether here.

modules/config/default/+evil-bindings.el
```diff
@@ -353,7 +353,9 @@
         :g "M-9"   #'+workspace/switch-to-8
         :g "M-0"   #'+workspace/switch-to-final
         :g "M-t"   #'+workspace/new
-        :g "M-T"   #'+workspace/display))
+        :g "M-T"   #'+workspace/display
+        (:map evil-window-map
```
hlissner: As of db1c1995f, the ui/workspaces module does this remapping internally, so this keybind can be removed.

hlissner:
Excellent. Thanks for your help!

PR #1645, opened 2019-08-06
Feature: Added support for jsonnet

vikigenius:
This PR adds support for jsonnet using [jsonnet-mode](https://github.com/mgyucht/jsonnet-mode).

### DIFF
--- a/modules/lang/data/config.el
+++ b/modules/lang/data/config.el
@@ -38,9 +38,9 @@
   (set-electric! 'json-mode :chars '(?\n ?: ?{ ?})))
 
 (use-package! jsonnet-mode
-  :mode "\\.jsonnet\\'"
+  :defer "\\.jsonnet\\'"
   :config
-  (set-electric! 'json-mode :chars '(?\n ?: ?{ ?})))
+  (set-electric! 'jsonnet-mode :chars '(?\n ?: ?{ ?})))
 
 ;;
 ;; Frameworks
--- a/modules/lang/data/config.el
+++ b/modules/lang/data/config.el
@@ -38,7 +38,7 @@
   (set-electric! 'json-mode :chars '(?\n ?: ?{ ?})))
 
 (use-package! jsonnet-mode
-  :defer "\\.jsonnet\\'"
+  :defer t
   :config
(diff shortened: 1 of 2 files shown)


modules/lang/data/config.el
```diff
@@ -37,6 +37,10 @@
   :config
   (set-electric! 'json-mode :chars '(?\n ?: ?{ ?})))
 
+(use-package! jsonnet-mode
+  :mode "\\.jsonnet\\'"
```
hlissner: The jsonnet-mode package [already creates an auto-mode-alist entry for itself](https://github.com/mgyucht/jsonnet-mode/blob/2b90b4e12a11c42df0f1e5db327a50555b6ff023/jsonnet-mode.el#L243), so `:mode` can be removed (and replaced with `:defer t`).

modules/lang/data/config.el
```diff
@@ -37,6 +37,10 @@
...
+  :config
+  (set-electric! 'json-mode :chars '(?\n ?: ?{ ?})))
```
hlissner: Shouldn't this be:

```diff
-  (set-electric! 'json-mode :chars '(?\n ?: ?{ ?})))
+  (set-electric! 'jsonnet-mode :chars '(?\n ?: ?{ ?})))

**Comment by vikigenius** [comment]
Yeah, sorry about the 2nd one, silly of me :) . I also made the change from mode to defer and it works fine. Can you review now, it should be fine I think.

File: modules/lang/data/config.el [hunk 3]
```diff
@@ -37,6 +37,10 @@
   :config
   (set-electric! 'json-mode :chars '(?\n ?: ?{ ?})))
 
+(use-package! jsonnet-mode
+  :defer "\\.jsonnet\\'"
```
hlissner: This should be `:defer t`. I don't think `:defer` will understand a string argument.

vikigenius:
yep, wasn't really that familiar with use-package mechanics, been using doom as just an end user so til. good to know, should be fixed now

hlissner:
No sweat! There's no better learning experience than to simply use Emacs. Thanks for the help, it's much appreciated!

PR #1646, opened 2019-08-07
Add symbol-overlay package to :ui

xuning97:
symbol-overlay is an easy-to-use package, with symbol-overlay-put, the
symbol under the cursor will be highlighted across the whole buffer, and
then when cursor if in the symbol, n/p can be used to navigate the symbol.

For more detailed description, please refer to
https://github.com/wolray/symbol-overlay

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -35,6 +35,7 @@
         +all             ; catch all popups that start with an asterix
         +defaults)       ; default popup rules
        ;;pretty-code       ; replace bits of code with pretty symbols
+       ;;symbol-overlay    ; highlight symbols in the same buffer with easy navigation
        ;;tabbar            ; FIXME an (incomplete) tab bar for Emacs
        ;;treemacs          ; a project drawer, like neotree but cooler
        ;;unicode           ; extended unicode support for various languages
--- a/modules/ui/symbol-overlay/packages.el
+++ b/modules/ui/symbol-overlay/packages.el
@@ -1,0 +1,4 @@
+;; -*- no-byte-compile: t; -*-
+;;; ui/symbol-overlay/packages.el
+
+(package! symbol-overlay)
--- a/modules/ui/symbol-overlay/config.el
+++ b/modules/ui/symbol-overlay/config.el
@@ -1,0 +1,8 @@
+;;; ui/symbol-overlay/config.el -*- lexical-binding: t; -*-
+
+(def-package! symbol-overlay
+  :config
+  (map! :leader
+        "c o" 'symbol-overlay-put)
+  (map! :map symbol-overlay-map
+        [escape] 'symbol-overlay-put))


init.example.el
```diff
@@ -41,6 +41,7 @@
        vi-tilde-fringe   ; fringe tildes to mark beyond EOB
        window-select     ; visually switch windows
        workspaces        ; tab emulation, persistence & separate workspaces
+       ;;symbol-overlay  ; hilight symbol in the same buffer with easy navigate
```
filalex77: Probably should be put in alphabetic order in the section.

modules/ui/symbol-overlay/packages.el
```diff
@@ -0,0 +1,4 @@
+;; -*- no-byte-compile: t; -*-
+;;; ui/symbol-overlay/packages.el
+
+(package! symbol-overlay)
```
filalex77: As I can tell, the package requires binding some keys for these functions:
- `symbol-overlay-put`
- `symbol-overlay-switch-forward`
- `symbol-overlay-switch-backward`
- `symbol-overlay-mode`
- `symbol-overlay-remove-all`

First, those bindings need to be decided upon - what is the appropriate prefix to bind them to? Then, you should create a `config.el` where you call the `map!` macro.

init.example.el
filalex77:
```suggestion
       ;;symbol-overlay    ; highlight symbols in the same buffer with easy navigation
```

modules/ui/symbol-overlay/packages.el
xuning97: sure, I think SPC c o could be an candidate for symbol-overlay-put

for others, there is no need of new key bindings, they are automatically enabled when the cursor is on the highlighted word

init.example.el
xuning97: @filalex77 thanks :-), probably I can make the change together with config.el

modules/ui/symbol-overlay/packages.el
filalex77: Sure, the `symbol-overlay-map` seems like enough, and `SPC c o` is free. I think you can go ahead and add it :)

xuning97: symbol-overlay-put is valid even when symbol-overlay-mode is not enabled.

i'm not sure if it's okay to occupy the keybinding globally.

filalex77:
Also, could you please rebase your PR on the `straight` branch? It's due to be merged in the couple days, and all the present developments are directed there.

> would you mind retargeting/rebasing this PR onto the straight branch? It will be merged into develop soon and it'd be a big help if PRs between now and then targeted it instead.
https://github.com/hlissner/doom-emacs/pull/1621#issuecomment-518787600

filalex77:
After playing a bit with `symbol-overlay` in my private config, I have a few things to add:

- We should probably bind `[escape]` in `symbol-overlay-map` to allow for  disabling of the overlay on par with `i` or `SPC c o` again. It might be either `symbol-overlay-put` (to toggle current symbol off) or `symbol-overlay-remove-all` (removes all symbols in buffer, might be overkill)
- The default keymap works pretty well with Evil, but might need some adjustments (e.g. `w` and `e`)

xuning97:
Sure, I’ll look into it

xuning97:
> Also, could you please rebase your PR on the `straight` branch? It's due to be merged in the couple days, and all the present developments are directed there.
> 
> > would you mind retargeting/rebasing this PR onto the straight branch? It will be merged into develop soon and it'd be a big help if PRs between now and then targeted it instead.
> > [#1621 (comment)](https://github.com/hlissner/doom-emacs/pull/1621#issuecomment-518787600)

rebased

xuning97:
> After playing a bit with `symbol-overlay` in my private config, I have a few things to add:
> 
> * We should probably bind `[escape]` in `symbol-overlay-map` to allow for  disabling of the overlay on par with `i` or `SPC c o` again. It might be either `symbol-overlay-put` (to toggle current symbol off) or `symbol-overlay-remove-all` (removes all symbols in buffer, might be overkill)
> * The default keymap works pretty well with Evil, but might need some adjustments (e.g. `w` and `e`)

please kindly have a look at the new update, I didn't change `w` and `e`,  it might be better to keep consistent with original author's choice.

hlissner:
To parrot and expand on @filalex77's sentiment, this config seems too trivial to warrant its own module. I'd happily reconsider if there were more to it, but I'd rather not add modules that could be recreated in a couple lines in one's private config. I also want to avoid setting a precedent for one-module-per-package; each module should represent a feature, rather than different implementations of a feature. An ideal that our modules don't quite live up to yet.

Exceptions to this rule might be features with no other decent (especially built-in) alternative. I'd argue that `symbol-overlay` has two: for evil users, the `*` and `#` highlight all instances of the symbol at point, not to mention `/` or `?` searches. Non-evil users have isearch (particularly `isearch-forward-symbol-at-point`) -- this may not technically be superior to symbol-overlay, but I think it falls under "acceptable built-in alternatives".

xuning97:
I recommend this package mainly because it's really useful while reading code. 
The purpose is different than * #, / ?. It can hight multiple symbols at the same time and keep them there.

For me, I always feel emacs is not that good for reading code comparing to source insight.
But with this package, with its highlight on symbols I cared, I now feel it's comparable to source insight's font style/highlight.

PR #1670, opened 2019-08-13
Added company-phpactor to packages

CodyWilson:
Previously this was included automatically with phpactor, but now the
phpactor package call is only pulling in phpactor.el and its
requirements.

So this is necessary for proper auto-completion.

### DIFF
--- a/modules/lang/php/packages.el
+++ b/modules/lang/php/packages.el
@@ -11,7 +11,8 @@
   (package! hack-mode :recipe (:host github :repo "hhvm/hack-mode")))
 
 (unless (featurep! +lsp)
-  (package! phpactor))
+  (package! phpactor)
+  (package! company-phpactor))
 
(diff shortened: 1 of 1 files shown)


modules/lang/php/packages.el
```diff
@@ -11,7 +11,8 @@
   (package! hack-mode :recipe (:host github :repo "hhvm/hack-mode")))
 
 (unless (featurep! +lsp)
-  (package! phpactor))
+  (package! phpactor)
+  (package! company-phpactor))
```
filalex77:
```suggestion
  (when (featurep! :completion company)
    (package! company-phpactor)))
```

filalex77:
```suggestion
  (when (featurep! :completion company)
    (package! company-phpactor)))
```

hlissner:

PR #1671, opened 2019-08-13
use emacsmirror for cmake-mode and protobuf-mode

fread2281:
To avoid giant git repos


hlissner:
Would this still work if you omitted the :files property? 

`package!` is configured to merge user-specified recipes with the default recipes; less configuration means less to maintain should the recipe change upstream.

fread2281:
The `:files` seems to be necessary (without it, opening the corresponding
file types doesn't activate the modes).

hlissner:
Understood. Thanks for looking into and addressing this!

PR #1675, opened 2019-08-15
Add org-mode keybindings

RBckmnn:
I added two key-bindings for adding attachments and setting properties
in org-mode.


hlissner:
Thanks for the help!

PR #1677, opened 2019-08-16
:tools/debugger: Add rust binding for dap-lldb

juliandemille:
Rust is supposed to be used with dap-lldb when using dap-mode. This patch adds an entry for that.


hlissner:
Thanks for the help!

PR #1678, opened 2019-08-16
:lang/rust: Add rustic-lsp

juliandemille:
Adding rustic-lsp allows users to select rust-analyzer as their language server.

### DIFF
--- a/modules/lang/rust/config.el
+++ b/modules/lang/rust/config.el
@@ -60,6 +60,11 @@
       (apply orig-fn args))))
 
 
+(use-package! rustic-lsp
+  :when EMACS26+
+  :after rustic)
+
+
 ;;
 ;;; Tools
 


filalex77:
Isn't `rustic-lsp` a part of `rustic` and enabld by default? And `rust-analyzer` can already be selected as the server of choice by following instructions in the `README` for `rustic` (source: tested myself).

juliandemille:
It is a part of `rustic`, but it doesn't seem to be loaded by default. Maybe I'm missing something. I went looking because changing that setting didn't enable `rust-analyzer` for me.

filalex77:
Thank you for your attention. I'll see if it's an issue on a fresh installation - there may be something to it. If you have anything to add, feel free to reopen this issue.

juliandemille:
Turns out I was being an idiot and forgot to add my `PATH` from `fish` to Emacs. It now works with `rust-analyzer`.

PR #1704, opened 2019-08-23
Fixing up csharp mode's keybinds (mostly unit-test ones)

Oreodave:
*TL;DR* Found that the unit test keybinds are using broken keybinds that don't work
anymore. Decided to fix them, along with any keybinds that seem to not work.

I personally found this to actually hinder my workflow, and added my
own keybinds to use the updated functions. However, the dead keybinds that don't
work anymore would always come up on which-key, almost trying to ridicule me for
my laziness in not fixing this issue myself. So here I am to fix it!

Not much of a big change so I doubt any real review is needed. Each commit has
decent information (well I suppose anyway).

### DIFF
--- a/modules/lang/csharp/config.el
+++ b/modules/lang/csharp/config.el
@@ -36,7 +36,6 @@
         :map omnisharp-mode-map
         "b" #'omnisharp-recompile
         (:prefix "r"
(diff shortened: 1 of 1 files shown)


hlissner:

PR #1711, opened 2019-08-26
emacs/dired: use diff-hl instead of dired-k

ztlevi:
`diff-hl` is more updated. It is also able to handle ignored `.` and `..` comes with `dired-omit-mode` while `dired-k` doesn't. 

There is a minor problem. `diff-hl-dired-mode` isn't updated (showing the wrong gutters) after calling `ranger-up-directory`. I believe if dired reuse the buffer, it doesn't work as well. I tried to hack it but failed.


hlissner:
Good work. Thanks again for your help!

> I tried to play around with diff-hl-dired-process-buffer and diff-hl-dired-update to make diff-hl updated after calling ranger-up-directory. It's kind of tricky and I didn't make it working. Would you mind taking a look at it?

Roger that. I'll look into it next weekend.

PR #1716, opened 2019-08-27
Add ln to +sh-builtin-keywords

flatwhatson:
Removed duplicate "cat" and added missing "ln", which IMO is common enough to be included.

### DIFF
--- a/modules/lang/sh/config.el
+++ b/modules/lang/sh/config.el
@@ -1,9 +1,9 @@
 ;;; lang/sh/config.el -*- lexical-binding: t; -*-
 
 (defvar +sh-builtin-keywords
-  '("cat" "cat" "cd" "chmod" "chown" "cp" "curl" "date" "echo" "find" "git"
-    "grep" "kill" "less" "ls" "make" "mkdir" "mv" "pgrep" "pkill" "pwd"
-    "rm" "sleep" "sudo" "touch")
+  '("cat" "cd" "chmod" "chown" "cp" "curl" "date" "echo" "find" "git" "grep"
+    "kill" "less" "ln" "ls" "make" "mkdir" "mv" "pgrep" "pkill" "pwd" "rm"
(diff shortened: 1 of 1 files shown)


chrunchyjesus:
Might as well add the whole collection of gnu coreutils:
b2sum
base32
base64
basename
basenc
cat
chcon
chgrp
chmod
chown
chroot
cksum
comm
cp
csplit
cut
date
dd
df
dir
dircolors
dirname
du
echo
env
expand
expr
factor
false
fmt
fold
head
hostid
id
install
join
link
ln
logname
ls
md5sum
mkdir
mkfifo
mknod
mktemp
mv
nice
nl
nohup
nproc
numfmt
od
paste
pathchk
pinky
pr
printenv
printf
ptx
pwd
readlink
realpath
rm
rmdir
runcon
seq
sha1sum
sha224sum
sha256sum
sha384sum
sha512sum
shred
shuf
sleep
sort
split
stat
stdbuf
stty
sum
sync
tac
tail
tee
test
timeout
touch
tr
true
truncate
tsort
tty
uname
unexpand
uniq
unlink
users
vdir
wc
who
whoami
yes

hlissner:
Thanks again for your help! I originally wanted to keep this command list short, but I'll consider the rest of the GNU coreutils to it soon.

chrunchyjesus:
Maybe there is some way to infer it from `$PATH`?

hlissner:
There is:

```elisp
(setq +sh-builtin-keywords
      (delete-dups
       (mapcar #'file-name-nondirectory
               (doom-files-in exec-path
                              :depth 0
                              :filter (lambda (f) (not (file-executable-p f)))))))
```

However, most PATHs contain thousands of binaries. That would be hell on Emacs' fontifier, so I'd rather restrict it to a much smaller subset of these utilities.

chrunchyjesus:
Ah okay. Didn't know that.

PR #1726, opened 2019-08-29
Fix defadvice bug of Chinese and Japanese org-html-paragraph

cnsunyour:
Previously submitted a fix in issues, which is the processing of \<br /\>\n in Chinese and English mixed text. Later, I found that this would be invalid for the case of \<br /\>\n between pure English, so I made the following changes: first delete only \<br /\>, then \n between multi-byte text Delete, so you can adapt to all situations.


hlissner:
Nice work. Thank you for addressing this!

PR #1727, opened 2019-08-29
Load hydra examples without using example hydras.

jjzmajic:
`+hydra/window-nav` uses functions from `hydra-examples.el` but the file
is never loaded. Adding this does not actually define any hydras (this
would require `hydra-examples-verbatim` to be `t`), it just load the
utility functions to make the hydras defined in doom work.

### DIFF
--- a/modules/ui/hydra/config.el
+++ b/modules/ui/hydra/config.el
@@ -1,4 +1,10 @@
 ;;; ui/hydra/config.el -*- lexical-binding: t; -*-
 
+(use-package! hydra-examples
+  :commands (hydra-move-splitter-up
+             hydra-move-splitter-down
+             hydra-move-splitter-right
+             hydra-move-splitter-left))
+
 ;;;###package hydra
 (setq lv-use-seperator t)


jjzmajic:
Should I add in a binding as well? Maybe `SPC w ,` with `,` being the convention for hydras (as doom does not have one yet)? I just assumed this was intentional and for the user to configure. But it would probably be more discoverable when bound.

filalex77:
Bindings in modules should be discussed with @hlissner. I personally think your suggestio would be a great fit!

hlissner:
I don't like that hydra-example is eagerly loaded (it pulls in hydra along with it at startup), and that its hydras aren't autoloaded. We'll have to define autoloads for them with :commands.

jjzmajic:
Just proposing it. Any convention is fine by me, I just think it should be easily reachable and uniform (and probably not a letter as it will likely clash with mnemonic key-bindings). Added the missing commands btw

hlissner:
I haven't yet decided a keybinding scheme for hydras yet, but I will merge this PR for the time being. Thanks for the help! (And sorry for the late response!)

jjzmajic:
No worries! PS: using `.` is fairly popular as well. Just food for thought while you're deliberating. I really don't have strong preferences when it comes to this as long as its consistent :)

PR #1729, opened 2019-08-29
ox-hugo for (:lang org)

zenlor:
add support for [ox-hugo](https://github.com/kaushalmodi/ox-hugo) to the org module. The change is straightforward as most of the module configuration is done in the org-files themselves.

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -127,6 +127,7 @@
        ;;ocaml             ; an objective camel
        (org              ; organize your plain life in plain text
         +dragndrop       ; drag & drop files/images into org buffers
+        ;+hugo           ; use Emacs for hugo blogging
         +ipython         ; ipython/jupyter support for babel
         +pandoc          ; export-with-pandoc support
         +present)        ; using org-mode for presentations
--- a/modules/lang/org/README.org
+++ b/modules/lang/org/README.org
@@ -61,6 +61,8 @@ https://www.mfoot.com/blog/2015/11/22/literate-emacs-configuration-with-org-mode
 + =+pandoc= Enables pandoc integration into the Org exporter.
 + =+present= Enables integration with reveal.js, beamer and org-tree-slide, so
   Emacs can be used for presentations.
++ =+hugo= Enables integration with [[https://gohugo.io][hugo]] to export from
+  Emacs well formed (blackfriday) markdown.
 
 ** Plugins
 + [[https://orgmode.org/][org-plus-contrib]]
@@ -99,6 +101,8 @@ https://www.mfoot.com/blog/2015/11/22/literate-emacs-configuration-with-org-mode
   + [[https://github.com/anler/centered-window-mode][centered-window]]
   + [[https://github.com/takaxp/org-tree-slide][org-tree-slide]]
   + [[https://github.com/yjwen/org-reveal][ox-reveal]]
++ =+hugo=
+  + [[https://github.com/kaushalmodi/ox-hugo][ox-hugo]]
 
 ** Hacks
 + The window is recentered when following links.
--- a/modules/lang/org/README.org
+++ b/modules/lang/org/README.org
@@ -62,7 +62,7 @@ https://www.mfoot.com/blog/2015/11/22/literate-emacs-configuration-with-org-mode
 + =+present= Enables integration with reveal.js, beamer and org-tree-slide, so
   Emacs can be used for presentations.
 + =+hugo= Enables integration with [[https://gohugo.io][hugo]] to export from
-  Emacs well formed (blackfriday) markdown.
+  Emacs well-formed ([[https://github.com/russross/blackfriday][blackfriday]]) markdown.
 
 ** Plugins
 + [[https://orgmode.org/][org-plus-contrib]]
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -868,6 +868,7 @@ compelling reason, so..."
 
   ;;; Custom org modules
   (if (featurep! +dragndrop) (load! "contrib/dragndrop"))
+  (if (featurep! +hugo)      (load! "contrib/+hugo"))
   (if (featurep! +ipython)   (load! "contrib/ipython"))
   (if (featurep! +present)   (load! "contrib/present"))
 
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -394,10 +394,6 @@ file isn't in `org-directory'."
             (mathjax . t)
             (variable . "revealjs-url=https://cdn.jsdelivr.net/npm/reveal.js@3/"))))
 
-  (when (featurep! +hugo)
-    (use-package! ox-hugo
-      :after ox)))
-
 
 
 (defun +org-init-habit-h ()
@@ -566,10 +562,7 @@ between the two."
           (:prefix ("b" . "from beamer")
             :desc "to latex"            "l" #'org-beamer-export-to-latex
             :desc "to latex & open"     "L" #'org-beamer-export-as-latex
-            :desc "as pdf"              "p" #'org-beamer-export-to-pdf)
-          (:when (featurep! +hugo)
-            :desc "to hugo"        "h" #'org-hugo-export-to-md
-            :desc "to hugo & open" "H" #'org-hugo-export-as-md))
+            :desc "as pdf"              "p" #'org-beamer-export-to-pdf))
         (:prefix ("g" . "goto")
           "g" #'org-goto
           (:when (featurep! :completion ivy)
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -392,8 +392,7 @@ file isn't in `org-directory'."
     (setq org-pandoc-options
           '((standalone . t)
             (mathjax . t)
-            (variable . "revealjs-url=https://cdn.jsdelivr.net/npm/reveal.js@3/"))))
-
+            (variable . "revealjs-url=https://cdn.jsdelivr.net/npm/reveal.js@3/")))))
 
 
 (defun +org-init-habit-h ()
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -392,7 +392,12 @@ file isn't in `org-directory'."
     (setq org-pandoc-options
           '((standalone . t)
             (mathjax . t)
-            (variable . "revealjs-url=https://cdn.jsdelivr.net/npm/reveal.js@3/")))))
+            (variable . "revealjs-url=https://cdn.jsdelivr.net/npm/reveal.js@3/"))))
+
+  (when (featurep! +hugo)
+    (use-package! ox-hugo
+      :after ox)))
+
 
 
 (defun +org-init-habit-h ()
@@ -561,7 +566,10 @@ between the two."
           (:prefix ("b" . "from beamer")
             :desc "to latex"            "l" #'org-beamer-export-to-latex
             :desc "to latex & open"     "L" #'org-beamer-export-as-latex
-            :desc "as pdf"              "p" #'org-beamer-export-to-pdf))
+            :desc "as pdf"              "p" #'org-beamer-export-to-pdf)
+          (:when (featurep! +hugo)
+            :desc "to hugo"        "h" #'org-hugo-export-to-md
+            :desc "to hugo & open" "H" #'org-hugo-export-as-md))
         (:prefix ("g" . "goto")
           "g" #'org-goto
           (:when (featurep! :completion ivy)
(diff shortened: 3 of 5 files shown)


init.example.el
```diff
@@ -128,7 +128,8 @@
         +dragndrop       ; file drag & drop support
         +ipython         ; ipython support for babel
         +pandoc          ; pandoc integration into org's exporter
-        +present)        ; using Emacs for presentations
+        +present         ; using Emacs for presentations
+        +hugo)           ; use Emacs for hugo blogging
```
filalex77: Alphabetic order can be preserved, because why not? :)

modules/lang/org/README.org
```diff
@@ -61,6 +61,8 @@ https://www.mfoot.com/blog/2015/11/22/literate-emacs-configuration-with-org-mode
 + =+pandoc= Enables pandoc integration into the Org exporter.
 + =+present= Enables integration with reveal.js, beamer and org-tree-slide, so
   Emacs can be used for presentations.
++ =+hugo= Enables integration with [[https://gohugo.io][hugo]] to export from
+  Emacs well formed (blackfriday) markdown.
```
filalex77:
```suggestion
  Emacs to well-formed ([[https://github.com/russross/blackfriday][blackfriday]]) markdown.
```

modules/lang/org/config.el
```diff
@@ -396,7 +396,12 @@ file isn't in `org-directory'."
     (setq org-pandoc-options
           '((standalone . t)
             (mathjax . t)
-            (variable . "revealjs-url=https://cdn.jsdelivr.net/npm/reveal.js@3/")))))
+            (variable . "revealjs-url=https://cdn.jsdelivr.net/npm/reveal.js@3/"))))
+
+  (when (featurep! +hugo)
+    (use-package! ox-hugo
+      :after ox)))
```
filalex77:
```diff
-   (when (featurep! +hugo)
-     (use-package! ox-hugo
-       :after ox)))
+   (use-package! ox-hugo
+     :when (featurep! +hugo)
+     :after ox))
```

modules/lang/org/README.org
zenlor: resolved in #921a8f8

init.example.el
zenlor: resolved in 921a8f8

modules/lang/org/config.el
zenlor: resolved in cc33707

init.example.el
```diff
@@ -126,9 +126,10 @@
        ;;ocaml             ; an objective camel
        (org              ; organize your plain life in plain text
         +dragndrop       ; file drag & drop support
+        +hugo            ; use Emacs for hugo blogging
         +ipython         ; ipython support for babel
         +pandoc          ; pandoc integration into org's exporter
-        +present)        ; using Emacs for presentations
+        +present)         ; using Emacs for presentations
```
filalex77:
```suggestion
        +present)        ; using Emacs for presentations
```

I hate to be nitpicky. Sorry :(

modules/lang/org/packages.el
```diff
@@ -49,5 +49,6 @@
   (package! centered-window :recipe (:host github :repo "anler/centered-window-mode"))
   (package! org-tree-slide)
   (package! ox-reveal))
-(when (featurep! +hugo)
-  (package! ox-hugo))
+(package! ox-hugo
+  :when (featurep! +hugo)
```
filalex77: I think `package!` does not support the `:when` property - only `use-package!`. Wrapping `package!` in a `when` is fine.

filalex77:
I just thought about having hugo stuff in `contrib/+hugo.el` (similar to how the other flags are implemented). What do you think about that? @zenlor

init.example.el
zenlor: woups! :D

zenlor:
> I just thought about having hugo stuff in `contrib/+hugo.el` (similar to how the other flags are implemented). What do you think about that? @zenlor

It's fine, I honestly didn't even see the folder, I'll move it

modules/lang/org/packages.el
```diff
@@ -49,3 +49,8 @@
   (package! centered-window :recipe (:host github :repo "anler/centered-window-mode"))
   (package! org-tree-slide)
   (package! ox-reveal))
+(when (featurep! +hugo)
+  (package! ox-hugo :recipe (:host github
+                             :repo "kaushalmodi/ox-hugo"
+                             :nonrecursive t
+                             :files (:defaults "*.el"))))
```
filalex77: Does this recipe work without `:nonrecursive` and `:files`? `ox-hugo` seems like a pretty normal Elisp repo for me, with `*.el` in the root and no submodules.

filalex77: Oh, I see.

> to avoid downloading the whole hugo configuration of the repo' documentation website

It's fine then.

zenlor:
I had to change the straight :recipe in e9c58f9, previously I forgot to add my whole private config to the commit

modules/lang/org/packages.el
zenlor: without the recipe it's going to download the whole hugo theme (https://github.com/kaushalmodi/ox-hugo/blob/master/.gitmodules). It's used by the dev for testing his package

zenlor: probably `:nonrecursive t` is enough, I'll have to test

modules/lang/org/contrib/+hugo.el
```diff
@@ -0,0 +1,12 @@
+;;; +hugo.el --- ox-hugo support -*- lexical-binding: t; -*-
+;;;###if (featurep! +hugo)
+
+(use-package! ox-hugo
+  :after ox
+  (map! :map org-mode-map
+        :localleader
+        (:prefix ("e" . "export")
+          :desc "to hugo"        "h" #'org-hugo-export-to-md
```
kaushalmodi: Hello! Thanks for working on adding `ox-hugo` to this config. My only suggestion would be to use the suggeste export functions .. may be `h` for `org-hugo-export-wim-to-md` (default binding `C-c C-e H H`) and `H` for `(org-hugo-export-wim-to-md :all-subtrees)` (default binding `C-c C-e H A`). 

It would be cool though if you can match the two-letter bindings for the export interface .. `HH` for the first function, and `HA` for the second. 

See https://ox-hugo.scripter.co/#export-bindings.

kaushalmodi: The reason to have Hugo export binding to be capital H is that by default the lower case h binding is for the inbuilt `ox-html`. Isn't that the case for the ox export interface in doom too.

zenlor: The export sub-menu doesn't have a binding for ox-html, it probably should. I never noticed as I usually simply use M-x.

In any case I can add a sub-menu to the local config mimicking the default bindings

kaushalmodi: > In any case I can add a sub-menu to the local config mimicking the default bindings

Can you do that to this PR? That way, there's a less of "context switch" for folks switching between a non-doom config and doom config i.e. `h` in the ox export menu should export using `ox-html`, and `H` using `ox-hugo`.

Thanks.

zenlor:
@kaushalmodi In c7b6f4f I added the :prefix keeping the same keys and descriptions as the `ox` export menu `C-c C-e H ...`.

It is a bit strange because the `org-hugo-export-wim-to-md` function is repeated and I had to add a couple of arguments to the package's interactive functions, please see if function names require a different name by doom convention.

kaushalmodi:
Thanks to this PR, I have improved the descriptions of the functions in the export menu in ox-hugo: https://github.com/kaushalmodi/ox-hugo/commit/6002903c21550244cb119bc7b265a733d578915f :)

If you do not mind, can you please port over those descriptions to here?

zenlor:
I just imported the new action descriptions from  kaushalmodi/ox-hugo@a8e0c6e (forced to change the commit message)

There might just be another minor issue with description's length where they are elided

Screenshot

There is an actual issue executing the `map!` macro in `use-package!`, I have been testing `:init` and `:config` and it seems it's not being executed

modules/lang/org/contrib/+hugo.el
```diff
@@ -0,0 +1,24 @@
...
+(use-package! ox-hugo
+  :after ox)
+
+(map! :map org-mode-map
+      :localleader
+      (:prefix "e"
+        (:prefix ("H" . "hugo")
+          :desc "Subtree or File to Md to file"        "H" #'org-hugo-export-wim-to-md
+          :desc "File to Md file"                      "h" #'org-hugo-export-to-md
+          :desc "Subtree or File to Md to file & open" "O" '(lambda ()
```
hlissner: Instead of quoted lambdas, please use the `λ!` macro (which is short for `(lambda () (interactive) ...)`, like so:

```elisp
(λ! (org-open-file (org-hugo-export-to-md)))
```

modules/lang/org/contrib/+hugo.el
```diff
@@ -0,0 +1,24 @@
+;;; +hugo.el --- ox-hugo support -*- lexical-binding: t; -*-
+;;;###if (featurep! +hugo)
+
+(use-package! ox-hugo
+  :after ox)
+
+(map! :map org-mode-map
```
hlissner: Please move this `map!` call into the above `use-package!` block for ox-hugo (under `:init`):

```elisp
(use-package! ox-hugo
  :after ox
  :init
  (map! ...))
```

zenlor:
I just rebased the whole PR, it should be clean now (at least on this side :))

modules/lang/org/contrib/+hugo.el
```diff
@@ -0,0 +1,18 @@
...
+  :init
+  (map! :map org-mode-map
+        :localleader
+        (:prefix "e"
+          (:prefix ("H" . "hugo")
+            :desc "Subtree or File to Md to file"        "H" #'org-hugo-export-wim-to-md
+            :desc "File to Md file"                      "h" #'org-hugo-export-to-md
+            :desc "Subtree or File to Md to file & open" "O" '(λ! (org-open-file (org-hugo-export-wim-to-md)))
+            :desc "File to Md file & open"               "o" '(λ! (org-open-file (org-hugo-export-to-md)))
+            :desc "All subtrees (or File) to Md file(s)" "A" '(λ! (org-hugo-export-wim-to-md :all-subtrees))
```
hlissner: Remove the leading quote on these lambdas. They shouldn't be quoted and are likely broken otherwise.

zenlor: I'm honestly having issues with `(!map` after putting it in the `use-package!` init section, the whole H prefix isn't loaded. With or without quotes.

hlissner: @zenlor It's likely that `org-mode-map` isn't defined when the `use-package!` block gets evaluated. Try this instead:

```diff
   :after ox
   :init
-  (map! :map org-mode-map
+  (map! :after org
+        :map org-mode-map
         :localleader
         (:prefix "e"
```

hlissner: This may be because `org-mode-map` isn't defined when this `use-package!` block is evaluated. Try this instead:

```diff
   :after ox
   :init
-  (map! :map org-mode-map
+  (map! :after org
+        :map org-mode-map
         :localleader
         (:prefix "e"

Comment (zenlor) [reply modules/lang/org/contrib/+hugo.el]: It seems that the package itself is not loaded. In my setup it would load because I'm using a `.dir-locals.el` file containing:

```
(("org/"
  . ((org-mode . ((eval . (org-hugo-auto-export-mode)))))))
```

In the `use-package!` macro, I have set the `:after ox`, but, of course, ox is not loaded before calling an export function. Using `:after org` it will load and execute the `!map` macro adding the prefix. The downside is that now, I guess, the package will load when opening any org file. I doesn't happen in other contrib modules like reveal since the changes are right in the main `config.el`.

hlissner: I'm not clear about what the problem is. The change I suggested will load the keybinds when org is loaded; not the whole package, which is what we want, no?

zenlor: Yes, the only issue is that the package will only load when the `ox` package is loaded, after `ox` is loaded the `map!` will appear. At first it won't show up (or even work) in the `local->e` prefix

hlissner: Ah, in that case, use `:preface` instead of `:init`. Its body is assured to run immediately.

Normally, `:init` has this role, but the rules seem to change when `:after` is involved.

zenlor: thanks for the tip, it's finally showing up as expected. I should definitely read more about package.el and the use-package! macro

hlissner:
Great work! Thanks so much for your help!

zenlor:
Thank you for all the tips!

PR #1731, opened 2019-08-29
Use straight for agda

rgrinberg:
We can fetch the mode directly from straight. There's no need to rely on
having this .el file installed for us. Now agda's el files can be edited
directly like all other emacs packages.


rgrinberg:
I was missing the actual `packages.el` for agda. I've added it.

hlissner:
This is certainly much nicer. Thanks for the help!

PR #1732, opened 2019-08-29
lang/org: smarter smartparens in org src blocks

UndeadKernel:
While point is in a babel src block, do not autoexpand emphasis markers.

Without this change, editing code in babel src blocks is cumbersome: every time you type `*` or `=` (among other characters), a pair is inserted that then needs to be manually deleted. This change checks if point is in a source block to decide if emphasis markers are duplicated.


hlissner:
Great work. That has certainly bothered me before. Thanks again for the help!

PR #1734, opened 2019-08-29
Fix/add add featurep condition to global keybindings

JakDar:
Added `(:when featurep! [...])` for popup, indent-guides, flycheck and flyspell bindings in `+evil-bindings.el` and  for popup bindings in `+emacs-bindings.el`


hlissner:
Sorry for the late response. Thank you for the PR!

PR #1735, opened 2019-08-30
cli/upgrade: add timeout flag

seanfarley:
Since there is a lot of logic in `bin/doom upgrade` than just running
two commands, I thought we should also add a timeout flag just like
`update` has.

I was having lots of timeout issues so I found this flag helpful.


hlissner:

PR #1741, opened 2019-08-31
Add helm-org (bound but not installed).

jjzmajic:
As the title states, commands from this package are bound, but the package is not installed. Let me know if I should fix anything.


hlissner:
Sorry for the late response and thanks for the PR!

jjzmajic:
De nada :)

PR #1742, opened 2019-09-01
Call renamed hook in the +cc/reload-compile-db autoload

valignatev:
It seems that during the renaming hooks to a new convention one call in the cc lang feature autoloads remained unchanged which causes `Symbol's function is void: +cc|irony-init-compile-options`. This PR fixes that.


hlissner:

Issue #1705, opened 2019-08-24
[QUESTION] Concerning which-key labels


RBckmnn:
Hi Henrik,

a few days ago we briefly talked on the discord channel about which-key labels for key-bindings and you explained to me why the labels that I added didn't work. (I added labels to bindings in minor-mode key-maps).

It seems that which-key has gained support for key-bindings in minor modes somewhat recently. At least this is my understanding of the following comment of which-key's maintainer:

https://github.com/justbur/emacs-which-key/issues/212#issuecomment-503368293

I don't know if you were aware of which-keys `which-key-enable-extended-define-key` option and if it can be used to add more labels to doom' key-bindings so I thought I bring it to your attention.

If you think it would be desirable to use the new which-key labeling option, I can try to make the neccessary changes, despite of my quite bad elisp skills :-).

Cheers!

hlissner:
This setting is on my radar, but I'm afraid is not a perfect solution. First, it would require a non-trivial amount of work to implement (setting `which-key-enable-extended-define-key` won't be enough; `map!` would need to be rewritten and we'd have to jury rig [general](https://github.com/noctuid/general.el/issues) to use the new way of declaring which-key labels, rather than the old, or replace general altogether, which is deeply woven into Doom).

Second, this new approach doesn't support remapped commands, which are used extensively (at least a third of all our commands are remapped, depending on what modules you have enabled, not counting any remapping done by the plugins we install). To demonstrate this issue:

```elisp
(defun command-a ()
  (interactive)
  (message "hi"))
(defun command-b ()
  (interactive)
  (message "bye"))

(define-key global-map (kbd "C-c .") '("A" . command-a))
(define-key global-map (kbd "C-c ,") '("B" . command-b))
```

Upon which-key opening, `C-c .` has the "A" label and `C-c ,` has the "B" label, as expected, until...

```elisp
(define-key global-map [remap command-a] #'command-b) 
```

`C-c .` no longer has a label, at all. This won't work either:

```elisp
(define-key global-map [remap command-a] '("Remapped B" . command-b))
```

-----

These are show stoppers and, between sacrificing which-key labels for keys on arbitrary keymaps and labels for our leader/localleader keys, I think the latter is less acceptable than the former. What's worse, I don't think there's an obvious or simple solution (on which-key's end) to this issue of remapped commands, besides simply associating labels with commands directly (i.e. rather than associating them with key sequences, keymaps or modes).

[...]

PR #1744, opened 2019-09-02
evil-mc binding descriptions

duog:
This PR:
Fixes interactive pattern for `+multiple-cursors:evil-mc` as noted on discord.

Adds descriptions for evil-mc key bindings;
These names were too long to be useful in my which-key. I think this is
an improvement?

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -371,19 +371,32 @@
 
       (:when (featurep! :editor multiple-cursors)
         ;; evil-mc
-        (:prefix "gz"
+        (:prefix ("gz" . "Multiple cursors")
+          :desc "Make and goto next match"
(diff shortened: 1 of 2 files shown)


hlissner:
Thanks for the PR! Sorry to say I cannot accept it for two reasons:

1. The interactive code was already corrected before this PR was submitted.
2. A limitation of which-key prevents us from properly labeling keybinds that aren't on a major mode keymap (e.g. `rust-mode-map`). normal+visual state keybinds are bound on `evil-normal-state-map` and `evil-visual-state-map`, neither of which are major mode keymaps. This may _appear_ to work, but you are effectively labeling _gz_ and all its subkeys, globally, wherever a gz keybind exists, regardless of keymap or mode. This was discussed in #1705. There is no workaround for this.

PR #1759, opened 2019-09-06
Add a module for the lean theorem prover

rgrinberg:
This isn't full working yet - I'm not sure why company is failing. I appreciate any help however.

### DIFF
--- a/modules/lang/lean/config.el
+++ b/modules/lang/lean/config.el
@@ -1,33 +1,34 @@
 ;;; lang/lean/config.el -*- lexical-binding: t; -*-
 
-(use-package! company-lean
-  :when (featurep :completion 'company)
-  :after lean-mode
-  :config (set-company-backend! 'lean-mode 'company-lean))
-
-(use-package! lean-mode
-  :config
+(after! lean-mode
   (set-lookup-handlers! 'lean-mode
     :definition #'lean-find-definition)
   (sp-with-modes 'lean-mode
     (sp-local-pair "/-" "-/")
-    (sp-local-pair "`'" nil :actions :rem)
+    (sp-local-pair "`" "`")
     (sp-local-pair "{" "}")
-    (sp-local-pair "«" "»"))
-  (map!
-   :map lean-mode-map
-   :localleader
-   "g" #'lean-toggle-show-goal
-   "n" #'lean-toggle-next-error
-   (:prefix "s"
-     "r" #'lean-server-restart
-     "s" #'lean-server-stop
-     "v" #'lean-server-switch-version)
-   (:prefix "p"
-     "t" #'lean-leanpkg-test
-     "b" #'lean-leanpkg-build
-     "c" #'lean-leanpkg-configure)
-   "f" #'lean-fill-placeholder
-   "h" #'lean-hole
-   "m" #'lean-message-boxes-toggle
-   "e" #'lean-execute))
+    (sp-local-pair "«" "»")
+    (sp-local-pair "⟨" "⟩")
+    (sp-local-pair "⟪" "⟫"))
+  (map! :map lean-mode-map
+        :localleader
+        "g" #'lean-toggle-show-goal
+        "n" #'lean-toggle-next-error
+        (:prefix ("s" . "server")
+          "r" #'lean-server-restart
+          "s" #'lean-server-stop
+          "v" #'lean-server-switch-version)
+        (:prefix ("p" . "leanpkg")
+          "t" #'lean-leanpkg-test
+          "b" #'lean-leanpkg-build
+          "c" #'lean-leanpkg-configure)
+        "f" #'lean-fill-placeholder
+        "h" #'lean-hole
+        "m" #'lean-message-boxes-toggle
+        "e" #'lean-execute))
+
+
+(use-package! company-lean
+  :when (featurep! :completion company)
+  :after lean-mode
+  :config (set-company-backend! 'lean-mode 'company-lean))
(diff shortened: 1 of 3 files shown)


rgrinberg:
It seems like I managed to fix this by getting rid of use-package

modules/lang/lean/config.el
```diff
@@ -0,0 +1,34 @@
+;;; lang/lean/config.el -*- lexical-binding: t; -*-
+
+(use-package! company-lean
+  :when (featurep! :completion company)
+  :after lean-mode
+  :config (set-company-backend! 'lean-mode 'company-lean))
+
+(map!
```
hlissner: For consistency with other uses of `map!` throughout Doom, please move make `:after` inline with `map!` (and reindent the rest of the map! block).

```diff
-(map!
- :after coq-mode
- :map coq-mode-map
+(map! :after coq-mode
+      :map coq-mode-map
```

hlissner: And please move this `map!` block to the end of the `(after! lean-mode ...)` block below (and remove `:after lean-mode`). I want config grouped as much as possible.

modules/lang/lean/config.el
```diff
@@ -0,0 +1,34 @@
+;;; lang/lean/config.el -*- lexical-binding: t; -*-
+
+(use-package! company-lean
```
hlissner: This is a minor issue, but please move this `use-package!` block to the end of the file. Conditional (and support) packages should follow more important packages.

And as per conventions, please ensure there are two blank lines separating lean-mode's config and company-lean's use-package! block.

modules/lang/lean/config.el
```diff
@@ -0,0 +1,34 @@
...
+ "h" #'lean-hole
+ "m" #'lean-message-boxes-toggle
+ "e" #'lean-execute)
+
+(after! lean-mode
+  (set-lookup-handlers! 'lean-mode
+    :definition #'lean-find-definition)
+  (sp-with-modes 'lean-mode
+    (sp-local-pair "/-" "-/")
+    (sp-local-pair "`'" nil :actions :rem)
```
hlissner: Does this work? Shouldn't it be: `(sp-local-pair "`" "'" :actions :rem)`?

rgrinberg: Ah, yeah I got rid of it.

hlissner:
Nice work! Is company still not working for you? And if not:

1. How does it fail? Are any errors or warnings emitted into \*Messages\*?
2. Could you report the value of `company-backends` in a lean-mode buffer?
3. Can you reproduce this behavior in a vanilla Doom instance ([using the `C-c C-p` option in the sandbox](https://gist.github.com/hlissner/3910fcf5ef14ab77657f269ba9b37f94))?

rgrinberg:
The company mode is now working fine, however, it seems like it's impossible to disable.

There's this pesky hook lean is enabling:

```
;;;###autoload
(defun company-lean-hook ()
  (set (make-local-variable 'company-backends) '(company-lean))
  (setq-local company-tooltip-limit 20)                      ; bigger popup window
  (setq-local company-minimum-prefix-length 5)
  (setq-local company-idle-delay nil)                        ; decrease delay before autocompletion popup shows
  ;(setq-local company-echo-delay 0)                          ; remove annoying blinking
  (setq-local company-begin-commands '(self-insert-command)) ; start autocompletion only after typing
  (company-mode t))

;;;###autoload
(add-hook 'lean-mode-hook #'company-lean-hook)
```

But otherwise, things are working fine.

hlissner:
That is frustrating. However, there is an easy fix for that. Try adding `:init (remove-hook 'lean-mode-hook #'company-lean-hook)` to company-lean's `use-package!` block. 

Since Doom loads all autoloads super early in Doom's startup process the hook will be ready to be removed by the time lang/lean is loaded, so that one line should be enough to reverse it.

hlissner:
Oh, and since it runs again once the package is loaded, I guess you'd have to add `(remove-hook 'lean-mode-hook #'company-lean-hook)` _again_, in `company-lean`'s `:config` block too.

Perhaps `(advice-add #'company-lean-hook :override #'ignore)` would be more convenient (we'd only need it once, in company-lean's :init hook)...

rgrinberg:
I went for the advice solution. However, now company mode stopped working. What is the correct place to put this line:

```
  (set-company-backend! 'lean-mode 'company-lean)
```

I put it in company-lean's `:init`, but it doesn't seem to set `company-backends` for lean mode files.

hlissner:
Hmm, this might be related to some company load order issues I'm trying to deal with. Does this only affect the first lean-mode buffer? Or does it affect all future lean-mode buffers too?

rgrinberg:
> Hmm, this might be related to some company load order issues I'm trying to deal with. Does this only affect the first lean-mode buffer? Or does it affect all future lean-mode buffers too?

I think I spoke too soon, everything is working fine. It's just the company mode for lean is a little finicky.

hlissner:
Understood. Oh, could you add lean to init.example.el? Below `latex`, above `ledger`? Then I think this will be ready to merge.

rgrinberg:
> Understood. Oh, could you add lean to init.example.el? Below latex, above ledger? Then I think this will be ready to merge.



Done. Although I can't think of a witty sentence to describe it. I'll give you the honours :)

rgrinberg:
Ping :) anything else I need to add?

hlissner:
Nope, it was fine as is. Just short on time. I'll get back to that witty sentence once inspiration strikes. Anyhow, thanks for the help!

PR #1760, opened 2019-09-07
Fix a regexp bug of (defadvice! +plantuml--fix-atstart-in-org-src-blocks-a ...)

cnsunyour:
Sorry for my negligence, the whitespace at the beginning of the regular expression should be [[:blank:]], and I mistakenly wrote \\w .


hlissner:
No worries! Thanks again for your help!

PR #1784, opened 2019-09-14
Fix the multi-face form of custom-set-faces!

flatwhatson:
We need to unwrap the outer list returned by the inner calls `doom--custom-theme-set-face`.

Also need to flatten the face with the rest of the spec list so that:
```
((foo bar baz) :weight bold)
```

Is equivalent to:
```
(foo :weight bold)
(bar :weight bold)
(baz :weight bold)
```

Instead of:
```
(foo (:weight bold))
(bar (:weight bold))
(baz (:weight bold))
```

### DIFF
--- a/core/autoload/themes.el
+++ b/core/autoload/themes.el
@@ -4,7 +4,7 @@
   (cond ((listp (car spec))
          (cl-loop for face in (car spec)
                   collect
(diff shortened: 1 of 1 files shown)


hlissner:
Good catch, thanks!

PR #1785, opened 2019-09-15
Add support for org-journal

rgrinberg:
As discussed in chat, there was an interest for adding support to this package.


rgrinberg:
Should I bother with removing bindings introduced by `org-journal` itself? For example: `C-c C-j` to create a new journal entry.

hlissner:
> Should I bother with removing bindings introduced by org-journal itself?

No, that's alright. So long as they aren't too imposing I prefer to keep a package's default keybinds intact.

PR #1786, opened 2019-09-15
Fixes wrong variable name: lv-use-seperator to lv-use-separator

cnsunyour:
The variable `lv-use-seperator` is not defined anywhere, but I found the variable `lv-use-separator` in the `lv.el` file, which is only one letter apart, so I guess the variable name is wrong.


hlissner:

PR #1794, opened 2019-09-19
[emacs lisp] Add bindings for debugging defuns

rgrinberg:
`, d f` - turn on debugging for defun
`, d F` - turn off debugging for defun


rgrinberg:
Would be nice to add some highlighting to debugged functions but I'm not sure how to do that.

rgrinberg:
Good call. I fixed the naming convention.

rgrinberg:
@hlissner How about just merging the functions without adding the keybindings? Those are already quite useful on their own since one does not have convenient functions for them.

I agree the keybindings could use a bit more work especially since it makes sense to create binding scheme that will be consistent between `debug` and `edebug`.

rgrinberg:
@hlissner ping?

hlissner:
Sorry for the delay and thanks for the help!

PR #1804, opened 2019-09-20
Fix package-installed-p hack for rustic

flatwhatson:
Fix override/around typo.

Stubbing package-installed-p to return true is sufficient, rustic happily activates the configured backend (lsp), which is then autoloaded properly.

### DIFF
--- a/modules/lang/rust/config.el
+++ b/modules/lang/rust/config.el
@@ -55,17 +55,12 @@
   (add-hook 'rustic-mode-hook #'rainbow-delimiters-mode)
 
   (defadvice! +rust--dont-install-packages-p (orig-fn &rest args)
-    :override #'rustic-setup-rls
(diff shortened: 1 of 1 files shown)


hlissner:
This is half right. Changing it to `:around` and aliasing package-installed-p -> identity is fine, but please restore the `rustic-install-rls-client-p` alias. If lsp/eglot are both legitimately unavailable, it will try to install lsp-mode -- which is what that was there to prevent.

hlissner:
Actually, considering recent upstream changes, `package-installed-p` should also be redefined to:

```elisp
              ((symbol-function #'package-installed-p)
               (lambda (pkg)
                 (require pkg nil t)))
```

So that a path where lsp/eglot is legitimately unavailable can even occur. Otherwise you'll just get an ugly missing-package error when it tries to require them.

PR #1810, opened 2019-09-22
Use mdfind for counsel-locate on a Mac

rgrinberg:
This is the more sensible option as locate isn't configured by default
and spotlight is turned on by default.


rgrinberg:
Similarly, I would like to set `locate-command` (for `M-x locate`) but i'm not sure where's the right place to do so.

hlissner:
We could set `locate-command` in the `:tools macos` module. I've also added mdfind support to helm-locate in 19ef10339.

Anyhow, thanks for the help and for bringing this to my attention!

PR #1811, opened 2019-09-22
cli: fix wrong argument and add docs

seanfarley:
----

A small fix for `doom upgrade -t TIMEOUT` that was recently broken by the new
code for `doom refresh --thread THREADS`.


hlissner:
Nice work. Thanks again for the help!

PR #1813, opened 2019-09-23
Set locate-command in tools macos

rgrinberg:
It should be mdfind since spot light is on by default.

I think you could still improve things further by not making users that prefer `locate` on a mac to override two options, but this is better already.


hlissner:
Thanks for the help!

PR #1824, opened 2019-09-26
Make +flyspell-immediately nil by default

rgrinberg:
This option can considerably slow down the handling of some buffers and
hence should not be on by default.

### DIFF
--- a/modules/tools/flyspell/config.el
+++ b/modules/tools/flyspell/config.el
@@ -1,6 +1,6 @@
 ;;; tools/flyspell/config.el -*- lexical-binding: t; -*-
 
-(defvar-local +flyspell-immediately t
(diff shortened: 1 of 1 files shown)


UndeadKernel:
I tend to agree with the feeling that "immediate" checking is not something really needed to be on by default.

A6GibKm:
Even on a few lines email it slows everyrhing..

HaoZeke:
Yeah it kills latex buffers for as long as 30 seconds

HaoZeke:
Yeah it kills latex buffers for as long as 30 seconds

hlissner:
I ended up just removing it altogether. Thanks for bringing it to my attention.

PR #1825, opened 2019-09-26
lang/latex: +pdf-viewers ordering and hook type-o

UndeadKernel:
Previously, the last viewer set in `+latex-viewers` would take
precedence. The list is now reversed before applying changes so that the
first set has precedence.

The wrong hook variable was used to revert view buffers after compilation.


hlissner:

PR #1827, opened 2019-09-26
Rust documentation change

A6GibKm:
`rustup component` doesn't find the mentioned packages in the toolchain, whereas `cargo install` builds the binaries required and install them in your `PATH`.


hlissner:
I've revised rust's README with a little more information about these dependencies and have included your changes. Thanks for bringing it to my attention.

PR #1829, opened 2019-09-27
doom/core: avoid the litter of session.* files

UndeadKernel:
When x-window commands emacs to quit, emacs saves a session in `.emacs.d/session.*` by default. There is no easy way to change the location of this session file. This PR overwrites the function responsible for creating the session file name `emacs-session-filename`. The new function sets the session file to reside in `doom-cache-dir`/sessions.

### DIFF
--- a/core/core.el
+++ b/core/core.el
@@ -196,6 +196,13 @@ users).")
       url-cache-directory          (concat doom-cache-dir "url/")
       url-configuration-directory  (concat doom-etc-dir "url/")
       gamegrid-user-score-file-directory (concat doom-etc-dir "games/"))
+;; HACK
+(with-eval-after-load 'x-win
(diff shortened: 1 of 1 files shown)


core/core.el
```diff
@@ -196,6 +196,17 @@ users).")
       url-cache-directory          (concat doom-cache-dir "url/")
       url-configuration-directory  (concat doom-etc-dir "url/")
       gamegrid-user-score-file-directory (concat doom-etc-dir "games/"))
+(with-eval-after-load 'x-win
+  ;; overwrite the `emacs-session-filename' function with a custom one
```
filalex77: Consider adding a `;; HACK` or `;; FIXME` label (this is *really* not that important).

hlissner:
Nice work. Thanks again for your help!

PR #1639, opened 2019-08-06
lang/plantuml: change mirror for +plantuml/install

filalex77:
This is the one they currently redirect to.

Fixes #1637.

Issue #1637, opened 2019-08-05
[BUG] M-x +plantuml/install Certificate validation failed


flyingns:
The develop branch was previously installed, no problem.

When switching to the straight branch, M-x +plantuml/install, it is reported:

`Contacting host: kent.dl.sourceforge.net:443`

`gnutls-negotiate: Certificate validation failed kent.dl.sourceforge.net, verification code 402`

hlissner:
#1639 was merged, which should resolve this issue. Let me know if that isn't the case and I will reopen this issue. Thanks for bringing it to my attention!

PR #1830, opened 2019-09-27
lang/plantuml: fix jar download path (again)

filalex77:
Now this should redirect to the auto-selected mirror.

Related: #1637 #1639


hlissner:
Thanks again!

PR #1832, opened 2019-09-27
Fix install command for clippy, uses rustup

webframp:
Trying to install with cargo fails with the following error:

```
Caused by:
  failed to run custom build command for `clippy v0.0.302`

Caused by:
  process didn't exit successfully: `/var/folders/66/180ynjt55z74l338gh0stq0c0000gn/T/cargo-installkNqJBB/release/build/clippy-bcde765a4c9acf93/build-script-build` (exit code: 1)
--- stderr

error: Clippy is no longer available via crates.io

help: please run `rustup component add clippy-preview` instead
```

Just a tiny doc update


hlissner:

PR #1836, opened 2019-09-30
tools/eval: change evil-append-line to evil-insert-state

MaskRay:
With some programs under term-mode (e.g. zsh, julia), evil-normal-state
at point-max moves to the beginning of the next blank line. When
switching to the REPL buffer, evil-append-line will append a new line,
which will cause a new prompt to be created. Don't move point and don't
call evil-append-line to fix the problem. Also, change to
evil-insert-state which is likely the most desired behavior.

------------

I have installed julia-repl and configured

```elisp
  (set-repl-handler! 'julia-mode #'julia-repl)
```

```
julia>
|              <-- evil-normal-state will move point here
```

After switching back to the REPL buffer with `SPC o r`, I will see a julia prompt with no command

```
julia>     # no command due to evil-append-line

julia> 
```

This PR fixes the problem. From my experience, `(call-interactively #'evil-append-line)` is not very useful, so I just deleted it and added `(evil-insert-state)` instead.

zsh (without rc file) under term-mode has a similar behavior, but bash (without rc file) does not. I haven't figured out the reason.

----------

I have considered another approach:

```elisp
(when (bolp) (forward-char -1))
```

It works if point is at the end of line. However, the approach will cause the evaluation of the line if the point is not at the end. This is weird.
```
julia> 1 |+ 2    # point is not at the end
```

After switching back to the REPL buffer:
```
julia> 1 + 2 + 2   # + 2 was appended. weird

julia> 
```

### DIFF
--- a/modules/tools/eval/autoload/repl.el
+++ b/modules/tools/eval/autoload/repl.el
@@ -32,12 +32,7 @@
                    (+eval-repl-mode +1))
                  (puthash key buffer +eval-repl-buffers)
                  buffer)))
-    (with-current-buffer buffer
-      (goto-char (if (and (derived-mode-p 'comint-mode)
-                          (cdr comint-last-prompt))
-                     (cdr comint-last-prompt)
-                   (point-max)))
-      buffer)))
+    buffer))
 
 (defun +eval-open-repl (prompt-p &optional other-window-p)
   (let ((command (cdr (assq major-mode +eval-repls))))
@@ -65,7 +60,7 @@
       (error "Couldn't find a valid REPL for %s" major-mode))
     (when (+eval--ensure-in-repl-buffer command other-window-p)
       (when (bound-and-true-p evil-mode)
-        (call-interactively #'evil-append-line))
+        (evil-insert-state))
       t)))
 
 ;;;###autoload


hlissner:
I can reproduce this in `julia-repl`, but not in any other repl. I've found an unrelated issue that prevents `+julia/repl` from working, but after fixing that, I do not see this behavior there either.

The root of the problem is that `julia-repl` uses `term-mode` in `term-char-mode`, which forwards every character inserted into the term buffer to the underlying process. Therefore, a better option for `+eval--ensure-in-repl-buffer` would be:

```diff
    (with-current-buffer buffer
-     (goto-char (if (and (derived-mode-p 'comint-mode)
-                         (cdr comint-last-prompt))
-                    (cdr comint-last-prompt)
-                  (point-max)))
+     (unless (or (derived-mode-p 'term-mode)
+                 (eq (current-local-map) term-raw-map)) ; detects term-char-mode
+       (goto-char (if (and (derived-mode-p 'comint-mode)
+                           (cdr comint-last-prompt))
+                      (cdr comint-last-prompt)
+                    (point-max))))
      buffer)))
```

> From my experience, (call-interactively #'evil-append-line) is not very useful, so I just deleted it and added (evil-insert-state) instead.

This appears to be a unrelated issue, so please revert it. The use of `evil-append-line` was intentional. I believe it makes the most sense to place the cursor at the end of the input, rather than the beginning. `evil-insert-state` can also leave the cursor on EOL-1 and EOL depending on the values of `evil-move-cursor-back`. With `evil-append-line` there is no such ambiguity.

If you'd like to make that behavior optional, perhaps a "after open repl" hook would be a better choice. Then we could attach `evil-append-line` or `evil-insert-state` there as desired

-----

Side question: is there anything that sets `julia-repl` apart from `inferior-julia-mode` included with julia-mode? I fixed `+julia/repl` like so:

```elisp
(defun +julia/repl ()
  "Run an inferior instance of `julia' inside Emacs."
  (interactive)
  (let ((buffer (get-buffer-create "*Julia*")))
    (unless (comint-check-proc "*Julia*")
      (apply #'make-comint-in-buffer "Julia" "*Julia*"
             julia-program nil julia-arguments))
    (pop-to-buffer buffer)
    (with-current-buffer buffer
      (inferior-julia-mode))
    buffer))
```

And there doesn't seem to be any difference.

hlissner:
With 4761a76 and 1aae5c9 I've resolved the issue. Let me know if that isn't the case and I'll reopen this. Thanks for bringing it to my attention.

PR #1840, opened 2019-09-30
Add doom-solarized-dark to +doom-solaire-themes

darth10:
This is need to call `solaire-mode-swap-bg` if the theme is `doom-solarized-dark`.


hlissner:
Thanks for the help!

PR #1842, opened 2019-10-01
[WIP] Experimental fixes for undo-tree

flatwhatson:
Increase `undo-limit`, `undo-strong-limit` and `undo-outer-limit` by a factor of ten, which should prevent emacs from prematurely truncating the undo history and corrupting the tree.

Restore `undo-tree-auto-save-history` which was disabled to avoid exacerbating the undo history truncation.

Change the `undo-tree` recipe to grab the latest version from the maintainer's repostiory instead of the outdated ELPA version.

Restore `undo-tree-enable-undo-in-region` which should be fixed in the latest version of `undo-tree`.

**WARNING:** There may still be a problem with `undo-tree-auto-save-history` causing a stack overflow on startup while deserializing the saved history.  This needs testing!  If you decide to test it, be prepared for undo history corruption!

### DIFF
--- a/core/core-editor.el
+++ b/core/core-editor.el
@@ -409,11 +409,13 @@ files, so we replace calls to `pp' with the much faster `prin1'."
   ;; Branching & persistent undo
   :after-call doom-switch-buffer-hook after-find-file
   :config
-  (setq undo-tree-auto-save-history nil ; disable because unstable
-        ;; undo-in-region is known to cause undo history corruption, which can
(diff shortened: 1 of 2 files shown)


hlissner:
I've so far been pretty happy with these changes. I haven't experienced the elusive undo-tree-canary error or stack overflow crashes so far, so I'll merge this now and see how it fares in the wild. Thanks for looking to this!

PR #1843, opened 2019-10-01
[WIP] Add org-sidebar (:lang org +sidebar)

filalex77:
This change is [Reviewable](https://reviewable.io/reviews/hlissner/doom-emacs/1843)

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -129,7 +129,8 @@
         +dragndrop       ; drag & drop files/images into org buffers
         +ipython         ; ipython/jupyter support for babel
         +pandoc          ; export-with-pandoc support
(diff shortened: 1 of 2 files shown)


hlissner:
I understand this is a WIP, but is there anything else that can be configured for org-sidebar? If it's only to install a package, I don't think this warrants its own flag.

filalex77:
@hlissner I think it needs some configuration to play nice with evil, but I'm not sure on both the configuration *and* the flag. Should it not be optional?

hlissner:
It certainly should be optional. The flag is fine. My issue is that it's so small to warrant being included in Doom. Evil config is nice, but evil-collection might be a better fit for that.

PR #1845, opened 2019-10-01
Api demos

A6GibKm:
Added demos for:

- use-package!
- disable-packages!
- map!
- quiet!
- setq-hook!
- unsetq-hook!

I would like some feedback on this.

### DIFF
--- a/docs/api.org
+++ b/docs/api.org
@@ -27,6 +27,7 @@ It is integrated into Helpful, in Doom.
     - [[#remove-hook][remove-hook!]]
     - [[#setq-hook][setq-hook!]]
     - [[#unsetq-hook][unsetq-hook!]]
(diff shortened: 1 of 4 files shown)


docs/api.org
```diff
@@ -276,9 +310,47 @@ It is integrated into Helpful, in Doom.
...
+#+END_SRC
+
+*** use-package!
+#+BEGIN_SRC elisp :eval no
+;; Use after-call to load package before hook
+(use-package! projectile
+  :after-call (pre-command-hook after-find-file dired-before-readin-hook))
+
+;; Defer recentf packages one by one
+(use-package! (use-package! recentf
```
hlissner: Typo: `use-package!` is repeated here.

hlissner:
Thanks for the help and quick response!

PR #1850, opened 2019-10-03
Update documentation of recognized latex viewers

alexandergraul:
The user documentation of +latex-viewers did not keep up with changes to
the variable. This commit realigns its documentation.


hlissner:
Thanks again!

PR #1851, opened 2019-10-03
Add openSUSE prerequisite instructions

alexandergraul:
These instructions were tested on openSUSE Tumbleweed and openSUSE Leap
15.1. There are some modules left that are not documented yet, but this
already improves the sitution for common openSUSE users.


hlissner:
I may revisit this later and combine the "simple" build instructions into a plain list (I'd like to avoid a future where we have a dozen one-line sections for this sort of thing), but in any case, thanks for the help!

PR #1853, opened 2019-10-04
Correct a requirement in README.org

oguzserbetci:
Quick fix to the README for python language module to correct the required python package for `pyimport.el` to `pyflakes` as mentioned in its own [README](https://github.com/Wilfred/pyimport).


hlissner:
Good catch! Thanks for the help.

PR #1865, opened 2019-10-09
Add keybindings for org-html-export

Porcupine96:
This commit adds keybindings for exporting org to HTML, analogous to those that are defined for markdown and reveal.js.


hlissner:

PR #1867, opened 2019-10-09
added company-files support for file links in org mode

JakDar:
Added regex for company-files to recognize file links in org-mode.
I can move changing `company-files-regexps` if there is better place to modify it.

### DIFF
--- a/modules/completion/company/config.el
+++ b/modules/completion/company/config.el
@@ -51,6 +51,11 @@
 ;;
 ;; Packages
 
+(after! company-files
+  (pushnew! company-files--regexps
+            "file:\\(\\(?:\\.\\{1,2\\}/\\|~/\\|/\\)[^\]\n]*\\)"))
+
+
 (use-package! company-prescient
   :hook (company-mode . company-prescient-mode)
   :config
--- a/modules/completion/company/config.el
+++ b/modules/completion/company/config.el
@@ -31,7 +31,9 @@
       :before #'company-begin-backend
       (company-abort)))
 
-  (add-hook 'company-mode-hook #'+company-init-backends-h)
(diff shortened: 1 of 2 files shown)


JakDar:
Moved adding regexp to org/config.el and removed escaping `:` ;)
I couldn't use `setq-hook!`, because `company-files` had to be required so that `company-files-regexps` is in scope.

hlissner:
> I couldn't use setq-hook!, because company-files had to be required so that company-files-regexps is in scope.

My bad, forgot about the variable reference in the new value. 

That said, I'd really like to avoid: 1) a quoted, closure for a hook, 2) eagerly loading company when org-mode loads, and 3) setting `company-files--regexp` globally, rather than buffer-locally.

Unfortunately, company's unusual lazy-loading makes achieving all of the above a bit difficult. I'll look into it.

JakDar:
If copying initialization of `company-files-regexps` is an option i might have a beginning of solution (removed eager loading and closure in a hook), but I still don't know how to make it buffer-local.

hlissner:
> but I still don't know how to make it buffer-local.

With `setq-local`, rather than `setq`.

> If copying initialization of company-files-regexps is an option

Perhaps we can avoid that with:

```elisp
(defun +org-init-company-h ()
  (after! company-files
    (setq-local company-files--regexps
                (cons "file\:\\(\\(?:\\.\\{1,2\\}/\\|~/\\|/\\)[^\]\n]*\\)"
                      company-files--regexps))))
```

JakDar:
Just made a change.

hlissner:
Great, thanks for the quick responses (and for putting up with me)!

Issue #1166, opened 2019-02-11
Meta issue for module documentation


Amatrelan:
I created this meta issue to track missing documentation. I added [x] even
thought documentation might not be complete. This point is just to make easier
to people to find what module needs documentation, so they can easier contribute
to documentation.

All these are from develop branch. 

* app
- [X] calendar
- [X] email
- [X] irc
- [ ] notmuch
- [ ] regex
- [ ] rss
- [ ] twitter
- [X] write

* collab
- [ ] floobits
- [ ] impatient-mode

* completion
- [X] Company
- [ ] helm
- [ ] ido
- [X] ivy

* editor
- [ ] format
- [ ] lispy
- [ ] multiple-cursors
- [ ] parifier
- [ ] potate-text

* emacs
- [ ] dired
- [ ] ediff
- [ ] electric
- [ ] hideshow
- [ ] imenu
- [ ] term
- [ ] vc

* feature
- [ ] debugger
- [X] eval
- [X] evil
- [ ] file-templates
- [ ] lookup
- [ ] snippets
- [ ] spellcheck
- [ ] syntax-checker
- [X] workspaces

* lang
- [X] agda
- [ ] assembly
- [X] cc
- [ ] clojure
- [ ] common-lisp
- [X] coq
- [ ] crystal
- [ ] csharp
- [ ] data
- [ ] elixir
- [ ] elm
- [ ] emacs-lisp
- [ ] erlang
- [X] ess
- [X] go
- [X] haskell
- [ ] hy
- [X] idris
- [ ] java
- [X] javascript
- [ ] julia
- [ ] latex
- [ ] ledger
- [ ] lua
- [ ] markdown
- [X] nim
- [ ] ocaml
- [ ] org | Needs more documentation.
- [X] perl
- [X] php
- [ ] plantuml
- [ ] purescript
- [ ] python
- [ ] qt
- [ ] racket
- [X] rest
- [ ] ruby
- [X] rust
- [ ] scala
- [X] sh
- [X] solidify
- [ ] swift
- [ ] vala
- [ ] web

* tools
- [ ] ansible
- [ ] docker
- [X] editorconfig
- [X] ein
- [ ] gist
- [ ] macos
- [ ] magit
- [ ] make
- [ ] password-store
- [ ] pdf
- [ ] prodigy
- [ ] rgb
- [ ] terraform
- [ ] tmux
- [ ] upload
- [X] wakatime

* ui
- [ ] deft
[...]

PR #1868, opened 2019-10-09
Reformats ui/doom/README.org to match __doom-readme template

zakkak:
Reformats ui/doom/README.org to match __doom-readme template as per #1166


hlissner:
And thanks a third time! These are a huge help.

PR #1869, opened 2019-10-09
Reformat ui/doom-dashboard/README.org to match __doom-readme template

zakkak:
----

Reformats ui/doom-dashboard/README.org to match __doom-readme template as per #1166


hlissner:
Thanks again!

PR #1870, opened 2019-10-09
Reformat app/write/README.org to match __doom-readme template

zakkak:
Reformat app/write/README.org to match __doom-readme template as per #1166

### DIFF
--- a/modules/app/write/README.org
+++ b/modules/app/write/README.org
@@ -1,29 +1,77 @@
-#+TITLE: :app write
+#+TITLE:   app/write
+#+DATE:    October 10, 2019
+#+SINCE:   v2.0.9
+#+STARTUP: inlineimages
 
+* Table of Contents :TOC_3:noexport:
+- [[#description][Description]]
+  - [[#module-flags][Module Flags]]
+  - [[#plugins][Plugins]]
+- [[#prerequisites][Prerequisites]]
+  - [[#language-tool][Language Tool]]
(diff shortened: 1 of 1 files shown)


modules/app/write/README.org
```diff
@@ -1,29 +1,77 @@
...
 
 - ~+langtool~ Enables language tool integration.
 - ~+wordnut~ Enables wordnet integration.
 
+** Plugins
++ synosaurus
++ mixed-pitch
++ langtool
++ wordnut
++ visual-fill-column
```
hlissner: Could you link these to their respective plugins' repos? You can find them on https://melpa.org/ or `SPC h d P PACKAGENAME` (for packages that aren't installed you'll be directed to their melpa page, which gets you halfway there).

modules/app/write/README.org
```diff
@@ -1,29 +1,77 @@
...
 This module provides two module flags:
 
 - ~+langtool~ Enables language tool integration.
 - ~+wordnut~ Enables wordnet integration.
 
+** Plugins
++ synosaurus
++ mixed-pitch
++ langtool
++ wordnut
```
hlissner: The convention for conditional packages in this list is to suffix them with an asterix and, in parentheses, the flag they depend on. e.g.

```
+ langtool* (=+langtool=)
+ wordnut* (=+wordnut=)
```

hlissner:

PR #1872, opened 2019-10-10
Set a jump point when using imenu

flatwhatson:
:ghost: :lollipop:


flatwhatson:
This sets a jump point *before* jumping, which is what was missing.  It seems we already set a jump point after, somehow.

hlissner:
Good point! Thanks again for the help!

PR #1873, opened 2019-10-10
[debugger] Check if evil-emacs-state-local-map is bound

darth10:
Looks like `realgud` errors out if the `:editor evil` feature is not enabled.
This fix simply checks if `evil-emacs-state-local-map` is bound before calling `define-key`.


hlissner:

PR #1877, opened 2019-10-10
Fixes comments for M and I in +emacs-bindings.el

zakkak:
Just fixes comments for IRC and mu4e bindings


hlissner:
Good catch, thanks!

PR #1878, opened 2019-10-11
Adds README for app:twitter

zakkak:
This PR adds as much info as I could infer from the code. There are some TODOs
left in the readme for hacks, configuration etc. since I felt they were out of
my reach.

### DIFF
--- a/modules/app/twitter/README.org
+++ b/modules/app/twitter/README.org
@@ -46,6 +46,11 @@ This module provides no flags.
 + ~twittering-icon-mode~ converts retrieved icons into XPM by default. To
   achieve this and for displaying icons in formats that are not supported by
   Emacs as well as for resizing icon images, [[http://www.imagemagick.org/][ImageMagick]] is required.
+
+  To build emacs with ImageMagick support the ~--with-imagemagick~ flag needs to
+  be passed to the ~configure~ script, e.g. ~./configure --with-imagemagick~.
+  For detailed instruction on how to build Emacs from source please refer to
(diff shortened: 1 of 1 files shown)


A6GibKm:
I suggest that you add install methods for the prerequisites in said section. Something reasonable would be `apt` command + the command you would use in your distro if this uses other method.

modules/app/twitter/README.org
```diff
@@ -0,0 +1,91 @@
...
++ Retweet
++ Follow and un-follow users
++ Favorite tweets
+
+** Module Flags
+This module provides no flags.
+
+** Plugins
+[[https://github.com/hayamiz/twittering-mode][twittering-mode]]
+[[https://github.com/abo-abo/avy][avy]]
```
hlissner: These should be list items, e.g.

```
+ [[https://github.com/hayamiz/twittering-mode][twittering-mode]]
+ [[https://github.com/abo-abo/avy][avy]]
```

zakkak:
I am thinking that maybe we need a file documenting how to install external dependencies and then linking to that. This way we don't have to replicate instructions on getting gzip etc. on each module that uses them. Same could be done for building Emacs with ImageMagick support.

A6GibKm:
I think it is not bad to assume for the time being that the user is in a distro that comes with gnupg. Most of the stuff left amounts just to a single `apt,dnf,pacman,...`command. While there might be duplicates I think that it would be even more convoluted to try to deduplicate that.

zakkak:
relates to #1166

zakkak:
Regarding gnupg @hlissner this module just needs it for symmetrical encryption (no need for any fancy setup).

PR #1880, opened 2019-10-11
Add support for org-pomodoro

flatwhatson:
This is a resurrection & review of #1484.

Adds `SPC t t` binding to toggle `org-pomodoro` (mnemonic "toggle timer").

Sends notifications using growl or libnotify if available.

### DIFF
--- a/modules/lang/org/contrib/pomodoro.el
+++ b/modules/lang/org/contrib/pomodoro.el
@@ -1,0 +1,18 @@
+;;; lang/org/contrib/pomodoro.el -*- lexical-binding: t; -*-
+;;;###if (featurep! +pomodoro)
+
+(use-package! org-pomodoro
+  :defer t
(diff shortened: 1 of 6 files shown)


modules/lang/org/contrib/pomodoro.el
```diff
@@ -0,0 +1,25 @@
+;;; lang/org/contrib/pomodoro.el -*- lexical-binding: t; -*-
+;;;###if (featurep! +pomodoro)
+
+(def-package! org-pomodoro
```
darth10: I'd recommend using `use-package!` here instead of `def-package!`

flatwhatson: Good catch, thanks.  Gotta keep with the times!

hlissner:

PR #1881, opened 2019-10-11
Optimize update package time

xosdy:
`shell-command-to-string` will run eg `fish -c git` instead of `git`. In my case, it takes a few seconds and lots of CPU to start the shell.

### DIFF
--- a/core/cli/packages.el
+++ b/core/cli/packages.el
@@ -173,11 +173,11 @@ a list of packages that will be installed."
                         ;; cause this async process to block indefinitely). We
                         ;; can't use `straight-normalize-package' because could
                         ;; create popup prompts too, so we do it manually:
-                        (shell-command-to-string "git merge --abort")
+                        (doom-sh "git" "merge --abort")
                         (straight--get-call "git" "reset" "--hard" branch)
                         (straight--get-call "git" "clean" "-ffd")
                         (unless nonrecursive
-                          (shell-command-to-string "git submodule update --init --recursive"))
+                          (doom-sh "git" "submodule update --init --recursive"))
                         (when upstream-repo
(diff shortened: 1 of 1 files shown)


rgrinberg:
Given that you are likely to encounter this problem with many other plugins, I recommend that you just set `shell-file-name`. I've done the following for example:

```
(setq shell-file-name "/usr/local/bin/dash")
```

dash is quite fast

core/cli/packages.el
```diff
@@ -173,11 +173,11 @@ a list of packages that will be installed."
                         ;; cause this async process to block indefinitely). We
                         ;; can't use `straight-normalize-package' because could
                         ;; create popup prompts too, so we do it manually:
-                        (shell-command-to-string "git merge --abort")
+                        (doom-sh "git" "merge --abort")
```
rgrinberg: Shouldn't you be providing the arguments as separate list elements? E.g. `(doom-sh "git" "merge" "--abort")`.

xosdy: Done

xosdy:
I just tried to set `shell-file-name`, but it seems to have no effect in the CLI tool.

rgrinberg:
My recommendation wasn't related to the CLI tool. I just meant that many Emacs plugins use `shell-command-to-string` indiscriminately so it's useful to have it set.

hlissner:
I approve of these changes, however, I'm a day or two away from publishing a rewrite Doom's CLI, which includes the replacement of `doom-sh` and a redesign of `doom upgrade` and `doom update`. I'll have to turn down this PR, but thank you for bringing it to my attention! It has inspired a few optimizations in the rewrite.

PR #1882, opened 2019-10-12
Faster prompt eshell

rgrinberg:
There's no need for a shell or for any complex output parsing. So this is simpler and more efficient.

Copy pasting `doom-sh-ignore-stderr` like that is ugly of course. I'm open to suggestions on how to refactor things.

The whole `sh` thing is confusing btw. Avoiding the shell seems like the whole point of using `doom-sh`.

### DIFF
--- a/core/core-cli.el
+++ b/core/core-cli.el
@@ -33,6 +33,19 @@ STATUS is a boolean"
                 (string-trim (buffer-string))))
       (kill-buffer output))))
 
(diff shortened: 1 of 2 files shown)


hlissner:
I approve of this change, but it conflicts with an upcoming rewrite of Doom's CLI, where `doom-sh` has been replaced with a generalized version. Inspired by this PR I will employ this optimization, but I must turn down the PR. In any case, thanks for bringing it to my attention!

PR #1883, opened 2019-10-12
Don't bind nav-flash keybind unless enabled

jseba:
The nav-flash keybind is enabled regardless of whether or not the module is enabled. This just adds a check to the map! in the default keybindings.


hlissner:
Sorry for the late response. I've removed the keybind in 9fa447a (not that `ga` has replaced its functionality), rendering this PR unnecessary. Thanks for the PR in any case!

PR #1891, opened 2019-10-13
Add GitHub actions

Emiller88:
Add CI for GitHub actions. 

I commented out compile since I believe that's being rewritten.

We might also consider removing `.travis.yml`

I also have not figured out how to allow `snapshot` to fail.

I'd also like to add macos testing, but I'm not sure if the action supports that.


hlissner:
Nice work! Thanks for looking into this!

PR #1892, opened 2019-10-13
Add syntax highlighting to ielm buffers

rgrinberg:
Taken from http://www.modernemacs.com/post/comint-highlighting/

Currently this uses `dash`, but that can be removed if the PR is approved.

### DIFF
--- a/modules/lang/emacs-lisp/config.el
+++ b/modules/lang/emacs-lisp/config.el
@@ -90,6 +90,40 @@ This marks a foldable marker for `outline-minor-mode' in elisp buffers.")
           "f" #'find-function
           "v" #'find-variable)))
 
+(use-package! ielm
+  :config
+  ;; Adapted from http://www.modernemacs.com/post/comint-highlighting/
+  (require 'dash)
+
+  (defun doom--kwd->comint-kwd (kwd)
+    "Converts a `font-lock-keywords' KWD for `comint-mode' input fontification."
+    (-let (((matcher . match-highlights) kwd))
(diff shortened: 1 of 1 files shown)


hlissner:
I'm not sure how useful ielm is with a scratch buffer, but I don't mind merging this PR.

That said, the original approach didn't work for me (the hook ran but the font-lock keywords were buffer-locally replaced by what looks like comint's keywords). So I tweaked it a bit:

```elisp
(after! ielm
  ;; Adapted from http://www.modernemacs.com/post/comint-highlighting/
  (add-hook! 'ielm-mode-hook
    (defun +emacs-lisp-init-syntax-highlighting-h ()
      (font-lock-add-keywords
       nil (cl-loop for (matcher . match-highlights)
                    in (append lisp-el-font-lock-keywords-2 lisp-cl-font-lock-keywords-2)
                    collect
                    `((lambda (limit)
                        (and ,(if (symbolp matcher)
                                  `(,matcher limit)
                                `(re-search-forward ,matcher limit t))
                             ;; Only highlight matches after the prompt
                             (> (match-beginning 0) (car comint-last-prompt))
                             ;; Make sure we're not in a comment or string
                             (let ((state (sp--syntax-ppss)))
                               (not (or (nth 3 state)
                                        (nth 4 state))))))
                      ,@match-highlights))))))
```

The hook is there to allow folks to disable it if they'd like.

rgrinberg:
Sure, feel free to push your tweak. I expect it's not useful for everyone as I know some people don't use the repl for elisp.

PR #1894, opened 2019-10-13
Fix tuareg-find-alternate-file binding

rgrinberg:
This binding should be available whenever tuareg is loaded. Previously,
it would loaded only after merlin.


rgrinberg:
@hlissner I pushed a commit to make `<localleader> t` work in lsp mode.

rgrinberg:
I got rid of the above commit since it was already bound to `K`.

hlissner:

PR #1901, opened 2019-10-14
Add emr to emacs lisp mode

rgrinberg:
While emr is sometimes buggy, it's basically the only choice for elisp.

I'm not a fan of this menu thing for common refactorings, so I've added some proper keybindings.


rgrinberg:
@hlissner ping. I wasn't aware this was controversial :)

hlissner:
Sorry for the insanely long delay. I've decided not to accept this PR and instead to later introduce a generalized `:editor refactor` module (powered by emr). In any case, thanks for bringing this to my attention!

PR #1902, opened 2019-10-14
Improve +word-wrap in non-code buffers

flatwhatson:
Adds `+word-wrap-text-modes`, a list of modes which shouldn't get any extra indentation. This is used for text and markdown modes, which should just indent to the parent depth.

Adds `+word-wrap-visual-modes`, a list of modes which shouldn't enable `adaptive-wrap-prefix-mode`. This is used to fix the prefix indentation in `org-mode`, which provides its own implementation.

Tweaks the indent behaviour to treat strings the same as comments so they don't receive additional indentation.

### DIFF
--- a/modules/editor/word-wrap/README.org
+++ b/modules/editor/word-wrap/README.org
@@ -12,10 +12,10 @@
 This module adds a minor-mode ~+word-wrap-mode~, which intelligently wraps long
 lines in the buffer without modifying the buffer content.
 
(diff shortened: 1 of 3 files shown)


flatwhatson:
Added some more text modes.

flatwhatson:
Refactored to avoid loading some modules unnecessarily.

hlissner:
Great work. Thanks again for all your help!

PR #1904, opened 2019-10-15
Add: evil bindings for 2 evil-mc functions

Nollo:
Two useful functions added for evil-mc

`evil-mc-make-cursor-in-visual-selection-beg`
Create cursors at the beginning of every visually selected line.

`evil-mc-make-cursor-in-visual-selection-end`
Create cursors at the end of every visually selected line.

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -397,7 +397,9 @@
           :nv "q" #'evil-mc-undo-all-cursors
           :nv "t" #'+multiple-cursors/evil-mc-toggle-cursors
           :nv "u" #'evil-mc-undo-last-added-cursor
-          :nv "z" #'+multiple-cursors/evil-mc-make-cursor-here)
+          :nv "z" #'+multiple-cursors/evil-mc-make-cursor-here
+          :v  "I" #'evil-mc-make-cursor-in-visual-selection-beg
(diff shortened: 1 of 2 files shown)


modules/config/default/+evil-bindings.el
```diff
@@ -397,7 +397,9 @@
           :nv "q" #'evil-mc-undo-all-cursors
           :nv "t" #'+multiple-cursors/evil-mc-toggle-cursors
           :nv "u" #'evil-mc-undo-last-added-cursor
-          :nv "z" #'+multiple-cursors/evil-mc-make-cursor-here)
+          :nv "z" #'+multiple-cursors/evil-mc-make-cursor-here
+          :v  "I" #'evil-mc-make-cursor-in-visual-selection-beg
+          :v  "A" #'evil-mc-make-cursor-in-visual-selection-end)
```
flatwhatson: I think these should be behind a feature flag, in case someone wants multiple-cursors support without replacing the built-in visual insert and append.  Not sure about flag name, maybe `(multiple-cursors +visual)`?

Nollo: Wouldn't it be too much to add a flag for only 2 functions?

Another way would be using "B" and "E" bindings (Begin, End) for the 2 functions.

hlissner:

PR #1905, opened 2019-10-15
lang/clojure: add +lsp support

UndeadKernel:
Configure LSP to work with clojure if the flag '+lsp' is provided.


UndeadKernel:
Cider and LSP do not play well together.
I'll close this for now and think how LSP could be used.
Maybe it's not needed because CIDER already does everything LSP.

PR #1906, opened 2019-10-15
[WIP] Add visual-fill-column support for +word-wrap

flatwhatson:
Adds support for soft-wrapping at `fill-column` using `visual-fill-column-mode`.

This isn't quite ready for prime-time.  Currently there is some problem with the margin calculation in `visual-fill-column-mode` causing it to wrap lines too short, leading to ugly results in files which were already hard-wrapped.

### DIFF
--- a/modules/editor/word-wrap/README.org
+++ b/modules/editor/word-wrap/README.org
@@ -24,6 +24,13 @@ The ~+word-wrap-extra-indent~ variable supports the following values:
 - a negative integer: dedent by this fixed amount
 - ~nil~: no extra indent
 
+Long lines are wrapped at the window margin by default, however soft-wrapping at
+~fill-column~ is supported by setting ~+word-wrap-fill-style~. When set to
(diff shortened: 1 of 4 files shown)


UndeadKernel:
I experimented using `visual-fill-column-mode` before but dropped it because I found several issues with projectile and the workspaces. Could you please test to see if the following problem is still here?

### Problem:
When `visual-fill-column-mode` was on and there were multiple workspaces, restoring workspaces would fail. For example, running `doom/restart-and-restore` would result in workspaces not being auto-loaded correctly. Manually loading saved sessions (a setup of many workspaces) would also result in undesirable results.

flatwhatson:
Thanks for the heads-up, I'll test that, should be fixable anyway :+1:

bdarcus:
Any progress on this @flatwhatson? This would be a nice addition.

hlissner:
Sorry for the late response, I've updated and merged this in 33db882. Thanks for the PR!

PR #1907, opened 2019-10-15
lang/clojure: start clojurescript repl correctly

UndeadKernel:
Before this PR, opening a REPL when in clojurescript-mode would result in an
infinite loop. With this PR, the REPL is started correctly.


hlissner:
Good catch. Thanks again for the help!

Issue #1910, opened 2019-10-16
[BUG] style files are not loaded in tex mode (causes and solution found)


jonathye:
**Describe the issue**
auctex will automatically load various style files (e.g., listings.el) for various latex packages, which will do a bunch of configurations for the macros and environments provided by these packages. However the style files are not loaded in doom-emacs.

**Steps to reproduce**
1. open any latex file.
2. add `\usepackage{listings}` in the preamble, and save the file.
3. edit `lstlisting` environment or `lstinline` macro. Observe that the fontification, filling instructions, and other configurations do not take effect.

**Causes and solution**
After a few hours of debugging, I finally figured out the causes of this problem. There are 3 of them.
1. auctex's subdir `style` is somehow missing in `straight/build/auctex` (and hence not byte-compiled). This is an upstream bug in straight that I just reported. See raxod502/straight.el#423.
For the time being, I set `TeX-data-directory` to `straight/repos/auctex` to get around this problem.

2. In `modules/lang/latex/config.el`, a few variables are set *after* loading tex package. See https://github.com/hlissner/doom-emacs/blob/957e1dd0809e353fa73e64a32e40cd3995b49376/modules/lang/latex/config.el#L31-L42.

However, these variables are defined by `defcustom`. Simply `setq` after loading tex is problematic. For example, if you inspect the value of `TeX-style-path`, you will find out it uses the old value of `TeX-auto-local`.

Instead, we should set them before loading tex. (alternatively we may set them by calling `customize-set-variable` after loading tex, but I have not tried this approach). I sent a pull request for this.

3. If we fix the above 2 issues, at least we can load the style files manually by calling `TeX-normal-mode` (C-c C-n). But it won't be done automatically when we open a tex file, although we set `TeX-parse-self` to `t` by default. That is because doom somehow removed a hook added by auctex: https://github.com/hlissner/doom-emacs/blob/957e1dd0809e353fa73e64a32e40cd3995b49376/modules/lang/latex/config.el#L61-L67

I am not sure why doom removed it. The comment says it is for not prompting for master files. But on the one hand, it breaks style loading function. On the other hand, doom also sets `TeX-master` to `nil`, which enables prompting and they seem pretty contradictory to me: https://github.com/hlissner/doom-emacs/blob/957e1dd0809e353fa73e64a32e40cd3995b49376/modules/lang/latex/config.el#L47-L48

I suggest 3 ways to fix this issue.
[...]

PR #1913, opened 2019-10-17
lang/latex: fix loading style files #1910.

jonathye:
See #1910


hlissner:
Looks good to me. Thanks for looking into this and for the help! I'm happy to get rid of that hacky hook.

PR #1919, opened 2019-10-19
Update PragmataPro Ligatures

aidangilmore:
Some new versions of PragmataPro have come out with an expanded ligature set. This updates the `pretty-code` module to take advantage of these new ligatures. It also fixes the ligature for the bind operator in Haskell.

As always thanks Henrik and I hope I'm not bothering you too much by upsteaming support for my edge case.


hlissner:
> I hope I'm not bothering you too much by upsteaming support for my edge case.

Not at all! I appreciate you looking into and fixing these issues. I use neither the font nor ligatures, so it would have slipped by me otherwise. Thanks for the help!

PR #1920, opened 2019-10-20
[coq] disable electric proof terminators

rgrinberg:
These send input to coq too agressively. It's often the case that this
causes delays in coq.


hlissner:
Thanks again!

PR #1921, opened 2019-10-21
Add some bindings for apropos and info-mode

flatwhatson:
Just a couple of useful bindings which were missing.


hlissner:
Thanks again for your help!

PR #1924, opened 2019-10-21
Restore save file binding on SPC f s

### LINKED ISSUE #1923: [BUG]  write-file is always asking for file name
**Describe the issue**
Before that change https://github.com/hlissner/doom-emacs/commit/5215b79f7a3521656d4f352ed61269a57ac5f815

image

`SPS f s` keybinding was a  convenient way to save changes. But now it 's always sking for file name, even if file exists.

Command label is still "Save file", while it behaves as "Save as".
I would propose to revert original behaviour and ass `SPS f S` keybinding that will works as "Save as"

Thanks!

flatwhatson:
Put `save-buffer` back on `SPC f s` to save without prompt, and add `write-file` as `SPC f S` for saving with prompt for filename.

Fixes #1923


myrgy:
thanks!

PR #1936, opened 2019-10-23
lang/latex: fix rainbow-delimiters-mode

jonathye:
Every time auctex does style update, it somehow messes up rainbow-delimiters-mode. See also https://tex.stackexchange.com/questions/272320/emacs-auctex-preserve-highlights-from-other-modes-after-apply-style-hooks

This pull request fixes rainbow-delimiters-mode, and also fixes a typo introduced in my previous pull request :/


hlissner:
Good catch. Thanks once again for your help!

PR #1939, opened 2019-10-23
[ocaml] disable more features with lsp

rgrinberg:
merlin, merlin-eldoc, merlin-flycheck are all unnecessary when lsp is
turned on

### DIFF
--- a/modules/lang/ocaml/packages.el
+++ b/modules/lang/ocaml/packages.el
@@ -9,7 +9,7 @@
 
 (package! ocp-indent)
 
-(when (and (featurep! :tools flycheck) (not (featurep :lsp)))
+(when (and (featurep! :tools flycheck) (not (featurep +lsp)))
   (package! flycheck-ocaml))
 
 (when (featurep! :tools eval)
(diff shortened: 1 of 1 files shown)


modules/lang/ocaml/packages.el
```diff
@@ -2,11 +2,14 @@
...
-(package! merlin-eldoc)
+
+(unless (featurep! +lsp)
+  (package! merlin)
+  (package! merlin-eldoc))
+
 (package! ocp-indent)
 
-(when (featurep! :tools flycheck)
+(when (and (featurep! :tools flycheck) (not (featurep :lsp)))
```
hlissner: That should be `+lsp`, not `:lsp`

hlissner: And may as well combine the two blocks:

```elisp
(unless (featurep! +lsp)
  (package! merlin)
  (package! merlin-eldoc)
  (when (featurep! :tools flycheck)
    (package! flycheck-ocaml)))
```

rgrinberg: done

hlissner:

PR #1940, opened 2019-10-23
Add comment to search loaded .el files

rgrinberg:
This is quit similar to `SPC h d l` except that it will only search for loaded files.

Here's a very common use case for this: you get an annoying `(message ..)` and you don't know where it's coming from. How do you find it? `SPC h d l` works but isn't so good because it might bring up a lot of false positive. This command is both faster and will avoid false positives.

I can factor out the copy-paste if you want, but I assume you have more plans for this anyway. Note that this requires `elisp-refs` uless you want to copy paste the function to get all loaded files.


hlissner:
> I can factor out the copy-paste if you want, but I assume you have more plans for this anyway.

That's fine, I'll deal with that later.

Anyhow, thanks yet again!

PR #1945, opened 2019-10-24
Expand notmuch features

mnick:
This PR adds various features to the notmuch module

- Filter read messages in notmuch show (by folding them automatically)
- Refresh all notmuch buffers after sync (via process sentinel)
- Support custom sync commands (introduces an additional defvar)
- Add command to select From email prior to composing via ivy

I wasn't sure if you prefer separate diff or one big diff. Please let me know if you want me to split this into separate PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,16 +63,21 @@ to learn more about our favorite OS.
 ## Doom's mantras
 
 - **Gotta go fast.** Startup and run-time performance are high priorities.
-  Expensive plugins and functionality is modified and optimized toward this end,
-  otherwise, they must be opt-in.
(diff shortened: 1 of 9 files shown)


Emiller88:
While we're here, it might be a good idea to add some default bindings. Ie I think `+notmuch/quit` should be bound similiarly to `mu4e` and other apps.

https://github.com/Emiller88/doom-emacs-private/blob/master/config.org#notmuch

mnick:
Accidentally committed this on the main branch. Will create a new clean PR in a separate branch

PR #1948, opened 2019-10-24
ui/modeline: fix ligature advice for hlissner#1216

facelesspanda:
The advice was pulling in an `icon-name` in the args list but the whole body considers `args` as containing everything. So now `args` contains everything.


hlissner:
Thanks for the help!

PR #1950, opened 2019-10-25
Improvements to notmuch module

mnick:
This PR adds various features to the notmuch module

- Filter read messages in notmuch show (by folding them automatically)
- Refresh all notmuch buffers after sync (via process sentinel)
- Support custom sync commands (introduces an additional defvar)
- Add command to select From email prior to composing via ivy

I wasn't sure if you prefer separate diff or one big diff. Please let me know if you want me to split this into separate PRs.

### DIFF
--- a/modules/email/notmuch/autoload.el
+++ b/modules/email/notmuch/autoload.el
@@ -87,12 +87,13 @@
   (notmuch-tree-next-message))
 
 ;;;###autoload
-(defun +notmuch/ivy-compose ()
+(defun +notmuch/compose ()
   "Compose new mail"
   (interactive)
-  (ivy-read "From: "
-            (notmuch-user-emails)
-            :action (lambda (selection) (notmuch-mua-mail nil nil (list (cons 'From selection))))))
+  (notmuch-mua-mail
+   nil
+   nil
+   (list (cons 'From  (completing-read "From: " (notmuch-user-emails))))))
 
 ;;;###autoload
 (defun +notmuch/open-message-with-mail-app-notmuch-tree ()
--- a/modules/email/notmuch/autoload.el
+++ b/modules/email/notmuch/autoload.el
@@ -57,6 +57,13 @@
   (notmuch-tree-next-message))
 
 ;;;###autoload
+(defun +notmuch/show-delete ()
+  "Mark email for deletion in notmuch-show"
+  (interactive)
+  (notmuch-show-add-tag (list "+trash" "-inbox" "-unread"))
+  (notmuch-show-next-thread-show))
+
+;;;###autoload
 (defun +notmuch/search-spam ()
   (interactive)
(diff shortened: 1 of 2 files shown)


modules/email/notmuch/autoload.el
```diff
@@ -68,6 +86,14 @@
   (notmuch-tree-add-tag (list "+spam" "-inbox" "-unread"))
   (notmuch-tree-next-message))
 
+;;;###autoload
+(defun +notmuch/ivy-compose ()
+  "Compose new mail"
+  (interactive)
+  (ivy-read "From: "
+            (notmuch-user-emails)
+            :action (lambda (selection) (notmuch-mua-mail nil nil (list (cons 'From selection))))))
```
hlissner: Could you generalize this to use `completing-read`? This way ivy, helm, ido or other users can use this command.

modules/email/notmuch/config.el
```diff
@@ -42,6 +45,9 @@
 
   ;; (setq-hook! 'notmuch-show-mode-hook line-spacing 0)
 
+  ;; only unfold unread messages in thread by default
+  (add-hook 'notmuch-show-hook '+notmuch-show-expand-only-unread-h)
```
hlissner: Minor issue: please sharp-quote function symbols. i.e. `#'+notmuch-show-expand-only-unread-h`

mnick: Resolved in e279409

modules/email/notmuch/autoload.el
mnick: Resolved in a4050a1

mnick:
Thanks for the review. I also added some basic key bindings under localleader.

modules/email/notmuch/config.el
```diff
@@ -56,7 +56,20 @@
...
+   :map (notmuch-search-mode-map notmuch-tree-mode-map notmuch-show-mode-map)
+   :desc "compose email" "c" #'+notmuch/compose
+   :desc "fetch new email" "u" #'+notmuch/update
+   :desc "quit notmuch" "q" #'+notmuch/quit
+   :map notmuch-search-mode-map
+   :desc "mark as deleted" "d" #'+notmuch/search-delete
+   :desc "mark as spam" "d" #'+notmuch/search-spam
+   :map notmuch-tree-mode-map
+   :desc "mark as deleted" "d" #'+notmuch/tree-delete
+   :desc "mark as spam" "d" #'+notmuch/tree-spam))
```
hlissner: It appears the `d` key is bound twice in each map. Was the second `d` meant to be `D`? Or a different key?

mnick: Good catch, thanks!

hlissner:
Looks good to me! Thanks for the help!

PR #1953, opened 2019-10-25
[org +gnuplot] Warn if gnuplot is not installed

darth10:
`doom doctor` should catch `gnuplot` not being installed when the `+gnuplot` feature is enabled.


hlissner:
Good idea. Thanks again!

PR #1954, opened 2019-10-25
Fix paths in the FAQ

gagbo:
Fix links in the faq which were absolute


hlissner:

PR #1955, opened 2019-10-25
Add an FAQ action to doom-dashboard

gagbo:
# Add FAQ to doom-dashboard

I keep forgetting that there's a searchable FAQ, and I thought that having it on the dashboard was the best place : first screen someone sees when opening Doom. Especially until the user manual is not done, I discovered that while adding the menu item

Feel free to change the wording/icon if you think something else should be used


hlissner:
Nice work. Thanks again for your help!

PR #1956, opened 2019-10-25
Add snippets to lsp completions by default

gagbo:
As suggested by company-yasnippet documentation https://github.com/company-mode/company-mode/blob/f6974e3103497afe01da2ad4d2ea3668ad68bc27/company-yasnippet.el#L118


gagbo:
It's indeed noisier than I thought it would be. I'll keep that in my private config because I don't really mind, but it's probably not a good default value

PR #1959, opened 2019-10-26
[config] which-key descriptions for prefixes

facelesspanda:
`SPC w m` did not have a prefix.
~~Neither did `SPC h d` and `SPC h r`~~

~~I used a variable to define the help key to avoid hard-coding to `h`. I don't know if it's the smartest choice.~~

~~(I also don't really know why the help-map is not mapped with `map!` but I left it like that)~~

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -627,3 +627,9 @@
         (:when (featurep! :lang org +pomodoro)
           :desc "Pomodoro timer"             "t" #'org-pomodoro)
         :desc "Word-wrap mode"               "w" #'+word-wrap-mode))
+
+(after! which-key
+  (let ((prefix-re (regexp-opt (list doom-leader-key doom-leader-alt-key))))
+    (cl-pushnew `((,(format "\\`\\(?:C-w\\|%s w\\) m\\'" prefix-re))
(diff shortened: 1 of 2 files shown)


facelesspanda:
@hlissner I rebased onto your latest changes. I still think there ought to be a which-key description to the window-map maximization prefix

hlissner:
> I still think there ought to be a which-key description to the window-map maximization prefix

Unfortunately, that will not work. Which-key does not support labels on non-major-mode keymaps (like `evil-window-map`). This change will have the unintended effect of labeling *all* `m` keys as "Maximize".

The only way around this is to do as we did with `SPC h {d,r}` and use absolute key sequences, e.g. `C-w m`, `<leader> w m` and `<alt-leader> w m`.

facelesspanda:
ah I see, thanks for the info. I'll see about that

facelesspanda:
I added a working config for maximize in `+evil-bindings.el` and changed the regexp for the help sub-prefixes, because `C-h` doesn't seem to correspond to `<help>`, or at least it doesn't work when I press `C-h`

hlissner:
I hate that this is necessary (no fault of your own; I which which-key was more versatile), but I think this is good enough. Thanks again for your help, and for putting up with me!

PR #1960, opened 2019-10-27
add key binding for commenting code

chrunchyjesus:
This PR adds a bindig to `<leader> c /` for comment/uncommenting code. I use this quite frequently so I thought it can be beneficial to others too.


gagbo:
Evil users can already use evil-nerd-commentary to do this, it's mapped on `gc <motion>` or `gc` if a region is selected or `gcc` to comment one line

chrunchyjesus:
Wow, didn't know that.

@hlissner Please close this if this PR does not provide any values in your eyes. :grin:

PR #1962, opened 2019-10-27
add crude version of docs/contributing.org

magnetophon:

### DIFF
--- a/docs/contributing.org
+++ b/docs/contributing.org
@@ -39,6 +39,7 @@ To contribute to doom, it's easiest to follow these steps:
     git commit
     git checkout -b my-first-feature
     git push origin my-first-feature
+    git checkout develop
     # hack hack hack
     git commit
     git checkout -b my-second-feature
--- a/docs/contributing.org
+++ b/docs/contributing.org
@@ -27,7 +27,9 @@ To contribute to doom, it's easiest to follow these steps:
    ```
     # hack hack hack (on develop branch)
     git commit
-    git push origin my-awesome-feature
+    git checkout -b my-feature
+    git push origin my-feature
+    git checkout develop
     # click link to make a PR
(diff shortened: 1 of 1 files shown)


facelesspanda:
I know it's obvious for some, but it wouldn't hurt to specify that `.emacs.d` in this scenario is cloned from a fork, and not this here repo (people blindly copying the commands assuming they're all safe and thought out would end up having two named remotes to hlissner/doom-emacs)

magnetophon:
Thanks @facelesspanda.  
I realize this problem, and would prefer to have an idiot-proof solution that can just be copy-pasted.
Idiot-proof, so I can use it myself!
Suggestions are welcome.

docs/contributing.org
```diff
@@ -0,0 +1,49 @@
...
+    git checkout develop
+    # click link to make a PR
+   ```
+
+ 4) In case you have multiple concurrent PR's going on:
+    ```
+    # hack hack hack
+    git commit
+    git checkout -b my-first-feature
+    git push origin my-first-feature
```
flatwhatson: Should `git checkout develop` after pushing.

magnetophon: @flatwhatson Thanks, done!

hlissner:
It isn't quite what was proposed, but 3f6decf is a start. What was proposed in this PR could go in `Contributing code > Your first code contribution`, but I'll have to think about it.

magnetophon:
@hlissner Thanks for writing these docs, I'll go read them right now.

In case the PR wasn't clear: what I need most is an explanation of how to work around the way ``doom up`` "breaks" after you contribute.
Another way to go would be for doom up to be changed so this doesn't happen.
Maybe by having it look for an upstream remote and pull from that?

PR #1967, opened 2019-10-27
doc - fix instruction on removing +org-cycle-only-current-subtree-h

JakDar:
`+org-cycle-only-current-subtree-h` is being added to hook in `evil-org`  config, not in `org`'s one.
Also snippet wouldn't work with `local` flag (last arg of  `remove-hook`).


hlissner:

PR #1971, opened 2019-10-27
[dired] add dired-git-info and hide dotfiles/details

mnick:
This PR makes a couple of simple changes to the dired module which I find useful:

- Add dired-git-info package and enable it automatically in git (sub)dirs
- Add support for toggling dotfiles
- Hide dotfiles and dired details by default

### DIFF
--- a/modules/emacs/dired/autoload.el
+++ b/modules/emacs/dired/autoload.el
@@ -10,7 +10,9 @@
 
 ;;;###autoload
 (defun +dired-enable-git-info-h ()
-  (if (locate-dominating-file "." ".git")
+  (if (and
+       (not (file-remote-p default-directory))
+       (locate-dominating-file "." ".git"))
       (dired-git-info-mode 1)))
 
 ;;;###autoload
--- a/modules/emacs/dired/autoload.el
+++ b/modules/emacs/dired/autoload.el
@@ -9,12 +9,12 @@
 
 
 ;;;###autoload
-(defun +dired/enable-git-info-h ()
+(defun +dired-enable-git-info-h ()
   (if (locate-dominating-file "." ".git")
       (dired-git-info-mode 1)))
 
 ;;;###autoload
-(defun +dired/dotfiles-hide ()
+(defun +dired-dotfiles-hide ()
   (set (make-local-variable '+dired-dotfiles-show-p) nil)
   (dired-mark-files-regexp "^\\\.")
   (dired-do-kill-lines))
@@ -24,6 +24,6 @@
   (interactive)
   (when (equal major-mode 'dired-mode)
     (if (or (not (boundp '+dired-dotfiles-show-p)) +dired-dotfiles-show-p) ; if currently showing
-        (+dired/dotfiles-hide)
+        (+dired-dotfiles-hide)
       (progn (revert-buffer) ; otherwise just revert to re-show
              (set (make-local-variable '+dired-dotfiles-show-p) t)))))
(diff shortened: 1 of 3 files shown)


modules/emacs/dired/config.el
```diff
@@ -168,3 +173,11 @@ we have to clean it up ourselves."
   :when (executable-find doom-projectile-fd-binary)
   :defer t
   :init (advice-add #'find-dired :override #'fd-dired))
+
+
+(use-package! dired-git-info
+  :bind (:map dired-mode-map (")" . dired-git-info-mode))
```
hlissner: We don't use :bind or the associated libraries in Doom. Please convert this into a `define-key` call in the `:config` block for `dired-git-info`.

```elisp
(define-key dired-mode-map ")" #'dired-git-info-mode)
```

modules/emacs/dired/config.el
```diff
@@ -168,3 +173,11 @@ we have to clean it up ourselves."
...
+  :after dired
+  :init
+  (progn
```
hlissner: Please remove this superfluous `progn`.

modules/emacs/dired/config.el
```diff
@@ -168,3 +173,11 @@ we have to clean it up ourselves."
...
+    (add-hook 'dired-after-readin-hook '+dired-enable-git-info-h)))
```
hlissner: On second thought, this is a mote too imposing. It is likely to break ranger/deer integration and slow TRAMP dired buffers to a crawl. I'd suggest that activating this package be opt-in, and manual.

modules/emacs/dired/autoload.el
```diff
@@ -6,3 +6,24 @@
   (interactive)
   (mapc #'kill-buffer (doom-buffers-in-mode 'dired-mode))
   (message "Killed all dired buffers"))
+
+
+;;;###autoload
+(defun +dired-enable-git-info-h ()
+  (if (locate-dominating-file "." ".git")
```
hlissner: `locate-dominating-file` will be fairly slow in TRAMP buffers. Perhaps wrap this in a `(file-remote-p FILE)` check, and no-op in that case.

modules/emacs/dired/autoload.el
```diff
@@ -6,3 +6,24 @@
...
+  (dired-do-kill-lines))
+
+;;;###autoload
+(defun +dired/dotfiles-toggle ()
+  (interactive)
+  (when (equal major-mode 'dired-mode)
+    (if (or (not (boundp '+dired-dotfiles-show-p)) +dired-dotfiles-show-p) ; if currently showing
+        (+dired-dotfiles-hide)
+      (progn (revert-buffer) ; otherwise just revert to re-show
+             (set (make-local-variable '+dired-dotfiles-show-p) t)))))
```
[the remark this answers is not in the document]
mnick: Good point. I removed those functions and just set `dired-omit-files` to `"^\\\."` in my config. Additionally added a localleader binding to toggle dired-omit

modules/emacs/dired/autoload.el
mnick: Good catch! I added a `file-remote-p` check in the hook and enable the package only if `+ranger` is not set for the module

mnick:
Just checking in to see if you want additional changes.

hlissner:
Sorry for the late response. This looks good to me. Thanks again for your help!

PR #1972, opened 2019-10-28
[popup] ignore send mail buffers

rgrinberg:
[C-x m] should not appear in a popup.

In fact, I think this rule should be applied to all `message-mode` buffers.


hlissner:

PR #1977, opened 2019-10-29
lang/ruby: auto enter inf-ruby when breakpoint detected

eltone:
This PR add a hook that auto switches to inf-ruby when the compilation buffer matches the `inf-ruby-breakpoint-pattern`.


hlissner:
Looks good to me. Thanks!

PR #1979, opened 2019-10-29
[FEATURE] Add :ui hydra README

gagbo:
This is mostly a stub, the keybindings are just the one I chose for my own usage, I don't know if there are better choices or not :(

- [x] ~~evil keybindings~~
- [ ] ~~emacs keybindings~~
- [x] README added

### DIFF
--- a/modules/ui/hydra/README.org
+++ b/modules/ui/hydra/README.org
@@ -1,0 +1,31 @@
+#+TITLE:   ui/hydra
+#+DATE:    october 29, 2019
+#+SINCE:   2.0
+#+STARTUP: inlineimages
+
+* Table of Contents :TOC_3:noexport:
+- [[#description][Description]]
+  - [[#module-flags][Module Flags]]
+  - [[#plugins][Plugins]]
+- [[#prerequisites][Prerequisites]]
+- [[#configuration][Configuration]]
+
+* Description
+This module adds hydra to Doom Emacs, as well as a few custom built hydras to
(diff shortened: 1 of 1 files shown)


hlissner:
I'm still on the fence about hydra keybinds, and not too keen to add a new leader prefix for a single key. For the time being, please remove them. There already exist font size commands on `C-=` and `C--` for buffer-local resizing, `M-C-=` and `M-C--` for frame-local. I'll happily accept the README though.

gagbo:
Fair enough, I wasn't sure since it was a little confidential

hlissner:
Thanks for the help!

myrgy:
@gagbo ,
> Fair enough, I wasn't sure since it was a little confidential
I realised, that bindings file is the best README for now)
https://github.com/hlissner/doom-emacs/blob/develop/modules/config/default/+evil-bindings.el

gagbo:
> @gagbo ,
> 
> > Fair enough, I wasn't sure since it was a little confidential
>
> I realised, that bindings file is the best README for now)
> https://github.com/hlissner/doom-emacs/blob/develop/modules/config/default/+evil-bindings.el

Oh yeah for sure, what I meant is that the hydras defined in the module (hydra/window-nav and hydra/text-zoom) aren't bound at all (in modules/hydra/config.el or +*-bindings.el). The only way to know about those was to actually go in the `config.el` file of the module :)

PR #1980, opened 2019-10-29
fix markdown formatting

ebv-gh:
Fix formatting for `export PATH="$HOME/.emacs.d/bin:$PATH"` to include the right type of quotes.


hlissner:
Ah, good catch. Thanks for the help!

PR #1983, opened 2019-10-30
[FEATURE] Add flycheck-golangci-lint

gagbo:
The added plugin runs
[golangci-lint](https://github.com/golangci/golangci-lint) in the
background and gathers the results in a checker.

I don't know if there's a macro already to add a checker to flycheck, I don't know which hook to aim for if I want to run `(flycheck-add-next-checker 'lsp-ui 'golangci-lint)` only in `go-mode` buffers (without the lsp-ui setup destroying the flycheck checkers priority list just after)


hlissner:
Thanks for your help! I'll look into the lsp-ui issue another time. No macro exists for it, but perhaps one should.

PR #1984, opened 2019-10-30
[clojure] Check for nrepl-server-buffer in advice

yuhan0:
When using `cider-connect` to an external server, this function would throw
an exception because nrepl-server-buffer is nil

### DIFF
--- a/modules/lang/clojure/config.el
+++ b/modules/lang/clojure/config.el
@@ -57,11 +57,12 @@
   (add-hook! 'cider-connected-hook
     (defun +clojure--cider-dump-nrepl-server-log-h ()
       "Copy contents of *nrepl-server* to beginning of *cider-repl*."
-      (save-excursion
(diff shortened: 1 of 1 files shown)


modules/lang/clojure/config.el
```diff
@@ -57,11 +57,12 @@
   (add-hook! 'cider-connected-hook
     (defun +clojure--cider-dump-nrepl-server-log-h ()
       "Copy contents of *nrepl-server* to beginning of *cider-repl*."
-      (save-excursion
-        (goto-char (point-min))
-        (insert
-         (with-current-buffer nrepl-server-buffer
-           (buffer-string))))))
+      (when (bufferp nrepl-server-buffer)
```
hlissner: May as well be extra careful and use `buffer-live-p` here instead.

hlissner:

PR #1985, opened 2019-10-30
[git] Add smerge-mode bindings under leader "gm"

yuhan0:
These are useful builtin commands for resolving merge conflicts, and don't seem to have an alternative bound elsewhere.

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -419,6 +419,14 @@
         :desc "Git revert file"             "R"   #'vc-revert
         :desc "Copy git link"               "y"   #'git-link
         :desc "Copy git link to homepage"   "Y"   #'git-link-homepage
+        (:prefix ("m" . "merge conflicts")
+          :desc "Keep all conflicts"        "a"   #'smerge-keep-all
+          :desc "Keep current conflict"     "c"   #'smerge-keep-current
+          :desc "Next conflict"             "j"   #'smerge-next
+          :desc "Previous conflict"         "k"   #'smerge-prev
+          :desc "Keep lower conflict"       "l"   #'smerge-keep-lower
+          :desc "Keep upper conflict"       "u"   #'smerge-keep-upper
+          :desc "Refine conflict diffs"     "r"   #'smerge-refine)
         (:when (featurep! :ui vc-gutter)
           :desc "Git revert hunk"           "r"   #'git-gutter:revert-hunk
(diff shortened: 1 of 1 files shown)


facelesspanda:
could use `smege-keep-other` too, on `SPC g m o` maybe ?

yuhan0:
I believe `other` and `mine` are equivalent to `upper` and `lower` and have been obsolete since Emacs 25? 

Although looking at the source, it looks like Smerge defines a customizable prefix (the awkward `C-c ^` by default), so we could simply defer to it by doing something like
```
(map! :leader "gm" smerge-basic-map)
```

Would that be preferable? I've personally never used the other commands and `C-m` for keep-current is a little awkward.

hlissner:
That said, I'm not sure the `<leader>` is the right place for smerge keybinds. They're not really global commands, they're local to specific buffers (i.e. they might make better localleader keybinds, or keybinds on the `g` or `[` and `]` prefixes.

I'll think more on it.

> (map! :leader "gm" smerge-basic-map)

Unless the package is loaded immediately and automatically by Emacs at startup, `smerge-basic-map` won't be available until the package is loaded, so it'd have to be in a `:after`, like so:

```elisp
(map! :leader
      (:after smerge-mode
        "gm" smerge-basic-map))
```

facelesspanda:
if it's not on `<leader>` it would make sense to have them under `g` indeed. I'm thinking `g M` maybe?

yuhan0:
I don't quite understand what the distinction is between global and buffer local keybinds - aren't currently bound `<leader> g` commands like `g ]` / `g [` and even the stage / unstage / revert / blame keybindings all specific to the local buffer? 

But feel free to close this PR and implement it your way, or suggest any changes I can make :)

yuhan0:
A hydra may also be useful for batch resolving of merge conflicts, see https://github.com/alphapapa/unpackaged.el#smerge-mode

facelesspanda:
no `<leader> g` is available whatever the buffer, whatever the mode.
Henrik was suggesting setting the smerge bindings up only when smerge-mode is enabled. Since it's a minor mode, having them under the evil normal state `g` submap (pressing g when in evil normal mode gives you a LOT of stuff, we could add it there) could make sense.

hlissner:
Sorry for the late response. I think binding `SPC g m` to a hydra would be best. However, I'm currently preoccupied with other work and won't be able to look into this for another week or two.

hlissner:
Sorry for the late response! This has been taken care of by #2512, so I will close this.

PR #1991, opened 2019-10-31
[lang/javascript] Add prefix names to JavaScript bindings

johsi-k:
These provide useful hints for `js2-refactor`'s deeply nested bindings.

`tide-rename-symbol` is remapped to `SPC r r s` to de-conflict with `js2-refactor`'s `r s` mapping.

### DIFF
--- a/modules/lang/javascript/config.el
+++ b/modules/lang/javascript/config.el
@@ -62,25 +62,7 @@
 
   (map! :map js2-mode-map
         :localleader
-        "S" #'+javascript/skewer-this-buffer
-        (:prefix ("n" . "npm"))
-        (:prefix ("r" . "refactor")
-          (:prefix ("a" . "add/arguments"))
-          (:prefix ("b" . "barf"))
-          (:prefix ("c" . "contract"))
-          (:prefix ("d" . "debug"))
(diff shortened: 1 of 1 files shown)


modules/lang/javascript/config.el
```diff
@@ -62,7 +62,25 @@
...
+          (:prefix ("l" . "localize/log"))
+          (:prefix ("o" . "organize"))
+          (:prefix ("r" . "rename"))
+          (:prefix ("s" . "slurp/split/string"))
+          (:prefix ("t" . "toggle"))
+          (:prefix ("u" . "unwrap"))
+          (:prefix ("v" . "var"))
+          (:prefix ("w" . "wrap"))
+          (:prefix ("3" . "ternary")))
+        (:prefix ("s" . "skewer"))))
```
hlissner: Could you move these into `js2-refactor`'s `:config` block instead?

johsi-k: Sure! Would it matter that the n and s prefixes don't belong to `js2-refactor`?

hlissner: Ah, good catch. Yes, I think that's appropriate.

hlissner:
Great. Thanks for your help!

PR #1998, opened 2019-11-01
Fix autoload cookie for doom/help-search-loaded-files

rgrinberg:
Signed-off-by: Rudi Grinberg <me@rgrinberg.com>


rgrinberg:
The second commit makes the searches a lot more useful. Symlinks are used extensively by straight, so without this option the results will usually be incomplete.

hlissner:

Issue #1994, opened 2019-10-31
[BUG] Terminal Emacs: cursor shape unchanged under insert mode


randomizedthinking:
**Describe the issue**
Under terminal emacs, the cursor shape is unchanged under both command and insert modes -- which leaves terminal emacs users confused at times.

The behavior is different under GUI Emacs. Even not many people use terminal Emacs, I'd think it is nice to correct the behavior.

**Steps to reproduce**
1. emacs -nw
2. open any file, then press `i` to enter insert mode
3. notice the cursor is unchanged.

**System information**
<details><pre>
emacs   version    26.1
        features   XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS NOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 THREADS LIBSYSTEMD LCMS2
        build      Feb 03, 2019
[...]

PR #1999, opened 2019-11-01
Add terminal-cursor-change snippet.

randomizedthinking:
This PR adds a snippet on docs/api.org page regarding cursor-shape-change under terminal Emacs, as referenced in #1994.


hlissner:
Thanks for your help!

PR #2000, opened 2019-11-01
[docs] correct some links

facelesspanda:
This un-hard-codes a few links I found in the docs. Most of those would work fine in most setups but were hard coded to `~/.emacs.d/...` so I changed them to `..`


hlissner:
Ah, my mistake. Thanks for your help!

PR #2001, opened 2019-11-01
Fix buttons in doom/help-packages

yuhan0:
Currently the generated lambda would place an unquoted list in the value position of `cl-destructuring-bind`, causing it to be interpreted as a function call.


hlissner:

PR #2003, opened 2019-11-02
Fix example to change org-directory in org README

bennofs:
org-directory must be set in `init.el` to make sure it is set before org
is loaded.


hlissner:
This is incorrect. Org is lazy loaded so that `org-directory` can be set from config.el (and it is preferred that configuration be done in config.el).

The only where this isn't true is if you eager-load `org` in your private config *before* setting `org-directory`.

PR #2005, opened 2019-11-02
Fixed keybindings to org-attach commands

RBckmnn:
Invoking the org-attach commands under the SPC-m-a prefix resulted in errors. It seems there was a plus sign omitted in the commands' names.


hlissner:

PR #2006, opened 2019-11-03
Fixed sudo edit on remote files

RBckmnn:
Hi there,

this pull request addresses a minor issue when sudo-editing remote files: The implementation of `doom/sudo-find-file` was adjusted in such a way, that the correct file is found, if the remote user is not given explicitely, i.e. if the file passed to the function is of the form:

`/ssh:otherhost:myfile`

instead of

`/ssh:alice@otherhost:myfile`

In that case, the previous implementation invoked `find-file` with a file name that contained an extra @ character.

 
Hope the code is not to obtuse or convoluted. I don't code much in elisp :-)

### DIFF
--- a/core/autoload/files.el
+++ b/core/autoload/files.el
@@ -319,9 +319,14 @@ file if it exists, without confirmation."
   (interactive "FOpen file as root: ")
   (when (file-writable-p file)
     (user-error "File is user writeable, aborting sudo"))
-  (find-file (if (file-remote-p file)
-                 (concat "/" (file-remote-p file 'method) ":" (file-remote-p file 'user) "@" (file-remote-p file 'host)  "|sudo:root@" (file-remote-p file 'host) ":" (file-remote-p file 'localname))
-               (concat "/sudo:root@localhost:" file))))
+  (let* ((host (or (file-remote-p file 'host) "localhost"))
+         (user-host (if-let (( user (file-remote-p file 'user)))
+                        (concat user "@" host)
+                      host))
+         (hop1 (when (file-remote-p file) (concat "/" (file-remote-p file 'method) ":" user-host)))
+         (sep (if hop1 "|" "/"))
+         (hop2 (concat "sudo:root@" host ":" (or (file-remote-p file 'localname) file))))
+    (find-file (concat hop1 sep hop2))))
 
 ;;;###autoload
 (defun doom/sudo-this-file ()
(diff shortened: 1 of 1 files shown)


RBckmnn:
Yes, that looks cleaner. I added another check to make sure that the current buffer's file is really the file to sudo-edit before getting rid of the buffer.

Did you remove the check for a writeable file in your code on purpose?

core/autoload/files.el
```diff
@@ -317,11 +317,19 @@ file if it exists, without confirmation."
...
+                                  (if-let (user (file-remote-p file 'user))
+                                      (concat user "@" host)
+                                    host)
+                                  "|"))
+                    "sudo:root@" host
+                    ":" (or (file-remote-p file 'localname)
+                            file))))
+    (if (and buffer-file-name (equal file (file-truename buffer-file-name)))
+        (find-alternate-file f)
+      (find-file f))))
```
hlissner: Use `(file-equal-p file buffer-file-name)` instead of `(equal file (file-truename buffer-file-name))`

RBckmnn: I did that at first but then I noticed that according to 

[Emacs manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Truenames.html#Truenames)

the result of `file-equal-p` is unspecified if either file does not exist. This might be an issue if

1. One wants to create the file via `doom/sudo-find-file`
2. The current buffer is not associated with a file

I considered testing for the files' existence but arrived at the analogous conclusion as you did:
I don't know how `file-exists-p` works with remote files :-).

So, where do we go from here?

hlissner:
Does this work for you?

```elisp
(defun doom--sudo-file (file)
  (let ((host (or (file-remote-p file 'host) "localhost")))
    (concat "/" (when (file-remote-p file)
                  (concat (file-remote-p file 'method) ":"
                          (if-let (user (file-remote-p file 'user))
                              (concat user "@" host)
                            host)
                          "|"))
            "sudo:root@" host
            ":" (or (file-remote-p file 'localname)
                    file))))

;;;###autoload
(defun doom/sudo-find-file (file)
  "Open FILE as root."
  (interactive "FOpen file as root: ")
  (find-file (doom--sudo-file file)))

;;;###autoload
(defun doom/sudo-this-file ()
  "Open the current file as root."
  (interactive)
  (find-alternate-file (doom--sudo-file buffer-file-name)))
```

RBckmnn:
Yes, works nicely. I updated the PR accordingly.

hlissner:
Great! Thanks for your help!

PR #2008, opened 2019-11-03
Mention $DOOMDIR/modules explictly in docs

bennofs:
The docs already mention in in the load path, but I found that to be a little bit hidden. I think it makes sense to explictly mention it when explaining how to write your own modules?


hlissner:
Thanks for the help!

PR #2009, opened 2019-11-03
Avoid nav-flash after each hydra action

bennofs:
Hydra displays the hints in a buffer created by lv-window, which
triggers nav-flash. By advicing lv-window, we can inhibit the nav-flash
hook.

### DIFF
--- a/modules/ui/hydra/config.el
+++ b/modules/ui/hydra/config.el
@@ -9,8 +9,7 @@
 ;;;###package hydra
 (setq lv-use-separator t)
 
-(after! hydra
-  (defadvice! +hydra/inhibit-window-switch-hooks-a (orig-fn)
-    :around #'lv-window
-    (let ((doom-inhibit-switch-window-hooks t))
-      (funcall orig-fn))))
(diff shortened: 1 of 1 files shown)


modules/ui/hydra/config.el
```diff
@@ -8,3 +8,9 @@
 
 ;;;###package hydra
 (setq lv-use-separator t)
+
+(after! hydra
```
hlissner: The `after!` is unnecessary for advice

modules/ui/hydra/config.el
```diff
@@ -8,3 +8,9 @@
...
+  (defadvice! +hydra/inhibit-window-switch-hooks-a (orig-fn)
```
hlissner: Please rename this function to `+hydra--inhibit-window-switch-hooks-a` to conform to [naming conventions](https://github.com/hlissner/doom-emacs/blob/develop/docs/contributing.org#naming-conventions).

hlissner:
Thanks for the help!

PR #2011, opened 2019-11-04
[ui/workspaces] Prompt to create new workspace

yuhan0:
When calling `gt` with a single workspace, DWIM by prompting to create a new workspace instead of throwing an error.


hlissner:
I disagree with this PR for several reasons:

1. The purpose of `+workspace/cycle` is to cycle through existing workspaces, not to DWIM, or create workspaces.
2. The purpose of `gt`/`gT` is to cycle between existing "tabs", as they do in vim. They do not create tabs there either.
3. This change would cause this command to prompt not just once, but twice -- once to confirm, and again to name the workspace. That's too imposing for motion keys.

I think this change would make `+workspace/cycle` do too much. The `SPC TAB n` or `C-t` (or `cmd+t` on mac) exist for this purpose.

yuhan0:
Alright, I'll close this and have it in my private config :)

PR #2016, opened 2019-11-05
Handle "'" local pair in python mode only when smartparens enabled

valignatev:
Otherwise, python-mode fails to start with "Symbol's value as variable is void": sp-local-pair

### DIFF
--- a/modules/lang/python/config.el
+++ b/modules/lang/python/config.el
@@ -74,10 +74,11 @@ called.")
         (setq-local flycheck-python-flake8-executable "flake8"))))
 
   (define-key python-mode-map (kbd "DEL") nil) ; interferes with smartparens
-  (sp-local-pair 'python-mode "'" nil
-                 :unless '(sp-point-before-word-p
-                           sp-point-after-word-p
-                           sp-point-before-same-p))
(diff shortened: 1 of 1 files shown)


hlissner:
I cannot reproduce this. Are you sure you didn't disable the smartparens package in your recent optimizing attempts? Smartparens is a core Doom package and is relied upon in a number of places. You can safely disable the mode, e.g.

```elisp
(after! smartparens
  (smartparens-global-mode -1))
```

But disabling the package, e.g. with `(package! smartparens :disable t)` isn't supported and will cause these sorts of errors.

valignatev:
Ah, got it. I definitely did `(package! smartparens :disable t)` because I've noticed that commenting out +smartparens from init.el didn't disable the mode.
```
(after! smartparens
  (smartparens-global-mode -1))
```
will work for me if disabling it altogether will result in breaking too much stuff, thanks!

hlissner:
I understand it isn't ideal. I will [eventually factor out smartparens](https://github.com/hlissner/doom-emacs/projects/3#card-10738874) from Doom, but it will be a while until I get to it. I'll close this for the time being.

PR #2017, opened 2019-11-05
Fix nil error in +ivy/jump-list for empty buffers

bennofs:
For empty buffers which don't have any lines, `(thing-at-point 'line)`
will return `nil`. This broke `+ivy/jump-list`.


hlissner:

PR #2019, opened 2019-11-05
Update Customize section links and a bit

sogaiu:
A couple of the links to the Customize section were 404-ing, so this PR attempts to address those.

Also, there was a reference to lispyville.el in the "Installing packages from external sources" which I found confusing.  Perhaps what it used to refer to was removed at some point?  The PR contains a suggestion for alternative text.


hlissner:
Thanks for your help!

sogaiu:
Thanks for doom-emacs :)

Issue #1952, opened 2019-10-25
[ROADMAP] rst plugins watch


gagbo:
# Plugins regarding ReST on melpa

I'm still not sure about what would be needed in a [:lang rst](https://github.com/hlissner/doom-emacs/projects/3#card-21772931) module, so I went ahead to look for the packages currently on melpa to see what they provide on top of emacs' bundled rst-mode.

## Org export

There's [ox-rst](https://github.com/msnoigrs/ox-rst/tree/master). Great README file which shows what to expects regarding the export, so reading that should be enough.

## Sphinx compilation

ReST documentation can be formatted into html or latex using this [Sphinx frontend package](https://github.com/kostafey/sphinx-frontend/tree/master).

Sphinx seems special since it has a python configuration file and all, but converting markup text to another markup text format seems also like something pandoc would handle well.

## Completion

There's only a completion source for auto-complete, [auto-complete-rst](https://github.com/tkf/auto-complete-rst/tree/master). I didn't find anything for company sadly.

That auto-complete plugin takes a list of path to Sphinx extension files, so I assume this provides project-aware completion depending on the extensions in use

hlissner:
What #2020 merged, I'll consider this resolved. Thanks for bringing it to my attention!

PR #2020, opened 2019-11-06
Add rst module

rgrinberg:
Uses sphinx-mode and adds a few common bindings

### DIFF
--- a/modules/lang/rst/config.el
+++ b/modules/lang/rst/config.el
@@ -1,3 +1,5 @@
+;;; lang/rst/config.el -*- lexical-binding: t; -*-
+
 (use-package! sphinx-mode
   :init
   (add-hook! 'rst-mode-hook #'sphinx-mode))
--- a/modules/lang/rst/config.el
+++ b/modules/lang/rst/config.el
@@ -8,8 +8,8 @@
   (map! :localleader
         :map rst-mode-map
         (:prefix ("a" . "adjust")
-          ("a" #'rst-adjust
-           "r" #'rst-adjust-region))
+          "a" #'rst-adjust
+          "r" #'rst-adjust-region)
         (:prefix ("t" . "table of contents")
           ("t" #'rst-toc
            "i" #'rst-toc-insert
--- a/modules/lang/rst/config.el
+++ b/modules/lang/rst/config.el
@@ -1,8 +1,7 @@
 ;;; lang/rst/config.el -*- lexical-binding: t; -*-
 
 (use-package! sphinx-mode
-  :init
-  (add-hook! 'rst-mode-hook #'sphinx-mode))
+  :hook (rst-mode . sphinx-mode))
 
 (use-package! rst
   :config
--- a/modules/lang/rst/config.el
+++ b/modules/lang/rst/config.el
@@ -1,0 +1,16 @@
+(use-package! sphinx-mode
+  :init
+  (add-hook! 'rst-mode-hook #'sphinx-mode))
+
+(use-package! rst
+  :config
+  (map! :localleader
+        :map rst-mode-map
+        (:prefix ("a" . "adjust")
+          ("a" #'rst-adjust
+           "r" #'rst-adjust-region))
+        (:prefix ("t" . "table of contents")
+          ("t" #'rst-toc
+           "i" #'rst-toc-insert
+           "u" #'rst-toc-update
+           "f" #'rst-toc-follow-link))))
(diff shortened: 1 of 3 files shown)


gagbo:
I looked for a few rst related plugins [here](https://github.com/hlissner/doom-emacs/issues/1952)

Does `sphinx-mode` provide the "compilation" that `sphinx-frontend` seems to provide ? Do you think that would be a good addition ?

rgrinberg:
Personally I have no use for any of these compilation commands since I've put the common ones that I use in a makefile and just invoke them via `M-x compile`. I don't have anything against them in case someone else wants to add them however.

I mostly use sphinx-mode because of the better syntax highlighting.

modules/lang/rst/config.el
```diff
@@ -0,0 +1,16 @@
+(use-package! sphinx-mode
```
hlissner: Missing comment headings.

```elisp
;;; lang/rst/config.el -*- lexical-binding: t; -*-


```

modules/lang/rst/config.el
```diff
@@ -0,0 +1,16 @@
...
+  :init
+  (add-hook! 'rst-mode-hook #'sphinx-mode))
```
hlissner: This package isn't lazy loaded.

hlissner: ~~Use `add-hook` when `add-hook!` isn't necessary.~~

Use `:hook (rst-mode . sphinx-mode)` for this. It will also solve the issue of lazy loading sphinx-mode.

modules/lang/rst/config.el
```diff
@@ -0,0 +1,16 @@
...
+  (map! :localleader
+        :map rst-mode-map
+        (:prefix ("a" . "adjust")
+          ("a" #'rst-adjust
+           "r" #'rst-adjust-region))
+        (:prefix ("t" . "table of contents")
+          ("t" #'rst-toc
+           "i" #'rst-toc-insert
+           "u" #'rst-toc-update
+           "f" #'rst-toc-follow-link))))
```
hlissner: This package isn't lazy loaded.

hlissner: One levels of parentheses too many.

Don't do:

```elisp
(:prefix ("a" . "adjust")
  ("a" #'rst-adjust
   "r" #'rst-adjust-region))
```

Do:

```elisp
(:prefix ("a" . "adjust")
  "a" #'rst-adjust
  "r" #'rst-adjust-region)
```

modules/lang/rst/packages.el
```diff
@@ -0,0 +1 @@
+(package! sphinx-mode)
```
hlissner: Missing comment headings.

```elisp
;;; lang/rst/packages.el


```

modules/lang/rst/config.el
hlissner: Does activating `sphinx-mode` throw an error if its dependencies aren't installed? If so, write a wrapper for `sphinx-mode` to demote the error to a warning.

hlissner:
Oh, forgot to mention. Perhaps this module's packages.el is missing a:

```elisp
(when (featurep! :lang org)
  (package! ox-rst))
```

modules/lang/rst/config.el
rgrinberg: What dependencies do you have in mind? Dependencies for running sphinx itself? If so, then sphinx-mode does not do that.

rgrinberg:
> Otherwise, lang/org/packages.el is missing a

I went for this.

modules/lang/rst/config.el
```diff
@@ -0,0 +1,18 @@
...
+  (map! :localleader
+        :map rst-mode-map
+        (:prefix ("a" . "adjust")
+          "a" #'rst-adjust
+          "r" #'rst-adjust-region)
+        (:prefix ("t" . "table of contents")
+          ("t" #'rst-toc
+           "i" #'rst-toc-insert
+           "u" #'rst-toc-update
+           "f" #'rst-toc-follow-link))))
```
hlissner: One levels of parentheses too many.

modules/lang/rst/config.el
hlissner: I assumed sphinx-mode depended on `sphinx`, but I'll take your response as though it doesn't.

rgrinberg:
> I assumed sphinx-mode depended on sphinx, but I'll take your response as though it doesn't.
Yup I use without having the Sphinx binary in my path. But perhaps there are some use cases which require it.

rgrinberg:
@hlissner should be ready now

hlissner:

PR #2022, opened 2019-11-06
replace go-add-tags with go-tag

babariviere:
go-add-tags isn't maintained and has a few issues.
This library will fix those issues by using a binary intended for this.


hlissner:

PR #2037, opened 2019-11-09
Add keybindings for commands that show keymaps

RBckmnn:
This PR adds a new prefix and keybindings under that prefix to the help map. The bindings provide access to several commands that allow for the inspection of keymaps.


hlissner:
Thanks for the help!

PR #2038, opened 2019-11-09
[WIP] add github-review

Emiller88:
Pretty simple and cool package.
@hlissner I'm going to request you do the keybinding for
`github-review-forge-pr-at-point`, or shoot me an example of binding something
in a magit buffer, and what you'd like the binding to be.

### DIFF
--- a/modules/tools/magit/config.el
+++ b/modules/tools/magit/config.el
@@ -110,6 +110,7 @@ ensure it is built when we actually use Forge."
           (dolist (hook forge-bug-reference-hooks)
             (add-hook hook #'forge-bug-reference-setup)))))))
 
+(use-package! github-review)
 
 (use-package! magit-todos
   :after magit
--- a/modules/tools/magit/config.el
+++ b/modules/tools/magit/config.el
@@ -110,7 +110,16 @@ ensure it is built when we actually use Forge."
           (dolist (hook forge-bug-reference-hooks)
             (add-hook hook #'forge-bug-reference-setup)))))))
 
-(use-package! github-review)
+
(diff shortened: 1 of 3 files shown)


darth10:
IMHO a binding for `github-review-forge-pr-at-point` would fit quite well under the `magit-merge` menu as it's essentially means "review a merge".
I use this in my config:

```lisp
(transient-append-suffix 'magit-merge "i"
  '("y" "Review pull request" github-review-forge-pr-at-point))
```

Emiller88:
@darth10 That looks good. My vote however would go under the `@` for forge, or I don't see what the harm in having both is.

modules/tools/magit/config.el
```diff
@@ -115,6 +115,7 @@ ensure it is built when we actually use Forge."
           (dolist (hook forge-bug-reference-hooks)
             (add-hook hook #'forge-bug-reference-setup)))))))
 
+(use-package! github-review)
```
hlissner: This unnecessary eager-loads the package. It should be both lazy loaded and where you place your transient bindings.

hlissner: There should also be two blank lines on either side of this `use-package!` block.

Emiller88:
@hlissner Your right it doesn't and that `+magit/start-github-review ` makes it work well outside a forge workflow now.

hlissner:
Could you rebase this PR onto `develop`? Besides the conflict I think this looks ready to merge.

Emiller88:
@hlissner should be good to go!

hlissner:

PR #2040, opened 2019-11-10
Add packages for Kubernetes

darth10:
Including a few useful Kubernetes packages as a new `kubernetes` module.
~~I've nested them under the `+k8s` flag under the `docker` module, but happy to change that.~~

Any other feedback is more than welcome!

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -80,6 +80,7 @@
        flycheck          ; tasing you for every semicolon you forget
        ;;flyspell          ; tasing you for misspelling mispelling
        ;;gist              ; interacting with github gists
+       ;;kubernetes
        (lookup           ; helps you navigate your code and documentation
         +docsets)        ; ...or in Dash docsets locally
        ;;lsp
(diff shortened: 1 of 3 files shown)


hlissner:
I don't think this warrants inclusion in the docker module. It appears to require zero configuration. A user could install it themselves. Also, is there a reason to couple kubernetes with docker? It can be its own thing as well.

darth10:
> It appears to require zero configuration. A user could install it themselves.

I agree it's simple enough, but I also feel Doom Emacs should include as many packages as possible that can be opted in through modules and flags. 

> Also, is there a reason to couple kubernetes with docker?

I haven't really seen `kubectl` being used without `docker`.
Like, Docker Desktop (https://www.docker.com/products/docker-desktop) includes `kubectl` along with the installation.
Keen to hear your thoughts on this.

Perhaps it can be made a part of the same module now, and later moved out into its own module if needed.

darth10:
Alrighty, I've moved these packages into a separate `kubernetes` module.

PR #2041, opened 2019-11-10
[doctor] Ignore files that don't exist

darth10:
Minor fix to ignore files that don't exist.
Without this change, the following error occurs on running `bin/doom doctor` if `savehist` or `projectile.cache` are not present.

```
! Attempt to load DOOM failed
  (number-or-marker-p nil)
```


hlissner:
I came up with a slightly better solution in 962459e which should resolve this issue. In any case, thanks for bringing it to my attention!

PR #2044, opened 2019-11-10
Add clock bindings to org agenda

rgrinberg:
I don't feel strongly about any of these bindings. Feel free to tweak them as you wish.


hlissner:
Thanks again for your help!

PR #2047, opened 2019-11-11
Add rst module to init.example.el

alarsyo:
PR #2020 adds support for a `lang/rst` module, but it wasn't added to the example `init.el`, which I believe is the one that's placed in `~/.doom.d/init.el` during a new installation ?

I think this greatly improves discoverability :)


hlissner:

PR #2055, opened 2019-11-13
lang/org: Bind org-agenda-show-and-scroll-up

jhacksworth:
Default Org binds 'SPC' to `org-agenda-show-and-scroll-up`, which
displays the original location of the Agenda item in another window
without moving focus from the Agenda window.

This is useful to preview items while staying in the Agenda window.
Since we can't use 'SPC', bind it to 'localleader SPC'.

See https://orgmode.org/manual/Agenda-commands.html

### DIFF
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -691,6 +691,7 @@ between the two."
 
   (map! :after org-agenda
         :map org-agenda-mode-map
+        :m "C-SPC" #'org-agenda-show-and-scroll-up
         :localleader
         "d" #'org-agenda-deadline
         (:prefix ("c" . "clock")


wedens:
There is also `org-agenda-follow-mode` which'll follow the cursor.

jhacksworth:
> bind `C-SPC` to it in `org-agenda-mode-map`

That's even better! 

Do I have to push a change or do you have it done in your repo? I see a diff in your comment, but I'm not sure where it's from.

hlissner:
It's an isolated diff I wrote to demonstrate the change I was describing. Feel free to apply it to your PR and I will merge it.

hlissner:
Great! Thanks for your help and prompt response!

PR #2056, opened 2019-11-13
+eval: do not trim when send-region-to-repl

randomizedthinking:
Remove string-trim in +eval/send-region-to-repl. The trim operation would negatively affect languages like python where spaces are used for defining code blocks.

Just accept whatever users decide to send to the repl.

### DIFF
--- a/modules/tools/eval/autoload/repl.el
+++ b/modules/tools/eval/autoload/repl.el
@@ -118,7 +118,7 @@ immediately after."
     (with-selected-window (get-buffer-window buffer)
       (when (bound-and-true-p evil-local-mode)
         (call-interactively #'evil-append-line))
-      (insert (string-trim selection))
+      (insert selection)
       (unless inhibit-auto-execute-p
         ;; `comint-send-input' isn't enough because some REPLs may not use
         ;; comint, so just emulate the keypress.


hlissner:
Wouldn't a better solution be to rigidly dedent the input rather than leaving it as is? Going with the python example, if you ran

```python
  def test():
    print("Hi")
```

You'd get an error. Instead, with `indent-rigidly` we'd dedent it to

```python
def test():
  print("Hi")
```

randomizedthinking:
One use case is that user might submit snippets of code to python repl which is part of a function. In other words, the codes are not indented rigidly. Not trimming the string won't cause issues for any repl users.

hlissner:
I'm not suggesting we trim the string. I agree that needs to change. I'm suggesting we change it from trimming the string to rigidly dedenting the snippet, which should address the use case you mentioned.

randomizedthinking:
I see. Dedenting the snippet would be nice. But I cannot think of a use-case which seriously requires it.

There are other two issues with the `send-region-to-repl` call. First, the call will add a `<RET>` in the end. Yet, the string might already carry a return by itself. In other words, an extra return is added. Mostly, it is cosmetic, yet it could alter the meaning of the program. For instance, if we have an I/O request to the repl, then the extra return is definitely not desirable. The solution is to trim the ending return in the string if there is one.

Second, when sending a multiple-line region to repl, the repl screen is mis-aligned somehow. I think we need to break the region and send it line-by-line to the repl.

With both, I believe people will have a much better repl experience.

hlissner:
> I see. Dedenting the snippet would be nice. But I cannot think of a use-case which seriously requires it.

I just mentioned one earlier. Imagine sending the body of this function:

```python
def test():
  if True:
    print("Look ma, an error")
```

Unsurprisingly yields:

```
  print("hello")    if True:
        print("Look ma, an error")
  File "<stdin>", line 1
    print("hello")    if True:
    ^
IndentationError: unexpected indent
>>>   File "<stdin>", line 1
    print("Look ma, an error")
    ^
IndentationError: unexpected indent

```

Easily resolved if the input was dedented, so that the repl received:

```python
if True:
  print("Look ma, an error")
```

> There are other two issues with the send-region-to-repl call. First, the call will add a <RET> in the end. Yet, the string might already carry a return by itself. In other words, an extra return is added. Mostly, it is cosmetic, yet it could alter the meaning of the program. For instance, if we have an I/O request to the repl, then the extra return is definitely not desirable. The solution is to trim the ending return in the string if there is one.

You can send code to the REPL without submitting it by passing the universal argument to `gr`: `SPC u g r`. I'll look into adding a variable flip this behavior.

> Second, when sending a multiple-line region to repl, the repl screen is mis-aligned somehow. I think we need to break the region and send it line-by-line to the repl.

I'll look into it.

randomizedthinking:
I noticed that you are using python so received all such error messages. The indentation scheme would be fine under `ipython`.

randomizedthinking:
I will close this now since a specific commit (f54d7a15cd18db18fafe9d6c4e496ebc1b99d7e1) was created to handle this issue. Thanks.

PR #2061, opened 2019-11-15
Add binding for setting effort

rgrinberg:
The org-clock-modify-effort-estimate is only useful after we've clocked
in a task.


hlissner:
LGTM. Thanks for the PR.

PR #2067, opened 2019-11-16
Disable hello company coq

rgrinberg:
It's just an annoying splash screen message. Users don't need to see it.


hlissner:
LGTM. Thanks again!

PR #2069, opened 2019-11-18
tools/eval: ensure final newline only if needed #2056

randomizedthinking:
This is a continuation of the work on the repl newline issue. The patch makes the newline is added only the last line starts with either space or tab char.

### DIFF
--- a/modules/tools/eval/autoload/repl.el
+++ b/modules/tools/eval/autoload/repl.el
@@ -124,19 +124,23 @@ immediately after."
                (delete-region (point-min) (point)))
              (indent-rigidly (point) (point-max)
                              (- (skip-chars-forward " \t")))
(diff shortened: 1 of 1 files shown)


hlissner:
What does this do that appending a literal `\n` to the trimmed selection doesn't? Perhaps two newlines should be appended?

randomizedthinking:
What this section of code does is to check whether the last line of the submitted code is indented, then a RET (or newline) would be added only if it is. I don't want to send in two newlines regardless. If the last line is not indented, then there is no need to send in an extra newline. Since I use ipython, the extra returns are very visible, and should be removed.

randomizedthinking:
I am going to close this PR for now. The REPL in doom still has some problem.. this one helps but didn't solve any fundamental problem.

PR #2086, opened 2019-11-20
Documentation: Make it clearer that JavaScript module also is for TypeScript

paldepind:
Currently, it is a bit hard to figure out that Doom supports TypeScript.

I looked for a `typescript` module in `init.el` but couldn't find one. I then looked for a `typescript` directory in `modules/lang` but couldn't find one. It took me a while to realize that the JavaScript module also supports TypeScript. This PR improves discoverability of this fact by:

* Mentioning in the default user `init.el` that the JS module supports TS such that people can find it when searcing for "typescript".
* Make it clearer in the JavaScript module documentation that it also works for TypeScript.

If other people have additional ideas for improving discoveability I'll update the PR. One could also add a `modules/lang/typescript` directory with a readme that simply points to the JavaScript module. But, that may be more controversial so I didn't do that in this PR.

### DIFF
--- a/docs/index.org
+++ b/docs/index.org
@@ -186,7 +186,7 @@ Modules that bring support for a language or group of languages to Emacs.
 + hy - TODO
 + [[file:../modules/lang/idris/README.org][idris]] - TODO
 + java =+meghanada +lsp= - TODO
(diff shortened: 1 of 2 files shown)


paldepind:
I've removed the modification to `init.example.el` and added a fix for capitalization.

hlissner:
Thanks for the help!

PR #2087, opened 2019-11-20
lang/latex: Support forward search with Skim

paldepind:
In LaTeX forward search doesn't work for me with the Skim PDF-reader. This fix appears to be necessary.

The format string for opening Skim is taken from [Skim's documentation](https://sourceforge.net/p/skim-app/wiki/TeX_and_PDF_Synchronization/).


hlissner:
Thanks for the help!

paldepind:
Thanks for merging and thanks for the great work you're doing. I've had a blast with doom-emacs such far.

PR #2101, opened 2019-11-21
lang/org: add support for org-cliplink

lovebug356:
This PR adds support for [org-cliplink](https://github.com/rexim/org-cliplink). org-cliplink takes a URL from the clipboard and inserts an org-mode link with a title of a page found by the URL into the current buffer.


lovebug356:
Thx for the quick reply, done!

hlissner:

PR #2106, opened 2019-11-23
emacs/dired: ranger catch up dired-git-info

ztlevi:
A simple PR for ranger using dired-git-info package.


hlissner:
Thanks for your help!

PR #2109, opened 2019-11-24
Add support for mu4e links in org

jakecoble:
This PR makes `mu4e:` links work in org-mode.

There's one remaining annoyance. Quitting the mu4e header view gives you this message: “`mu4e~headers-quit-buffer: Symbol’s function definition is void: mu4e~main-view`”. I'm still figuring out why this is, but, if anybody else likes to link to emails from their todo list like I do, this PR should improve their workflow regardless.


jakecoble:
Looks like it doesn't work as advertised – it breaks `mu4e~main-view` everywhere. Working on it.

jakecoble:
Going to close this pull request and resubmit when I'm sure I've got something that works.

PR #2114, opened 2019-11-26
Fix dap-debug-template-conf.. void-variable

gagbo:
The variable changed name in
emacs-lsp/dap-mode@7768e145be56c9c6f2575ef9090b763bc9fb3730


hlissner:

PR #2122, opened 2019-11-29
Add LSP support to Kotlin

juliandemille:
`lsp-mode` has built in support for Kotlin. This makes it automatically used if you use the `+lsp` flag.

### DIFF
--- a/modules/lang/kotlin/config.el
+++ b/modules/lang/kotlin/config.el
@@ -1,5 +1,10 @@
 ;;; lang/kotlin/config.el -*- lexical-binding: t; -*-
 
+(use-package! kotlin-mode
+  :init 
+  (when (featurep! +lsp)
+    (add-hook 'kotlin-mode-local-vars-hook #'lsp!)))
+
 (after! kotlin-mode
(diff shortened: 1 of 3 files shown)


modules/lang/kotlin/config.el
```diff
@@ -1,5 +1,10 @@
 ;;; lang/kotlin/config.el -*- lexical-binding: t; -*-
 
+(use-package! kotlin-mode
+  :init 
+  (when (featurep! +lsp)
+    (add-hook 'kotlin-mode-local-vars-hook #'lsp!)))
```
hlissner: Please move these two lines inside the `after! kotlin-mode` block just below.

This `use-package!` call eager-loads kotlin-mode at startup.

hlissner:
Oh, also, please add the following to lang/kotlin/doctor.el:

```elisp
(assert! (or (not (featurep! +lsp))
             (featurep! :tools lsp))
         "This module requires (:tools lsp)")
```

hlissner:
Nice work. Thanks for the help and quick response!

juliandemille:
Thanks for your quick response too!

PR #2124, opened 2019-11-30
email/mu4e: fix refile target

dsdshcym:
- before this change: refile would use mu4e-trash-folder as the target
- after this change: refile would use mu4e-refile-folder as expected

Question: 
- Is a fancy icon needed for refile `:char`?
- What does `prompt` mean here? Do I need to change it to something else?

### DIFF
--- a/modules/email/mu4e/config.el
+++ b/modules/email/mu4e/config.el
@@ -166,9 +166,9 @@
                 :action #'+mu4e--mark-seen)
           ;; Refile will be my "archive" function.
           (alist-get 'refile mu4e-marks)
-          (list :char '("d" . "▼")
-                :prompt "dtrash"
-                :dyn-target (lambda (_target msg) (mu4e-get-trash-folder msg))
+          (list :char '("r" . "▼")
+                :prompt "rrefile"
+                :dyn-target (lambda (_target msg) (mu4e-get-refile-folder msg))
                 :action #'+mu4e--mark-seen))
 
     ;; This hook correctly modifies gmail flags on emails when they are marked.


hlissner:
It's been a long time since I've used mu4e, but refiling to the trash folder was necessary specifically for gmail, otherwise you'd get duplicates (because gmail copies the mail server-side, and mu redownloads it to your refile folder), so we'd just delete the local one and let gmail do its thing.

Is this no longer the case? I'd be surprised if google changed something so fundamental.

> Is a fancy icon needed for refile :char?

I don't know anymore. I'll let you decide.

> What does prompt mean here? Do I need to change it to something else?

It's just the message displayed to you when you use the command interactively. I'll leave it up to you.

dsdshcym:
> so we'd just delete the local one and let gmail do its thing.

That's what `(setq mu4e-sent-messages-behavior 'delete)` does, i guess?
Below is from its doc:
> Note, when using GMail/IMAP, you should set this to either
> trash or delete, since GMail already takes care of keeping
> copies in the sent folder.

Refiling to the trash folder won't delete the local one, but just move it to another folder. If gmail would copy the mail server-side, then moving an email to trash folder would still duplicate it. But from my daily usage, refiling an email to `all mail` folder won't cause any problem.

hlissner:
> That's what (setq mu4e-sent-messages-behavior 'delete) does, i guess?
> Below is from its doc:

I meant when you "archive" (refile) a message, rather than sending them.

Archiving them creates a copy in the all mail folder, so it seemed sensible to delete it from the local inbox and let mu redownload the copy on next sync.

Also, iirc, `:action #'+mu4e--mark-seen` prevents the emails from being moved at all, so these emails are only moved virtually (in Emacs), but the change isn't written anywhere (and doesn't survive a re-syncing or updating mu4e).

Can you confirm this is not the case anymore?

dsdshcym:
Yes, after applying this PR, I can refile a message to `all mail` folder without a duplicate.

PR #2125, opened 2019-12-01
lang/ruby: add featurep for rspec and minitest

luisenrike:
The ruby's testing frameworks should be optional. When both are installed rspec's keybindings take precedence over minitest's.

### DIFF
--- a/modules/lang/ruby/config.el
+++ b/modules/lang/ruby/config.el
@@ -127,6 +127,10 @@
   :init
   (when (featurep! :editor evil)
     (add-hook 'rspec-mode-hook #'evil-normalize-keymaps))
+  (advice-add 'rspec-enable-appropriate-mode :around
+              (lambda (orig-fun)
+                (when (file-exists-p (expand-file-name "spec/" (rspec-project-root)))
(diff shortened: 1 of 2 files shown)


hlissner:
I want to avoid adding more flags than necessary, especially on a per-package basis. Instead, I think the more appropriate solution would be to advise `rspec-enable-appropriate-mode` to no-op if `{project-root}/spec` doesn't exist. Can you make this change? If not, I'll look into it sometime within the next week or two.

luisenrike:
Could it be something like this?

```
(advice-add 'rspec-enable-appropriate-mode :around
            (lambda (orig-fun)
              (when (file-exists-p (expand-file-name "spec/" (rspec-project-root)))
                (funcall orig-fun))))
```

luisenrike:
> I want to avoid adding more flags than necessary, especially on a per-package basis. Instead, I think the more appropriate solution would be to advise `rspec-enable-appropriate-mode` to no-op if `{project-root}/spec` doesn't exist. Can you make this change? If not, I'll look into it sometime within the next week or two.

PR #2126, opened 2019-12-01
Add evil-quick-diff

rgrinberg:
This introduces the `g o d` text object for comparing regions. It works
like evil-exchange. Select two regions in sequence with `g o d` and an
ediff buffer of the selections will pop up.


hlissner:
Thanks for the PR! I can't say I'm a big fan of the keybind (`go` is used in various places by evil-collection) and that the package isn't lazy loaded. In spirit of the original linediff.vim plugin I'd like to give it a `:l[ine]diff` ex command instead. I'm completely swamped at the moment however, so it'll have to wait until another weekend.

rgrinberg:
Ah well, I can help out with this. I'll add an ex command and get rid of the default binding here.

rgrinberg:
@hlissner The default key binding is removed, `:evil-quick-diff` now works, and the plugin is loaded lazily.

hlissner:
Thanks again for your help!

PR #2127, opened 2019-12-01
Easier creation of named workspaces

rgrinberg:
When a user tries to switch to a named workspace that doesn't exist,
create it, instead of just failing.

This is consistent with bookmarks for example

### DIFF
--- a/modules/ui/workspaces/autoload/workspaces.el
+++ b/modules/ui/workspaces/autoload/workspaces.el
@@ -299,13 +299,10 @@ workspace, otherwise the new workspace is blank."
              (+workspace/display)))
     ((debug error) (+workspace-error (cadr e) t))))
 
-;;;###autoload
-(defun +workspace/switch-to (index)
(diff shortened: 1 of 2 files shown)


modules/ui/workspaces/autoload/workspaces.el
```diff
@@ -319,7 +319,7 @@ end of the workspace list."
                  (+workspace-switch dest)))
               ((stringp index)
                (unless (member index names)
-                 (error "No workspace named %s" index))
+                 (+workspace/new index))
                (+workspace-switch index))
```
hlissner: That seems sensible. Could you do away with the whole `unless` form and instead pass `'auto-create` as `+workspace-switch`'s second argument? This will do the same thing more concisely.

rgrinberg: done

hlissner:
Thanks for the help!

rgrinberg:
A little typo sneaked in here. Will put up a fix

PR #2129, opened 2019-12-01
Add factor module

rgrinberg:
Factor is a stack based programming language with an awesome emacs mode :)

### DIFF
--- a/modules/lang/factor/README.org
+++ b/modules/lang/factor/README.org
@@ -1,0 +1,43 @@
+#+TITLE:   lang/factor
+#+DATE:    December 3, 2019
+#+SINCE:   v3.0.0
+#+STARTUP: inlineimages
+
+* Table of Contents :TOC_3:noexport:
+- [[#description][Description]]
+  - [[#module-flags][Module Flags]]
+  - [[#plugins][Plugins]]
+  - [[#hacks][Hacks]]
+- [[#prerequisites][Prerequisites]]
+- [[#configuration][Configuration]]
+- [[#troubleshooting][Troubleshooting]]
+
+* Description
+This module adds support to the [[https://github.com/factor/factor][factor]] programming language and its associated
+_fuel_ emacs plugin.
+
++ If possible, include a brief list of feature highlights here
++ Like code completion, syntax checking or available snippets
++ Include links to packages & external things where possible
+
+** Module Flags
+This module provides no flags.
+
+** Plugins
+{A list of linked plugins}
+
+** Hacks
+{A list of internal modifications to included packages}
+
+* Prerequisites
+You must install [[https://github.com/factor/factor][factor]] to use the advanced functionality of this module.
+
+* Configuration
+This module requires the installation of factor to be available at
+=fuel-factor-root-dir=. Here's an example of how to set it:
+#+BEGIN_SRC emacs-lisp
+(setq fuel-factor-root-dir "/Applications/factor")
+#+END_SRC
+
+* Troubleshooting
+Common issues and their solution, or places to look for help.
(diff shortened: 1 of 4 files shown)


modules/lang/factor/config.el
```diff
@@ -0,0 +1,51 @@
+;;; lang/factor/config.el -*- lexical-binding: t; -*-
+
+(use-package fuel-mode
```
hlissner: Replace `use-package` with `use-package!`.

hlissner: Also: `fuel-mode` is not lazy loaded. If :mode, :hook or :after aren't needed, use `:defer t`

modules/lang/factor/config.el
```diff
@@ -0,0 +1,51 @@
...
+      (:prefix ("r" . "refactor")
+        ("s" #'fuel-refactor-extract-sexp
+         "w" #'fuel-refactor-extract-region
+         "v" #'fuel-refactor-extract-vocab
+         "i" #'fuel-refactor-inline-word
+         "g" #'fuel-refactor-make-generic
+         "u" #'fuel-update-usings
+         "r" #'fuel-refactor-rename-word)))
+
+(map! :after fuel-listener-mode
```
hlissner: The package is named `fuel-listener`, rather than `fuel-listener-mode`

modules/lang/factor/config.el
```diff
@@ -0,0 +1,51 @@
...
+         "t" #'fuel-scaffold-tests))
+
+      (:prefix ("r" . "refactor")
+        ("s" #'fuel-refactor-extract-sexp
+         "w" #'fuel-refactor-extract-region
+         "v" #'fuel-refactor-extract-vocab
+         "i" #'fuel-refactor-inline-word
+         "g" #'fuel-refactor-make-generic
+         "u" #'fuel-update-usings
+         "r" #'fuel-refactor-rename-word)))
```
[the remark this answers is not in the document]
hlissner: Also, please remove the blank newline between each prefix.

rgrinberg:
Thanks for the thorough review @hlissner. I added a README, but haven't filled out everything yet.

Do we need a doctor script to debug a missing installation of factor?

hlissner:
> Do we need a doctor script to debug a missing installation of factor?

If that can be easily determined from elisp (e.g. with a `executable-find` check), definitely.

modules/ui/workspaces/autoload/workspaces.el
```diff
@@ -340,7 +340,7 @@ end of the workspace list."
   (interactive
    (list (or current-prefix-arg
              (completing-read "Switch to workspace: " (+workspace-list-names)))))
-  (+workspace/switch-to index))
+  (+workspace--switch-to index))
```
hlissner: Whoops. Looks like this snuck into the PR. Please remove it.

modules/lang/factor/README.org
```diff
@@ -0,0 +1,40 @@
+#+TITLE: :lang factor
```
hlissner: Please add the full module readme header:

```org
#+TITLE:   lang/factor
#+DATE:    December 3, 2019
#+SINCE:   v3.0.0
#+STARTUP: inlineimages
```

rgrinberg: done

rgrinberg:
> If that can be easily determined from elisp (e.g. with a executable-find check), definitely.

Hmm, it's a bit hard to do with `executable-find`. The issue is that factor does not need to be in the path (and it's not for me).

hlissner:
> Hmm, it's a bit hard to do with executable-find. The issue is that factor does not need to be in the path (and it's not for me).

In that case I don't think it's necessary. That said, thanks again for the help!

PR #2130, opened 2019-12-01
Fix +workspace/switch-to

rgrinberg:
This function was calling itself infinitely instead of the helper.


hlissner:
I decided to merge the two into one command, which indirectly fixes this issue. Thanks for bringing it to my attention!

PR #2133, opened 2019-12-02
Fix typos

bwignall:
Should be non-semantic.

Uses https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings/For_machines to find likely typos.

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -78,7 +78,7 @@
        ;;ein               ; tame Jupyter notebooks with emacs
        (eval +overlay)     ; run code, run (also, repls)
        flycheck          ; tasing you for every semicolon you forget
-       ;;flyspell          ; tasing you for misspelling mispelling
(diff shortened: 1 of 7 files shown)


init.example.el
```diff
@@ -78,7 +78,7 @@
        ;;ein               ; tame Jupyter notebooks with emacs
        (eval +overlay)     ; run code, run (also, repls)
        flycheck          ; tasing you for every semicolon you forget
-       ;;flyspell          ; tasing you for misspelling mispelling
+       ;;flyspell          ; tasing you for misspelling misspelling
```
hlissner: This was an intentional misspelling.

bwignall: Sorry! Clearly I missed this joke, including "tasing".

I have reverted this.

hlissner: No problem! Thanks for the help!

PR #2143, opened 2019-12-03
Popup rules for more builtin types

rgrinberg:
process, timer, abbreviations, output, occur buffers are all better
displayed in a maximized fashion. Otherwise, they aren't really useful.


rgrinberg:
1. I followed your example for unsent mail where you didn’t escape it. But will fix.
2. Fixed
3. Sure
4. list-registers

hlissner:

PR #2145, opened 2019-12-04
ui/popup: respect :select nil when reusing a window

dsdshcym:
- before this change: when reusing a window, select option would get ignored
- after this change: when reusing a window, select option would be used
  just as a new popup


hlissner:
Thanks for the help!

PR #2158, opened 2019-12-06
Fix interactive call of doom/narrow-buffer-indirectly

yuhan0:
The optional 3rd arg to the function was removed in f9190c08bfb908df1df3cc49ee2ecec4b360d42c


hlissner:

Issue #2160, opened 2019-12-06
[REQUEST] LSP-backed jump-to-random-symbol


gagbo:
# Completion for LSP-backed "jump to symbol"

Since we're using [lsp-mode](https://github.com/emacs-lsp/lsp-mode) for the `:tools lsp` module, we could use the packages they provides for an interface with ivy/helm :
- [:completion ivy (lsp-ivy)](https://github.com/emacs-lsp/lsp-ivy)
- [:completion helm (helm-lsp)](https://github.com/emacs-lsp/helm-lsp)

The goal would be to map these commands to something like `SPC c {j/J}` (for code jump ?)

robbert-vdh:
There's also another package for [treemacs integration](https://github.com/emacs-lsp/lsp-treemacs).

PR #2162, opened 2019-12-06
Add lsp-ivy and helm-lsp when :tools lsp

gagbo:
As described in #2160 

- Not sure if my `:init` block is correct (namely the `after lsp`)
- Not sure if using `doom-leader-code-map` is the way to go
- I think all those settings belong to the completion modules though

### DIFF
--- a/modules/completion/helm/config.el
+++ b/modules/completion/helm/config.el
@@ -183,3 +183,13 @@ be negative.")
         (lambda (buf &optional _resume) (pop-to-buffer buf)))
   (global-set-key [remap swiper] #'swiper-helm)
   (add-to-list 'swiper-font-lock-exclude #'+doom-dashboard-mode nil #'eq))
+
+(use-package! helm-lsp
+  :when (featurep! :tools lsp)
+  :commands (helm-lsp-workspace-symbol helm-lsp-global-workspace-symbol)
+  :init
+  (after! lsp
+    (map!
+     :map doom-leader-code-map
+     :desc "Jump to symbol in current workspace" "j" #'helm-lsp-workspace-symbol
+     :desc "Jump to symbol in any workspace"     "J" #'helm-lsp-global-workspace-symbol)))
--- a/modules/completion/helm/config.el
+++ b/modules/completion/helm/config.el
@@ -183,13 +183,3 @@ be negative.")
         (lambda (buf &optional _resume) (pop-to-buffer buf)))
   (global-set-key [remap swiper] #'swiper-helm)
   (add-to-list 'swiper-font-lock-exclude #'+doom-dashboard-mode nil #'eq))
-
-(use-package! helm-lsp
-  :when (featurep! :tools lsp)
-  :commands (helm-lsp-workspace-symbol helm-lsp-global-workspace-symbol)
-  :init
-  (after! lsp
-    (map!
-     :map doom-leader-code-map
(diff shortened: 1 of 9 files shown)


modules/completion/helm/config.el
```diff
@@ -182,3 +182,13 @@ be negative.")
         (lambda (buf &optional _resume) (pop-to-buffer buf)))
   (global-set-key [remap swiper] #'swiper-helm)
   (add-to-list 'swiper-font-lock-exclude #'+doom-dashboard-mode nil #'eq))
+
+(use-package! helm-lsp
+  :when (featurep! :tools lsp)
+  :commands (helm-lsp-workspace-symbol helm-lsp-global-workspace-symbol)
```
hlissner: Remove the parentheses around the :commands list. I omit them when the list is short.

modules/completion/ivy/config.el
```diff
@@ -368,6 +368,15 @@ evil-ex-specific constructs, so we disable it solely in evil-ex."
   (setq prescient-save-file (concat doom-cache-dir "prescient-save.el"))
   (prescient-persist-mode +1))
 
+(use-package! lsp-ivy
+  :when (featurep! :tools lsp)
+  :commands (lsp-ivy-workspace-symbol lsp-ivy-global-workspace-symbol)
```
hlissner: Same as above.

modules/completion/helm/config.el
gagbo: Will do

modules/completion/ivy/config.el
gagbo: Same as above

gagbo:
>     1. Please move these ivy/helm packages to the `:tools lsp` module. In general, ivy/helm plugins should be added to the module they extend.

Ok, so I just check for `(featurep! :completion xxxx)` in `:tools lsp` module ?

>     2. In general, do not modify leader keybinds from modules. Modify them from `modules/config/default/+{evil,emacs}-bindings.el`.

Yes, when I pushed it, I remember all the `:when` in `map!`
 
>     3. This is a lot like imenu on `SPC s i`. Perhaps it should be on `SPC s {j,J}` instead, or perhaps remap `SPC s i` when lsp-mode is enabled?

I didn't know about `SPC s i`. I thought about `SPC c` prefix because a few LSP related commands (rename and format iirc) are already there, and "**s**earch **j**ump" feels less mnemonic than "**c**ode **j**ump"

hlissner:
> Ok, so I just check for (featurep! :completion xxxx) in :tools lsp module ?

Yes.

> I didn't know about SPC s i. I thought about SPC c prefix because a few LSP related commands (rename and format iirc) are already there, and "search jump" feels less mnemonic than "code jump"

True. I was thinking of `SPC s s` for search symbols, but it's already taken by swiper. Perhaps I'll replace `SPC s b` with swiper-isearch so that this could take `SPC s s`?

gagbo:
>     3. This is a lot like imenu on `SPC s i`. Perhaps it should be on `SPC s {j,J}` instead, or perhaps remap `SPC s i` when lsp-mode is enabled?

One (big) difference between the 2 is that imenu works in the current buffer, whereas those lsp completions work on all symbols in the workspace

gagbo:
`doom-search-map j` is already taken to search the jumplist, so I think `c j/J` is better at least for now

hlissner:
Looks good to me. Thanks again for the help!

PR #2164, opened 2019-12-07
Add ibuffer filter group for the workspace

rgrinberg:
This introduces the +ibuffer-current-workspace function which open an
ibuffer with only buffers from the current workspace.

This proof of concept already works and is proving to be quite useful for me, but there are still some imperfections:

* The workspace filter is displayed in an ugly way in the ibuffer window. (it's pretty printed as an emacs lisp struct)

* We should be using something like `ivy-read` to read the workspace in `:reader`.


rgrinberg:
@hlissner friendly ping

hlissner:

PR #2168, opened 2019-12-08
add common paths to be ignored in most js projects

chrunchyjesus:


hlissner:

PR #2123, opened 2019-11-30
email/mu4e: add featurep check for workspaces

dsdshcym:
- if module workspaces is not used, an error would be raised when calling `=mu4e`
  `=mu4e: Symbol’s function definition is void: +workspace-switch`
- if module workspaces is not used, an error would be raised when quiting mu4e
  `=mu4e: Symbol’s function definition is void: +workspace/delete`

This PR fixes these two issues.

I think there might be a better fix:
Like extracting these workspaces tweeks to a new file and adding a `;;;###if (featurep! :ui workspaces)` check at the beginning?

PR #2170, opened 2019-12-08
Use featurep! in mu4e/autoload/email.el

alexandergraul:
`featurep` works a bit different from` featurep!`, which is required in this
case.

Fixes problem reported in the PR that introduced `featurep` in [this comment](https://github.com/hlissner/doom-emacs/pull/2123#issuecomment-563003555)


hlissner:
Thanks for fixing this!

PR #2172, opened 2019-12-09
mu4e is only available from the AUR

AxelBohm:
mu4e is not available from the main repos in Arch Linux, so I fixed the documentation


hlissner:

PR #2173, opened 2019-12-09
Changed to correct default location

adbrown101:
https://git-scm.com/docs/git-credential-cache

I believe that the default location for git-credential-cache should be
$XDG_CACHE_HOME/git/credential/socket

Which is where I found it running git from the command line.


hlissner:

PR #2176, opened 2019-12-09
vterm: adapt to new function signature

seanfarley:
emacs-libvterm@376db7cf416 had a breaking change for
`vterm-exit-functions` which sends the event as the second argument.


hlissner:

PR #2180, opened 2019-12-11
Add py-isort feature to the python module

rakanalh:
This PR introduces a new feature to the python module for `isort`. 

`isort` is a python tool which sorts imports which is actively maintained and used extensively by the python community. Emacs integration for this tool is done using the py-isort emacs package available on melpa.

The only thing that is missing from this PR is registering the hook to sort on save but i think this can be left to user-config rather than having it done automatically by doom.

### DIFF
--- a/modules/lang/python/packages.el
+++ b/modules/lang/python/packages.el
@@ -31,7 +31,4 @@
 
 ;; Import managements
 (package! pyimport)
-(package! pyimpsort)
-
-(when (featurep! +isort)
-  (package! py-isort))
+(package! py-isort)
--- a/modules/lang/python/packages.el
+++ b/modules/lang/python/packages.el
@@ -32,3 +32,6 @@
 ;; Import managements
 (package! pyimport)
 (package! pyimpsort)
+
+(when (featurep! +isort)
+  (package! py-isort))
--- a/modules/lang/python/README.org
+++ b/modules/lang/python/README.org
@@ -31,7 +31,7 @@ Adds Python support to Doom Emacs.
 ** Plugins
 + [[https://github.com/pythonic-emacs/anaconda-mode][anaconda-mode]]*
 + [[https://github.com/Wilfred/pyimport][pyimport]]*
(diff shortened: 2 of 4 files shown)


modules/lang/python/packages.el
```diff
@@ -32,3 +32,6 @@
 ;; Import managements
 (package! pyimport)
 (package! pyimpsort)
```
rakanalh: Should this be ignored if the user has `+isort` added to the python module?

hlissner:
How is this different from py-impsort? Can py-isort replace it? And if so, why one over the other?

rakanalh:
I wouldn't know about py-impsort if i wasn't using doom-emacs. isort is a mainstream tool that python developers use in their IDEs as well as on CI. As stated previously, it's also more actively maintained and more sophisticated when it comes to customizability when compared to py-impsort.

I would say isort can very easily replace py-impsort but i didn't want to do that in this PR unless you agree.

hlissner:
I see. I suppose the allure of py-impsort is it works out of the box; i.e. has no additional dependencies, but it certainly doesn't hold a candle to isort. Then the decision is to either keep both and fall back to py-impsort in isort's absence, or just go full-in on isort. Give me a couple days to sit on that.

hlissner:
I've decided to go the isort-exclusive route. Please go ahead and remove the pyimpsort package and [this stray keybind](https://github.com/hlissner/doom-emacs/blob/develop/modules/lang/python/config.el#L142).

Could you also add install instructions for isort in the lang/python module's README.org (under "Prerequisites")?

rakanalh:
Sounds great. Will do.

rakanalh:
@hlissner all done.

hlissner:
Looks good to me! Thanks for the prompt response and for your help!

danlamanna:
Should a check for `isort` be added to the doctor for the python module? I can take a stab at that in a couple days if it makes sense.

hlissner:
@danlamanna Please do; it'd be much appreciated!

PR #2190, opened 2019-12-13
YT Rwnd '18 is still the most disliked YT Rwnd Yet

avronr:
Mentioned that YT Rewind '18 is still the most disliked YT Rewind Yet. I really don't think it going to get any better any time soon.
I'd really like to see a [Peertube](https://joinpeertube.org/) instance which could afford not to put data caps on creators.


hlissner:
Haha, good to know that it hasn't changed, but I cannot accept this PR. The YT Rewind '18 reference is only an analogy, and is secondary to the point that is being made in this paragraph (that there are many ways to set up your environment). This parenthetical doesn't add value to that point.

Thanks for the PR, in any case!

PR #2196, opened 2019-12-16
Fix regex for detecting compilation buffer

rgrinberg:
*Compilation* and *compilation* are both allowed


hlissner:
Makes sense to me, but please rebase this onto `develop`.

rgrinberg:
Oh yeah. Fixed

hlissner:
Thanks again!

PR #2199, opened 2019-12-16
fix doc for enabling rust analyzer

chrunchyjesus:
Using the previous config did not change the language server to rust analyzer for me. I think it is because setting it after lsp-rust has been initialized is too late.


hlissner:
Thanks again for the help!

PR #2205, opened 2019-12-17
Update plantuml doctor prescription

niklascarlsson:
This patch changes the recommendation from +plantuml/install to
plantuml-dowload-jar, when the plantuml.jar can't be found. The command
+plantuml/install has been removed in a previous patch.

----

# 

When running bin/doom doctor and plantuml.jar is not found the recommendation is
to use the +plantuml/install command. But this command has been removed some
time ago. Instead it works to run the commnand plantuml-dowload-jar. This commit
updates the subscription by the doctor.

### DIFF
--- a/modules/lang/plantuml/doctor.el
+++ b/modules/lang/plantuml/doctor.el
@@ -7,4 +7,4 @@
     (warn! "Couldn't find java. PlantUML preview or syntax checking won't work"))
   ;; plantuml.jar
   (unless (file-exists-p plantuml-jar-path)
-    (warn! "Couldn't find plantuml.jar. Install it with-x +plantuml/install")))
+    (warn! "Couldn't find plantuml.jar. Install it with M-x plantuml-download-jar")))


modules/lang/plantuml/doctor.el
```diff
@@ -7,4 +7,4 @@
     (warn! "Couldn't find java. PlantUML preview or syntax checking won't work"))
   ;; plantuml.jar
   (unless (file-exists-p plantuml-jar-path)
-    (warn! "Couldn't find plantuml.jar. Install it with-x +plantuml/install")))
+    (warn! "Couldn't find plantuml.jar. Install it with-x plantuml-download-jar")))
```
hlissner: Please change `with-x plantuml-download-jar` to `with M-x plantuml-download-jar`

niklascarlsson: Oops! I thought I created a new fork but I must have accidentally used an old one :O. I will try to fix this branch or create a new pull request.

hlissner:
Looks good to me! Thanks for the help and the quick response!

PR #2208, opened 2019-12-17
Fix #2177

### LINKED ISSUE #2177: In dired copy multiple files to another directory which wrongly popup a fresh buffer to dispaly the fies
Working in dired to copy a singular file to another directory, it works properly to prompt in the mini-buffer of the path guide.

However, if %m to marked multiple files then perform Copy them to another directory, the doom-emacs displayed all the marked files in a freshly opened buffer, rather than prompt a path in the mini-buffer.

What's the problem, it might be?

RBckmnn:
Hi Henrik,

this PR addresses [2177](https://github.com/hlissner/doom-emacs/issues/2177).

The code to select a target directory for a copy/move operation was not executed because of an error. That error was due to `+popup--delete-other-windows` expecting `+popup/raise` to return a window, whereas it returned a buffer instead.

### DIFF
--- a/modules/ui/popup/autoload/popup.el
+++ b/modules/ui/popup/autoload/popup.el
@@ -415,8 +415,9 @@ the message buffer in a popup window."
 
 ;;;###autoload
 (defun +popup/raise (window &optional arg)
-  "Raise the current popup window into a regular window.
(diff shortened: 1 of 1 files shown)


hlissner:
Great catch! Thanks for looking into this!

PR #2209, opened 2019-12-18
fix: gruvbox theme with solaire-mode

ragone:
I believe that `doom-gruvbox` supports solaire, which means that the background colors are swapped. This PR adds the `doom-gruvbox` theme to the list of supported solaire themes.


hlissner:

PR #2212, opened 2019-12-19
Fix comment about dired-hide-details-hide-symlink-targets

naiua:
Just a small fix for comment.


hlissner:
I realized that more work was needed to revise the comments in that area, so I did them in 995031f. I also didn't think that `dired-hide-details-hide-symlink-targets` really needed a comment for it; the variable is self-explanatory. In any case, thanks for the PR and for bringing it to my attention!

Issue #2213, opened 2019-12-19
(wrong-number-of-arguments (3 . 3) 2) error when installing or updating haskell


qcfu-bu:
I'm trying to install the Haskell mode through doom.d init and I keep running into two consecutive lines of

`Eager macro-expansion failure: (wrong-number-of-arguments (3 . 3) 2)`
`Eager macro-expansion failure: (wrong-number-of-arguments (3 . 3) 2)`

when doom refresh attempts Installing & building packages. Quoting out haskell and refresh works normally.

The full debug message is quite long, not sure if posting that here will be of much help.

jdnavarro:
Upstream issue: https://github.com/haskell/haskell-mode/issues/1691

qcfu-bu:
Thank you, I'll close the issue

hlissner:
I'll keep this open until it is resolved upstream.

tchajed:
...and it's fixed upstream: https://github.com/haskell/haskell-mode/commit/38c512b8859c02fe13e11c558ef9feefee1674b4.

hlissner:
Thank goodness. @tchajed Thanks for letting us know!

asymmetric:
How do we recover from this error?

tchajed:
You just need the latest version of the haskell-mode package. `doom update` should work for that.

asymmetric:
`doom` update with the `haskell` module hits the error. If I remove the module, the package doesn't get updated.

tchajed:
Then maybe manually do a `git pull` in `~/.emacs.d/.local/straight/repos/haskell-mode` followed by `doom refresh`?

You should also be able to remove `haskell`, refresh (which will delete the repo), and then add it back (which will clone it at the latest release), but that's a bit more complicated.

asymmetric:
Thanks, `doom refresh -p` did the trick - a simple `refresh` doesn't purge orphaned repos.

tchajed:
Ah, I didn't realize that (perhaps a change since I last needed to purge a repo). Glad you sorted that out!

So to summarize, the cleanest way to recover from a broken upstream package is to remove the relevant module from your config, delete the clone with `doom refresh -p`, then add it back to your config and run `doom refresh` to install it.

hlissner:
@asymmetric That, or delete these two directories manually: `~/.emacs.d/.local/straight/{build,repos}/haskell-mode`

PR #2226, opened 2019-12-22
docs/faq: add broken upstream pkg to common issues

asymmetric:
See #2213.


hlissner:
Thanks for your help!

PR #2230, opened 2019-12-23
irc: turn off smartparens mode for circe buffers

seanfarley:
Users probably don't want ":-(" to be autocompleted as ":-()" so let's
turn of smartparens for chatting.


hlissner:
Sounds good to me. Thanks for the help (and Merry Christmas)!

seanfarley:
Thanks! (Merry Christmas, too!)

PR #2232, opened 2019-12-23
Fix deprecated ESS variables

ashiklom:
- `ess-smart-S-assign-key` and `ess-expression-offset` are deprecated variables
- `ess-default-style` has been replaced with `ess-style`

None of these are super important, but a bit of cleanup can never hurt 😄


hlissner:
Thanks for the help (and Merry Christmas)!

PR #2242, opened 2019-12-24
dap-mode: set extensions variable to doom-cache

seanfarley:
Previously, this would put the .extension directory in
`user-emacs-directory' which in doom's case will be in the root of the
doom directory.

So, we move the directory to doom's cache dir and also name it something
descriptive instead of 'extension'.

### DIFF
--- a/modules/tools/debugger/config.el
+++ b/modules/tools/debugger/config.el
@@ -23,8 +23,10 @@
 (use-package! dap-mode
   :when (featurep! :tools lsp)
   :after lsp-mode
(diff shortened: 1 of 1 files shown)


seanfarley:
Will do!

hlissner:
Thanks for the quick response! Although, it hadn't occurred to me until now: does dap use `dap-utils-expansion-path` for cache files? Or would they be better suited to `doom-etc-dir`?

seanfarley:
Ah, good point. These are files it installs via `dap-gdb-lldb-setup` so I think you're right that `etc` is a better place. I fixed up the code and commit message to reflect that.

PR #2244, opened 2019-12-25
Add dune-mode to lispy modes

rgrinberg:
Dune uses an sexpression based syntax as well


hlissner:
Sounds good to me. Thanks again!

PR #2245, opened 2019-12-25
Remove lispy from evil-collections

rgrinberg:
The bindings that it creates are weird and conflict with lispyville's
own bindings.

### DIFF
--- a/modules/editor/evil/+everywhere.el
+++ b/modules/editor/evil/+everywhere.el
@@ -35,7 +35,8 @@
     package-menu
     ruby-mode
     simple
-    slime)
+    slime
+    lispy)
   "A list of `evil-collection' modules to ignore. See the definition of this
 variable for an explanation of the defaults (in comments). See
 `evil-collection-mode-list' for a list of available options.")


JSDurand:
Adding 
```lisp
(after! (evil-collection lispy)
  (lispy-set-key-theme '(special lispy)))
```
to the .doom.d/config.el fixes the problem for me.

rgrinberg:
What problem does it fix for you?

JSDurand:
I can use the lispy keybindings again.

rgrinberg:
What does it have to do with disabling evil collections for lispy?

JSDurand:
Since you said the bindings by evil-collection are weird, I think it may help to set the theme to use the bindings not created by evil-collection.

rgrinberg:
Okay, thanks, now I see what you mean. However, I still think that this isn't optimal. It seems like evil collections for lispy essentially conflicts with lispyville. Consider bindings like `M-(` which are bound both by lispyville and evil-collection's lispy. I think that these styles are mutually exclusive and we should either:

* pick one style in the lispy module
* give users the choice using some config

Having this workaround doesn't give a good user experience out of the box.

hlissner:
The relationship between evil-collection-lispy, lispy and lispyville isn't clear to me. I'll look into it after the holidays. In the meantime you can add symbols to `+evil-collection-disabled-list` to disable their respective evil-collectio module.

```elisp
(add-to-list '+evil-collection-disabled-list 'lispy)
```

rgrinberg:
done

hlissner:
Great! Thanks again!

PR #2254, opened 2019-12-26
Set the directory for the Java LSP workspace

r-darwish:


hlissner:
Thanks again for the help!

PR #2256, opened 2019-12-26
Make C-x b work when workspaces are disabled

r-darwish:


hlissner:
The real source of this issue is that the workspace commands were bound for vanilla users whether or not the workspaces module was enabled. I've fixed this at the source in a488c30. Thanks for bringing it to my attention!

PR #2257, opened 2019-12-26
[WIP] init.example.el module docs

MarioRicalde:
@hlissner this is basically a continuation from our conversation in Discord about a more to the point documentation for the modules.

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -19,30 +19,30 @@
        ;;japanese
 
        :completion
-       company           ; the ultimate code completion backend
-       ;;helm              ; the *other* search engine for love and life
-       ;;ido               ; the other *other* search engine...
-       ivy               ; a search engine for love and life
+       company           ; completion backend
+       ;;helm              ; search engine
+       ;;ido               ; search engine
+       ivy               ; search engine
 
        :ui
-       ;;deft              ; notational velocity for Emacs
-       doom              ; what makes DOOM look the way it does
-       doom-dashboard    ; a nifty splash screen for Emacs
-       doom-quit         ; DOOM quit-message prompts when you quit Emacs
+       ;;deft              ; notational velocity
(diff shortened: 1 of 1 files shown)


hlissner:
I'm conflicted.

I get it. Levity does not a good description make. The grammar, capitalization and punctuations fixes are needed -- and the new comments are welcome -- but I struggle to accept the rest.

I'd preach the less-is-more mantra anywhere else, but here we are not simply slimming down excess, we're sacrificing personality to save on characters in what are already <50 character blurbs. I don't think it justifies the cost. Especially when documentation is a hop away (and documented within the same file). Perhaps a better approach would be clearer jokes.

Tone is important to me. The humor doesn't have to sit well with everyone, but it sets a precedent for interacting with the project (and me). The curt professionalism of other open source projects does not suit us (or me -- or Emacs, IMO). Configuring Emacs is a Sisyphean task and we're madmen to be doing it -- that mad scientist vibe is Doom's shtick.

Anyhow, I'll sit on it for a couple days and revisit this after the holidays.

MarioRicalde:
Valid points. If you could comment on them on a case per case basis, that'd help a lot. Feel free to mark whole blocks if it's too much, or whatever way that makes sense.

Let me know.

itai33:
As some rando that has been getting into doom emacs over the past week or two, I'm heavily against this pull request. I think that the humorous tone of the descriptions is actually a really good idea pragmatically speaking. Because of it I started to go over each module and check them out, because the levity in the descriptions caught my attention. I don't think these descriptions even clarify anything, they just make it less fun.

I don't know what the discord conversation was, but if it was about how these descriptions make it hard for new users to understand what the modules are about or whatever, I think that the existing documentation is about as informative as you're going to get with a single sentence, while the eye-catching humor will increase the chance of people looking in to it in the first place.

PR #2265, opened 2019-12-29
Add ivy-yasnippets

r-darwish:


hlissner:
I don't think this adds enough value over `yas-insert-snippet` to be included in Doom, so I cannot accept this PR.

I'd rather we add `C-h` support to company-yasnippet and an ivy action for `yas-insert-snippet` to preview the original snippet. I'll look into that instead.

PR #2266, opened 2019-12-29
Add LSP bindings

r-darwish:


modules/lang/java/config.el
```diff
@@ -46,3 +46,7 @@ If the depth is 2, the first two directories are removed: net.lissner.game.")
   :mode "\\.g\\(?:radle\\|roovy\\)$"
   :config
   (set-eval-handler! 'groovy-mode "groovy"))
+
+(when (featurep! +lsp)
+  (map! :map java-mode-map :leader
+        :desc "Organize imports" "l o" #'lsp-java-organize-imports))
```
hlissner: Please move these inside the `:config` blog of `lsp-java` in lang/java/+lsp.el`.

modules/lang/java/config.el
```diff
@@ -46,3 +46,7 @@ If the depth is 2, the first two directories are removed: net.lissner.game.")
   :mode "\\.g\\(?:radle\\|roovy\\)$"
   :config
   (set-eval-handler! 'groovy-mode "groovy"))
+
+(when (featurep! +lsp)
+  (map! :map java-mode-map :leader
```
hlissner: Move this keybind to `:localleader`, where buffer/mode-local keybinds belong. `:leader` is for global keys.

modules/lang/java/config.el
```diff
@@ -46,3 +46,7 @@ If the depth is 2, the first two directories are removed: net.lissner.game.")
   :mode "\\.g\\(?:radle\\|roovy\\)$"
   :config
   (set-eval-handler! 'groovy-mode "groovy"))
+
+(when (featurep! +lsp)
```
hlissner: Once this has been moved to the lsp-java use-package block, this `(when (featurep! +lsp) ...)` will no longer be necessary.

modules/tools/lsp/config.el
```diff
@@ -156,3 +156,11 @@ Also logs the resolved project root, if found."
...
   :when (featurep! :completion ivy)
   :commands lsp-ivy-workspace-symbol lsp-ivy-global-workspace-symbol)
+
+(map! :map lsp-mode-map :leader
+      :desc "Rename symbol" "l r" #'lsp-rename
+      :desc "Restart LSP workspace" "l R" #'lsp-workspace-restart
+      :desc "Shutdown LSP workspace" "l q" #'lsp-workspace-shutdown
+      :desc "Find symbol" "l s"  #'lsp-ui-find-workspace-symbol
+      :desc "Perform code action" "l c" #'lsp-ui-sideline-apply-code-actions
+      :desc "Toggle documentation window" "l d" #'lsp-ui-doc-mode)
```
hlissner: Please remove these keybinds.

A couple of these are already [under the `<leader> c` prefix](https://github.com/hlissner/doom-emacs/blob/develop/modules/config/default/%2Bevil-bindings.el#L359-L367)

Also, `:leader` keybinds do not belong in modules, they belong in `:config default` (in the `+emacs-bindings.el` or `+evil-bindings.el` files.

I will look into adding more LSP keybinds on he leader at a later time.

r-darwish: Is it possible that the binding you showed me don't bind when not using evil?
I'm not using evil and looking at `lsp-rename` documentation I see:
```
Key Bindings
This command is not in any keymaps.
```

hlissner: That is true. It would need to be added to `+emacs-bindings.el` instead then. I know little about appropriate prefixes for non-evil users however. Perhaps @UndeadKernel can suggest a prefix for LSP keybinds?

r-darwish: Good. I'll fix these comments once a prefix has been decided. I personally suggest `C-c l`

hlissner: `C-c l` is our localleader key for vanilla users, unfortunately.

r-darwish: Hmm... Maybe `C-c a` ("a" as in actions), `C-c r` ("r" as in refactor) or `C-c i` ("i" as in "intelligent" stuff)

UndeadKernel: It is true that there are no LSP bindings in the default emacs bindings. We should definitely add them. The tricky part, however, is to figure out a good spot for them.

I was just working on adding the prefix `C-c c` for "code" related things. How about setting `C-c c l` for LSP bindings?

UndeadKernel: I've submitted a PR (#2287) that includes many LSP bindings.
As soon as it is merged, I will add your bindings into the default emacs bindings.

r-darwish:
Superseded by #2287

PR #2267, opened 2019-12-29
Enable zen mode keybind only if zen module present

JingYenLoh:
Previously it would error with `command-execute: Wrong type argument: commandp, writeroom-mode`.


hlissner:
I will not be accepting PRs to feature-gate keybinds in `:config default` because:

1. It sets a precedent for feature-gating *all* commands in the module, which would:
2. Ruin the readability of the module (*so* many commands would need to be feature gated -- it defeats the purpose of having a centralized, readable keybind reference).
3. Add many `featurep!` calls, which imposes a non-trivial amount of overhead.

`commandp` errors are the lesser evil between lost performance and spaghetti code. I will one day get around to addressing this a better way, but it is low priority at the moment.

PR #2276, opened 2019-12-30
fix flyspell-correct

brettlyons:
Yesterday, flyspell-correct had a new release, version 0.6.1 which removed aliases used in modules/tool/flyspell/config.el and modules/editor/evil/config.el

  https://github.com/d12frosted/flyspell-correct/commit/7b4cf8c9ba5ac65e3bb2b62f5b72d45f4c9cf7b6

This patch updates those calls.


hlissner:
Thanks you for addressing this! The help is much appreciated.

PR #2287, opened 2020-01-01
bindings: default emacs up to the state of the art

UndeadKernel:
Added `toggle` (C-c t), `code` (C-c c), `open` (C-c o) and `insert` (C-c i) bindings.
Moved org bindings from the `C-c o` to the `C-c n` prefix.
Updated the `file` (C-c f) and `quit` (C-c q) bindings.
Removed the `terminal` bindings, now under `open` (C-c o).
Removed standalone bindings for the REPL (now under `open` (C-c o)).

@hlissner, if merged, it would be good to mark the merge as "breaking". That is, users that use the default emacs bindings would need to be aware of the changes. Maybe even a notice in the #announcements Discord channel would also be good.

### DIFF
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -25,28 +25,54 @@
       (:prefix ("l" . "<localleader>")) ; bound locally
       (:prefix ("!" . "checkers"))      ; bound by flycheck
 
+      ;;; <leader> c --- code
+      (:prefix-map ("c" . "code")
+        :desc "Compile"                               "c"   #'compile
+        :desc "Recompile"                             "C"   #'recompile
+        :desc "Jump to definition"                    "d"   #'+lookup/definition
+        :desc "Jump to references"                    "D"   #'+lookup/references
+        :desc "Evaluate buffer/region"                "e"   #'+eval/buffer-or-region
+        :desc "Evaluate & replace region"             "E"   #'+eval/region-and-replace
+        :desc "Format buffer/region"                  "f"   #'+format/region-or-buffer
+        :desc "Jump to documentation"                 "k"   #'+lookup/documentation
+        :desc "Send to repl"                          "s"   #'+eval/send-region-to-repl
+        :desc "Delete trailing whitespace"            "w"   #'delete-trailing-whitespace
+        :desc "Delete trailing newlines"              "W"   #'doom/delete-trailing-newlines
+        :desc "List errors"                           "x"   #'flymake-show-diagnostics-buffer
+        (:when (featurep! :tools flycheck)
+          :desc "List errors"                         "x"   #'flycheck-list-errors)
+        (:when (featurep! :tools lsp)
+        :desc "LSP Code actions"                      "a"   #'lsp-execute-code-action
+        :desc "LSP Format buffer/region"              "F"   #'+default/lsp-format-region-or-buffer
+        :desc "LSP Organize imports"                  "i"   #'lsp-organize-imports
+        :desc "LSP Rename"                            "r"   #'lsp-rename
+        (:when (featurep! :completion ivy)
+          :desc "Jump to symbol in current workspace" "j"   #'lsp-ivy-workspace-symbol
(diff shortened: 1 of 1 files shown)


modules/config/default/+emacs-bindings.el
```diff
@@ -25,28 +25,53 @@
...
+        :desc "Jump to definition"                    "d"   #'+lookup/definition
+        :desc "Jump to references"                    "D"   #'+lookup/references
+        :desc "Evaluate buffer/region"                "e"   #'+eval/buffer-or-region
+        :desc "Evaluate & replace region"             "E"   #'+eval/region-and-replace
+        :desc "Format buffer/region"                  "f"   #'+format/region-or-buffer
+        :desc "Jump to documentation"                 "k"   #'+lookup/documentation
+        :desc "Send to repl"                          "s"   #'+eval/send-region-to-repl
+        :desc "Delete trailing whitespace"            "w"   #'delete-trailing-whitespace
+        :desc "Delete trailing newlines"              "W"   #'doom/delete-trailing-newlines
+        :desc "List errors"                           "x"   #'flymake-show-diagnostics-buffer
```
r-darwish: Could you please add a binding to `lsp-execute-code-action` as well?

seanfarley:
I had to remap a few of my custom bindings but overall I like this. I even discovered a few new tricks! 😬

hlissner:
Great work! I'll add an announcement to the discord. Thanks for working on this!

ckruse:
This breaks with the convention <q>keep `C-c` and a letter reserved for the user</q> (<https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html>). Now my personal keybindings no longer work and I have to get used to a different prefix. :(

dive:
@UndeadKernel, hi. May I ask about the reasons behind these changes?

UndeadKernel:
@ckruse and @dive, the reason why Doom uses the prefixes `C-c <letter>` is precisely because these are the only free prefixes available to us. We also considered using something like `C-c C-<letter>` but, because Doom includes a lot of packages, almost all `C-c C-<letter>` prefixes are already in use.

This PR introduced new bindings into the prefixes `C-c <letter>`. Note than we were already using many of these prefixes before this PR.

We iterated diverse ideas regarding the default emacs bindings and came to the following conclusion: if the user keeps the "+bindings" flag of the "default" module, it's usually because the user is starting from scratch. Doom is opinionated in diverse ways, key bindings is one of such ways. However, if you are an experienced Emacs user, I concluded, then it's easier for you (in comparison to a beginner) to disable the default bindings and include your own.

I took the side of the newbie in this decision: new users that wish to use a non-evil Emacs should not be concerned about setting up hundreds of bindings to be able to experience all that Doom can offer!

Note that all this is always up to discussion. If you guys have suggestions or ideas, bring them up to the table and let's discuss them! Maybe there can be a better way.

For example, some time ago, there was a discussion that the "local-leader" binding for non-evil users was to cumbersome `C-c l`. I was thinking on moving this to `C-z` because that binding is usually more annoying that useful (and can be also triggered with `C-c C-z`. Another person then suggested using `C-z <letter>` as the Doom way to do things and `C-c <letter>` as the "local-leader". I'm still playing around with the idea to see if it would work. I have to admit, however, that I've struggled getting used to that binding scheme.

ckruse:
I'm sorry, I didn't want to be so negative. I was just a bit grumpy because the keybindings I've been using for over 10 years don't work any longer. Your reasoning is comprehensible; I didn't know about dropping `+bindings`, I will look into this option. Thanks for the pointer.

UndeadKernel:
@ckruse I didn't interpret your comment negatively. I understand that personal bindings are sacred!

I recommend that you take a look at those bindings defined in the default config so that you can adapt them to your own config. I'm sure you will discover one or two things that Doom does awesomely this way!

dive:
Thank you, @UndeadKernel, for the clarification. As @ckruse said, your reasoning is comprehensible.

PR #2292, opened 2020-01-02
Enable smartparens-mode during pp-eval-expression

howsiwei:


hlissner:

PR #2295, opened 2020-01-02
irc: fix logic for creating a workspace

seanfarley:
I think this got accidentally flipped in 436d36fcd when `unless` was changed to `when`.


hlissner:
Thanks again!

PR #2296, opened 2020-01-03
ui/doom-dashboard: De-hardcode dashboard faces

Foldex:
Doom's dashboard faces are currently hardcoded to font-lock faces set by the current theme. These changes let you customize the faces independently of the current theme without having to change the widget functions themselves.


hlissner:

PR #2301, opened 2020-01-03
Fix typo in env.el

wnikola:
I noticed a typo when I ran `doom help env`, so I fixed it in core/cli/env.el.


hlissner:
Good catch. Thanks!

Issue #2303, opened 2020-01-04
[BUG] doom --help doesn't work without an argument


rohitpaulk:
**What did you expect to happen?**

Expected help to be printed out. 

**What actually happened?**

```sh
$ doom --help
--------------------------------------------------

There was an unexpected error:
  Wrong type argument (wrong-type-argument)
  stringp
  nil

Run the command again with the -d (or --debug) option to enable debug
mode and, hopefully, generate a stack trace. If you decide to file a bug
report, please include it!

Emacs outputs to standard error, so you’ll need to redirect stderr to
stdout to pipe this to a file or clipboard!

  e.g. doom -d install 2>&1 | clipboard-program

Error in Doom Emacs core: wrong-type-argument, stringp, nil
```

**Steps to reproduce:**

1. Run `doom --help`

(Running `doom --help <command>` works just fine)

PR #2304, opened 2020-01-04
cli: Handle --help without an argument

rohitpaulk:
Fixes https://github.com/hlissner/doom-emacs/issues/2303.


hlissner:
Makes sense to me. Thanks!

PR #2307, opened 2020-01-04
editor/evil: fix: outdated log

bennyyip:
These commands were deleted [here](https://github.com/hlissner/doom-emacs/commit/263c82def11ec7cd90f950b086b2c4e32e1fce1e#diff-8a6b6a4d5e97dc53317340b597c5352eL63), but not all docs have been updated.


hlissner:

PR #2309, opened 2020-01-05
Suppress delete-window in org related functions

dsdshcym:
- If there's only one other window, delete-window would be called,
  instead of delete-other-windows
- when there is only one window other than the popup, this other window
  is the main window in this frame, delete it would raise an error:
  `delete-window: Attempt to delete main window of frame <frame name>`


hlissner:
Looks good to me. I'll give it a spin. Thanks for looking into this!

PR #2312, opened 2020-01-05
Fix name of solarized package

laMudri:
The example of using a third party theme previously didn't work, because there is no package `solarized`. Instead, `solarized-theme` provides the theme.


hlissner:
Good catch! Thanks for the help.

PR #2318, opened 2020-01-06
Use newer cider-ns-refresh in Clojure bindings

jcf:
Cider deprecated `cider-refresh` in favour of `cider-ns-refresh` in version 0.18.

https://github.com/clojure-emacs/cider/blob/master/CHANGELOG.md#changes-6


hlissner:
Thanks for the help, and for staying on top of these changes!

jcf:
@hlissner no thanks necessary! Thank you so much for Doom!!

PR #2320, opened 2020-01-07
evil: use featurep! on word-wrap/zen bindings

Foldex:
Just a quick fix for those who don't have those modules enabled.


hlissner:
This is a duplicate of #2267.

PR #2324, opened 2020-01-07
Add lsp support to sh module.

dinkonin:

### DIFF
--- a/modules/lang/sh/README.org
+++ b/modules/lang/sh/README.org
@@ -9,6 +9,7 @@ This module adds support for shell scripting languages.
 * Table of Contents :TOC:
 - [[#install][Install]]
   - [[#dependencies][Dependencies]]
+  - [[#module-flags][Module Flags]]
 
 * Install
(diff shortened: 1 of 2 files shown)


gagbo:
Can you add the dependencies for the server too in the README ? (thanks for adding that by the way !)

I don't know which servers are replying to requests for shell files

dinkonin:
I need a little clarification about this change request, what exactly should I add to README. In the original PR I've already added `bash-language-server` to the list of of dependencies. Should I add a link to the `bash-language-server` repo?

hlissner:
Ah, I've just realized that lang/sh's README hasn't been updated to the new format. In that case I'll accept the PR as is and reformat it myself later.

I'll merge once I get home later tonight.

gagbo:
>  I've already added `bash-language-server` to the list of of dependencies.

My bad, I definitely missed the line since I was expecting a hefty org link in the change

hlissner:
Thanks for your help!

PR #2329, opened 2020-01-07
fix: use org-capture-projects-file

Emiller88:
Fixes centralized org-capture for projects


hlissner:

PR #2344, opened 2020-01-09
Make sudo-this-file usable in dired-buffers

RBckmnn:
Hi Henrik,

this PR allows doom/sudo-this-file to be called from dired-buffers, opening the directory as root. Previously such invocations would lead to an error, since buffer-file-name is nil for dired-buffers.


hlissner:
Looks good to me. Thanks again for the contribution!

PR #2345, opened 2020-01-09
Use composite to call harfbuzz for ligatures

gagbo:
# Use Harfbuzz to natively display ligatures
--

This is based on code from microsoft/cascadia-code#153, which sets a
proper composition-function-table according to specific ligature regexp.

Basically, for a list of characters, we check a regex pattern to see if we need to ask harfbuzz (through `font-shape-gstring`) to re-render the whole pattern. The current pattern list is taken from this `Cascadia Code` issue, so it might not be exhaustive and will need to be curated.

**This is a bit experimental** I did not thoroughly test a big number of setups, this is mostly to enable early ~~27~~ 28 adopters to test the module before emacs ~~27.1~~ 28.1 comes out so Doom's ready.

I expect that having a list with too much characters/patterns is not going to cause bugs (if a font has no ligature for a given pattern it'll just render the characters as is), but it *might* have an impact on performance.

The code has been tested on Emacs 27.0.50 (commit hash d57bb0c323) and FiraCode (from bionic packages) by me, and by a few other on Discord. This code has also been tested on a flurry of commits of emacs source code, until I found the commit that prevents this PR from freezing emacs in an infinite loop (See [Update](#Update)) 

## Compatibility with "old" pretty-code

Since : 
-  the `+font` files aren't loaded because of that `cond`,
- and those `+font` files only handle "programming" ligatures and not pretty symbols

It should merge the nicer ligatures and the replacement symbols

## Feature-gating

The only place I could find information about having built with harfbuzz is in `system-configuration-features`, hence `(string-match-p "HARFBUZZ" system-configuration-features)`.
But the docstring of the variable states it's not a really good idea. I have no other idea though (and there is a `--with-harbuzz` configuration flag so that'd be the only thing to do imo)

Also, currently this is only working with "emacs 28" (a bug in emacs-27 branch doesn't exist on `master`, see [Update](#Update))

## Major-mode-gating

The list of enabled major-modes is hardcoded in the module for now. Finding a cleaner way to do it would be nice. Since the config.el file is loaded early, I suppose we need to define a new hook which will check the list each time. Or we can enable them everywhere (including `text-mode`). Your call.

# Testing status

## Linux tests

It "just works" when the font has the expected ligatures (like Fira Code). Currently I am using `Hack` font, and this patch tries to compose `>=` and `~>` for example into weird characters (I need to setup a screenshot tool on my testing setup properly before I can show it).

So the "what happens to *other* otf fonts ?" question needs to be monitored.

## MacOS tests

Did not test (and we would need to tests for each flavor or emacs on mac too :( )

## Windows tests

Did not test (I do not want to build emacs from source on that OS)

# Update

Users (including myself) reported freezes when trying this code out in emacs sessions. This is a bug which https://github.com/emacs-mirror/emacs/commit/fe903c5ab7354b97f80ecf1b01ca3ff1027be446 fixes.

Currently this commit is on `master` branch but not `emacs-27` branch, meaning that emacs-27 users should not use the code this PR provides.

The summary of the emacs 27 specific bug is under [bug #40864](https://lists.gnu.org/archive/html/bug-gnu-emacs/2020-04/msg01121.html)


gagbo:
Pics or it didn't happen.

Screenshot_20200109_105425

gagbo:
It'll need a doctor check too, to explain that harfbuzz builds are needed to get ligatures working this way.

ocharles:
Could you clarify the benefit of this? It looks amazing - is the goal to allow people to use stock fonts, or do we still have to use custom fonts?

gagbo:
> Could you clarify the benefit of this? It looks amazing - is the goal to allow people to use stock fonts, or do we still have to use custom fonts?

Neither custom font nor `+font` feature flag should be needed after this is merged.

### Currently
Getting ligatures to render is a font-by-font hack, where we register a replacement table for each group of characters needing ligatures, and mapping them to a character of the "symbol font".
So you need to patch the private zone of the font and making a prettify-symbols replacement table like done in `+fira` `+iosevka` `+hasklig` etc.

### After
If you get a font with ligatures (thinking all of nerd fonts, Iosevka, Pragmata....), you'll get them rendered properly automagically.
We just match common ligature patterns in regex (`!=` `==`...) and ask emacs to ask harfbuzz to ask the font to render it.

### tl;dr
That means if you use a new ligatured font with specific stylistic sets that we don't know about, it'll still work as you expect.
And less maintenance to do (like the patched iosevka font @ar1a gently gave)

ocharles:
Very nice!

gagbo:
Note : it got reported on Discord that `?*` ligatures conflict in org-mode when using level-3 headings

gagbo:
I strongly suspect that `org-bullets` is incompatible with this PR because of its calls to `compose-region` into `font-lock-flush`. But unsure about how these interact together.

gagbo:
For the time being, in *my* configuration of doom, the maybe incompatible packages seem to be :
```
$ rg -uuul -g \*el font-lock-flush | xargs rg -l compose-region                                                                                                             
.local/straight/repos/nix-mode/nix-prettify-mode.el
.local/straight/repos/org-bullets/org-bullets.el
.local/straight/repos/markdown-mode/markdown-mode.el
```

gagbo:
I thinks it's safe to put this on hold. After [asking emacs-devel](https://lists.gnu.org/archive/html/emacs-devel/2020-01/msg00626.html), Eli Zaretski answered :

> More importantly, please don't use prettify-symbols-mode, which are
based on static compositions, together with automatic compositions.
Static compositions are an obsolete feature, it lacks support for some
modern Emacs features (e.g., bidirectional text), and we should remove
it from Emacs at some future point -- but not before we implement a
replacement for it using composition-function-table and related
machinery.  Mixing these two incompatible compositions is asking for
trouble.  If you turn off prettify-symbols-mode, does the problem go
away?

gagbo:
Investigated a little more, here's progress (everything in Commentary section).

Looks like I found a small package incompatible with these changes (`eros-mode`), I'll try to repro / continue investigation later.
```lisp
;;; ligature-bug.el --- Reproduce an autocomposition bug -*- lexical-binding: t; -*-
;;
;; Copyright (C) 2020 Gerry Agbobada
;;
;; Author: Gerry Agbobada <http://github/gagbo>
;; Maintainer: Gerry Agbobada <gagbobada+git@gmail.com>
;; Created: April 14, 2020
;; Modified: April 14, 2020
;; Package-Requires: ((emacs 27.0.90) (cl-lib "0.5"))
;;
;; This file is not part of GNU Emacs.
;;
;;; Commentary:
;;
;;  Reproducer for the composition table bug
;;
;;;; Method
;;
;;  Hit C-x C-e on the Fira line (change it for any font, whether you have ligatures or not)
;;
;;;; Vocabulary
;;  - Bugs = Emacs is stuck in an infinite loop, showing only
;;    "Attempt to shape unibyte text" in the minibuffer when sent a SIGUSR2 from outside
;;  - Does not bug = Emacs changes the font, I can evaluate the rest of the buffer.
;;    Evaluating the rest of the buffer, I put in parens where *Ligatures* or *No Ligatures*
;;    are appearing
;;
;;;; Current state :
;;  - Does not bug with Vanilla Emacs (Ligatures)
;;  - Does not bug with Vanilla Doom (Ligatures)
;;  - Does not bug without pretty-code and with eval (Ligatures)
;;  - Does not bug without pretty-code and with (eval +overlay) (Ligatures)
;;  - Does not bug with pretty-code and without eval
;;    (Ligatures appear with evaling only the set-frame-font line)
;;  - Does not bug with pretty-code and with eval
;;    (Ligatures appear with evaling only the set-frame-font line)
;;  - Bugs with pretty-code and with eval +overlay
;;
;;;; Commit information :
;; commit e1f0e0892232221e6333b24788b97942c83ec738 (HEAD -> emacs-27, origin/emacs-27)
;; Author: Eli Zaretskii <eliz@gnu.org>
;; Date:   Wed Mar 25 21:54:21 2020 +0200
;;
;; * lisp/files.el (directory-files-recursively): Doc fix.  (Bug#40202)
;;
;;
;;; Code:

(set-frame-font "Fira Code 11" nil t)

(defvar other-composition-ligature-table (make-char-table nil))
(require 'composite)

(let ((alist
       '(
         (?! . "!\\(?:\\(==\\|[!=]\\)[!=]?\\)") ; Matches !== != !! !!= ! and a few others
         (?= . "=\\(?:\\(!=\\|[=]\\)[=]?\\)") ; Matches =!= =!== =!=== == === and a few others
         )))
  (dolist (char-regexp alist)
    (set-char-table-range other-composition-ligature-table (car char-regexp)
                          `([,(cdr char-regexp) 0 font-shape-gstring]))))

(set-char-table-parent other-composition-ligature-table composition-function-table)

(setq-local composition-function-table other-composition-ligature-table)

;; !===
;; ===!
;; !=
;; =!=====

(provide 'ligature-bug)
;;; ligature-bug.el ends here

```

gagbo:
Vanilla emacs with `eros.el` loaded and `eros-mode` enabled goes in the `Does not bug (Ligatures)` category though....

gagbo:
Well, in the sandbox I can make all modules coexist in all versions and get the `Does not bug (Ligatures appear with evaling only the set-frame-font line)` result.

So I suppose it's again something in my private config that triggered the bug. I need to find out what though, because I wasn't alone having those freezes, and it's important to have an answer for those incoming bugs

gagbo:
Found the issue, it's overlays that seems not working after all. https://gist.github.com/gagbo/13239f12cfd6b90cce32b1c255f2553d

I've put it on emacs-devel, I'll see how I should handle that

gagbo:
The composition bug is fixed by https://github.com/emacs-mirror/emacs/commit/fe903c5ab7354b97f80ecf1b01ca3ff1027be446#diff-9114056b032516fa2de10ffaaa8e8ea7

This is safe to merge *if emacs has been compiled with this patch in*

hlissner:

PR #2347, opened 2020-01-10
Fix flycheck migration void-function

gagbo:
A bunch of `(void-function +flycheck-init-popups-h)`


hlissner:
Thanks for the help!

PR #2351, opened 2020-01-10
Correct behaviour of normal mode 'o' and 'O' in haskell-mode

nam3less:
Overwrite the 'o' and 'O' normal mode commands to correctly indent
new lines.
Code was adapted from: https://github.com/haskell/haskell-mode/issues/1265

This adresses an issue where new lines created by 'o' and 'O' where not correctly indented.


hlissner:
Sorry for the late response. Is this still an issue? If so I will merge this tonight.

hlissner:
I will assume that is the case. Thanks for the PR!

PR #2353, opened 2020-01-11
[ocaml] support eliom files

facelesspanda:
Eliom/eliomi is ocsigen's extensions for its files using the eliom ppx.

More info at http://ocsigen.org/eliom/6.6/manual/ppx-syntax

:warning: untested in this exact form. I have this in my own config.el:
```lisp
(after! tuareg
  (add-hook! tuareg-mode
    (let ((ext (file-name-extension buffer-file-name t)))
      (cond ((equal ext ".eliom")
             (setq-local ocamlformat-file-kind 'implementation))
            ((equal ext ".eliomi")
             (setq-local ocamlformat-file-kind 'interface))))))
```
So I assume putting in the function associated with the `tuareg-mode` hook loaded after tuareg in the ocamlformat `use-package!` would run that at the same time.


hlissner:
Thanks for the help!

PR #2358, opened 2020-01-12
Fix pdf-outline error

howsiwei:
Runnning `pdf-outline` resulting in the error `Invalid side 'right specified`. This PR fixes it.


hlissner:

Issue #2343, opened 2020-01-09
org-capture-journal-file is not expanded with org-directory


game-of-drones:
**What did you expect to happen?**
I added `(setq org-directory "/path/to/my/org")` in `.doom/config.el` and run `~/.emacs.d/bin/doom refresh`. After opening emacs and press
```SPC X j```
I want it to capture a journal entry in `/path/to/my/org/journal.org`

**What actually happened?**
I got a message saying "Heading not found on level 1: Inbox". And when I list ibuffer with `SPC b I`, I found a new file is created `mycurrentdir/journal.org`, where `mycurrentdir` is the path where I started emacs. I.e., it didn't expand "journal.org" with org-directory.

What's weird is that if I capture a todo by `SPC X t`, it open the correct file to add entry which is `/path/to/my/org/todo.org`. capture notes also works. Only journal doesn't work.

**Additional details:**
1. When I grep "org-capture-todo-file", I can see following lines  in `modules/lang/org/autoload/org-capture.el`
```
79:(defun +org-capture-todo-file ()
80:  "Expand `+org-capture-todo-file' from `org-directory'.
81:If it is an absolute path return `+org-capture-todo-file' verbatim."
82:  (expand-file-name +org-capture-todo-file org-directory))
```

2. When I grep "org-capture-notes-file", similar expand-file-name is called

3. However, when I grep "org-capture-journal-file", I don't see such expand-file-name called.

So my guess is that in ".emacs.d/modules/lang/org/config.el", something needs to be done for org-capture-journal-file to expand it correctly.

Thanks!

**Comment by game-of-drones** [comment]
** Update
I did the following settings in `.doom.d/config.el`. Now `SPC X j` can find the correct file `~/Documents/jrnl/my_journal.org`. 
```
;; If you intend to use org, it is recommended you change this!
(setq org-directory "~/Documents/jrnl")
(after! org
  (setq +org-capture-journal-file
        (expand-file-name "my_journal.org"  org-directory)
  )
)
```

However, it still complains `Heading not found on Level 1: Inbox`.  This seems to be related to the fact that in variable `org-capture-templates`, the entry for journal is like this.
```
 ("j" "Journal" entry
  (file+olp+datetree +org-capture-journal-file "Inbox")
```

game-of-drones:
** Update 2
[...]

PR #2360, opened 2020-01-12
Fix for https://github.com/hlissner/doom-emacs/issues/2343

harritaylor:
https://github.com/hlissner/doom-emacs/issues/2343 seems to describe an issue where the correct `journal.org` isn't found. I experienced the same problems. Based off https://github.com/hlissner/doom-emacs/issues/2343#issuecomment-572719940, I figured that setting `(setq +org capture-journal-file (expand-file-name "my_journal.org"  org-directory)` in `.../org/config.el` would fix the issue, which it did. 

If there is a more elegant solution you'd prefer, I'd be happy to change (I'm still a relative noob when it comes to elisp!)

### DIFF
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -263,6 +263,8 @@ I like:
    vimperator, dmenu or a global keybinding."
   (setq org-default-notes-file
         (expand-file-name +org-capture-notes-file org-directory)
+        +org-capture-journal-file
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the help!

harritaylor:
@hlissner No worries - thanks for the approval 👍

PR #2361, opened 2020-01-12
Update lsp-server-install-dir location

niklascarlsson:
Update the lsp-server-install-dir to doom-etc-dir. This aligns the installation
location with other lsp servers.

----

# 

Other lsp servers are installed under doom-etc-dir. When the user installs a
server through the command lsp-install-server it make sense if those servers
are also installed under the doom-etc-dir. At the moment they are instead
installed in the default location which is .emacs.d/.cache/lsp.


hlissner:
Now that LSP has a central install dir variable, I'll revisit all the server dir variables for the individual clients. Better to have them under a common root. Thanks for the help and bringing it to my attention!

PR #2363, opened 2020-01-13
Minor updates to documentation

niklascarlsson:
- Update the keybinding for +doom/help-autodefs
- Remove silver_search from the lookup module. Ripgrep is the supported
dependency.

----

# 

Just some minor updates to the documentation that I ran into and thought I might
as well address.


hlissner:
Good catches. Thanks again!

PR #2364, opened 2020-01-13
Update feature check of flycheck in evil-bindings

niklascarlsson:
This patch updates the check for enabling flycheck bindings to the new
module location.

----

# 

the evil-bindings check for enabling flycheck bindings is not synced with the
new module location.


hlissner:
Ah, good catch. Thanks for the fix!

PR #2366, opened 2020-01-14
Update feature check of flycheck in elixir mode

ckruse:
This patch updates the check for enabling flycheck bindings to the new
module location.


hlissner:
Thanks!

PR #2368, opened 2020-01-14
Use Docker docset for dockerfile-mode

niklascarlsson:
This patch setups the appropriate docset to use in dockerfile-mode.
Similar to how it is done for other major-modes that have associated
docsets.

----

# Add docset configuration for Docker

When dockerfile-mode is active it make sense to activate the Docker docset.


hlissner:
Makes sense to me. Thanks!

PR #2371, opened 2020-01-14
Add LSP Support to Erlang

pekist:
Add LSP support to the erlang module.

Should probably make a Readme.md too...

### DIFF
--- a/modules/lang/erlang/config.el
+++ b/modules/lang/erlang/config.el
@@ -26,4 +26,4 @@
   :hook (erlang-mode . company-erlang-init))
 
 (when (featurep! +lsp)
-  (add-hook 'erlang-mode-hook #'lsp!))
+  (add-hook 'erlang-mode-local-vars-hook #'lsp!))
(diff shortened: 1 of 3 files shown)


modules/lang/erlang/config.el
```diff
@@ -22,4 +22,8 @@
 
 (use-package! company-erlang
   :when (featurep! :completion company)
+  :unless (featurep! +lsp)
   :hook (erlang-mode . company-erlang-init))
+
+(when (featurep! +lsp)
+  (add-hook 'erlang-mode-hook #'lsp!))
```
hlissner: Could you change `erlang-mode-hook` to `erlang-mode-local-vars-hook`. It's customary in Doom to start LSP as late as possible in the mode activation process.

modules/lang/erlang/config.el
```diff
@@ -22,4 +22,8 @@
 
 (use-package! company-erlang
   :when (featurep! :completion company)
+  :unless (featurep! +lsp)
```
hlissner: If `company-erlang` is loaded conditionally, please wrap the `package!` declaration for company-erlang in this module's packages.el file in a `+lsp` check as well.

hlissner:
Looks good to me! Thanks for the help and prompt response!

PR #2374, opened 2020-01-15
Doctor: Check whether rbenv is enabled before complaining

valrus:
This fixes an issue where `doom doctor` was giving me the error:

```
      ! No versions of ruby are available via rbenv, did you forget to install one?
```

even though I have `(ruby +rvm)` in my `init.el` and don't use rbenv.


hlissner:
Thanks for the help!

Issue #2335, opened 2020-01-08
[BUG] direnv argument list to long


bkchr:
Hey,
after running emacs for some time, at some point I get the following error printed when opening a file:
```
Open buffer ‘*direnv*’ for full output.
Warning (emacs): Error running direnv (exit code 1):
direnv: loading ../../../.envrc
/nix/store/xb062l4b76zyhq6grqf4iyfdikkpg8fl-bash-4.4-p23/bin/bash: /nix/store/n42ddrk1lklv65b0ayip3kqik9nin5q1-direnv-2.20.1-bin/bin/direnv: Argument list too long
direnv: error exit status 126
```

Restarting emacs fixes it.

This is the output of the *direnv* buffer:
```
{
  "DIRENV_DIFF": "eJw0yr1ygjAAAOB3ySxH_EkR7jogaEWUfxVZcgFjRSpESA3E8907df9egAHj9R6BGhgvYDvR0jtg24mAART12typmpOOl6RWWdvcaME7lZG25AMY_eujmVjrZQwMQDc9lN3K7y0zNL0Fvqe-3MNHHfpzmuQit8VOnpMvLQu9k1dkWJ9oxXp6RVLHog_gzlUslC72N44Q6scdzxH7EP0znW6xYAGPlkKKo5vUj_nmMVjwF_puakYzpVA6b_yjXfQjnYUIxZP8Mi-Z22rflwEdHAUGvDpHMdXiXK4qQqjTwH1tVqSB26xp_Wp13l5vd5ZS0hzKLDoN6aB32FngRtJn5ZmmjTGuSBzAT_B-_wUAAP__obJmjg==",
  "DIRENV_DIR": "-/home/bastian/projects/parity",
  "DIRENV_WATCHES": "eJx0zsFOxCAQANB_mXOzU0qnQO8eTbwbDwMzdTG7ZQNYNcZ_927cH3h5z9_wxP0MK-C5XBUjt555x1stb5p6wxvX3L_wpPtREwzwWKTnq8JqyC0u0OKXAR4-c-sN1l7f9We4Q55S2bf8ipKr7gfy5VI-0PtkdRSe7SbzFkaaeIo0UnAkao0LZorOkFdLhjmpXeaoViZRYyXy9s_IB_ozevkNAAD___kaSP0="
}

direnv: loading ../../../.envrc
[...]

PR #2384, opened 2020-01-17
Potential fix for direnv argument list too long

bkchr:
Potential fix for https://github.com/hlissner/doom-emacs/issues/2335

The removed code looks similar to what already is done in direnv-emacs: https://github.com/wbolster/emacs-direnv/commit/bb0de5e8f662c3f522b931c6e2c2340e4ed682af

The commit that introduced this feature to doom, also has a comment on it that complains about missing environment variables: https://github.com/hlissner/doom-emacs/commit/4860bb86ce34c17ae1b064e029843a4931b1847c

@ocharles could you test this as well?

### DIFF
--- a/modules/tools/direnv/config.el
+++ b/modules/tools/direnv/config.el
@@ -21,13 +21,6 @@ Emacs."
         (add-hook 'after-change-major-mode-hook
                   #'direnv--maybe-update-environment))))
 
-  (defadvice! +direnv--make-process-environment-buffer-local-a (items)
-    :filter-return #'direnv--export
-    (when items
-      (mapc 'kill-local-variable '(process-environment exec-path))
-      (mapc 'make-local-variable '(process-environment exec-path)))
-    items)
-
   ;; Fontify special .envrc keywords; it's a good indication of whether or not
   ;; we've typed them correctly.
   (add-hook! 'direnv-envrc-mode-hook


ocharles:
Ooo! Yes, I will give this a try!

bkchr:
I'm using it since saturday and did not run into the problem anymore.

bkchr:
@hlissner what do you think of this patch? I did not encounter the error anymore.

hlissner:
Without that advice direnv.el is mutating the global `process-environment`. The purpose of [these lines](https://github.com/hlissner/doom-emacs/blob/develop/modules/tools/direnv/config.el#L13-L22) is to invoke `direnv--maybe-update-environment` once, when a major mode is first activated in a buffer, rather than on every window focus change. Making the `process-environment` buffer-local is necessary for this to work safely, otherwise the new environment will contaminate other buffers.

The `kill-local-variable` calls next to it exists to clear the local environment so stale state doesn't persist into the new environment.

I have yet to reproduce what was discussed in #2335.

bkchr:
direnv-emacs is also setting [exec-path](https://github.com/wbolster/emacs-direnv/blob/master/direnv.el#L235), maybe we need to make this local as well? Something grows and direnv doesn't detect that it was already executed. In a terminal you can see that calling direnv multiple times, doesn't change anything after the first call. Maybe `exec-path` is really the thing that grows.

hlissner:
@bkchr Aye, it is made buffer-local, along with `process-environment`: <https://github.com/hlissner/doom-emacs/blob/develop/modules/tools/direnv/config.el#L27-L28>

bkchr:
Ahh, to late. Sorry!

ocharles:
@hlissner What is the need to so heavily modify the direnv package?

hlissner:
@ocharles Without these modifications, direnv updates every time you switch buffers or windows. This can be dreadfully expensive, especially for nix users. This aggressive behavior is necessary because direnv affects the global `process-environment`. The newly loaded state could contaminate other buffers/projects if the changes weren't undone when switching buffers.

By making `process-environment` buffer-local we need only trigger direnv once; when you first open the buffer (and perhaps when you save or move it), and never again. That environment naturally remains isolated to that buffer.

hlissner:
@bkchr @ocharles Does this change the situation for either of you?

```elisp
(after! direnv
  (advice-remove #'direnv--export #'+direnv--make-process-environment-buffer-local-a)
  (defadvice! +direnv--make-process-environment-buffer-local-a (orig-fn &rest args)
    :around #'direnv-update-directory-environment
    (let ((vars '(process-environment exec-path)))
      (mapc #'kill-local-variable vars)
      (mapc #'make-local-variable vars)
      (condition-case-unless-debug e
          (apply orig-fn args)
        (error
         (mapc #'kill-local-variable vars))))))
```

ocharles:
> @ocharles Without these modifications, direnv updates every time you switch buffers or windows. This can be dreadfully expensive, especially for nix users (adding 2-5 seconds to buffer management in general; closing a window, killing/switching a buffer, etc). 

IMO, This is Nix+direnvs problem, not Emacs. I don't think working around it in Elisp is correct. The correct solution is to just use lorri, which makes dirnev usage instantaneous: https://github.com/target/lorri

> This aggressive behavior is necessary because direnv affects the global process-environment. The newly loaded state could contaminate other buffers/projects if the changes weren't undone when switching buffers.

But doesn't `direnv-mode` already take care of that? If `dirnenv-mode` isn't unloading itself appropriately, that seems like an upstream bug.

hlissner:
> IMO, This is Nix+direnvs problem, not Emacs. I don't think working around it in Elisp is correct. The correct solution is to just use lorri, which makes dirnev usage instantaneous: https://github.com/target/lorri

I agree that it _is_ an upstream issue, but I don't see this being fixed anytime soon. With lorri I still see a noticeable (albeit far more tolerable) delay when switching buffers, but I'm not happy that I need a yet-another-tool:tm: just to make the first run tolerably. 

Since it can be solved in Emacs (I'd hope), I'd like to do so while we await a better solution.

> But doesn't direnv-mode already take care of that? If dirnenv-mode isn't unloading itself appropriately, that seems like an upstream bug.

I apologize for not explaining clearly enough. I'm not saying direnv-mode doesn't work. It works correctly OOTB.

The issue is that it's implementation is sub-optimal, IMO. It calls `direnv--maybe-update-environment` on `pre-command-hook`, which checks if the global environment should be updated before every command and keypress. This brute force approach ensures that global state is always synced with the selected buffer, but a side effect is that it triggers every time we switch between windows or buffers that are in different directory environments, introducing an undesirable delay.

My implementation makes the environment buffer-local, so direnv-mode only needs to be updated once, when the buffer is first opened.

hlissner:
I've reverted my modifications in a12a54b until I can figure out a better solution.

bkchr:
@hlissner could you reproduce this as well?

hlissner:
@bkchr I have some (untested) ideas, but I haven't reproduced it yet.

PR #2390, opened 2020-01-19
Use conditional binding for 'SPC n d' "Open deft"

Mishkun:
Deft is not enabled by default and trying to open it with the shortcut
is confusing newbie users (including me) with weird error.

Remove this shortcut if deft is not enabled by using :when key.


hlissner:
This is a duplicate of #2267 (and #2320). See explanation here: https://github.com/hlissner/doom-emacs/pull/2267#issuecomment-569536183

Issue #2397, opened 2020-01-22
Emacs minimizes itself upon closing a Helm posframe [Resolved, changes to defaults required]


snyball:
**What did you expect to happen?**
I expected focus to return to the main Emacs window.

**What actually happened?**
The Emacs window got minimized to the tray.

Really scratching my head here as to why this happens.

**Additional details:**
Using the latest development branches of both Doom, and Emacs itself.

Able to work around the issue by switching to i3 temporarily when I have to use Emacs, because the i3 WM does not support minimizing windows.

Doing a `(posframe-show ...)` to get a posframe, and closing that posframe with `(posframe-hide-all)` did not minimize Emacs, it is specifically something in helm.

Did a couple of complete re-installs, and disabled my own configuration, but the problem has persisted.

Posting this here to see if it can be reproduced, next I'll try installing doom with Emacs 26.3 to see if there's an Emacs bug causing all this, because I've been unable to find anything on Google.

Also, the latest version of posframe has a problem with later versions of Emacs on Linux after  https://github.com/tumashu/posframe/commit/d75dc1547a6a1cc2b385c736880eee77d7981aec#commitcomment-35754593, for which the workaround is this patch that I load in my config:

<details>
<pre>
(with-eval-after-load 'posframe
 (defun posframe--set-frame-size (posframe height min-height width min-width)
   "Set POSFRAME's size.
It will set the size by the POSFRAME's HEIGHT, MIN-HEIGHT
WIDTH and MIN-WIDTH."
   (if (and width height)
       (unless (equal posframe--last-posframe-size
                      (cons width height))
         (set-frame-size posframe width height)
         ;; (fit-frame-to-buffer
         ;;  posframe height min-height width min-width)
         (setq-local posframe--last-posframe-size
                     (cons width height)))
     (fit-frame-to-buffer
      posframe height min-height width min-width))))
</pre>
</details>

This patch does not have any effect on the minimization bug, I've reproduced the bug with and without the patch.

I attempted to turn `minimize-window` into a no-op:
```elisp
(defun minimize-window ()
  "no-op"
  (interactive))
(defun window-resize ()
  "no-op"
  (interactive))
(defun window-resize-pixelwise ()
  "no-op"
  (interactive))
```
Did not work, the Emacs window still got minimized.
Is there a way to trace the path of execution that occurs after the helm selection is made?

Update: The bug is also present on Emacs 26.3, which means that the bug might have been introduced in helm.

**Steps to reproduce:**
[...]

PR #2402, opened 2020-01-22
Helm posframe fixes

snyball:
- Fixes https://github.com/hlissner/doom-emacs/issues/2397
- Displays Helm input in header line on top of posframe if `helm-echo-input-in-header-line` is set to true.

### DIFF
--- a/modules/completion/helm/autoload/posframe.el
+++ b/modules/completion/helm/autoload/posframe.el
@@ -10,6 +10,9 @@ bottom, which is easier on the eyes on big displays."
           (truncate (/ (frame-pixel-height parent-frame)
                        2)))))
 
+(when helm-echo-input-in-header-line
+  (add-hook 'helm-minibuffer-set-up-hook #'helm-hide-minibuffer-maybe))
+
 (defvar +helm--posframe-buffer nil)
 ;;;###autoload
 (defun +helm-posframe-display-fn (buffer &optional _resume)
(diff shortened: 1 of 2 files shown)


snyball:
Made an oops in the #2397 fix, going to fix that

modules/completion/helm/autoload/posframe.el
```diff
@@ -10,6 +10,9 @@ bottom, which is easier on the eyes on big displays."
           (truncate (/ (frame-pixel-height parent-frame)
                        2)))))
 
+(when helm-echo-input-in-header-line
+  (add-hook 'helm-minibuffer-set-up-hook #'helm-hide-minibuffer-maybe))
```
hlissner: This could be a race condition in waiting, because users are encouraged to customize packages by wrapping configuration in `(after! helm ...)`, if an autoloaded function (in autoload/posframe.el) is called, this condition will run before the user has an opportunity to change `helm-echo-input-in-header-line`.

Since it appears `helm-hide-minibuffer-maybe` [does this check already](https://github.com/emacs-helm/helm/blob/master/helm.el#L5282) I think it's safe to unconditionally add this hook (in which case, please move it to helm's `:config` block in this module's config.el.

snyball: Yeah, that was a bit messy. Fixed now.

hlissner:
Thanks for looking into and fixing this!

Issue #2381, opened 2020-01-17
flyspell/ispell error on nixos


codygman:
**What did you expect to happen?**
Flyspell to work normally in the scratch buffer after I run `SPC t s`.

**What actually happened?**
From `*doom:scratch*` I run `SPC t s` and I see improper usage error output from ispell in the message area.

If I look in the messages buffer I see the full output:

```
Starting new Ispell process /home/cody/.nix-profile/bin/ispell with default dictionary...
Error enabling Flyspell mode:
(?       Show this help screen.
^Z      Suspend program.
^L      Redraw screen.
!       Shell escape.
        Leaves file unchanged.
Q       Quit immediately.  Asks for confirmation.
        and start next file.
X       Write the rest of this file, ignoring misspellings,
L       Look up words in system dictionary.
0-n     Replace with one of the suggested words.
U       Accept and add lowercase version to private dictionary.
I       Accept the word, and put it in your private dictionary.
A       Accept the word for the rest of this session.
Space   Accept the word this time only.
R       Replace the misspelled word completely.

Commands are:

completely, or choosing one of the suggested words.
next to each one.  You have the option of replacing the word
contains any similar words, they are listed with a number
it is printed on the first line of the screen.  If the dictionary
Whenever a word is found that is not in the dictionary,
       /home/cody/.nix-profile/bin/ispell -v
       /home/cody/.nix-profile/bin/ispell [-dfile] [-wchars] -D
       /home/cody/.nix-profile/bin/ispell [-dfile] [-wchars] -e[1-4]
       /home/cody/.nix-profile/bin/ispell [-dfile] [-wchars | -Wn] -c
       /home/cody/.nix-profile/bin/ispell [-dfile | -pfile | -ffile | -Wn | -t | -n | -H | -B | -C | -P | -m | -Ttype | -ktype kws | -Fpgm] [-a | -A]
       /home/cody/.nix-profile/bin/ispell [-dfile | -pfile | -wchars | -Wn | -t | -n | -H | -Ttype | -ktype kws | -Fpgm] -l
Usage: /home/cody/.nix-profile/bin/ispell [-dfile | -pfile | -wchars | -Wn | -t | -n | -H | -x | -b | -S | -B | -C | -P | -m | -Lcontext | -M | -N | -Ttype | -ktype kws | -Fpgm | -V] file .....)
```

[...]

PR #2410, opened 2020-01-23
Move an aspell specific argument to appropriate block

Geospace:
Fix for https://github.com/hlissner/doom-emacs/issues/2381

A CLI argument specific to `aspell` was also passed to `ispell`, which was then unable to start and printed its help message instead.


hlissner:
Thanks for the help!

PR #2418, opened 2020-01-24
Make `ls` aliases work with subdirectories (Eshell)

dive:
Hello there,

`ls` aliases for Eshell do not take any params and it is a bit annoying to use different commands between the current directory and subdirectories. Let's allow them to take params.


hlissner:

PR #2423, opened 2020-01-25
use updated brew package name to get clang

korantu:
I believe llvm package is correct way to get clang:

https://formulae.brew.sh/formula/llvm
https://github.com/Homebrew/homebrew-core/blob/4f169c6375b1bc8c23ddf374a78d33a0caba8ec9/Formula/llvm.rb#L72


hlissner:
Thanks for the help!

PR #2425, opened 2020-01-25
make +org--insert-item respect emacs-mode

nikodemus:
No change of behaviour in non-emacs modes, but in emacs-mode stays there
   instead of switching to insert-mode.

--

In upstream Doom Emacs hitting C-RET in org-mode switches to insert-mode regardless
even if the editor was in emacs-mode, which I at least find quite disorienting.


hlissner:

nikodemus:
Thank you for the superfast merge!

PR #2193, opened 2019-12-14
added doctor warning for pyflakes

lelionmusic:
Added this warning to doctor, after I found out I had to do pip install pyflakes for pyimport.

PR #2437, opened 2020-01-27
lang/python: add additional checks

danlamanna:
Related: #2180, #2193 

A few questions and a comment:

1) Should line 8 be an assertion? Is `python` not being installed a blocker?
2) I realize these doctor checks are all checking for global installations of various packages. How does this conflict with the various virtualenv solutions people are using?
3) How do you determine where to put intermixed dependencies? For example, should the `editor` module also check if `python` is enabled and warn about `black`?
4) Finally, I settled on using the format of "couldn't find $executable. $feature will not work". Doom seems to be a little bit inconsistent here, but this format seemed to be most common.

### DIFF
--- a/modules/lang/python/doctor.el
+++ b/modules/lang/python/doctor.el
@@ -5,7 +5,7 @@
          "This module requires (:tools lsp)")
 
 (if (not (executable-find "python"))
-    (warn! "Python isn't installed.")
+    (assert! "Python isn't installed.")
   (unless (featurep! +lsp)
     (unless (zerop (shell-command "python -c 'import setuptools'"))
       (warn! "setuptools wasn't detected, which anaconda-mode requires"))))
--- a/modules/lang/python/doctor.el
+++ b/modules/lang/python/doctor.el
@@ -20,6 +20,30 @@
   (unless (executable-find "conda")
     (warn! "Couldn't find conda in your PATH")))
 
+(when (featurep! +cython)
+  (unless (executable-find "cython")
+    (warn! "Couldn't find cython. cython-mode will not work.")))
+
 (when (featurep! +ipython)
   (unless (executable-find "ipython")
     (warn! "Couldn't find ipython in your PATH")))
+
+(if (and (featurep! :editor format)
+         (not (executable-find "black")))
+    (warn! "Couldn't find black. Code formatting will not work."))
+
+(if (not (executable-find "pytest"))
+    (warn! "Couldn't find pytest. Running tests through pytest will not work."))
+
+(if (not (executable-find "nosetests"))
+    (warn! "Couldn't find nosetests. Running tests through nose will not work."))
+
+(if (not (executable-find "pipenv"))
+    (warn! "Couldn't find pipenv. pipenv support will not work."))
+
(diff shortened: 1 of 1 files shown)


modules/lang/python/doctor.el
```diff
@@ -20,6 +20,30 @@
...
+
+(if (not (executable-find "pipenv"))
+    (warn! "Couldn't find pipenv. pipenv support will not work."))
+
+(if (not (executable-find "isort"))
+    (warn! "Couldn't find isort. Import sorting will not work."))
+
+(if (and (featurep! :editor format)
+         (not (executable-find "pyflakes")))
+    (warn! "Couldn't find pyflakes. Import management will not work."))
```
hlissner: Please combine the two `(featurep! :editor format)` checks.

modules/lang/python/doctor.el
```diff
@@ -20,6 +20,30 @@
...
+(if (not (executable-find "nosetests"))
+    (warn! "Couldn't find nosetests. Running tests through nose will not work."))
+
+(if (not (executable-find "pipenv"))
+    (warn! "Couldn't find pipenv. pipenv support will not work."))
+
+(if (not (executable-find "isort"))
+    (warn! "Couldn't find isort. Import sorting will not work."))
+
+(if (and (featurep! :editor format)
```
hlissner: Please use `(unless (executable-find ...) ...)` instead of `(if (not (executable-find ...)) ...)`

hlissner:
>     1. Should line 8 be an assertion? Is `python` not being installed a blocker?

It probably should be, yeah.

>     2. I realize these doctor checks are all checking for global installations of various packages. How does this conflict with the various virtualenv solutions people are using?

I'm afraid there is no simple solution to this problem. We can only check for global installs and rely on heuristics. User's can ignore its warnings if they know better. I'm open to suggestions.
 
>     3. How do you determine where to put intermixed dependencies? For example, should the `editor` module also check if `python` is enabled and warn about `black`?

Good question. The provider should do the checks, rather than the consumer. i.e. It should be checked for in the `:editor format` module, wrapped in `(when (featurep! :lang python) ...)`. I've put it in the python module because `:editor format` is overdue a rewrite (possibly with the reformatter.el package instead).

>     4. Finally, I settled on using the format of "couldn't find $executable. $feature will not work". Doom seems to be a little bit inconsistent here, but this format seemed to be most common.

Very true. I have a rewrite of the doctor system in the works to standardize these checks (and make following up on specific warnings/errors easier, through a `doom doctor --describe ERRCODE` command or similar). In any case, this is just fine for the time being.

modules/lang/python/doctor.el
```diff
@@ -5,7 +5,7 @@
          "This module requires (:tools lsp)")
 
 (if (not (executable-find "python"))
-    (warn! "Python isn't installed.")
+    (assert! "Python isn't installed.")
```
hlissner: Actually, this should be an `error!` call. This isn't how `assert!` is used and will throw an error.

hlissner:
Sorry for the late response. This looks good to me. Thanks for your help!

Issue #2439, opened 2020-01-27
Tramp session gets re-established over and over when traversing git repo directory structure


mihai-dinu:
**What did you expect to happen?**
`C-x C-f` to open an ssh connection to a remote server to a directory. The directory contains a git repo. Expecting a dired buffer. Expecting no slowness while going through the directories of the repo. 

**What actually happened?**
Once you get into a remote git directory, and start going into different subdirectories, dired responds slow and I'm seeing a lot of logs about re-establishing the connection in the \*Messages\* buffer:

```
Tramp: Sending command ‘exec ssh -q   -o ControlMaster=auto -o ControlPath='tramp.%C' -o ControlPersist=no -e none clouddesk’
Tramp: Waiting for prompts from remote shell...done
Tramp: Found remote shell prompt on ‘clouddesk’
Tramp: Opening connection for clouddesk using ssh...done
Tramp: Opening connection for clouddesk using ssh...
Tramp: Sending command ‘exec ssh -q   -o ControlMaster=auto -o ControlPath='tramp.%C' -o ControlPersist=no -e none clouddesk’
Tramp: Waiting for prompts from remote shell...done
Tramp: Found remote shell prompt on ‘clouddesk’
Tramp: Opening connection for clouddesk using ssh...done
Tramp: Opening connection for clouddesk using ssh...
Tramp: Sending command ‘exec ssh -q   -o ControlMaster=auto -o ControlPath='tramp.%C' -o ControlPersist=no -e none clouddesk’
Tramp: Waiting for prompts from remote shell...done
Tramp: Found remote shell prompt on ‘clouddesk’
Tramp: Opening connection for clouddesk using ssh...done
```

**Additional details:**
This seems to be happening only in dired and only for git repos. I'm assuming dired tries to fetch some git specific info from the remote server, but for doing that it creates a new tramp connection every time. 
Tried the same thing with vanilla Emacs and there are no problems with the connection and no reconnects.

**Steps to reproduce:**
1. Connect to remote machine using Tramp
2. Open a dired buffer to a git repo on the remote machine
3. Browse the sub-directories
4. Check the minibuffer or \*messages\*  for attempts to re-establish the ssh connection

**System information:**
<details><pre>
```
emacs   version    26.3
        features   NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES THREADS
        build      Sep 02, 2019
[...]

PR #2440, opened 2020-01-27
Don't use dired git highlighting if remote

mihai-dinu:
Fixes https://github.com/hlissner/doom-emacs/issues/2439

### DIFF
--- a/modules/emacs/dired/config.el
+++ b/modules/emacs/dired/config.el
@@ -65,7 +65,7 @@ only variant that supports --group-directories-first."
 
 
 (use-package! diff-hl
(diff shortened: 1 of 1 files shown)


mihai-dinu:
@hlissner Should I merge this with the test failing? It's a `nconcq!` failure that doesn't have anything to do with the change.

hlissner:
You can ignore the failing tests. Those failures are unrelated; I'll deal with them soon. In any case, I wasn't aware `diff-hl-dired-mode-unless-remote` was a thing. Thanks for looking into this and fixing it!

PR #2442, opened 2020-01-28
Add {little,greater}-than to +web-entities-list

dsdshcym:
< and > were missing from `+web-entities-list`


hlissner:
Good catch. Thanks!

PR #2452, opened 2020-01-29
Bump to brotzeit/rustic@124e028

kid:
Rustic was broken because of brotzeit/rustic#88


hlissner:
My bad. Thanks for the PR!

PR #2454, opened 2020-01-29
Improve solidity layer

sei40kr:
## Description

- [x] Enable `+javascript-npm-mode` in `solidity-mode` when `package.json` exists.
- [x] Set Dash docsets for solidity-mode.
- ~[ ] Enable flycheck checkers by default~
  https://github.com/ethereum/emacs-solidity/blob/master/solidity-flycheck.el#L89
  if we set it active by default, it throws errors when the executables were not found.
- [x] Use locally-installed solium (Ethlint)
- ~[ ] Format a Solidity file with Prettier + prettier-plugin-solidity~
  -> don't do in this PR because it's still experimental.

### DIFF
--- a/modules/lang/javascript/config.el
+++ b/modules/lang/javascript/config.el
@@ -284,7 +284,13 @@ to tide."
 ;;; Projects
 
 (def-project-mode! +javascript-npm-mode
-  :modes '(html-mode css-mode web-mode markdown-mode js-mode typescript-mode)
+  :modes '(html-mode
+           css-mode
+           web-mode
+           markdown-mode
+           js-mode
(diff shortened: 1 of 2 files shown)


sei40kr:
@hlissner 
To describe more strictly, we can do it for `solc`, but `solium` may be provided by local node_modules.
In addition, checking the executable is not originally what `solidity-flycheck` should do. Flycheck already does it and disable unusable checkers. So I think this is upstream issue.

hlissner:
Understood. Thanks for looking into this and for the PR (and all the others as well; it is a big help)!

PR #2459, opened 2020-01-30
[purescript] add +psc-ide as a feature

CGenie:
I prefer LSP, I don't want psc-ide to be turned on automatically for my
PureScript projects.

### DIFF
--- a/modules/lang/purescript/+psc-ide.el
+++ b/modules/lang/purescript/+psc-ide.el
@@ -1,0 +1,8 @@
+;;; lang/purescript/+psc-ide.el -*- lexical-binding: t; -*-
+;;;###if (featurep! +psc-ide)
+
+(use-package! psc-ide
+  :hook (purescript-mode . psc-ide-mode)
+  :config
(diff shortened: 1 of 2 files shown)


CGenie:
My LSP code isn't that straighforward as I use yarn so I don't provide the `+lsp` feature in this PR. I just want `+psc-ide` disabled, that's all :)

CGenie:
OK then, I'll disable `psc-ide` for now and try to come up with `+lsp` later, my config for it is quite customized.

LoveSponge:
For anyone that comes here wondering why lsp mode isn't giving any of it's lsp goodness, it should be noted that `(package psc-ide :disable t)` **must** be added to your `~/.doom.d/packages.el` in order for lsp to actually work **at all.**

You can see that both modes attempt to start in the **lsp-logs** buffer where it explicitly says `Starting purs IDE server` **which is NOT the lsp server and does not give the same level of error handling.**

Ps. Doing the above will not effect your `.purs` buffers, syntax highlighting still works perfectly, just now, so does lsp mode.

PR #2462, opened 2020-01-31
Add persistent-soft to core/packages.el

alexandergraul:
core/autoload/cache.el defines doom-cache-* functions that wrap around
the persistent-soft package.

I am not sure if this is the correct place to add the `package!` statement to but I figured you'll tell me if it isn't. I stumbled upon this when I wanted to try https://github.com/hlissner/doom-emacs/blob/develop/docs/api.org#persist-emacs-initial-frame-position-dimensions-andor-full-screen-state-across-sessions. I've pinned it to the latest commit and don't have any issue with it (I just checked with `doom-cache-set` and `doom-cache-get`)


hlissner:
Good catch. Thanks!

Issue #2461, opened 2020-01-31
Add better support do Rails


otavioschwanck:
I'm a pure emacs user and i really miss some plugins:

- projectile-rails (https://github.com/asok/projectile-rails)
- rspec-mode (https://github.com/pezra/rspec-mode)
- Robe (https://github.com/dgutov/robe)

Projectile-rails is excelent for navigation and file creation.  Can navigate between test \ real file, search for models (C-c r m), controllers (C-c r c), etc.  its super complete and make all diffrence.
Rspec-mode is good to run tests.
Robe for auto-complation.

Maybe we need a Rails layer.  Its to hard to implement ?  If not, i can contribute.

hlissner:
> I'm a pure emacs user and I really miss some plugins

I don't understand. So you don't use Doom, but you want those plugins? Or you use Doom and you want those plugins?

If the former, how is this relevant to Doom?

If the latter, Doom's `:lang ruby` module already provides robe and rspec-mode, but not rails.

> Maybe we need a Rails layer. Its to hard to implement ? If not, i can contribute.

I'd welcome rails support so long as it's lazy loaded. Go ahead and add it to the `:lang ruby` module under the `+rails` flag.

alexandergraul:
There is also this issue open for projectile-rails support: https://github.com/hlissner/doom-emacs/issues/2275

Issue #2275, opened 2019-12-30
[REQUEST] projectile-rails


luisenrike:
https://github.com/asok/projectile-rails

luisenrike:
I could make a PR if this is something you think we should have.

hlissner:
Sure, I'll accept a PR for it.

I have far-off plans to remove projectile (and replace it with project.el) once we drop emacs 26 support -- but that's quite a while away; likely late 2021 or 2022 -- so I was conflicted about including it. That said, there doesn't seem to be any alternative to projectile-rails, so we'll cross that bridge when we get there.

otavioschwanck:
projetile-rails is the only reason that i dont use doom emacs.  Would be very nice!

PR #2463, opened 2020-02-01
Add Rails support to Ruby module

alexandergraul:
This is WIP, I've picked the upstream keyindings with minor changes for now.

Fixes https://github.com/hlissner/doom-emacs/issues/2461, https://github.com/hlissner/doom-emacs/issues/2275

### DIFF
--- a/modules/lang/ruby/config.el
+++ b/modules/lang/ruby/config.el
@@ -63,7 +63,7 @@
         "'"  #'robe-start
         ;; robe mode specific
         "h"  #'robe-doc
-        "rr" #'robe-rails-refresh
+        "R" #'robe-rails-refresh
         ;; inf-enh-ruby-mode
         :prefix "s"
         "d"  #'ruby-send-definition
@@ -168,3 +168,20 @@
         "a" #'minitest-verify-all
         "s" #'minitest-verify-single
         "v" #'minitest-verify))
+
+
+(use-package! projectile-rails
+  :when (featurep! +rails)
+  :hook (enh-ruby-mode . projectile-rails-mode)
+  :init
+  (when (featurep! :lang web)
+    (add-hook 'web-mode-hook #'projectile-rails-mode))
+  :config
+  (when (featurep! :editor evil)
+    (add-hook 'projectile-rails-mode-hook #'evil-normalize-keymaps))
+  (map! :localleader
+        :map projectile-rails-mode-map
+        "r" #'projectile-rails-command-map)
+  (push '((nil . "projectile-rails-\\(.+\\)") . (nil . "\\1"))
+        which-key-replacement-alist)
+  (set-popup-rule! "^\\*\\(projectile-\\)?rails" :ttl nil))
--- a/docs/modules.org
+++ b/docs/modules.org
@@ -131,7 +131,7 @@ Modules that bring support for a language or group of languages to Emacs.
 + qt - TODO
 + racket - TODO
 + [[file:../modules/lang/rest/README.org][rest]] - TODO
-+ ruby =+lsp +rvm +rbenv= - TODO
++ ruby =+lsp +rvm +rbenv +rails= - TODO
 + [[file:../modules/lang/rust/README.org][rust]] =+lsp= - TODO
 + scala =+lsp= - TODO
 + [[file:../modules/lang/scheme/README.org][scheme]] - TODO
--- a/modules/lang/ruby/packages.el
+++ b/modules/lang/ruby/packages.el
@@ -27,3 +27,8 @@
 ;; Testing frameworks
 (package! rspec-mode :pin "c4353a1bff")
 (package! minitest :pin "6d9f6233b7")
+
+;; Rails
+(when (featurep! +rails)
+  (package! projectile-rails :pin "228f631")
+  (package! inflections :pin "e4f1372"))


modules/lang/ruby/config.el
```diff
@@ -163,3 +164,55 @@
...
+        :desc "Find Job" "b" #'projectile-rails-find-job
+        :desc "Extract Region" "x" #'projectile-rails-extract-region
+        ;; Go to files
+        :desc "Goto Gemfile" "gg" #'projectile-rails-goto-gemfile
+        :desc "Goto Routes" "gr" #'projectile-rails-goto-routes
+        :desc "Goto Schema" "gd" #'projectile-rails-goto-schema
+        :desc "Goto Seeds" "gs" #'projectile-rails-goto-seeds
+        :desc "Goto Spec Helper" "gh" #'projectile-rails-goto-spec-helper)
+        ;; Open consoles etc.
+        (:prefix ("o" . "open")
```
hlissner: Please remove these which-key labels. They do not work on minor mode keymaps (the keys are bound absolutely, meaning all `<localleader> r` keybinds will have these labels, in all modes.

I will soon replace which-key labels with docstrings to get around this, but until then we're stuck without which-key support.

modules/lang/ruby/config.el
```diff
@@ -58,7 +58,8 @@
         "'"  #'robe-start
         ;; robe mode specific
         "h"  #'robe-doc
-        "rr" #'robe-rails-refresh
+        :prefix "r"
+        "r" #'robe-rails-refresh
```
hlissner: The `;prefix "r"` inadvertantly applies to the keybinds below it. Please revert this change.

modules/lang/ruby/config.el
alexandergraul: I see, the lables are one of the reasons I labled it WIP as I as not happy with it anyway. I'll remove them.

modules/lang/ruby/config.el
alexandergraul: I thought the `:prefix "s"` would override my `:prefix "r"`a for the keybindings below it. I don't mind removing it, just thought it was cleaner to write the prefix explicit instead of "rr"

hlissner: Prefixes stack. `:prefix "a" :prefix "b" "c" #'do-something` binds to the `"a b c"` key sequence.

You'll have to enclose them to isolate them: `(:prefix "a" ...) (:prefix "b" ...)`

modules/lang/ruby/config.el
```diff
@@ -163,3 +163,58 @@
...
+  :hook (enh-ruby-mode . projectile-rails-mode)
+  :config
+  (map! :localleader
+        :map projectile-rails-mode-map
+        ;; projectile-rails also has a projectile-rails-command-map
+        ;; There are two reasons for manual keybindings instead of mapping
+        ;; prefix "r" to it:
+        ;; - prefix "r" is also used by robe
+        ;; - Some of the keybindings differ
+        (:prefix ("r" . "Rails")
```
hlissner: The which-key labels have to be removed from :prefix as well:

```diff
-(:prefix ("r" . "Rails")
+(:prefix "r"
```

modules/lang/ruby/config.el
```diff
@@ -163,3 +163,58 @@
...
+        "b" #'projectile-rails-find-job
+        "x" #'projectile-rails-extract-region
+        ;; Go to files
+        "gg" #'projectile-rails-goto-gemfile
+        "gr" #'projectile-rails-goto-routes
+        "gd" #'projectile-rails-goto-schema
+        "gs" #'projectile-rails-goto-seeds
+        "gh" #'projectile-rails-goto-spec-helper)
+        ;; Open consoles etc.
+        (:prefix ("o" . "open")
```
hlissner: Same as above

otavioschwanck:
Very nice work, love the doom emacs community <3

modules/lang/ruby/config.el
otavioschwanck: This prefix will work on entire project or only on enh-ruby-mode ?

In spacemacs, you only can use projectile-rails shortcuts on enh-ruby-mode files, and it is very bad.  On pure emacs with projectile-rails, you can use the projectile-rails shortcuts on any mode (any file).

otavioschwanck:
image

When search for projectile-commands with M-x, the shortcuts appears wrong

otavioschwanck:
image

Ruby on Rails have many type of files on your workflow.  html, css, JS, .rb, .arb, .yml.

Its super necessary to have access to projectile-rails on those files too.  The projectile-rails have commands to search JS, CSS, etc.  

I tried here, the SPC m r * only works on .rb files.  Would be very nice enable projectile-rails commands on whole project.  Or at least, the possibility to do this.

modules/lang/ruby/config.el
alexandergraul: ah, good to know!

alexandergraul:
> When search for projectile-commands with M-x, the shortcuts appears wrong

For me they show up correctly , but I've also just removed those prefix descriptions.

> Ruby on Rails have many type of files on your workflow. html, css, JS, .rb, .arb, .yml.

That is true, I'll see what modes to hook it into additionally

otavioschwanck:
> > When search for projectile-commands with M-x, the shortcuts appears wrong
> 
> For me they show up correctly , but I've also just removed those prefix descriptions.
> 
> > Ruby on Rails have many type of files on your workflow. html, css, JS, .rb, .arb, .yml.
> 
> That is true, I'll see what modes to hook it into additionally

Some modes that i use on rails project:

coffee-mode, json-mode, yaml-mode, js2-mode, css-mode, scss-mode, javascript-mode, projectile-rails-server-mode, inf-ruby-mode, rspec-mode

alexandergraul:
@otavioschwanck I've added a hook into web-mode (if you enable the :lang web module). That way you will have it at least in .html and .css files. For yaml I am not sure if it is good idea to hook it into, yaml is used for a lot more than just rails. You can always do it manually `(add-hook 'yaml-mode-hook #'projectile-rails-mode)`

js2-mode already uses `<localleader> r`, I don't know what to do about that :(

otavioschwanck:
> @otavioschwanck I've added a hook into web-mode (if you enable the :lang web module). That way you will have it at least in .html and .css files. For yaml I am not sure if it is good idea to hook it into, yaml is used for a lot more than just rails. You can always do it manually `(add-hook 'yaml-mode-hook #'projectile-rails-mode)`
> 
> js2-mode already uses `<localleader> r`, I don't know what to do about that :(

maybe we can skip js-mode for this, sadly =\.  Good work anyway.

otavioschwanck:
I forget one mode, magit-status-mode.  Many times i use the SPC m r * on this.

otavioschwanck:
I found something strange:
image

Opened two .rb files, one gets enhruby and the other enhruby 2.6.3.  The SPC m r * only works on enhruby 2.6.3.

i'm using +rbenv.

otavioschwanck:
Another problem:

The Rails Server and Console are not listed on SPC , (buffer-list).

If you run the server, close, and try to open again (SPC m o s), it trys to open another server instead opening the buffer of last created server (like projectile-rails on vanilla emacs). it gives a error (port already in use)

modules/lang/ruby/config.el
```diff
@@ -163,3 +163,61 @@
...
+        "gr" #'projectile-rails-goto-routes
+        "gd" #'projectile-rails-goto-schema
+        "gs" #'projectile-rails-goto-seeds
+        "gh" #'projectile-rails-goto-spec-helper)
+        ;; Open consoles etc.
+        (:prefix "o"
+        "d" #'projectile-rails-dbconsole
+        "c" #'projectile-rails-console
+        "s" #'projectile-rails-server
+        "g" #'projectile-rails-generate)))
```
hlissner: Please correct the indentation of these two :prefix blocks (by one space).

hlissner:
> Another problem:
> 
> The Rails Server and Console are not listed on SPC , (buffer-list).

This is because Doom doesn't register them as "real" buffers. Add `doom-mark-buffer-as-real-h` to their mode's hooks.

modules/lang/ruby/config.el
otavioschwanck: would be nice rename the projectile-rails to something else:

image


its impossible to see what the commands do.

Maybe change it to rails ?

hlissner: That's not really up to us. Perhaps `(setq which-key-show-docstrings 'docstring-only)` may help?

alexandergraul:
If you want projectile-rails to always be available you can also call `(projectile-rails-global-mode)` in your private config.

modules/lang/ruby/config.el
luisenrike: There's a way to remove `projectile-rails` 
https://github.com/justbur/emacs-which-key#key-and-description-replacement

```
(push '((nil . "projectile-rails-\\(.+\\)") . (nil . "\\1"))
        which-key-replacement-alist)
```

luisenrike:
Some of commands are missing, this is one of the cons of having to manually map them, I would just assign `projectile-rails-command-map` to a different prefix and just use the default keybindings. (see [this](https://github.com/hlissner/doom-emacs/pull/2463#discussion_r374496846) for how to remove the `projectile-rails` prefix)

```
projectile-rails-find-component
projectile-rails-find-current-javascript
projectile-rails-find-current-serializer
projectile-rails-find-current-stylesheet
projectile-rails-find-rake-task
projectile-rails-find-serializer
projectile-rails-find-webpack
projectile-rails-goto-file-at-point
projectile-rails-mode-goto-map
projectile-rails-mode-run-map
```

alexandergraul:
Thanks for double checking! Yes, it seems that doing it manually is not a good way if I already missed a few. I'd really like to stick to "r" as the prefix, I can't find any other that is as intuitive to use. robe only uses `<localleader> r r` for `robe-rails-refresh`. While this is a breaking change, maybe we can change that single keybinding and keep "r" for `projectile-rails-command-map`?

luisenrike:
Yeah, `r` is the best option for `projectile-rails`, maybe move robe to `R`?

alexandergraul:
That might work, we can lose the prefix as well for a single command then

otavioschwanck:
Sometimes, the rails-server (SPC m o s) closes when press q, instead of only hiding it.

otavioschwanck:
> I found something strange:
> image
> 
> Opened two .rb files, one gets enhruby and the other enhruby 2.6.3. The SPC m r * only works on enhruby 2.6.3.
> 
> i'm using +rbenv.
> 
> It occurs on the first file opened with projectile.

About this, it is related to +rbenv flag.  Maybe the +rbenv change the mode and this PR changes doen't work.



> Sometimes, the rails-server (SPC m o s) closes when press q, instead of only hiding it.
> 
> Keep pressing SPC m o s, then, q. On second tries, it close and start a new server.

Also related to +rbenv flag.  Dont know why.

otavioschwanck:
Since this PR improves ruby support, add a shortcut to robe-jump.  Its a very nice feature.

modules/lang/ruby/config.el
alexandergraul: I've included this change in  97ee4b4

alexandergraul:
@otavioschwanck robe-jump is already on `<leader> c d` (`+lookup/definition`)

alexandergraul:
@otavioschwanck  I am not sure I understand the issue - are you saying that the rails console and rails server processes are restarted sometimes? Is this only a problem when using them with Doom (this PR) or does it also happen with projectile-rails without Doom?

otavioschwanck:
> @otavioschwanck I am not sure I understand the issue - are you saying that the rails console and rails server processes are restarted sometimes? Is this only a problem when using them with Doom (this PR) or does it also happen with projectile-rails without Doom?

it only occurs on Doom.  I've my .emacs: https://github.com/otavioschwanck/emacs_on_rails and it never happened.

The expected behaviour is after start console \ server and try to open again, it just show the buffer instead starting a new server \ console.

luisenrike:
Hi @otavioschwanck, does it happen only with this PR and not with the latest doom update? If not that's a different issue (seems like it) and not related to this one. It would be good to know this so @alexandergraul can get this one merged 🤞 , really need this stuff

otavioschwanck:
> Hi @otavioschwanck, does it happen only with this PR and not with the latest doom update? If not that's a different issue (seems like it) and not related to this one. It would be good to know this so @alexandergraul can get this one merged  , really need this stuff

i've my own config with projectile-rails with doom (installed manually) and it happens too.  I dont know  why it happen.  Without server and console, projectile-rails lose so much productivity.

alexandergraul:
For me personally, server and console are nice but not the main reason I'm using projectile-rails (which is navigation). I think we should create an issue for the closing of the console/server processes and investigate why that happens after this PR is merged. I don't mind living on this branch for my personal use, but others might want to use projectile-rails without setting it up themselves.

What do you think @hlissner ?

hlissner:
@otavioschwanck wrt the missing `SPC m r` prefix, does this fix your issue?

```elisp
(add-hook 'projectile-rails-mode-hook #'evil-normalize-keymaps)
```

As for the missing rails issue, this is because the popup manager auto-kills buffers (and their associated processes) after you close the popup by default. A popup rule needs to defined for the console buffer. e.g.

```elisp
(after! projectile-rails
  (set-popup-rule! REGEXP :ttl nil))
```

You must replace `REGEXP` with a regex that matches the name of the rails console buffer (open it and run `M-: (buffer-name)` in it).

otavioschwanck:
> @otavioschwanck wrt the missing `SPC m r` prefix, does this fix your issue?
> 
> ```emacs-lisp
> (add-hook 'projectile-rails-mode-hook #'evil-normalize-keymaps)
> ```
> 
> As for the missing rails issue, this is because the popup manager auto-kills buffers (and their associated processes) after you close the popup by default. A popup rule needs to defined for the console buffer. e.g.
> 
> ```emacs-lisp
> (after! projectile-rails
>   (set-popup-rule! REGEXP :ttl nil))
> ```
> 
> You must replace `REGEXP` with a regex that matches the name of the rails console buffer (open it and run `M-: (buffer-name)` in it).

i think it need be added in this PR. @alexandergraul

modules/lang/ruby/config.el
```diff
@@ -163,3 +163,17 @@
...
+
+
+(use-package! projectile-rails
+  :when (featurep! +rails)
+  :hook (enh-ruby-mode . projectile-rails-mode)
+  :init
+  (when (featurep! :lang web)
+    (add-hook 'web-mode-hook #'projectile-rails-mode))
+  :config
+  (map! :localleader
```
otavioschwanck: (add-hook 'projectile-rails-mode-hook #'evil-normalize-keymaps)

otavioschwanck: It works, but maybe have a better way (with regex):

  (after! projectile-rails
    (set-popup-rule! "*rails*" :ttl nil)
    (set-popup-rule! "*projectile-rails-server*" :ttl nil))

otavioschwanck:
I added in comments the fixes for the two bugs that i listed.  Thanks @hlissner !

modules/lang/ruby/config.el
alexandergraul: I've added the hook to the `:hook` section, afaik that should do the same

alexandergraul:
@hlissner I've added the two fixes, can you take another look please?

modules/lang/ruby/config.el
otavioschwanck: nice

otavioschwanck:
> @hlissner I've added the two fixes, can you take another look please?

already tested here, working flayless.  Good job @alexandergraul !

image

One idea:

On default init.el, change from:

;; ruby

to

;; (ruby
;;  +rails)

99% of ruby programmers use Rails

modules/lang/ruby/config.el
hlissner: Please remove `evil-normalize-keymaps` from `:hook` and move it into an `(add-hook 'projectile-rails-mode-hook #'evil-normalize-keymaps)` under :config.

`use-package` has a side-effect where it associates :hook functions with the containing package, as if they belonged to them. i.e. `(autoload 'evil-normalize-keymaps "projectile-rails")`, which can cause errors when Emacs tries to load `projectile-rails` looking for `evil-normalize-keymaps`. 

Since, in this particular case, evil is always eager loaded at startup, this isn't a problem here, but it will come back to bite us if a) `:lang (ruby +rails)` were loaded before `:editor evil` (for whatever reason) or b) we one day start lazy loading evil.

hlissner: Also: the evil-normalize-keymaps call should be conditional, otherwise it will throw errors for non-evil users.

alexandergraul:
Thanks for the info about `use-package`. I also forgot about non-evil users :see_no_evil: 

I've done the requested changes.

otavioschwanck:
Tested the changes, still works!

hlissner:
Looks good to me. Thanks for all your help!

PR #2465, opened 2020-02-01
Add LSP support for ESS R

sei40kr:
# Description

- Add LSP support for ESS R
- Refer R Dash doc for the symbols in an R buffer
- Add `lintr` to prequisites in the module doc

screenshot

### DIFF
--- a/modules/lang/ess/README.org
+++ b/modules/lang/ess/README.org
@@ -4,9 +4,16 @@ This module adds support for various statistics languages, including R, S-Plus,
 SAS, Julia and Stata.
 
 * Table of Contents :TOC:
-- [[Appendix][Appendix]]
-  - [[Keybindings][Keybindings]]
+- [[#prequisites][Prequisites]]
+- [[#appendix][Appendix]]
+  - [[#keybindings][Keybindings]]
 
+* Prequisites
+This module has sevral optional dependencies:
+
++ [[https://github.com/jimhester/lintr][lintr]]: Enables R linting.
++ [[https://github.com/REditorSupport/languageserver][languageserver]]: Enables LSP support in an R buffer (with =+lsp= flag).
+ 
 * Appendix
 ** Keybindings
 *** :map ess-doc-map
--- a/modules/lang/ess/README.org
+++ b/modules/lang/ess/README.org
@@ -9,7 +9,7 @@ SAS, Julia and Stata.
   - [[#keybindings][Keybindings]]
 
 * Prequisites
-This module has sevral optional dependencies:
+This module has several optional dependencies:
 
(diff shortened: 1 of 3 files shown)


modules/lang/ess/config.el
```diff
@@ -19,9 +19,15 @@
         ess-style 'DEFAULT
         ess-history-directory (expand-file-name "ess-history/" doom-cache-dir))
 
+  (set-docsets! 'ess-r-mode "R")
+  (if (featurep! +lsp)
+      (add-hook 'ess-r-mode-hook #'lsp!)
+    (set-lookup-handlers! 'ess-r-mode
+      :documentation #'ess-display-help-on-object))
```
hlissner: Should this be conditional? [We register a lookup handler for 'lsp-mode](https://github.com/hlissner/doom-emacs/blob/develop/modules/tools/lsp/config.el#L33-L36), which will take precedence. Failing that it will fall back to `ess-display-help-on-object`. Is that not what we want?

modules/lang/ess/README.org
```diff
@@ -4,9 +4,16 @@ This module adds support for various statistics languages, including R, S-Plus,
...
 
 * Table of Contents :TOC:
-- [[Appendix][Appendix]]
-  - [[Keybindings][Keybindings]]
+- [[#prequisites][Prequisites]]
+- [[#appendix][Appendix]]
+  - [[#keybindings][Keybindings]]
 
+* Prequisites
+This module has sevral optional dependencies:
```
hlissner: s/sevral/several

sei40kr:
@hlissner 
Thanks for your review. I've updated the PR.

modules/lang/ess/config.el
sei40kr: No, you're right.
There're some cases the LSP features are not usable even when `+lsp` given (e.g. the server executable not found).

modules/lang/ess/config.el
```diff
@@ -19,6 +19,10 @@
         ess-style 'DEFAULT
         ess-history-directory (expand-file-name "ess-history/" doom-cache-dir))
 
+  (set-docsets! 'ess-r-mode "R")
+  (when (featurep! +lsp)
+    (add-hook 'ess-r-mode-hook #'lsp!))
```
hlissner: Ah, sorry. I missed this the first time around. Could you change `ess-r-mode-hook` to `ess-r-mode-local-vars-hook`? This is a custom Doom hook for major modes that runs later than `ess-r-mode-hook`, in time to read file-local or dir-local variables.

sei40kr: I've fixed.

hlissner:
Looks good to me. Thanks again for all your help and prompt responses!

PR #2466, opened 2020-02-01
Bump to Fuco1/smartparens@be8d5c9

sei40kr:
# Description

Bump to Fuco1/smartparens@be8d5c9.

### DIFF
--- a/core/packages.el
+++ b/core/packages.el
@@ -19,7 +19,7 @@
 (when IS-MAC
   (package! ns-auto-titlebar :pin "1efc30d385"))
 (package! pcre2el :pin "0b5b2a2c17")
-(package! smartparens :pin "9449ae0859")
+(package! smartparens :pin "be8d5c9a63")
 (package! so-long
   :built-in 'prefer ; included in Emacs 27+
   ;; REVIEW so-long is slated to be published to ELPA eventually, but until then


alexandergraul:
I don't use Rust much and don't know if this is a worthy trade-off: [chars](https://doc.rust-lang.org/std/primitive.char.htm) use paired single-quotes `'a'`. Again, I don't know how often one would use chars compared to how often one uses lifetimes, just did not want this to be forgotten ;)

sei40kr:
@alexandergraul 
Yes, this is trade-off.
Perhaps I can make it behave better since life specifiers `'` always come after `&` or in generics types.
- `struct SomeStruct<'a> {`
- `&'a i32`

sei40kr:
@hlissner 
Thanks. I raised a PR in upstream. But we still need to update smartparens in this PR.
https://github.com/Fuco1/smartparens/pull/1003

sei40kr:
@hlissner 
I replaced commits with the one to bump to Fuco1/smartparens@be8d5c9. Please review.

hlissner:
Thanks for looking into this!

bkchr:
@sei40kr does it work for you now with `<'a>`? It still autocompletes `<'a'>` for me :( And for some reason the second `>` is only added when I do this:
```
(after! rustic
  (require 'smartparens-rust))
```

sei40kr:
@bkchr 
https://github.com/Fuco1/smartparens/blob/master/smartparens-rust.el#L62
It seems to try to handle it, but it didn't work for me.
This might be an upstream issue.

bkchr:
Ahh I thought that this fixed it for you. If this doesn't fix it, what was the reason for this pr? ^^

sei40kr:
@bkchr 
I tested only for `&'a mut` and this worked for me.
I should've confirmed `<'a>` too.

bkchr:
Ahh okay, fwiw the `&` use case worked for me before as well 🤷‍♂️

PR #2471, opened 2020-02-02
docs: update mu macos install cmd

webframp:
`--with-emacs` arg no longer exists in the formula

Just a docs PR, skipped a bunch of the template steps that aren't needed.


hlissner:

PR #2473, opened 2020-02-02
Add fstar module

tchajed:
Add a module for [F*](https://fstar-lang.org/), a language for program verification from Microsoft Research.

I roughly followed https://github.com/kyoDralliam/fstar-layer/ to get an idea of what aspects of `fstar-mode.el` should be supported but more closely mirrored the existing Coq module as far as keybindings. This way all three people that use F* and Coq won't get confused (when the other two of them switch to doom).

### DIFF
--- a/modules/lang/fstar/README.org
+++ b/modules/lang/fstar/README.org
@@ -1,3 +1,58 @@
-#+TITLE: :lang F*
+#+TITLE: :lang fstar
 
+#+TITLE:   lang/fstar
+#+DATE:    February 2, 2020
+#+SINCE:   2.0.10
+#+STARTUP: inlineimages nofold
+
+* Table of Contents :TOC_3:noexport:
+- [[#description][Description]]
+  - [[#maintainers][Maintainers]]
+  - [[#module-flags][Module Flags]]
+  - [[#plugins][Plugins]]
+- [[#prerequisites][Prerequisites]]
+  - [[#f][F*]]
+    - [[#macos][macOS]]
+    - [[#arch-linux][Arch Linux]]
+- [[#features][Features]]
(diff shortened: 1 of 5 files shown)


tchajed:
@cpitclaudel might have some thoughts as the author of `fstar-mode.el`.

cpitclaudel:
LGTM, though is there a way to put these keybindings in the upstream package, as an evil-mode config?

modules/lang/fstar/README.org
```diff
@@ -0,0 +1,3 @@
+#+TITLE: :lang F*
```
hlissner: s/F\*/fstar

modules/lang/fstar/README.org
```diff
@@ -0,0 +1,3 @@
...
+
+This module adds [[https://fstar-lang.org/][F*]] support, powered by [[https://github.com/FStarLang/fstar-mode.el][fstar-mode.el]].
```
hlissner: Please conform to and fill out our our README template: https://github.com/hlissner/doom-emacs/blob/develop/modules/editor/file-templates/templates/org-mode/__doom-readme

If you have the `:editor file-templates` module enabled, typing `__doom-readme` then TAB will expand it.

modules/lang/fstar/packages.el
```diff
@@ -0,0 +1,4 @@
+;; -*- no-byte-compile: t; -*-
+;;; lang/fstar.el
```
hlissner: Should be

```elisp
;;; lang/fstar/packages.el
```

hlissner:
Looks good to me. Thanks for the PR!

PR #2477, opened 2020-02-03
[WIP] lang/org: Improve Jupyter support

sei40kr:
## Description

* Remove `text/html` since it's not human readable
* Enable documentation lookup in in Jupyter code blocks
* Enable Jupyter code completion in Jupyter code blocks
* Integrate Jupyter code completion + company-box
* Make dwim in Jupyter code blocks jump to a next block
* Fix Jupyter + R deferred loading

### Todos

- [ ] Fix default params for Jupyter code blocks won't be loaded at the first run (existing bug, seems upstream issue)
  -> https://github.com/hlissner/doom-emacs/issues/3171
- [x] ~Enable `jupyter-org-interaction-mode` for the buffers created before `ob-jupyter` loaded~ -> done
  - It's normally enabled in `org-mode-hook`

### DIFF
--- a/modules/lang/org/contrib/jupyter.el
+++ b/modules/lang/org/contrib/jupyter.el
@@ -1,11 +1,26 @@
 ;;; lang/org/contrib/jupyter.el -*- lexical-binding: t; -*-
 ;;;###if (featurep! +jupyter)
 
+(defconst +org--company-box-icons-jupyter-alist '(("class"     . Class)
+                                                  ("function"  . Function)
+                                                  ("instance"  . Variable)
+                                                  ("keyword"   . Keyword)
+                                                  ("module"    . Module)
+                                                  ("statement" . Variable)
+                                                  ("param"     . Property)
+                                                  ("path"      . File)))
+
+(set-lookup-handlers! 'org-mode
+  :documentation '+org/jupyter-documentation-lookup-handler)
+
 (use-package! ob-jupyter
   :defer t
   :init
   (after! ob-async
-    (pushnew! ob-async-no-async-languages-alist "jupyter-python" "jupyter-julia"))
+    (pushnew! ob-async-no-async-languages-alist
+              "jupyter-python"
+              "jupyter-julia"
+              "jupyter-R"))
 
   (after! org-src
     (dolist (lang '(python julia R))
@@ -24,4 +39,12 @@
                (add-to-list 'org-src-lang-modes (cons lang-name (intern lang-tail)))))
         (with-demoted-errors "Jupyter: %s"
           (require lang nil t)
-          (require 'ob-jupyter nil t))))))
+          (require 'ob-jupyter nil t)))))
+  :config
+  (advice-add 'org-babel-jupyter-initiate-session
+              :after #'+org--ob-jupyter-initiate-session-a)
+  ;; Remove text/html since it's not human readable
+  (delq! :text/html jupyter-org-mime-types))
+
+(after! company-box
+  (push #'+org--company-box-icons-jupyter company-box-icons-functions))
--- a/modules/lang/org/autoload/contrib-jupyter.el
+++ b/modules/lang/org/autoload/contrib-jupyter.el
@@ -1,0 +1,29 @@
+;;; lang/org/autoload/contrib-jupyter.el -*- lexical-binding: t; -*-
+;;;###if (featurep! +jupyter)
+
+;;;###autoload
+(defun +org--in-jupyter-src-block-p ()
+  (when-let ((info (org-babel-get-src-block-info))
+             (lang (car info)))
+    (string-prefix-p "jupyter-" lang)))
+
+;;;###autoload
+(defun +org--ob-jupyter-initiate-session-a (&rest _)
+  (unless (bound-and-true-p jupyter-org-interaction-mode)
+    (jupyter-org-interaction-mode)))
+
+;;;###autoload
+(defun +org/jupyter-documentation-lookup-handler ()
+  (interactive)
+  (when (and (+org--in-jupyter-src-block-p)
+             (call-interactively #'jupyter-inspect-at-point))
+    (display-buffer (help-buffer))
+    t))
+
+;;;###autoload
+(defun +org--company-box-icons-jupyter (candidate)
+  (when (+org--in-jupyter-src-block-p)
+    (when-let* ((type (get-text-property 0 'annot candidate))
+                (type (string-trim type)))
+      (alist-get type +org--company-box-icons-jupyter-alist nil nil
+                 #'string-equal))))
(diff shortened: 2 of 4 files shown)


modules/lang/org/contrib/jupyter.el
```diff
@@ -1,6 +1,16 @@
...
 
+(use-package! jupyter-org-client
+  :commands jupyter-org-interaction-mode
+  :init
+  (add-hook 'org-mode-hook 'jupyter-org-interaction-mode)
+  :config
+  (delq! :text/html jupyter-org-mime-types))
+
+(after! (jupyter-org-client jupyter-org-extensions)
+  (setq jupyter-org-interaction-mode-map (make-sparse-keymap)))
```
hlissner: Could this be simplified to a single package? Perhaps the package that defines `jupyter-org-interaction-mode-map`, or even set in `(after! jupyter ...)` to keep things simple for the end user, if they want to customize this keymap for some reason or another.

Which of these two packages defines this variable? Perhaps it'd be simplest to put this in the `use-package!` in `(after! jupyter ...)`, becaus

modules/lang/org/contrib/jupyter.el
```diff
@@ -1,6 +1,16 @@
 ;;; lang/org/contrib/jupyter.el -*- lexical-binding: t; -*-
 ;;;###if (featurep! +jupyter)
 
+(use-package! jupyter-org-client
+  :commands jupyter-org-interaction-mode
+  :init
+  (add-hook 'org-mode-hook 'jupyter-org-interaction-mode)
```
hlissner: Use `:hook (org-mode . jupyter-org-interaction-mode)` here and remove `:commands jupyter-org-interactive-mode` (the former implies the latter).

modules/lang/org/contrib/jupyter.el
```diff
@@ -1,6 +1,16 @@
...
+  :config
+  (delq! :text/html jupyter-org-mime-types))
```
hlissner: Could you add a comment here to explain what this is for and html output isn't valid? (I presume because the output is difficult to read)

modules/completion/company/autoload.el
```diff
@@ -4,6 +4,7 @@
 (defvar +company-backend-alist
   '((text-mode company-dabbrev company-yasnippet company-ispell)
     (prog-mode company-capf company-yasnippet)
+    (org-mode company-capf company-dabbrev company-files company-yasnippet)
```
hlissner: Instead of this, use `(set-company-backend! 'org-mode 'company-capf)` below this line: https://github.com/hlissner/doom-emacs/blob/develop/modules/lang/org/config.el#L980. This will prepend `company-capf` to the default `text-mode` backends, and is easier to keep track of when org-specific configuration is kept together.

sei40kr:
@hlissner 
I've updated the PR, please review.

sei40kr:
I made this wip again,

It throws error `Need a valid session and a kernel to form a key` on the first run

sei40kr:
@hlissner 
Could you review again?

modules/lang/org/contrib/jupyter.el
```diff
@@ -1,11 +1,22 @@
 ;;; lang/org/contrib/jupyter.el -*- lexical-binding: t; -*-
 ;;;###if (featurep! +jupyter)
 
+(set-company-backend! 'org-mode
+  'company-capf
+  'company-dabbrev
+  'company-files
+  'company-yasnippet)
+(set-lookup-handlers! 'org-mode
+  :documentation '+org-jupyter-lookup-documentation)
```
hlissner: Remove these for now; they are too opinionated. I think a more general solution is needed for the documentation handler. And the company backends should probably be set in lang/org/config.el.

I'm already working on babel-aware eval and lookup handlers for org-mode, and will look into better code completion in org, so you can leave that to me.

modules/lang/org/contrib/jupyter.el
```diff
@@ -24,4 +35,12 @@
                (add-to-list 'org-src-lang-modes (cons lang-name (intern lang-tail)))))
         (with-demoted-errors "Jupyter: %s"
           (require lang nil t)
-          (require 'ob-jupyter nil t))))))
+          (require 'ob-jupyter nil t)))))
+  :config
+  (setq jupyter-org-interaction-mode-map (make-sparse-keymap))
```
hlissner: Could you add a comment here to explain why this is useful/necessary?

modules/lang/org/contrib/jupyter.el
sei40kr: > I'm already working on babel-aware eval and lookup handlers for org-mode, and will look into better code completion in org, so you can leave that to me.
@hlissner Any progress on this?

adder:
Just letting know that this PR would be very useful to me as some default functionality from org-jupyter is not working.
As I gather from above there is only a comment missing from the author.
Any change this PR could be just accepted?

sei40kr:
> Just letting know that this PR would be very useful to me as some default functionality from org-jupyter is not working.
> As I gather from above there is only a comment missing from the author.
> Any change this PR could be just accepted?

@hlissner 
> I'm already working on babel-aware eval and lookup handlers for org-mode, and will look into better code completion in org, so you can leave that to me.
Since it seems to take much time, can't we merge this PR now?

drchsl:
>     * Remove `text/html` since it's not human readable

I'd rather keep the output handling as close to "vanilla" emacs-jupyter as possible (and it's docs) - this change would confuse me.

benneti:
slightly of topic:
does `(set-company-backend! 'org-mode 'company-capf)` work for you as expected?
If I add it to my config.el file and then try `\alpha` TAB (in a jupyter julia block) it should expand to the unicode alpha, but it does not. Inside the jupyter repl it works.

sei40kr:
@drchsl 
Thank you for your opinion. I'm updating the PR.

@hlissner 
lsp-mode supports lsp-mode in a Org source block.
I think it'd be better to use LSP for completion if available, because it completes without executing a source block at once.
But for documentation lookup handler, I prefer the emacs-jupyter's one because it shows more information we can't get at compile-time. How do you think?

visini:
Looking forward to these changes!

drchsl:
Looking great! Would love to see this merged soon, as +jupyter is in a broken state right now.

> lsp-mode supports lsp-mode in a Org source block.
Am i mistaken, or does this only work when the blocks are tangled?

sei40kr:
@drchsl 
Currently we manually have to call `lsp-org` in a code block.
It works without `:tangle` but then the LSP will process a code block as one file, unlike Jupyter that treats code blocks which has same `:session` as one file.

robbyoconnor:
Rebase needed.

hlissner:
Sorry for the tremendously late response. As of bf90423 the fixes suggested in this PR are now merged into `develop` (with credit). Thanks for bringing it to my attention!

PR #2495, opened 2020-02-06
Fix recenter after evil search

dsdshcym:
This PR adds after-advice to following evil search functions:
- evil-ex-search-word-forward
- evil-ex-search-next    
- evil-ex-search-previous

### DIFF
--- a/modules/editor/evil/config.el
+++ b/modules/editor/evil/config.el
@@ -159,7 +159,7 @@ directives. By default, this only recognizes C directives.")
   ;; Recenter screen after most searches
   (dolist (fn '(evil-visualstar/begin-search-forward
                 evil-visualstar/begin-search-backward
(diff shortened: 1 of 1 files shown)


hlissner:
Could you try `(setq scroll-conservatively 10)` instead? I'm wondering if this is a better solution to all these advice functions.

dsdshcym:
@hlissner I did a test after removing these recenter advice and set `scroll-conservatively` to 10. But the behavior is different: recenter only happens when the search occurrence is off screen by at least 10 lines.
I don't think it's the desired behavior. So I'd suggest we either advice all the evil search functions (this PR), or not advice any functions to get a consistent behavior. WDYT?

hlissner:
Understood. I'll go with advising all the evil search functions. At least until a better solution pops up. Thanks for the help!

PR #2501, opened 2020-02-08
Also look for 'fdfind' executable

yoavm448:
This makes `+ivy--counsel-file-jump-use-fd-rg-a` also search for the `fdfind`
executable, and not only `fd`. (On debian its called 'fdfind'.)


yoavm448:
Did it. works as expected.

hlissner:
Great! Thanks again!

PR #2502, opened 2020-02-09
Add LSP support for YAML

sei40kr:
## Description

* Move `yaml-mode` in `data` to a new module `yaml`
* Enable LSP for `yaml-mode` if `+lsp` is present, which enables:
  * code completion
  * documentation
  * diagnostics
    (these are based on schemas downloaded by [JSON Schema Store](http://schemastore.org/json)).

### DIFF
--- a/modules/lang/data/config.el
+++ b/modules/lang/data/config.el
@@ -43,7 +43,10 @@
   (set-electric! 'jsonnet-mode :chars '(?\n ?: ?{ ?})))
 
 (after! yaml-mode
-  (setq-hook! 'yaml-mode-hook tab-width yaml-indent-offset))
+  (setq-hook! 'yaml-mode-hook tab-width yaml-indent-offset)
+
+  (when (featurep! +yaml-lsp)
+    (add-hook 'yaml-mode-local-vars-hook #'lsp!)))
 
 
 ;;


hlissner:
Is there any reason this couldn't have been done within the `:lang data` module? I don't think this yaml module is large enough to warrant its own module.

sei40kr:
@hlissner 
It's because we may have LSP on another data language (e.g. json) in the future.
Also I thought giving `+lsp` to `data` module to enable LSP on `yaml-mode` is a little bit confusing.

yoavm448:
Maybe enable lsp for yaml with a `+lsp-yaml` flag under `:lang data`?

sei40kr:
@yoavm448 Thanks for your opinion.

@hlissner 
I've updated the PR. Please review.

yoavm448:
Should probably also update that `:lang data` accepts a `+yaml-lsp` flag under `docs/modules.org`. It should also be in the table in `:tools lsp`'s docs.

hlissner:
Sorry for the delay. I am reconsidering moving yaml (and json) to their own modules. Possibly the others, but I haven't quite decided how. I'll post here again once I have.

Emiller88:
Just some input both `json` and `yaml` are built-in to lsp-mode so if you have `:tools lsp`, so install the language server `yaml` works out of the box for me, with no extra flags.

hlissner:
Sorry for the late response. As of 4a8c481 I've split the yaml and json packages into their own modules, including the changes you've proposed here. Thanks for bringing it to my attention!

PR #2504, opened 2020-02-09
add benchmark bindings (,tbs and ,tba)

nl5887:
go supports running benchmarks much similar to running tests,
the new key bindings will run single or all benchmarks.


hlissner:
Thanks for the PR!

PR #2508, opened 2020-02-09
Fix list passed as argument to executable-find

johanthoren:
Previously, a list would be passed as an argument to 'executable-find' in '+ledger--check-version-a'.

This instead passes "hledger" as an argument if found, else "ledger".

### DIFF
--- a/modules/lang/ledger/config.el
+++ b/modules/lang/ledger/config.el
@@ -10,7 +10,7 @@
   (setq ledger-binary-path
         (if (executable-find "hledger")
             "hledger"
-            "ledger"))
+          "ledger"))
 
   (defadvice! +ledger--check-version-a (orig-fn)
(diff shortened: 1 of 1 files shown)


modules/lang/ledger/config.el
```diff
@@ -8,7 +8,8 @@
 
   :config
   (setq ledger-binary-path
-        (or (cl-delete-if-not #'executable-find (list "hledger" "ledger"))
+        (if (executable-find "hledger")
+            "hledger"
```
hlissner: Indent this by two more spaces.

hlissner:
Derp, forgot I could just do it myself in github. Thanks for the help!

johanthoren:
> Ah, my bad. One small fix and I will merge this.

Sorry, I'm used to clojure convention. Thanks for fixing it!

PR #2518, opened 2020-02-11
Make doom purge -g more effective

bepvte:
Reflog keeps the history of every reset + pull, so expiring it allows us to remove more older commits. git gc by default only cleans for 2 week old objects, so I made it more aggressive.


hlissner:
Great idea! Thanks for the PR

PR #2527, opened 2020-02-12
Change org-re-reveal reveal.js root path to https://revealjs.com

soedirgo:
In org-re-reveal, a reveal.js plugin will look for things relative to the path specified by the variable `org-re-reveal-root`. E.g. the `notes` plugin will look for `.../plugin/notes/notes.html`. But the current URL gives the `notes.html` file in .txt form, rendering the plugin unusable.

Instead, consider using `https://revealjs.com`, which is one of the customization option of the variable [[1]](https://gitlab.com/oer/org-re-reveal/-/blob/master/org-re-reveal.el#L228).

### DIFF
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -442,7 +442,7 @@ file isn't in `org-directory'."
     (setq org-pandoc-options
           '((standalone . t)
             (mathjax . t)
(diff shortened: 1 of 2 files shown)


hlissner:
Thanks for the help!

PR #2529, opened 2020-02-12
Enable flyspell in git-commit-mode

bixuanzju:
Not sure why flyspell was not enabled in text-mode before. For me, it's a bit annoying when I write commit messages in magit, but it's NOT spell-checked.

### DIFF
--- a/modules/checkers/spell/config.el
+++ b/modules/checkers/spell/config.el
@@ -51,7 +51,8 @@
                TeX-mode-hook
                rst-mode-hook
                mu4e-compose-mode-hook
-               message-mode-hook)
+               message-mode-hook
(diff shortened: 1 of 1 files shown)


bixuanzju:
> Please replace `text-mode-hook` with `git-commit-mode`. It was excluded from `text-mode-hook` intentionally, because there are quite a few markup/config major modes that are derived from text-mode, where flyspell-mode isn't desired.
> 
> Instead, I'd rather add flyspell-mode to the modes where it is wanted.

Thanks, use `git-commit-mode` instead.

hlissner:
Great! Thanks for the prompt response and PR!

PR #2537, opened 2020-02-14
Fix `+file-template-p` to use `:when` keyword

licebmi:
Based on the documentation of `set-file-template!`, it appears to me that the `:when` keyword should be used as a secondary predicate after matching based on the regex, based on commit 3f759dde2354fb81e91407c8df21d23b99203fe5, it seems like it was previously the case.

As it is, `+file-template-p` will match based on the first predicate, either a symbol or a regex, and will ignore any predicates specified with the keyword `:when`

This commit fixes said behavior.


hlissner:
Ah, I've been meaning to get to this. Thanks for the help!

PR #2553, opened 2020-02-18
Add LSP for Terraform

howdoicomputer:
Emacs LSP has support for Terraform these days.

### DIFF
--- a/modules/tools/terraform/config.el
+++ b/modules/tools/terraform/config.el
@@ -1,5 +1,9 @@
 ;;; tools/terraform/config.el -*- lexical-binding: t; -*-
 
+(when (featurep! +lsp)
+  (add-hook 'terraform-mode-local-vars-hook #'lsp!)
+
+
 (map! :after terraform-mode
       :map terraform-mode-map
       :localleader
--- a/modules/tools/terraform/config.el
+++ b/modules/tools/terraform/config.el
@@ -13,3 +13,8 @@
   :after terraform-mode
   :config
   (set-company-backend! 'terraform-mode 'company-terraform))
+
+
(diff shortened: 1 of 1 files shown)


modules/tools/terraform/config.el
```diff
@@ -13,3 +13,8 @@
   :after terraform-mode
   :config
   (set-company-backend! 'terraform-mode 'company-terraform))
+
+
+(after! terraform-mode
+    (if (featurep! +lsp)
+      (add-hook 'terraform-mode-local-vars-hook #'lsp!))
```
hlissner: 1. No need for an after! block for a hook.
2. Move it to the top, above the `map!` call, to keep like config together.
3. Use `when` when there is no else clause.
4. The indentation is off.

e.g.

```elisp
(when (featurep! +lsp)
  (add-hook 'terraform-mode-local-vars-hook #'lsp!))
```

howdoicomputer: Alright! Learning elisp.

hlissner:
Looks good to me. Thanks for the help and quick response!

modules/tools/terraform/config.el
```diff
@@ -1,5 +1,9 @@
 ;;; tools/terraform/config.el -*- lexical-binding: t; -*-
 
+(when (featurep! +lsp)
+  (add-hook 'terraform-mode-local-vars-hook #'lsp!)
```
robsonpeixoto: Missing a `)`

howdoicomputer: https://github.com/hlissner/doom-emacs/pull/2566

PR #2555, opened 2020-02-18
bindings: correct description type-o

UndeadKernel:
Fix the description of the binding `C-c f X` binding for emacs users. The correct and new description is "Switch to scratch buffer".


hlissner:
Thanks for the help!

PR #2557, opened 2020-02-18
Add install instructions for scoop emacs

howdoicomputer:
I noticed the conversation in Discord. Updated docs for installing emacs via Scoop.

Emacs is in the 'extras' repo/bucket because Scoop only keeps CLI tools in the 'main' bucket.

https://github.com/lukesampson/scoop/wiki/Criteria-for-including-apps-in-the-main-bucket
https://github.com/lukesampson/scoop-extras/blob/master/bucket/emacs.json

### DIFF
--- a/docs/getting_started.org
+++ b/docs/getting_started.org
@@ -268,9 +268,15 @@ choco install git emacs ripgrep fd llvm
 #+END_SRC
 
 #+begin_quote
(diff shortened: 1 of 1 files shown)


hlissner:
Good to know. Thanks for the correction!

PR #2558, opened 2020-02-18
Update to latest lsp-mode, lsp-ui

sebastiansturm:
This updates `lsp-mode` and `lsp-ui`, pulling in 2 performance improvements (https://github.com/emacs-lsp/lsp-mode/commit/da406084aef627eedaadeb94d96e1ddd53860673 and https://github.com/emacs-lsp/lsp-mode/commit/2fc0963a50488c47a5be3836ad042380b71d1cfb). There was one small breaking change introduced by https://github.com/emacs-lsp/lsp-mode/commit/4d59bded6bbba6d3bc73032305d1b3c071eb73c4 which I have tried to address; noticed no other breaking changes.
Opening this PR so other people can test it as well; I've only given it a few hours of superficial testing so far (using C++ and TypeScript language servers)

### DIFF
--- a/modules/tools/lsp/config.el
+++ b/modules/tools/lsp/config.el
@@ -27,8 +27,9 @@ This can be a single company backend or a list thereof. It can be anything
         lsp-intelephense-storage-path (concat doom-cache-dir "lsp-intelephense/"))
 
   :config
-  (when (and lsp-auto-configure lsp-auto-require-clients)
-    (require 'lsp-clients))
+  (when lsp-auto-configure
+    (mapc (lambda (package) (require package nil t))
+          lsp-client-packages))
 
   (set-lookup-handlers! 'lsp-mode :async t
     :documentation 'lsp-describe-thing-at-point
--- a/modules/tools/lsp/packages.el
+++ b/modules/tools/lsp/packages.el
@@ -1,8 +1,8 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; tools/lsp/packages.el
 
-(package! lsp-mode :pin "e7c7abf236")
-(package! lsp-ui :pin "6caacc8c93")
+(package! lsp-mode :pin "2fc0963a50")
+(package! lsp-ui :pin "e8200e3b72")
 (when (featurep! :completion company)
   (package! company-lsp :pin "f921ffa0cd"))
 (when (featurep! :completion ivy)


modules/tools/lsp/config.el
```diff
@@ -27,8 +27,9 @@ This can be a single company backend or a list thereof. It can be anything
         lsp-intelephense-storage-path (concat doom-cache-dir "lsp-intelephense/"))
 
   :config
-  (when (and lsp-auto-configure lsp-auto-require-clients)
-    (require 'lsp-clients))
+  (when (and lsp-auto-configure)
+    (seq-do (lambda (package) (require package nil t))
+              lsp-client-packages))
```
hlissner: Please use `mapc`. Doom does not (explicitly) load the `seq` library in interactive sessions, so this is not guaranteed to work.

hlissner: Also, is there any possibility that a package in `lsp-client-packages` will fail to load? If not, `(mapc #'require lsp-client-packages)` is more concise.

modules/tools/lsp/config.el
```diff
@@ -27,8 +27,9 @@ This can be a single company backend or a list thereof. It can be anything
         lsp-intelephense-storage-path (concat doom-cache-dir "lsp-intelephense/"))
 
   :config
-  (when (and lsp-auto-configure lsp-auto-require-clients)
-    (require 'lsp-clients))
+  (when (and lsp-auto-configure)
```
hlissner: Remove the `(and ...)` if there is only one argument.

sebastiansturm: ouch, that's a dumb one. Thanks for all your suggestions, will do that

modules/tools/lsp/config.el
sebastiansturm: as to whether loading may fail: I think so; as far as I can see, `lsp-client-packages` contains several packages that are not part of `lsp-mode` but distributed separately (`lsp-python-ms`, `lsp-java`, `lsp-haskell`, maybe more)

modules/tools/lsp/config.el
sebastiansturm: have made the changes (didn't squash commits though, should I?)

hlissner: @wildsebastian Yes, please do.

sebastiansturm: done

hlissner:
Thanks for the help!

PR #2561, opened 2020-02-19
README: remove dangling link for example Doom configurations page.

vincentbernat:
The target does not exist and I didn't find anything related in the
current repository or in past iterations.


hlissner:
Ah, the file hasn't been pushed yet. I've held off on it because I haven't figured out the rules for submission and how I'll present the listing. Since the file will eventually exist, I will refuse this PR, but thanks for reminding me!

PR #2562, opened 2020-02-19
Fix missing parenthesis in terraform config

babariviere:
A parenthesis was missing in PR #2553, this makes doom fail to load his config.


rohitpaulk:
:+1: Just ran into this. Workaround: disable `terraform` in `init.el`

PR #2572, opened 2020-02-20
lang/ruby: Add detection for additional files

msanders:
This adds support for podspecs used by Cocoapods, Brewfiles from Homebrew Bundle, and Fastfiles from Fastlane.


hlissner:
Looks good to me. Thanks!

PR #2574, opened 2020-02-20
fix conflicting golang import bindings

nl5887:
the import "i" binding will jump to imports, instead of opening import prefix


hlissner:
I do not see a conflict. The original keybind was `ri`, not `i`.

nl5887:
I had issues with the keybind `ri`, but probably that was caused by something different. Quite new to lisp / emacs incorrectly assumed the prefix didn't work.

hlissner:
No worries! In that case I'll consider this resolved.

PR #2580, opened 2020-02-22
Change clojure linter to clj-kondo

yuhan0:
[Clj-kondo](https://github.com/borkdude/clj-kondo) is a much more full-featured and actively developed linter compared to Joker, and is set as the (recommended) default on Spacemacs and the VScode Calva plugin.


hlissner:
Makes sense to me. Thanks!

PR #2582, opened 2020-02-22
completion/ivy: sort projectile files with prescient

bricewge:
The behavior is explained by prescient author in https://github.com/raxod502/prescient.el/issues/35#issuecomment-471206879.


hlissner:
Makes sense to me. Thanks!

PR #2585, opened 2020-02-23
VTerm remembers insert point when leaving normal mode

mbriggs:
It is useful to be able to enter normal mode in vterm, but going back to insert
gets weird, because there is only one place where inserting makes sense in
vterm; where you left it.

Setting some buffer local hooks for insert entry/exit, and storing `(point)` in
a buffer local variable.

This is my first PR to doom :) Thank you for this awesome project, brought me
back to emacs <3

### DIFF
--- a/modules/term/vterm/config.el
+++ b/modules/term/vterm/config.el
@@ -16,11 +16,10 @@
   (setq vterm-kill-buffer-on-exit t)
 
   (when (featurep! :editor evil)
-    (defvar +vterm-insert-point nil)
-    (add-hook 'vterm-mode-hook
-      (lambda ()
-        (add-hook 'evil-insert-state-exit-hook '+vterm-remember-insert-point-h nil t)
-        (add-hook 'evil-insert-state-entry-hook '+vterm-goto-insert-point-h nil t))))
+    (add-hook! 'vterm-mode-hook
+      (defun +vterm-init-remember-point-h ()
+        (add-hook 'evil-insert-state-exit-hook #'+vterm-remember-insert-point-h nil t)
+        (add-hook 'evil-insert-state-entry-hook #'+vterm-goto-insert-point-h nil t))))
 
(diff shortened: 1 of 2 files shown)


modules/term/vterm/config.el
```diff
@@ -15,6 +15,13 @@
 
   (setq vterm-kill-buffer-on-exit t)
 
+  (when (featurep! :editor evil)
+    (defvar +vterm-insert-point nil)
```
hlissner: Please move this `defvar` to `vterm/autoload.el`. Implementation details should be kept together. Also, rename it to `+vterm--insert-point` to communicate that it is an internal variable.

modules/term/vterm/config.el
```diff
@@ -15,6 +15,13 @@
...
+    (add-hook 'vterm-mode-hook
+      (lambda ()
+        (add-hook 'evil-insert-state-exit-hook '+vterm-remember-insert-point-h nil t)
+        (add-hook 'evil-insert-state-entry-hook '+vterm-goto-insert-point-h nil t))))
```
hlissner: Please use `add-hook!` and sharp-quote the function symbols:

```elisp
(add-hook! 'vterm-mode-hook
  (defun +vterm-init-remember-point-h ()
    (add-hook 'evil-insert-state-exit-hook #'+vterm-remember-insert-point-h nil t)
    (add-hook 'evil-insert-state-entry-hook #'+vterm-goto-insert-point-h nil t)))
```

This makes it easier for users to disable, if they prefer.

modules/term/vterm/config.el
mbriggs: fixed. I was a bit hesitant about putting non autoload stuff in autoload.el, but what you are saying makes sense :)

modules/term/vterm/config.el
mbriggs: fixed. Didn't know that :)

hlissner:
Great work! Thanks for the help and prompt responses!

PR #2590, opened 2020-02-23
Modernize lsp integration

sebastiansturm:
I'm opening this PR without thorough testing, though I've tried every change I introduced at least once and so far it seemed to work with my setup. What this PR does (currently in separate commits, in case some changes appear desirable and others don't):

1. `company-lsp` is replaced by lsp-mode's newly introduced `capf` functionality. According to the `lsp-mode` performance guide, this improves performance (and it reduces the number of required packages by one). Since I don't know of any capf downsides compared to `company-lsp`, I have decided not to make `company-lsp` a user-selectable alternative. This commit also bumps `lsp-mode` and `lsp-ui` versions and pulls in one `capf`-related improvement (https://github.com/emacs-lsp/lsp-mode/commit/1b5963dc04c6aa2f29c508ed3863f99b932974ac)

2. By default, doom-emacs uses `lsp-ui`'s peek interface for semantic navigation whenever `lsp-ui` is in use. While I find the peek interface very pretty to look at, the `xref`-based interface is more useful to me, especially when used together with `ivy` and `ivy-occur`. I therefore bound the peek interface to a new feature `+peek` (though, on second thought, this change would require users to change their configs to preserve former behavior. I should probably change this to `+no-peek` or something?)

3. recent builds of Emacs 27+ define a variable `read-process-output-max` that, when increased beyond its default value of 4KB, is said to offer significant performance benefits to `lsp-mode` users (see the `lsp-mode` performance guide). Whenever this value is available, this commit increases it to 1 MiB

4. `lsp-mode`'s flycheck/flymake integration has recently been moved from `lsp-ui` to lsp proper. This commit follows suit; since `lsp-ui-flycheck` remains available for backwards compatibility, this shouldn't affect the user experience at all but allows users to completely exclude `lsp-ui` without losing flycheck/flymake integration

### DIFF
--- a/modules/tools/lsp/config.el
+++ b/modules/tools/lsp/config.el
@@ -1,11 +1,20 @@
 ;;; tools/lsp/config.el -*- lexical-binding: t; -*-
 
 (defvar +lsp-company-backend 'company-lsp
-  "What backend to prepend to `company-backends' when `lsp-mode' is active.
+  "What backend to use for lsp-driven autocompletion, unless
+overridden by `+lsp-capf-blacklist'.
 
-This can be a single company backend or a list thereof. It can be anything
-`company-backends' will accept.")
+While `company-capf' does not require the `company-lsp' package
+and should offer better performance, it has been integrated into
+lsp only recently and as of 02/25/2020 is known to cause issues
+with some language servers. If you wish to use `company-capf' in
+general but fall back to `company-lsp' for specific language
+servers, set `+lsp-company-backend' to `company-capf' and add the
+excluded servers' identifiers to `+lsp-capf-blacklist'.")
 
+(defvar +lsp-capf-blacklist '(ts-ls gopls)
+  "Language servers listed here will always use the `company-lsp' backend,
+irrespective of what `+lsp-company-backend' is set to.")
 
 ;;
 ;;; Packages
@@ -20,6 +29,42 @@ This can be a single company backend or a list thereof. It can be anything
   ;; Auto-kill LSP server once you've killed the last buffer associated with its
   ;; project.
   (setq lsp-keep-workspace-alive nil)
+  (add-hook! 'lsp-mode-hook
+    (defun +lsp-init-company-h ()
+      (if (not (bound-and-true-p company-mode))
+          (add-hook 'company-mode-hook #'+lsp-init-company-h t t)
+        (let ((preferred-backend +lsp-company-backend))
+          (lsp-foreach-workspace
+           (when (memq (lsp--client-server-id (lsp--workspace-client lsp--cur-workspace))
+                       +lsp-capf-blacklist)
+             (setq preferred-backend 'company-lsp)))
+          (if (eq 'company-capf preferred-backend)
+              ;; use capf backend
+              (progn
+                (setq-local lsp-enable-completion-at-point t)
+                (setq-local lsp-prefer-capf t)
+                (setq-local company-backends
+                            (cons 'company-capf (remq 'company-capf company-backends))))
+            ;; use company-lsp backend (may need to be loaded first)
+            (require 'company-lsp)
+            (setq-local lsp-enable-completion-at-point nil)
+            (setq-local lsp-prefer-capf nil)
+            (setq-local company-backends
+                        (cons 'company-lsp (remq 'company-capf company-backends)))
+            (setq-default company-lsp-cache-candidates 'auto))
+          (remove-hook 'company-mode-hook #'+lsp-init-company-h t))))
+    (defun +lsp-init-flycheck-or-flymake-h ()
+      "Set up flycheck-mode or flymake-mode, depending on `lsp-diagnostic-package'."
+      (cond ((eq :none lsp-diagnostic-package))
+            ((or (eq :flymake lsp-diagnostic-package)
+                 (eq t lsp-diagnostic-package))
+             (lsp--flymake-setup))
+            ((require 'flycheck nil t)
+             (let ((old-checker flycheck-checker))
+               (lsp-flycheck-enable t)
+               (when old-checker
+                 (setq-local flycheck-checker old-checker)
+                 (kill-local-variable 'flycheck-check-syntax-automatically)))))))
 
   ;; For `lsp-clients'
   (setq lsp-server-install-dir (concat doom-etc-dir "lsp/")
@@ -79,7 +124,7 @@ modules.
 
 Also see:
 + `+lsp-init-company-h' (on `lsp-mode-hook')
-+ `+lsp-init-ui-flycheck-or-flymake-h' (on `lsp-ui-mode-hook')
++ `+lsp-init-flycheck-or-flymake-h' (on `lsp-mode-hook')
 
 This also logs the resolved project root, if found, so we know where we are."
     :override #'lsp
@@ -92,6 +137,10 @@ This also logs the resolved project root, if found, so we know where we are."
               (lsp--try-project-root-workspaces
                (equal arg '(4))
                (and arg (not (equal arg 1))))))
+         ;; read-process-output-max is only available on recent
+         ;; development builds of Emacs 27 and above
+         (when (boundp 'read-process-output-max)
+           (setq-local read-process-output-max (* 1024 1024)))
          (prog1 (lsp-mode 1)
            (setq-local lsp-buffer-uri (lsp--buffer-uri))
            ;; Announce what project root we're using, for diagnostic purposes
@@ -110,24 +159,9 @@ This also logs the resolved project root, if found, so we know where we are."
 
 (use-package! lsp-ui
   :hook (lsp-mode . lsp-ui-mode)
-  :init
-  (add-hook! 'lsp-ui-mode-hook
-    (defun +lsp-init-ui-flycheck-or-flymake-h ()
-      "Sets up flymake-mode or flycheck-mode, depending on `lsp-prefer-flymake'."
-      (cond ((eq :none lsp-prefer-flymake))
-            (lsp-prefer-flymake
-             (lsp--flymake-setup))
-            ((require 'flycheck nil t)
-             (require 'lsp-ui-flycheck)
-             (let ((old-checker flycheck-checker))
-               (lsp-ui-flycheck-enable t)
-               (when old-checker
-                 (setq-local flycheck-checker old-checker)
-                 (kill-local-variable 'flycheck-check-syntax-automatically)))))))
 
   :config
-  (setq lsp-prefer-flymake nil
-        lsp-ui-doc-max-height 8
+  (setq lsp-ui-doc-max-height 8
         lsp-ui-doc-max-width 35
         lsp-ui-sideline-ignore-duplicate t
         ;; lsp-ui-doc is redundant with and more invasive than
@@ -138,27 +172,10 @@ This also logs the resolved project root, if found, so we know where we are."
         ;; errors flash briefly and then disappear).
         lsp-ui-sideline-show-hover nil)
 
-  (set-lookup-handlers! 'lsp-ui-mode :async t
-    :definition 'lsp-ui-peek-find-definitions
-    :references 'lsp-ui-peek-find-references))
-
-
-(use-package! company-lsp
-  :when (featurep! :completion company)
-  :defer t
-  :init
-  ;; Make sure that `company-capf' is disabled since it is incompatible with
-  ;; `company-lsp' (see lsp-mode#884)
-  (add-hook! 'lsp-mode-hook
-    (defun +lsp-init-company-h ()
-      (if (not (bound-and-true-p company-mode))
-          (add-hook 'company-mode-hook #'+lsp-init-company-h t t)
-        (setq-local company-backends
-                    (cons +lsp-company-backend
-                          (remq 'company-capf company-backends)))
-        (remove-hook 'company-mode-hook #'+lsp-init-company-h t))))
-  :config
-  (setq company-lsp-cache-candidates 'auto)) ;; cache candidates for better performance
+  (when (featurep! +peek)
+    (set-lookup-handlers! 'lsp-ui-mode :async t
+      :definition 'lsp-ui-peek-find-definitions
+      :references 'lsp-ui-peek-find-references)))
 
 
 (use-package! helm-lsp
--- a/docs/modules.org
+++ b/docs/modules.org
@@ -162,7 +162,7 @@ Small modules that give Emacs access to external tools & services.
 + gist - TODO
 + [[file:../modules/tools/lookup/README.org][lookup]] =+dictionary +docsets= - Universal jump-to & documentation lookup
   backend
-+ [[file:../modules/tools/lsp/README.org][lsp]] - TODO
++ [[file:../modules/tools/lsp/README.org][lsp]] =+peek= - TODO
 + macos - TODO
 + magit - TODO
 + make - TODO
--- a/modules/tools/lsp/README.org
+++ b/modules/tools/lsp/README.org
@@ -41,7 +41,7 @@ As of this writing, this is the state of LSP support in Doom Emacs:
 | [[../../lang/haskell/README.org][:lang haskell]]    | haskell-mode                                            | haskell-ide-engine                                            |
 | [[../../lang/java/README.org][:lang java]]       | java-mode                                               | lsp-java                                                      |
 | [[../../lang/javascript/README.org][:lang javascript]] | js2-mode, rjsx-mode, typescript-mode                    | typescript-language-server                                    |
-| [[../../lang/ocaml/README.org][:lang ocaml]]      | taureg-mode                                             | ocaml-language-server                                         |
+| [[../../lang/ocaml/README.org][:lang ocaml]]      | tuareg-mode                                             | ocaml-language-server                                         |
 | [[../../lang/php/README.org][:lang php]]        | php-mode                                                | php-language-server                                           |
 | [[../../lang/python/README.org][:lang python]]     | python-mode                                             | lsp-python-ms                                                 |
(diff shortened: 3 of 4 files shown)


modules/tools/lsp/config.el
```diff
@@ -102,32 +101,37 @@ This also logs the resolved project root, if found, so we know where we are."
...
+                              lsp--buffer-workspaces))))
+         (when (featurep! :completion company)
+           (defun +lsp-init-company-h ()
+             (if (not (bound-and-true-p company-mode))
+                 (add-hook 'company-mode-hook #'+lsp-init-company-h t t)
+               ;; make sure company-capf is the first backend to be tried
+               (setq-local company-backends
+                           (cons 'company-capf
+                                 (remq 'company-capf company-backends)))
+               (remove-hook 'company-mode-hook #'+lsp-init-company-h t))))
```
hlissner: This only defines the function, without calling it. Add this to `lsp-before-initialize-hook` instead, to make it easier for users to disable, if they prefer.

modules/tools/lsp/config.el
```diff
@@ -102,32 +101,37 @@ This also logs the resolved project root, if found, so we know where we are."
...
+         (cond ((eq :none lsp-diagnostic-package))
+               ((or (eq :flymake lsp-diagnostic-package)
+                    (eq t lsp-diagnostic-package))
+                (lsp--flymake-setup))
+               ((require 'flycheck nil t)
+                (let ((old-checker flycheck-checker))
+                  (lsp-flycheck-enable t)
+                  (when old-checker
+                    (setq-local flycheck-checker old-checker)
+                    (kill-local-variable 'flycheck-check-syntax-automatically)))))))
```
hlissner: This should also be added to `lsp-before-initialize-hook`, in a `+lsp-init-flycheck-or-flymake-h` function.

modules/tools/lsp/config.el
```diff
@@ -138,27 +142,10 @@ This also logs the resolved project root, if found, so we know where we are."
...
-    (defun +lsp-init-company-h ()
-      (if (not (bound-and-true-p company-mode))
-          (add-hook 'company-mode-hook #'+lsp-init-company-h t t)
-        (setq-local company-backends
-                    (cons +lsp-company-backend
-                          (remq 'company-capf company-backends)))
-        (remove-hook 'company-mode-hook #'+lsp-init-company-h t))))
-  :config
-  (setq company-lsp-cache-candidates 'auto)) ;; cache candidates for better performance
+  (when (featurep! +peek)
```
hlissner: Please document this new flag in modules/tools/lsp/README.org (under Description > Module Flags) and in docs/modules.org (see other modules in that list with flags listed next to them).

modules/tools/lsp/config.el
sebastiansturm: oops, thanks for the heads up! I'll do as you suggest, though I wonder if it wouldn't be sufficient (and more readable) to just add the `company-capf` backend when `(featurep! :completion company)` instead of going through a separately defined function?

modules/tools/lsp/config.el
sebastiansturm: should I also change it to something like `+no-peek` or `+xref`, so that users don't suddenly see a different frontend when updating doom-emacs without modifying their local config?

modules/tools/lsp/config.el
sebastiansturm: not sure why it should be `lsp-before-initialize-hook` instead of `lsp-mode-hook` though; I believe only the latter will be executed when opening another buffer sharing the same lsp backend. Or do you want to only defer the definition of `+lsp-init-company-h` to `lsp-before-initialize-hook` (though it will then get redefined on every workspace start)?
I think I'll add it to `lsp-mode-hook` for now as that is what the `company-lsp` config used to do, and later move it to its correct location once I've understood the rationale for doing so

sebastiansturm:
> I welcome this change. Thanks for looking into it for me!
> 
> Although, I'm nervous about changing `read-process-output-max` globally, since I can't anticipate the possible side-effects. Perhaps this could be done buffer-locally? Either in process buffers that lsp creates or in `lsp-mode` buffers? Or both?

good suggestion, thanks. I've done that and squashed my commits. Hope I didn't mess up anything else, otherwise I'll fix that tomorrow.
Adding a `printf` statement to `process.c` showed that the buffer-local value of `read-process-output-max` is indeed used in `lsp-mode` buffers, whereas other processes such as ripgrep now continue to work with the 4K default, so I guess the change should work. Didn't do any actual benchmarks though (I wonder if `helm` and `ivy` would benefit from such a change, too?)

modules/tools/lsp/config.el
hlissner: I think `+peek` is best. Disabling lsp-ui's lookup handlers does not guarantee it'll fall back to xref, so `+xref` is misleading.

Side note: flags that disable functionality should be prefixed with a minus, instead. e.g. `-peek`

sebastiansturm: ok, I'll change it back to `+peek` then

sebastiansturm: done

sebastiansturm:
I just found a case where I'm getting bad completions with the `ts-ls` backend (the completions returned by the server seem fine though). This might be an upstream issue; in any case the PR shouldn't be merged until that is resolved. Did someone else try this PR already?

numkem:
@sebastiansturm I'm gonna look into it today. I've very keen on a faster LSP integration!

numkem:
It's very impressive. I've tested with with Go (gopls `master`) and Python (pyls) and they both work flawlessly. I've even tested using `company-idle-delay` set to `0.0` and it doesn't cause any slowdowns or anything of the sort.

This PR even took care of removing a lot of my custom config through the `+peek` flag.

Awesome job!

sebastiansturm:
thanks :) (though all the heavy lifting has been done by @yyoncho and other `lsp-mode` contributors, can't take credit for that)

As for the bad completions, a possibly related issue was opened just today over at the `lsp-mode` repository (https://github.com/emacs-lsp/lsp-mode/issues/1451). That issue also mentions a variable `completion-styles` that we might wish to configure as well. Will probably have a look at it this evening

numkem:
I'd like to see how `lsp-treemacs` could be integrated as well. I've started to look at some of it's functions and they are pretty slick. Especially for the [error reporting part](https://github.com/emacs-lsp/lsp-treemacs#lsp-treemacs-errors-list).

sebastiansturm:
treemacs is great, yes. I've long wanted to implement support for clangd's typeHierarchy, but have been too lazy so far. Now that `callHierarchy` support has recently been implemented there might be pretty little left to do though.

sebastiansturm:
my TypeScript issue seems to be caused by the lsp server returning a completion candidate starting with '.'. Not sure how to best resolve this, so I've added my observation to the `lsp-mode` issue

sebastiansturm:
also, I found that `cc-mode`'s type detection heuristics can cause severe performance issues. I'm currently using an extremely bare-bones custom c/c++ mode, plus semantic highlighting provided by lsp-mode. In the longer run it might make sense to use this in general, or to use the tree-sitter dynamic module to provide syntax highlighting (though with the latter, adding semantic highlighting on top might be overkill).
Unfortunately the semantic highlighting implementation can currently cause performance issues of its own that will only be resolved when the `noverlay` branch is merged into master (guess `ccls`'s font-lock backend is better in this regard). With tree-sitter I've also seen performance issues that may or may not boil down to the same underlying issue (haven't had a closer look yet).

Maybe it would make sense to open a sticky issue tracking the progress of the `noverlay` branch?

modules/tools/lsp/config.el
hlissner: From the sound of it, `lsp-mode-hook` does sound like the better choice. Thanks!

hlissner:
Would it make sense to keep `company-lsp` (and disable `lsp-prefer-capf`) for ts-ls, and other servers not quite prepared for this change?

sebastiansturm:
> Would it make sense to keep `company-lsp` (and disable `lsp-prefer-capf`) for ts-ls, and other servers not quite prepared for this change?

probably yes. I've reintroduced `company-lsp` as a selectable option and added a list of mode-specific overrides. Though I realize the latter isn't ideal (should bind the decision to language servers instead of major modes), but I'll first have to look up if that information is available when we set up completion

sebastiansturm:
guess we'll want to wait for this as well: https://github.com/emacs-lsp/lsp-mode/pull/1454. If and when it gets merged, I'll bump `lsp-mode` accordingly

sebastiansturm:
the capf/snippet fix has been merged, so I'll bump `lsp-mode`. Given that several different capf issues have surfaced during the past few days, I guess I should also change the default to `company-lsp` and leave the `capf` setting to adventurous users for the time being?

sebastiansturm:
changed the default back to `company-lsp` for now so other users' lsp configs don't break without due warning.

numkem:
Say I want to activate `capf`, how would I do that?

modules/tools/lsp/config.el
```diff
@@ -20,6 +34,42 @@ This can be a single company backend or a list thereof. It can be anything
...
+      (cond ((eq :none lsp-diagnostic-package))
+            ((or (eq :flymake lsp-diagnostic-package)
+                 (eq t lsp-diagnostic-package))
+             (lsp--flymake-setup))
+            ((require 'flycheck nil t)
+             (let ((old-checker flycheck-checker))
+               (lsp-flycheck-enable t)
+               (when old-checker
+                 (setq-local flycheck-checker old-checker)
+                 (kill-local-variable 'flycheck-check-syntax-automatically)))))))
```
hlissner: Please combine these two `add-hook` calls into one `add-hook!` call, like so:

```elisp
(add-hook! 'lsp-mode-hook
  (defun ...)
  (defun ...))
```

Passing `defun`s to `add-hook` won't work, since `defun`'s aren't guaranteed to return a symbol (see `defun`'s docstring). `add-hook!` does a little magic to grab the name from the `defun` form (and accepts multiple hooks).

hlissner: Also, `setq-local` only accepts a single var->value pair. You cannot assign to two variables in one call.

hlissner: Also also, the else clause of an elisp if statement does not need a `progn`; `if` is variadic, the first argument is its `then` clause, the rest of its arguments are its `else` clause. i.e.

```lisp
(if t
    a
  x
  y
  z)
```

modules/tools/lsp/config.el
```diff
@@ -30,6 +80,7 @@ This can be a single company backend or a list thereof. It can be anything
   (when lsp-auto-configure
     (mapc (lambda (package) (require package nil t))
           lsp-client-packages))
+  (setq lsp-enable-completion-at-point t)
```
hlissner: Just double checking; should this be set buffer-locally, along with lsp-prefer-capf?

hlissner:
> should bind the decision to language servers instead of major modes

This would make sense. Perhaps something like

```elisp
(defvar company-lsp-servers '(ts-ls ...))

(add-hook! 'lsp-after-initialize-hook
  (defun lsp-init-capf-or-company-h ()
    (if (memq (lsp--client-language-id (lsp--workspace-client lsp--cur-workspace))
              company-lsp-servers)
        (progn
          (setq-local lsp-prefer-capf t)
          (setq-local company-backends (cons 'company-lsp (remq 'company-lsp company-backends))))
      (setq-local lsp-prefer-capf t)
      (setq-local company-backends (cons 'company-capf (remq 'company-lsp company-backends))))))
```

modules/tools/lsp/config.el
sebastiansturm: wow, plenty of bugs there. Will do so, thanks!

sebastiansturm:
> > should bind the decision to language servers instead of major modes
> 
> This would make sense. Perhaps something like
> 
> ```emacs-lisp
> (defvar +lsp-company-clients '(ts-ls ...))
> 
> (add-hook! 'lsp-after-initialize-hook
>   (defun +lsp-init-capf-or-company-h ()
>     (if (memq (lsp--client-language-id (lsp--workspace-client lsp--cur-workspace))
>               +lsp-company-clients)
>         (progn
>           (setq-local lsp-prefer-capf nil)
>           (setq-local company-backends (cons 'company-lsp (remq 'company-lsp company-backends))))
>       (setq-local lsp-prefer-capf t)
>       (setq-local company-backends (cons 'company-capf (remq 'company-capf company-backends))))))
> ```
> 
> (would also make it easier for users to change)

sort of, but by binding it to `lsp-after-initialize-hook` it won't get reexecuted when opening another buffer sharing the same lsp backend so my buffer-local variables won't take effect? Also I think `company-capf` should be removed from `company-backends` when using `company-lsp` (as the latter shouldn't be enabled by default whereas the former is; this is also what `lsp-mode` and the master branch of `doom-emacs` currently do AFAICS).
Replacing the mode-specific association list with a list of non-capf modes or servers is something I didn't do initially because that would make it less convenient for users who'd like to use `company-lsp` most of the time and only opt into `capf` for a few servers. But doing it your way will get more users to test the new backend and is biased towards the speedier alternative, so I think that's a good approach to take

modules/tools/lsp/config.el
sebastiansturm: when omitting the `:completion company` check, wouldn't it at least create the otherwise nonexistent `company-mode-hook` then if `company` is disabled? If so, that would seem a bit untidy to me though I realize it would be irrelevant in practice

modules/tools/lsp/config.el
sebastiansturm: ugh, yes. This is actually the important part, thanks again

sebastiansturm:
> Say I want to activate `capf`, how would I do that?

with the current commit you would either set your default backend to capf:
```
(after! lsp-mode (setq +lsp-company-backend 'company-capf))
```
or add mode-specific overrides to `+lsp-company-overrides` (which should really be server-specific but aren't yet). The default value of `+lsp-company-overrides` demonstrates how to do this.

If you change your default to `company-capf`, note that the 2 overrides that are currently defined by default will always cause `company-lsp` to be selected in `typescript-mode` and `go-mode` buffers, so if you want to use capf for one of these modes, make sure to remove the corresponding overrides (but if you do so, you may want to have a look at this issue: https://github.com/emacs-lsp/lsp-mode/issues/1451)

modules/tools/lsp/config.el
hlissner: >  wouldn't it at least create the otherwise nonexistent company-mode-hook then if company is disabled?

It would, yeah. I'm willing to forgive that untidiness for now, and return to it once I'm certain the lsp module is running without issues.

sebastiansturm:
I've now changed `+lsp-init-company-h` to iterate over all of a buffer's associated workspaces and apply all corresponding overrides (if any). The association list has been replaced by a `capf` blacklist, and `lsp-mode` has been bumped to a newer commit that includes another `capf`-related bug fix

sebastiansturm:
so far this seems to be working for me, so I'll squash my commits (except for the unrelated typo fix)

modules/tools/lsp/config.el
```diff
@@ -20,6 +29,42 @@ This can be a single company backend or a list thereof. It can be anything
   ;; Auto-kill LSP server once you've killed the last buffer associated with its
   ;; project.
   (setq lsp-keep-workspace-alive nil)
+  (add-hook! 'lsp-mode-hook
+    (defun +lsp-init-company-h ()
+      (let ((preferred-backend +lsp-company-backend))
+        (lsp-foreach-workspace
+         (when (memq (lsp--client-server-id (lsp--workspace-client lsp--cur-workspace))
+                     +lsp-capf-blacklist)
+           (setq preferred-backend 'company-lsp)))
```
hlissner: Should this be done if `company-mode` is enabled? Above `(if (eq 'company-capf preferred-backend) ...`?

Thus ensuring this work is done as late as possible.

modules/tools/lsp/config.el
```diff
@@ -1,11 +1,20 @@
...
+overridden by `+lsp-capf-blacklist'.
 
-This can be a single company backend or a list thereof. It can be anything
-`company-backends' will accept.")
+While `company-capf' does not require the `company-lsp' package
+and should offer better performance, it has been integrated into
+lsp only recently and as of 02/25/2020 is known to cause issues
+with some language servers. If you wish to use `company-capf' in
+general but fall back to `company-lsp' for specific language
+servers, set `+lsp-company-backend' to `company-lsp' and add the
```
yoavm448: Shouldn't `+lsp-company-backend` be set to `company-capf`?

sebastiansturm: ah yes, thanks

yoavm448: ...And then the blacklist would fall back to `company-lsp` for certain buffers

modules/tools/lsp/config.el
sebastiansturm: sure, that makes sense. I've pushed a commit that should fix this but didn't squash it yet, in case I messed up something. If there's anything else to fix, I'll do so tomorrow morning

sebastiansturm:
> I think this looks about ready to merge, save one thing that I've annotated. Thanks so much for all your hard work on this! It is a huge help, and I look forward to merging it!

thank you for doom-emacs and all the helpful advice!

sebastiansturm:
noticed no further issues today, so I squashed my commits. I also just learned about `yas-inhibit-overlay-modification-protection` which according to https://github.com/emacs-lsp/lsp-mode/issues/1451 we may want to set to `t` if company is enabled and company-posframe is not. But I haven't tried that yet and it's more of a company thing anyway so I wouldn't want to include it in this PR

hlissner:
And done! Thanks again!

> yas-inhibit-overlay-modification-protection

Thanks for letting me know. I'll look into that as well.

PR #2592, opened 2020-02-24
Construct org-id-locations-file using expand-file-name

karlicoss:
Otherwise if you forgot a trailing slash on your `org-directory`, you end up with a wrong file


hlissner:
Good catch, thanks!

PR #2593, opened 2020-02-24
mdempsky doesnt work with modules, updating to stamblerre

esoterick:
github.com/mdempsky/gocode is an unmaintained fork of github.com/nsf/gocode which is also unmaintained.

github.com/stamblerre/gocode seems to be the current working module, which is also in "maintenance" mode as people are recommended to switch to `gopls`.

This is get the doom go module in working order until it can be switched to use gopls.


hlissner:

PR #2598, opened 2020-02-25
Fix syntax error in Clojure module

darth10:
Looks like this bug was introduced in a65403011e930fceaa59664dceabdee27b4c4cb4.
It causes Doom Emacs to fail on start-up if the `:lang clojure` module is enabled.
The workaround would be to disable this module until this change is merged in.


hlissner:
Ack, sorry about that. Thanks!

PR #2603, opened 2020-02-25
add `open-in-iterm` shortcut for macos module

digizeph:
`+macos/open-in-iterm` opens the current directory in `iTerm`. 
[`iTerm`](https://iterm2.com/) is a popular alternative to the macos's default terminal emulator `term`.


hlissner:
Thanks for the help!

PR #2614, opened 2020-02-26
Fix file template :when logic to apply to modes

tylerware:
When a file template rule is evaluated against a particular buffer it is
supposed to call the `:when` function if it is defined. Currently, the
function is ignored if the `PRED` of the template rule is a mode. This
fixes this issue.

## Testing
To test the fix:
1. Before checking this out, register the following file template:
```
(set-file-template! 'org-mode
  :when #'(lambda (file)
            (message "I was called"))
  :trigger "__"
  :mode 'org-mode)
```
2. Open an `'org-mode` buffer. Note that there is no `"I was called"` message. 
3. Checkout this branch
4. Again, open up an `'org-mode` buffer. Note that there is now a `"I was called"` message.


hlissner:
Good catch, thanks!

PR #2615, opened 2020-02-26
Set elisp lookups for helpful-mode

yoavm448:
I often read the extracted source code in helpful. This would allow using
`+lookup/documentation` and co. in helpful, which saves a few key presses
compared to the ivy/helm stuff on `SPC h f/v`.


yoavm448:
Should I also add helpful to the repl & eval handlers that are set for emacs-lisp? 
Or maybe its an overkill, since you could just go to the linked source file at this point.

hlissner:
I think that'd be overkill. For now the lookup handlers are good enough.

yoavm448:
Yeah, you are probably correct. Moved it to emacs-lisp/config.el.

hlissner:
Thanks!

PR #2622, opened 2020-02-27
Improve haskell documentation

quangIO:
Add how to use `ghcide` in lsp mode.

### DIFF
--- a/modules/lang/haskell/README.org
+++ b/modules/lang/haskell/README.org
@@ -60,7 +60,8 @@ Depending on whether you use Intero, Dante or LSP, your dependencies will
 differ:
 
 + Dante users need =cabal=, =ghc= and =ghc-mod=
-+ LSP users need the =haskell-ide-engine= LSP server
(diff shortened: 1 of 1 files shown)


modules/lang/haskell/README.org
```diff
@@ -85,26 +86,41 @@ sudo zypper install cabal-install ghc
...
-*** Arch Linux
+**** Arch Linux
 =haskell-ide-engine-git= is available on the AUR
 
-#+BEGIN_SRC emacs-lisp
+#+BEGIN_SRC sh
 yay -S haskell-ide-engine-git
 #+END_SRC
+*** ghcide
+If you prefer [[https://github.com/digital-asset/ghcide#using-it][ghcide]], you can add the following snippet to your =config.el=
```
ocharles: Rather than documenting this, I wonder if we should just have `+ghcide`?

hlissner:
As of 00a21e9 the lang/haskell module now has a `+ghcide` flag. Thanks for bringing this to my attention!

PR #2628, opened 2020-02-28
Fix clojure link in modules.org

yoavm448:
Now the link is relative and not dependent of the absolute location of `doom-emacs-dir`.


hlissner:
Good catch, thanks!

PR #2630, opened 2020-02-28
Use s-titleized-words (if defined) for title in org template

yoavm448:
`s-titleized-words` of the file name is generally a much better heuristic for
the title than just the file name.

I didn't want to add s.el as a dependency for file-templates, so if not defined,
the template uses the normal `buffer-file-name`. If you think its legit to add
s.el as a dependency, I'll do so.

### DIFF
--- a/modules/editor/file-templates/templates/org-mode/__
+++ b/modules/editor/file-templates/templates/org-mode/__
@@ -1,6 +1,20 @@
 # -*- mode: snippet -*-
 # name: Org template
 # --
-#+TITLE: ${1:`(file-name-base buffer-file-name)`}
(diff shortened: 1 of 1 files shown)


hlissner:
e.g.

```lisp
(string-join (mapcar #'capitalize (split-string buffer-file-name "[^[:word:]0-9]+" t)) " ")
```

yoavm448:
Your snippet doesn't handle CamelCase... so I extracted the heart of `s-titleized-words`, and used it instead.

hlissner:
Hoo, that's a bit of a mess. In any case, thanks for the PR!

PR #2631, opened 2020-02-28
Set org-babel-python-command to be same as python-shell-interpreter

karlicoss:
By default it's `python` so suffers from the same issues as `python-shell-interpreter`.

tbh, not sure if the way I'm suggesting is a good one, maybe it make more sense to simply copy-paste the [bit of logic](https://github.com/hlissner/doom-emacs/blob/56cba7828c295acafbbd0e14938f62dd23080a3d/modules/lang/python/config.el#L51-L55) for python-shell-interpreter. Let me know what you think is the right way!


hlissner:
I think we can use this for now. Thanks for the help!

PR #2632, opened 2020-02-28
Add aliases in doom help

gagbo:
Feel free to reformat and/or change the way aliases are added, I just wanted to add the basic if-let call

### DIFF
--- a/core/cli/help.el
+++ b/core/cli/help.el
@@ -89,14 +89,14 @@
        (if (null (car group))
            (dolist (cli (cdr group))
              (print! "%-16s %s"
-                     (doom-cli-name cli)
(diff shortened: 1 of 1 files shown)


gagbo:
The change in width from `16` to `21` was to accomodate the many, long aliases of `doom sync`.

hlissner:
I cannot accept PRs to doom's CLI for the time being. A rewrite of `bin/doom` is in the works and due to be pushed soon. Plus, the new version will list each command's first alias in `doom help`. But thanks for the PR in any case!

gagbo:
Fair enough :D I knew that `bin/doom` is about to change a lot :+1:

PR #2634, opened 2020-02-29
Extend backend-agnostic vc functionality

sebastiansturm:
This PR extends support for non-git version control systems in the following ways:

1. `git-gutter:handled-backends` is extended to `(git hg svn bzr)`. I'm using the `hg` backend with no issues so far, though I haven't tried `svn` or `bzr`. In the latter two cases I'm relying on git-gutter's claim of support but if that is deemed to risky we might as well reduce it to `(git hg)` for now

2. if both `hydra` and `vc-gutter` are enabled, `<leader> g .` is bound to the (IMO very useful) `+vc/gutter-hydra`.

3. `+vc/gutter-hydra`'s quit functions are extended to automatically close the `git-gutter` popup (not sure if that is what other users want, as it doesn't currently differentiate between popups opened through the hydra itself and a preexisting git-gutter popup)

4. I believe the popup rules defined for `vc-annotate` were slightly incorrect before (no leading asterisk and the too-unspecific `vc-d` prefix matched anything starting with `*vc-dir` as well)

5. the various `log-view` modes invoked by, e.g., `vc-print-log` now start up in emacs state, with added `j`/`k` keybindings to navigate between entries. Although meta-leader works out of the box, the functionality shadowed by enabling doom's usual `SPC` leader is not useful enough, IMO, to break muscle memory, so I'm reenabling the default leader within those mode maps. The way I currently do this is rather ugly though; is there some helper function or macro that sets a mode's initial state to `emacs` but leaves the leader key in effect?

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -395,15 +395,17 @@
         :desc "Sudo this file"              "U"   #'doom/sudo-this-file
         :desc "Yank filename"               "y"   #'+default/yank-buffer-filename)
 
-      ;;; <leader> g --- git
+      ;;; <leader> g --- git/version control
       (:prefix-map ("g" . "git")
-        :desc "Git revert file"             "R"   #'vc-revert
+        :desc "Revert file"                 "R"   #'vc-revert
         :desc "Copy link to remote"         "y"   #'+vc/browse-at-remote-kill-file-or-region
         :desc "Copy link to homepage"       "Y"   #'+vc/browse-at-remote-kill-homepage
         (:when (featurep! :ui hydra)
           :desc "SMerge"                    "m"   #'+vc/smerge-hydra/body)
         (:when (featurep! :ui vc-gutter)
-          :desc "Git revert hunk"           "r"   #'git-gutter:revert-hunk
+          (:when (featurep! :ui hydra)
+            :desc "VCGutter"                "."   #'+vc/gutter-hydra/body)
+          :desc "Revert hunk"               "r"   #'git-gutter:revert-hunk
           :desc "Git stage hunk"            "s"   #'git-gutter:stage-hunk
           :desc "Git time machine"          "t"   #'git-timemachine-toggle
           :desc "Jump to next hunk"         "]"   #'git-gutter:next-hunk
--- a/modules/emacs/vc/config.el
+++ b/modules/emacs/vc/config.el
@@ -3,18 +3,31 @@
 (when IS-WINDOWS
   (setenv "GIT_ASKPASS" "git-gui--askpass"))
 
+(after! log-view
+  (set-evil-initial-state!
+    '(log-view-mode
+      vc-git-log-view-mode
+      vc-hg-log-view-mode
+      vc-bzr-log-view-mode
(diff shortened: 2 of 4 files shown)


sebastiansturm:
just noticed there's a bug with the keymap modification, hold on

sebastiansturm:
think it should be ok now. Also, I think the vc log benefits much more from vertical space than from horizontal space. Is there a builtin way to have a popup open vertically if the current buffer is wide (say, at least 80% of the total frame width or so), and in its usual configuration otherwise? Also, should popup configuration be left entirely to the user, or would you be fine with setting the log buffer's default height to 30-40% or so?

sebastiansturm:
one more thing: personally, I'd find it useful to have `<leader> g L` map to magit's buffer log if and only if git is the currently active vc backend, and to either `vc-print-log` or `vc-print-root-log` otherwise. Similarly, the blame keybinding might switch between magit blame in git-backed modes and `vc-annotate` elsewhere. Not sure if usefulness outweighs inconsistency in this case, but I think I'd consider that change a net usability improvement

sebastiansturm:
> I'll have to remove the leader key in log-view-mode-map; it causes some problems placing that there, especially that it doesn't take into account leader keys for non-evil users (the leader is accessible via `M-SPC` in emacs state in any case). I may look into conditionally adding hg, svn and bzr to `git-gutter:handled-backends` (e.g. depending on the presence of their executables) as well.
> 
> Otherwise, this seems fine. Thanks for your help and for looking into this!

wow, that was quick. Thanks a lot! What do you think about having some vc keybindings like magit blame call generic vc functions if the current backend is something other than git?

hlissner:
> What do you think about having some vc keybindings like magit blame call generic vc functions if the current backend is something other than git?

I think that's a great idea. Would you like to make a PR, or can you list those vc commands and what magit command they'd replace, so I can look into them?

sebastiansturm:
I can make a PR; I think magit-blame would best be replaced by `vc-annotate`, `magit-status` might map to `vc-dir`, etc. I yet have to come up with a proper list, will look into it.

On a related note, I thought it might be useful to have doom's formatting methods remap to lsp-provided formatting if supported by the current lsp backend (and unless overruled by some user config). If you agree, I could look into that afterwards

hlissner:
> On a related note, I thought it might be useful to have doom's formatting methods remap to lsp-provided formatting if supported by the current lsp backend (and unless overruled by some user config). If you agree, I could look into that afterwards

This needs a larger solution. The `:editor format` module, in general, is in dire need of a rewrite, to fix its issues and integrate LSP into it, without sacrificing a way to customize formatters. I think this'll have to wait until I get around to that.

sebastiansturm:
> > On a related note, I thought it might be useful to have doom's formatting methods remap to lsp-provided formatting if supported by the current lsp backend (and unless overruled by some user config). If you agree, I could look into that afterwards
> 
> This needs a larger solution. The `:editor format` module, in general, is in dire need of a rewrite, to fix its issues and integrate LSP into it, without sacrificing a way to customize formatters. I think this'll have to wait until I get around to that.

even better, didn't know it was on your todo list. I'll just do the vc PR then, thanks!

PR #2641, opened 2020-03-01
bin/org-capture: Suppress output for initial emacsclient call

msanders:
Avoids an extraneous `nil` log when invoking `org-capture` CLI.


hlissner:
Thanks!

PR #2642, opened 2020-03-01
add systemd-mode

chrunchyjesus:


hlissner:
I have to refuse this PR, because:

1. It's not really a "data" format and doesn't really belong in this module.
2. systemd is a tad too specialized for this module,
3. This is a trivial amount of configuration. It can be done by the end user.

chrunchyjesus:
No problem. :)

PR #2650, opened 2020-03-02
Fix Spring bug on Ruby Rspec Mode

otavioschwanck:
Without this, Rails 6 (new version) always try to use spring.  Adding this modification, fix it for all rails versions.  If the user really want, they can add on config.el.


hlissner:
Thanks for the help!

PR #2657, opened 2020-03-03
Spotify for Doom Emacs

otavioschwanck:


modules/app/spotify/packages.el
```diff
@@ -0,0 +1,17 @@
+;;; app/spotify/config.el -*- lexical-binding: t; -*-
+
+;;
+;;; Packages
+
+(package! helm-spotify-plus :pin "3383b1399a"
```
otavioschwanck: I dont now what is this pin, i put a random number.  Can anyone help me ?

modules/app/spotify/config.el
```diff
@@ -0,0 +1,14 @@
+;;; app/spotify/config.el -*- lexical-binding: t; -*-
+
+(use-package! helm-spotify-plus
+  :when (featurep! helm)
+  :config
+  (map! :leader
+        :map global-map
+        :prefix "5"
```
otavioschwanck: i accept suggestions on this prefix.

modules/app/spotify/packages.el
yoavm448: Put your cursor inside those parentheses and execute `M-x doom/update-pinned-package-form`

otavioschwanck: Thanks!

hlissner:
I cannot accept this PR because it's not enough to warrant a module. A user could trivially set it up themselves. It also only caters to helm users and does not accommodate ivy.

I don't mind a spotify module, but there has to be more to it.

Side notes: 
+ Modules should not define their own leader keybinds (leader keybinds should be bound in `:config default`)
+ `(featurep! helm)` should be `(featurep! :completion helm)`
+ The `helm-spotify-plus` isn't lazy loaded

PR #2659, opened 2020-03-03
Make +evil:copy-this-file autoload interactively

yoavm448:
`:cp` was not autoloaded correctly beforehand. I just changed `+evil:copy-this-file`'s autoload
signature it so it would match `+evil:move-this-file`, with which I had no issue - and it works!


hlissner:
Good catch. Thanks!

PR #2668, opened 2020-03-04
bump rustic

brotzeit:
I moved some code to lsp-mode because there are issues with requiring lsp-mode related code in doom. An error will be thrown when starting the first lsp session. I've tried the fix in doom, but it would be good if somebody can confirm this.

### DIFF
--- a/modules/lang/rust/config.el
+++ b/modules/lang/rust/config.el
@@ -15,7 +15,6 @@
   (set-popup-rule! "^\\*rustic-compilation" :vslot -1)
 
   (setq rustic-indent-method-chain t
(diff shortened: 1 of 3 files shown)


hlissner:
Thanks so much for the PR! Would you mind rewording the commit to:

```gitcommit
Bump lsp-mode & rustic

brotzeit/rustic@9939691 -> brotzeit/rustic@373f5a1
emacs-lsp/lsp-mode@2e6b748 -> emacs-lsp/lsp-mode@6a94356
```

I try to make sure our bump commits are machine-readable for future CI/CD use.

hlissner:
Thanks again!

PR #2669, opened 2020-03-05
Add polymode to handle RMarkdown and RNoweb files

even4void:
As of ESS 19+, ess-noweb and ess-swv libraries are now obsolete, and
polymode should be used instead (https://emacs.stackexchange.com/a/17065).


hlissner:
Should ESS not be bumped as well?

hlissner:

PR #2670, opened 2020-03-05
Associate xhtml files with web-mode

yoavm448:
Although xhtml files are indeed xml, they usually have embedded JavaScript and CSS like
normal HTML, so `web-mode` fits them better than `nXML-mode` (the default)


hlissner:
Thanks!

PR #2677, opened 2020-03-07
Extend backend-agnostic vc functionality

sebastiansturm:
this PR defines a few interactive commands that automatically resolve to magit-provided functionality if magit is enabled, and the current repository is backed by git, or vc-provided functionality otherwise. Existing evil/emacs keybindings have been changed to map to the new backend-agnostic functions. Seems to work, though I yet have to try out `+vc/file-delete`.

When calling magit-specific functionality, I currently `(require 'magit)` explicitly at function call time to make sure the desired command is actually available. While this works, it's definitely not elegant. I guess there is some better way to achieve the same result?

### DIFF
--- a/core/core-modules.el
+++ b/core/core-modules.el
@@ -357,15 +357,16 @@ This value is cached. If REFRESH-P, then don't use the cached value."
   ;; HACK Fix `:load-path' so it resolves relative paths to the containing file,
   ;;      rather than `user-emacs-directory'. This is a done as a convenience
   ;;      for users, wanting to specify a local directory.
-  (defadvice! doom--resolve-load-path-from-containg-file-a (orig-fn _label arg &optional _recursed)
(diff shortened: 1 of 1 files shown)


hlissner:
This needs some more thought. If `:emacs vc` is disabled, then nothing on the `SPC g` prefix will work, regardless of `:tools magit`'s status.

The dispatchers should also be autoloaded and be renamed to follow [our naming conventions](https://github.com/hlissner/doom-emacs/blob/develop/docs/contributing.org#naming-conventions). `X/Y` is reserved for interactive commands.

I'm swamped at the moment, so I'll get back to this when I can.

sebastiansturm:
alright, I'll have a look and I understand the workload required to maintain doom must be pretty massive. Thanks for your comments!

PR #2680, opened 2020-03-08
emacs: vc: browse-at-remote: bump to 771a3079e2

asymmetric:
It now defaults to adding the line number. Controlled by [this variable](https://github.com/rmuslimov/browse-at-remote/blob/771a3079e27f397d2f5a9470b945980fa68ee048/browse-at-remote.el#L79-L85).


hlissner:
The purpose of these commands was to do the equivalent of `browse-at-remote-add-line-number-if-no-region-selected = nil`. a25dea9 does this and bump the package. Thanks for bringing it to my attention.

Issue #2039, opened 2019-11-09
[BUG] eshell cursor moves when running gcc compiled program.


DesantBucie:
Hello Doomacsers, I have a problem with eshell.

When i open it in second window and compile program in gcc, then run it i have to really fight when trying to enter input. The cursor moves like someone was pressing Tab key on the other keyboard, well if I had one. It only happens using my compiled program, and the error i get is: `Company: backend company-pcomplete error " No catch for tag: pcompleted, t" with args (prefix)`. I don't know if it is a problem with emacs or system dependecies.

My system is Manjaro XFCE   
I DO NOT use Evil.

Of course the easiest way to solve the problem would probably be to  comment company from init, but it is usefull.
<details><pre>
```
emacs   version    26.3
        features   XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS GLIB NOTIFY ACL GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS LIBSYSTEMD LCMS2
        build      Aug 29, 2019
        buildopts  (--prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib --localstatedir=/var --with-x-toolkit=gtk3 --with-xft --with-modules 'CFLAGS=-march=x86-64 -mtune=generic -O2 -pipe -fno-plt' CPPFLAGS=-D_FORTIFY_SOURCE=2 LDFLAGS=-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now)
        windowsys  x
        daemonp    server-running
doom    version    2.0.9
        build      HEAD -> develop, origin/develop, origin/HEAD 8db3b7dca 2019-11-04 17:21:56 -0500
system  type       gnu/linux
        config     x86_64-pc-linux-gnu
        shell      /bin/bash
        uname      Linux 4.19.80-1-MANJARO #1 SMP PREEMPT Fri Oct 18 12:28:02 UTC 2019 x86_64
[...]

PR #2683, opened 2020-03-08
Fix for the incorrect quotes handling within Eshell

dive:
Hello there,

There is an annoying issue with company mode and Eshell. When you type quote/double-quotes (aka `'` or `"`) then Emacs inserts a tabulation to the input and the following error happens:
```
Company: An error occurred in auto-begin
Company: backend company-pcomplete error "No catch for tag: pcompleted, t" with args (prefix)
```
I saw the same issue before, here #2039. The fix proposed there works in this case as well.

Let me know if it is possible to fix it in a different way or there are problems with the proposed fix.

Thanks!


hlissner:

PR #2685, opened 2020-03-09
Replace obsolete elisp function in dante-mode

kenranunderscore:
`dante-mode` was using `haskell-process-load-or-reload` on `, l` and `haskell-mode` complained about it being obsolete.
This PR changes it to `haskell-process-load-file`, same as bound to `C-c C-l`.


hlissner:
Thanks!

PR #2692, opened 2020-03-09
Fix internal documentation links containing "%20"

league:
These links to org headers didn't work for me when they contain `%20` instead of space. Should they have?


hlissner:
With them, it breaks the links in org-mode. Without them, it breaks the links on github. Wonderful. I guess org-mode is the priority here. Thanks.

league:
Ah, didn't think about that angle. Was wondering if there's some org patch I was missing that would support these links.

PR #2694, opened 2020-03-09
Ruby Lang improvements

otavioschwanck:


modules/lang/ruby/config.el
```diff
@@ -183,6 +183,7 @@
   :when (featurep! +rails)
   :hook (enh-ruby-mode . projectile-rails-mode)
   :init
+  (setq inf-ruby-console-environment "development")
```
otavioschwanck: With this, emacs launch robe without asking the env.

hlissner:
Thanks!

PR #2712, opened 2020-03-13
Add a :lang puppet stub

arrdem:
Does what it says on the tin. Pulls in puppet-mode and tells DOOM how to load it.


hlissner:
Thanks for the PR, but I'm afraid I cannot accept it as it is. It isn't enough to warrant a new module. As a rule of thumb, if it can be trivially installed by the user in their private config with two or three lines, it is likely to be a poor candidate for a new module.

At the very least, it needs documentation, localleader keybinds, lookup/eval/repl handlers (if applicable), and a doctor file.

arrdem:
Cool! Noted for the future.

PR #2716, opened 2020-03-13
Remove enh-ruby and improve ruby support

otavioschwanck:


hlissner:
I've already removed these references locally, in unpushed commits. Thanks for the PR in any case.

PR #2719, opened 2020-03-13
Fix robe on first rb file and extend rails projectile

otavioschwanck:


hlissner:
As mentioned in #2716, this was already dealt with in some unpushed commits. They've been pushed in 4b20c72, which should resolve these issues. Thanks for bringing them to my attention!

Issue #2429, opened 2020-01-26
Support =org-download-screenshot= and bind to key


codygman:
**Describe the feature**

`org-download-dnd` is already available in doom to provide drag and drop. I also find it useful to use `org-download-screenshot` while taking notes in org mode to quickly screenshot some information that is then inserted into the org mode document.

Insanely useful if your default mode of working is a literate journal of notes/source code and tons of screenshots of things you'd otherwise have to type.

However when I call `org-download-screenshot` (which is available after calling `org-download-dnd` since it's deferred I guess) it by default uses `attach:` instead of `file:`. This matters because org displays images with `file:` but by default it doesn't do that with `attach:`. Here is a screenshot of what I mean:

![screen28](https://user-images.githubusercontent.com/225847/73131356-27fe9f80-3fcf-11ea-9931-7ce329b862fe.png)

I'm note sure how t

- If the feature involves new plugins, include links to them
- If the feature involves replacement of current functionality, how the new
  functionality is better?

The new functionality is better because you can actually see the images inline with the notes. 

If this isn't quite aligned with the doom philosophy and it's preferred not to show notes inline, I think doom should still implement what's necessary to show `attach:` images inline if `org-startup-with-inline-images` is set to `t`.

hlissner:
Doom will display `attach:...` links like it does `file:...` links, if the end file is an image. It appears that it org-download didn't insert the `attach:` as a link. I'll look into it later tonight.

hlissner:
As of 827225e and 743c03f, I've replaced `attach` with `download` since org-mode already has its own `attachment` links, and our custom attachment system was recently replaced with org's built-in one; it'd be confusing to have both. Still, I wanted to keep the shorter link format for files added via org-download.

In any case, `download` links should be rendered inline now. If that isn't the case and I'll reopen it. Thanks for bringing it to my attention!

codygman:
When I start doom fresh and do `SPC m a c` I get an error about `org-download-screenshot` not being defined. If I do something to trigger loading org-download like calling `org-download-dnd` it works fine, but I think we want to make sure `org-download` is loaded so we can call `SPC m a c` without error.

hlissner:
Ah, good catch. I've autoloaded them in b9f823b.

jsravn:
Hello, I'm running into a small issue with this. It seems to insert the wrong link (it includes `.attach` when it shouldn't):

[...]

PR #2721, opened 2020-03-14
Fix inserted filename for org-download-screenshot

jsravn:
Fix https://github.com/hlissner/doom-emacs/issues/2429#issuecomment-599085171


hlissner:

PR #2723, opened 2020-03-14
Minor changes to the helm module

RBckmnn:
Hey Henrik,

this PR addresses some minor issues with the helm module. Currently, helmized command (i.e. commands that will use helm's completion interface because `helm-mode` is active) as well as some native helm commands will provide only very basic pattern matching behaviour. This is because they will match according to `completion-styles` which is equal to `(basic partial-completion emacs22)`, at least for my doom/emacs installation.

I've also removed code that sets values of variables that have, in the meantime, been removed from helm's source code.


hlissner:
Thanks for the PR and for looking into this!

Issue #1862, opened 2019-10-08
[BUG] IRC openssl 1.1


Renzix:
**Describe the issue**
IRC (doom module that uses circe) doesnt connect to freenode. I believe its because openSSL 1.1 is broken for emacs. See this issue https://github.com/jorgenschaefer/circe/issues/340

Following the thread there was a fix and it did work for me. Also the thread said it was unlikely that a fix will come in upstream because tls.el has been obsoleted. 
```lisp
(setq tls-end-of-info
      (concat
       "\\("
       ;; `openssl s_client' regexp.  See ssl/ssl_txt.c lines 219-220.
       ;; According to apps/s_client.c line 1515 `---' is always the last
       ;; line that is printed by s_client before the real data.
       "^    Verify return code: .+\n\\(\\|^    Extended master secret: .+\n\\)---\n\\|"
       ;; `gnutls' regexp. See src/cli.c lines 721-.
       "^- Simple Client Mode:\n"
       "\\(\n\\|"                           ; ignore blank lines
       ;; According to GnuTLS v2.1.5 src/cli.c lines 640-650 and 705-715
       ;; in `main' the handshake will start after this message.  If the
       ;; handshake fails, the programs will abort.
       "^\\*\\*\\* Starting TLS handshake\n\\)*"
       "\\)"))
```

Note nothing errors it just says `Package tls is depreicated`

*Messages*
```
Doom loaded 274 packages across 55 modules in 0.578s
Package cl is deprecated
Loading /home/genzix/.emacs.d/.local/cache/recentf...done
Quit
Debug mode on
Starting new Ispell process aspell with default dictionary...
Package tls is deprecated
Decrypting /home/genzix/.password-store/irc/freenode.net.gpg...done
```

My config https://git.renzix.com/Renzix/Dotfiles/src/branch/master/.doom.d/config.org
Mirror https://gitlab.com/renzix/Dotfiles/blob/master/.doom.d/config.org

**Steps to reproduce**
1. Install DOOM emacs
2. Enable IRC
3. Setup IRC config
4. Make sure you are on openssl 1.1 (I think)

**System information**
<details><pre>
```
emacs   version    27.0.50
[...]

PR #2727, opened 2020-03-15
IRC/circe fixes: TLS connection workaround and default server settings

hperrey:
This PR addresses the problem mentioned in issue #1862 where the IRC module circe is hanging on TLS connections when using OpenSSL versions > 1.1.0. In this case, `tls.el` does not correctly determine the end of the info block and hangs.

This workaround was first proposed here: https://github.com/jorgenschaefer/circe/issues/340#issuecomment-454521683
Since it looks like it will not be fixed upstream, this might be a good place to incorporate the workaround.

Additionally, this PR fixes `circe-network-defaults' being overwritten on init: by default, this list is filled with predefined server details for e.g. Freenode.net which is rather useful.

Please note that this code works fine on my setup and with my OpenSSL setup (1.1.1a on Ubuntu Mate 19.10) connecting to Freenode.net but I have not yet been able to test it on another system/server connection.

### DIFF
--- a/modules/app/irc/config.el
+++ b/modules/app/irc/config.el
@@ -93,7 +93,7 @@ playback.")
   (add-hook 'circe-channel-mode-hook #'turn-on-visual-line-mode)
   (add-hook 'circe-mode-hook #'+irc--add-circe-buffer-to-persp-h)
   (add-hook 'circe-mode-hook #'turn-off-smartparens-mode)
-  (add-hook 'circe-mode-hook (lambda () (setq-local tls-end-of-info
+  (setq-hook! 'circe-mode-hook tls-end-of-info
       (concat
        "\\("
        ;; `openssl s_client' regexp.  See ssl/ssl_txt.c lines 219-220.
@@ -107,8 +107,7 @@ playback.")
        ;; in `main' the handshake will start after this message.  If the
        ;; handshake fails, the programs will abort.
        "^\\*\\*\\* Starting TLS handshake\n\\)*"
-       "\\)")
-                                         )))
+       "\\)"))
 
 
   (defadvice! +irc--circe-run-disconnect-hook-a (&rest _)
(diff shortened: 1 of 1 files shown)


modules/app/irc/config.el
```diff
@@ -94,6 +93,23 @@ playback.")
   (add-hook 'circe-channel-mode-hook #'turn-on-visual-line-mode)
   (add-hook 'circe-mode-hook #'+irc--add-circe-buffer-to-persp-h)
   (add-hook 'circe-mode-hook #'turn-off-smartparens-mode)
+  (add-hook 'circe-mode-hook (lambda () (setq-local tls-end-of-info
```
hlissner: Please use `setq-hook!` here instead.

```elisp
(setq-hook! 'circe-mode-hook
  tls-end-of-info ...)
```

modules/app/irc/config.el
```diff
@@ -94,6 +93,23 @@ playback.")
...
+       "^    Verify return code: .+\n\\(\\|^    Extended master secret: .+\n\\)\\(\\|^    Max Early Data: .+\n\\)---\n\\|"
+       ;; `gnutls' regexp. See src/cli.c lines 721-.
+       "^- Simple Client Mode:\n"
+       "\\(\n\\|"                           ; ignore blank lines
+       ;; According to GnuTLS v2.1.5 src/cli.c lines 640-650 and 705-715
+       ;; in `main' the handshake will start after this message.  If the
+       ;; handshake fails, the programs will abort.
+       "^\\*\\*\\* Starting TLS handshake\n\\)*"
+       "\\)")
+                                         )))
```
hlissner: Please join these hanging parentheses onto the same line.

hperrey:
Thanks for the feedback -- fixed! First elisp-based PR, so I really appreciate someone looking over even these rather trivial changes :)

hlissner:
Looks good to me! Thanks for the help and the quick response!

PR #2731, opened 2020-03-16
Disable namespace indentation in :lang cc

judemille:
Most C++ style guides require no indentation in namespaces, and it's *really* bugging me when an extra indentation is added for no reason just because I'm in a namespace. This change would fix that. If you feel that changing the global settings for that isn't the right move, perhaps we could add a config option to customize the C++ style. It also appears there is a hack to force electric to use clang-format to fix indentation, so maybe that can be configurable too. Whatever the case, I would like there to be reasonable defaults.

### DIFF
--- a/modules/lang/cc/config.el
+++ b/modules/lang/cc/config.el
@@ -107,7 +107,8 @@ This is ignored by ccls.")
              ;; another level
              (access-label . -)
              (inclass +cc-c++-lineup-inclass +)
-             (label . 0))))
+             (label . 0)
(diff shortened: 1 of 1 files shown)


juliandemille:
I've tested that clang-format hack and can't get it to work, but I haven't tested extensively.

juliandemille:
I also tested this change and couldn't get it to work either, even disabling lsp-enable-on-type-formatting.

judemille:
Ok, I think I was having a different issue.

PR #2738, opened 2020-03-18
Fix and improve ido module

belak:
- ido will actually be loaded at startup - I'm hoping there's a better
  way to optimize this.
- ido-sort-mtime-mode was fixed
- Up/Down were added to get closer to
  `(setq ido-vertical-define-keys 'C-n-C-p-up-down-left-right)`
- Faces were re-enabled
- The number of results are now displayed with ido

### DIFF
--- a/modules/completion/ido/config.el
+++ b/modules/completion/ido/config.el
@@ -1,11 +1,13 @@
 ;;; completion/ido/config.el -*- lexical-binding: t; -*-
 
-(defun +ido-init-h ()
+(use-package! ido
+  :after-call pre-command-hook
+  :config
   (setq ido-ignore-buffers
         '("\\` " "^\\*ESS\\*" "^\\*Messages\\*" "^\\*Help\\*" "^\\*Buffer"
           "^\\*.*Completions\\*$" "^\\*Ediff" "^\\*tramp" "^\\*cvs-"
           "_region_" " output\\*$" "^TAGS$" "^\*Ido")
-        ido-use-faces nil
+        ido-auto-merge-work-directories-length -1
         ido-confirm-unique-completion t
         ido-case-fold t
         ido-enable-tramp-completion nil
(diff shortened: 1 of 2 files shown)


dive:
I am not sure if this problem applies to this, but recently I noticed that [default keybindings](https://github.com/hlissner/doom-emacs/blob/develop/modules/config/default/+emacs-bindings.el) for Emacs are too Ivy/Swiper oriented. And even when `ido` package is enabled (and Ivy disabled with Helm) key bindings still prefer Swiper for search. Like this:

```
(:prefix-map ("s" . "search")
        :desc "Search buffer"                "b" #'swiper
        :desc "Search current directory"     "d" #'+default/search-cwd
        :desc "Search other directory"       "D" #'+default/search-other-cwd
        ...
        :desc "Search buffer"                "s" #'swiper-isearch
        :desc "Search buffer for thing at point" "S" #'swiper-isearch-thing-at-point)
```

So, maybe this is a good idea to fix it within the "Fix and improve ido module" task :)

belak:
I also wanted to start a discussion around this module in particular - I really like ido and would be interested in further changes/improvements, but I don't know how well they'd line up with the ivy/helm modules which seem to be better tested.

modules/completion/ido/config.el
```diff
@@ -19,41 +22,30 @@
...
-                  (time-less-p
-                   (sixth (file-attributes (concat ido-current-directory b)))
-                   (sixth (file-attributes (concat ido-current-directory a)))))))
-    (ido-to-end  ;; move . files to end (again)
-     (cl-loop for x in ido-temp-list
-              if (char-equal (string-to-char x) ?.)
-              collect x)))
-  (add-hook! '(ido-make-file-list-hook ido-make-dir-list-hook)
-             #'ido-sort-mtime)
+  (defun +ido-init-h ()
```
hlissner, line 25: Since this is now a `use-package!` block, the contents of `+ido-init-h` don't need to be in a function anymore (and shouldn't be, as it makes it harder for users to customize from their config.el).

Please move it out into `:config`.

modules/completion/ido/config.el
```diff
@@ -19,41 +22,30 @@
...
+    (define-key! (ido-common-completion-map ido-completion-map ido-file-completion-map)
+      "\C-n"   #'ido-next-match
+      "\C-p"   #'ido-prev-match
+      "<down>" #'ido-next-match
+      "<up>"   #'ido-prev-match
+      "TAB"    #'ido-exit-minibuffer
+      "\C-w"   #'ido-delete-backward-word-updir
```
hlissner, line 32: General implicitly wraps key strings with `kbd`, so escaping these keys (while not exactly wrong) is unnecessary. Please remove the backslashes on `C-n`, `C-p` and `C-w`.

modules/completion/ido/config.el
```diff
@@ -1,11 +1,14 @@
 ;;; completion/ido/config.el -*- lexical-binding: t; -*-
 
-(defun +ido-init-h ()
+(use-package! ido
+  :defer t
```
hlissner, line 4: Remove `:defer t`, as it is implied by `:after-call ...`.

modules/completion/ido/config.el
belak: Alright, good to know.

modules/completion/ido/config.el
belak: Good catch. I think this was a leftover from some other experiments.

hlissner:
Thanks for your help!

PR #2741, opened 2020-03-18
Adds documentation to map! on how to add bindings to existing prefixes

docbiz:


hlissner:
Thanks for the PR, but I cannot accept it because the documented use case does not match your description of it, and is technically already shown in other `map!` examples.

i.e. This is not an example of how to add a keybind to an existing prefix, it's an example of how to add a keybinding to a plugin's keymap, which is already documented in other examples, e.g.

```elisp
(map! :prefix "C-x"
      :i "C-l" #'+company/whole-lines)))
```

I'll look into adding an example for `:after`.

In any case, this particular snippet may be better suited to modules/editor/evil/README.org in a section under Configuration for adding evil-easymotion keybinds.

PR #2755, opened 2020-03-22
Only use the vterm popup rule for Doom vterm helpers

brettatoms:
The popup rule for vterm applies to all vterm buffers but sometimes I want to open vterm buffers to take up the full screen or window.  This PR make the Doom popup rule for vterm specific to the Doom vterm helpers.


brettatoms:
Realize this can be accomplished better in an `(after! vterm ....)` call.

PR #2760, opened 2020-03-23
Fix +evil:defun-txtobj docstring

oskarkv:
The docstring for +evil:defun-txtobj was obviously wrong.

Btw what a "defun" is is hard to describe. In Lisp it seems to be the top-level form, but in Python it's the closest `def`, not necessarily at the top-level.


hlissner:
By "defun" I am referring to the defun "thing" as defined in the thingatpt.el package, whose bounds are defined with:

```elisp
(put 'defun 'beginning-op 'beginning-of-defun)
(put 'defun 'end-op       'end-of-defun)
(put 'defun 'forward-op   'end-of-defun)
```

More specifically, it depends on what the current major mode has set `beginning-of-defun-function` and `end-of-defun-function` to.

PR #2761, opened 2020-03-23
dragndrop: insert #+attr_org: :width if specified

myshevchuk:
In line with latex and html attributes. Essential for high resolution displays.


hlissner:

myshevchuk:
Thank you!

Issue #2759, opened 2020-03-23
add roam flag to org module


bdarcus:
The new `org-roam` minor mode is now on MELPA. It would be great if it could be included as a flag on the `org` module, much like the `brain` flag for the similar `org-brain`.

Installation instructions are [here](https://org-roam.readthedocs.io/en/latest/installation/#doom-emacs), based on input from @hlissner.

hlissner:
As of f28a304 Doom's `:lang org` module now has a `+roam` flag to install org-roam, so I'll consider this resolved. Let me know if you encounter any issues using it. Thanks for bringing it to my attention!

bdarcus:
Thank you!

I'm running into a problem, though I don't understand what's going on behind the scenes to understand where that problem lies. Hopefully this screenshot demonstrates.

I added it over at [the org-roam repo issue also](https://github.com/jethrokuan/org-roam/issues/311).

![Screenshot from 2020-03-30 09-18-27](https://user-images.githubusercontent.com/1134/77916683-721e2e80-7267-11ea-9748-c939d2849a29.png)

bdarcus:
Looks like this is the problem, and solution.

https://github.com/hlissner/doom-emacs/commit/f28a304abd1b9e13c4e1217211bef305c5f3badc#r38132446

bdarcus:
One other question: how do you deal with iterating these packages? When the org-roam repo is updated past the pinned commit in doom, what happens?

PR #2764, opened 2020-03-24
Add basic module for org-roam

jethrokuan:
This is just a reference module for Org-roam, not sure if it's fully appropriate. #2759

### DIFF
--- a/modules/lang/org/contrib/roam.el
+++ b/modules/lang/org/contrib/roam.el
@@ -1,0 +1,29 @@
+;;; lang/org/contrib/roam.el -*- lexical-binding: t; -*-
+;;;###if (featurep! +roam)
+
+(use-package! org-roam
+  :commands (org-roam-insert
+             org-roam-find-file
+             org-roam-switch-to-buffer
+             org-roam-show-graph
+             org-roam)
+  :hook
+  (after-init . org-roam-mode)
+  :init
+  (map! :leader
+        :prefix "n"
+        :desc "org-roam" "l" #'org-roam
+        :desc "org-roam-insert" "i" #'org-roam-insert
+        :desc "org-roam-switch-to-buffer" "b" #'org-roam-switch-to-buffer
+        :desc "org-roam-find-file" "f" #'org-roam-find-file
+        :desc "org-roam-show-graph" "g" #'org-roam-show-graph
+        :desc "org-roam-insert" "i" #'org-roam-insert)
+  (setq org-roam-directory org-directory)
+  :config
+  (require 'org-roam-protocol))
+
+(use-package company-org-roam
+  :when (featurep! :completion company)
+  :after org-roam
+  :config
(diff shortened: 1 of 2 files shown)


hlissner:
Hey! I've been meaning to add a `+roam` flag, but haven't found time to look into it yet. I have a few issues with this PR, however:

1. It initializes org-roam-mode at startup, via `:hook (after-init . org-roam-mode)`. This pulls in all of org; a very expensive web of packages to load. Can this be avoided?
2. It binds its own leader keys. This should be relegated to the `:config default` module. Modules should only bind their own localleader keys.
3. Since I [lazy load org-protocol](https://github.com/hlissner/doom-emacs/blob/develop/modules/lang/org/config.el#L680-L710), it should be safe to lazy load org-roam-protocol separately, e.g.

```elisp
(use-package! org-roam-protocol
  :after org-protocol)
```

hlissner:
Oh, forgot to mention, could you also add `+roam` to the list of `:lang org`'s flags in a) [the module's README.org](https://github.com/hlissner/doom-emacs/blob/develop/modules/lang/org/README.org#module-flags) (in alphabetical order) under description > module flags and b) in [docs/modules.org](https://github.com/hlissner/doom-emacs/blob/develop/docs/modules.org#lang)

jethrokuan:
Ah no problem, I just copied whatever I had in my config over as a reference. This is especially since there is little information about the company backend for org-roam. It might be easier if you made a separate PR.

1. Yes, but the mode needs to be on for the rest of the functions to work properly. Calling the interactive functions don't load the mode, so some patching would need to be done here.

2. I see. I'm not using the evil mode, so the mapping stuff are still a bit confusing to me.

3. Ah okay!

Wetlize:
> Yes, but the mode needs to be on for the rest of the functions to work properly. Calling the interactive functions don't load the mode, so some patching would need to be done here.

Just place `(org-roam-mode +1)` at the end of `:config` section and it should activate the mode on the actual demand (i.e. when any of the autoloaded `:commands` will be called).

hlissner:
As of f28a304 the lang/org module now has a `+roam` flag, and should be sufficiently lazy loaded without compromising it. Let me know if that isn't the case and I'll have another look at it. Thanks for bringing it to my attention!

modules/lang/org/contrib/roam.el
```diff
@@ -0,0 +1,29 @@
...
+        :desc "org-roam-insert" "i" #'org-roam-insert
+        :desc "org-roam-switch-to-buffer" "b" #'org-roam-switch-to-buffer
+        :desc "org-roam-find-file" "f" #'org-roam-find-file
+        :desc "org-roam-show-graph" "g" #'org-roam-show-graph
+        :desc "org-roam-insert" "i" #'org-roam-insert)
+  (setq org-roam-directory org-directory)
+  :config
+  (require 'org-roam-protocol))
+
+(use-package company-org-roam
```
zeronone: Is it supposed to be `use-package!`?

PR #2765, opened 2020-03-24
Enable `mu4e-use-fancy-chars` by default

vqhuy:


hlissner:

PR #2777, opened 2020-03-26
Fix region formatting

sei40kr:
## Description

Fix `+format/buffer` not to take an argument because it results in an error:

```
let: Wrong number of arguments: (0 . 0), 1
```


hlissner:
Thanks for the PR!

PR #2779, opened 2020-03-26
irc: only use the switch buffer logic during creation

seanfarley:
This fixes a problem where calling IRC (e.g. C-c I I) will switch to
a (seemingly random but is ordered by load order) buffer.

----

#

### DIFF
--- a/modules/app/irc/autoload/irc.el
+++ b/modules/app/irc/autoload/irc.el
@@ -7,12 +7,14 @@
              (not inhibit-workspace))
     (+workspace-switch +irc--workspace-name 'auto-create))
   (let ((buffers (doom-buffers-in-mode 'circe-mode nil t)))
-    (if buffers
-        (ignore (switch-to-buffer (car buffers)))
-      (require 'circe)
-      (delete-other-windows)
-      (switch-to-buffer (doom-fallback-buffer))
-      t)))
+    (if (not (member (window-buffer) buffers))
+        (if buffers
+            (ignore (switch-to-buffer (car buffers)))
+          (require 'circe)
+          (delete-other-windows)
+          (switch-to-buffer (doom-fallback-buffer))
+          t)
+      (user-error "IRC buffer is already active and selected"))))
 
 ;;;###autoload
 (defun =irc (&optional inhibit-workspace)


modules/app/irc/autoload/irc.el
```diff
@@ -7,12 +7,13 @@
              (not inhibit-workspace))
     (+workspace-switch +irc--workspace-name 'auto-create))
   (let ((buffers (doom-buffers-in-mode 'circe-mode nil t)))
-    (if buffers
-        (ignore (switch-to-buffer (car buffers)))
-      (require 'circe)
-      (delete-other-windows)
-      (switch-to-buffer (doom-fallback-buffer))
-      t)))
+    (when (not (member (window-buffer) buffers))
```
hlissner, line 10: Please replace `(window-buffer)` with `(current-buffer)`

hlissner: And replace `(when (not ...))` with an if-else statement, whose else statement should announce that an IRC buffer is already available. e.g. `(user-error "IRC buffer is already active and selected")`, so the user knows the command "failed".

seanfarley: > Please replace `(window-buffer)` with `(current-buffer)`

That was my first implementation but it didn't seem to work. It always selected the `*server*` buffer for some reason (I guess it's the first buffer in the list?).

seanfarley: > And replace `(when (not ...))` with an if-else statement, whose else statement should announce that an IRC buffer is already available. e.g. `(user-error "IRC buffer is already active and selected")`, so the user knows the command "failed".

Ah, sure, that makes sense.

hlissner: > That was my first implementation but it didn't seem to work. It always selected the *server* buffer for some reason (I guess it's the first buffer in the list?).

Ah, that makes sense. In that case, disregard that request.

seanfarley: > > Please replace `(window-buffer)` with `(current-buffer)`
> 
> That was my first implementation but it didn't seem to work. It always selected the `*server*` buffer for some reason (I guess it's the first buffer in the list?).

Forgot to ask: is there an order of selecting the `current-buffer` that I'm missing / doing incorrectly?

seanfarley: Ok, I think that does what you asked!

hlissner: > Forgot to ask: is there an order of selecting the current-buffer that I'm missing / doing incorrectly?

No, Emacs changes the "current" buffer more often than it changes the selected window, so the selected window is a more reliable representation of what the user is actually looking at (in scenarios where you can't control the executing context -- which is usually the case when you're dealing with workspace switching).

hlissner:
Thanks for the help!

PR #2783, opened 2020-03-28
Improve org pretty-symbols for begin_src/end_src

jsravn:
When using `org-insert-structure-template`, it inserts the code blocks
as lowercase. This change tweaks the pretty-symbols to match the
lowercase version of `begin_src`/`end_src`.

### DIFF
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -179,7 +179,9 @@ background (and foreground) match the current theme."
   (set-pretty-symbols! 'org-mode
     :name "#+NAME:"
     :src_block "#+BEGIN_SRC"
-    :src_block_end "#+END_SRC"))
+    :src_block_end "#+END_SRC"
+    :src_block "#+begin_src"
+    :src_block_end "#+end_src"))
 
 
 (defun +org-init-babel-h ()


hlissner:
This breaks pretty-symbols for the uppercase versions, however. I'd rather fix `org-insert-structure-template`, or encourage folks to use our snippet library (which has `<s` and `src` snippets for `#+BEGIN_SRC lang` blocks).

jsravn:
@hlissner it seems to work okay w/ uppercase:

```
+pretty-code-symbols-alist is a variable defined in autoload.el.

Value
((org-mode
  ("#+end_src" . "«")
  ("#+begin_src" . "»")
  ("#+END_SRC" . "«")
  ("#+BEGIN_SRC" . "»")
  ("#+NAME:" . "»"))
 (t))
```

I can insert both upper/lower and it replaces it with the symbol.

I have snippets enabled but never used them! I've been using orgmode's `C-c C-,`. I can swap over if you don't want to accept this PR.

jsravn:
The snippets workflow is much nicer... so you can close this, unless you want to consider merging for anachronisms like `C-c C-,`.

jsravn:
I am noticing that pretty-code doesn't seem to replace the snippet block reliably:

Peek 2020-03-29 12-50

Here I can reproduce it pretty easily with a new org file. In other org files it seems to randomly not replace either BEGIN_SRC or END_SRC sometimes.

PR #2784, opened 2020-03-28
Add default gnome support for org +dragndrop

jsravn:
By default Gnome based distros (like Ubuntu) have `gnome-screenshot`
installed. This change modifies `+dragndrop` to support gnome-screenshot
by default.


hlissner:
Makes sense to me. Thanks!

PR #2785, opened 2020-03-28
Add dart to `init.example.el`

MYDavoodeh:
Thought it will be added, but I see it's absent after the merge so I added it myself.


hlissner:
Thanks!

PR #2786, opened 2020-03-28
Inhibit ivy-erlang-completion when lsp is active

andyjac:
I noticed this issue while trying to get `erlang_ls` up and running in doom. Basically, prior to this fix, if you have `ivy` enabled in your `init.el` and you add `(erlang +lsp)` to your config, you get the following message when attempting to open a `.erl` file:

```
File mode specification error: (file-missing Cannot open load file No such file or directory ivy-erlang-complete)
```

Like with `company` completion, this fix simply disables `ivy-erlang-completion` if the `+lsp` flag is active for erlang.

### DIFF
--- a/modules/lang/erlang/config.el
+++ b/modules/lang/erlang/config.el
@@ -14,6 +14,7 @@
 
 (use-package! ivy-erlang-complete
   :when (featurep! :completion ivy)
+  :unless (featurep! +lsp)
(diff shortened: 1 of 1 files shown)


hlissner:
I've removed ivy-erlang-completion in cf180be because it is redundant with company-erlang (which uses ivy-erlang-completion as a backend anyhow). Thanks for bringing this to my attention!

PR #2791, opened 2020-03-29
core-packages.el: silence compile warning for unused var `pin'

bandresen:
tiny PR :-)


hlissner:
There's nothing wrong with the PR, but there is a big rewrite of Doom's CLI and package manager in the works, which makes the modified code (and this problem) obsolete, so I'll have to turn it down, but thanks for the help anyway!

PR #2792, opened 2020-03-29
enable comint mode for terraform apply

xeijin:
terraform apply requires typing 'yes' to actually execute the change


hlissner:
I thought `compile` could handle stdin prompts, but I guess not? Thanks in any case!

PR #2794, opened 2020-03-30
Move away from obsolete org-roam function in org +roam

pnivlek:
The command org-roam-show-graph was marked obsolete in favor of org-roam-graph-show by org-roam commit 22b9d4b (before org-roam's pinning at b86d2c8), see https://github.com/jethrokuan/org-roam/pull/363.


hlissner:
Thanks for catching that. It should be reported to [org-roam that its docs are outdated, in that case](https://org-roam.readthedocs.io/en/latest/installation/).

PR #2799, opened 2020-03-31
grammar/langtool: support setting the path to the binary i/o jar

peterhoeg:
On nix we create a wrapper script that handles java for us, so allow setting `langtool-bin` instead of forcing looking for the `jar`.


hlissner:
Thanks!

PR #2803, opened 2020-03-31
Make uuidgen consistent for org-id-new

jsravn:
On macos/bsd systems, uuidgen produces all uppercase output. On
Linux/GNU systems it produces all lowercase output. This leads to
problems when managing org files across both systems (see e.g.
https://lists.gnu.org/archive/html/emacs-orgmode/2019-07/msg00081.html).

This change adds a defadvice for `org-id-new` to always lowercase the
output of `uuidgen` so the behavior is consistent across platforms.

### DIFF
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -514,12 +514,12 @@ current workspace (and clean them up)."
                     (apply old-create-image file-or-data type data-p props)))))
       (apply orig-fn args)))
 
-  (defadvice! +org--fix-inconsistent-uuidgen-case (orig-fn &rest args)
+  (defadvice! +org--fix-inconsistent-uuidgen-case-a (uuid)
     "Ensure uuidgen always produces lowercase output regardless of system."
-    :around #'org-id-new
-    (if (equal org-id-method 'uuid)
-        (downcase (apply orig-fn args))
-      (apply orig-fn args))))
+    :filter-return #'org-id-new
+    (if (eq org-id-method 'uuid)
+        (downcase uuid)
+      uuid)))
 
 
 (defun +org-init-keybinds-h ()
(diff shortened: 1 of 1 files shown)


jsravn:
To be honest I wasn't sure if I should open a PR for this. But this behavior in orgmode basically broke all my attachments when I swapped between Mac/Linux editing the same orgfiles. So I think it is something that could benefit other people doing something similar, and ideally should be fixed in org itself.

jsravn:
I don't think this should break existing behavior for users either. Existing IDs will be preserved - this just affects the creation of new IDs. Unfortunately, as in my case, the attach folder remains broken due to a mixture of lowercase/uppercase and MacOS/Dropbox's case insensitivity.

modules/lang/org/config.el
```diff
@@ -512,6 +512,13 @@ current workspace (and clean them up)."
                 (lambda (file-or-data &optional type data-p &rest props)
                   (let ((type (if (plist-get props :width) type)))
                     (apply old-create-image file-or-data type data-p props)))))
+      (apply orig-fn args)))
+
+  (defadvice! +org--fix-inconsistent-uuidgen-case (orig-fn &rest args)
+    "Ensure uuidgen always produces lowercase output regardless of system."
+    :around #'org-id-new
+    (if (equal org-id-method 'uuid)
+        (downcase (apply orig-fn args))
```
hlissner, line 521: Please use `:filter-return` instead of `:around`. I prefer that advice be as surgical as possible. e.g.

```elisp
(defadvice! +org--fix-inconsistent-uuidgen-case-a (uuid)
  "Ensure uuidgen always produces lowercase output regardless of system."
  :filter-return #'org-id-new
  (if (eq org-id-method 'uuid)
      (downcase uuid)
    uuid))
```

hlissner:
Looks good to me! Thanks!

PR #2806, opened 2020-04-01
Enable pipenv-with-projectile

### LINKED ISSUE #1666: [REQUEST] Automatically activate a pipenv environment if available
It would be really nice if `doom-emacs` tried to run `pipenv-activate` on Python files.

Rationale:

1. Spacemacs [does it](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/python/funcs.el#L14)
2. It's a noop if a `pipenv` is not available
3. (most important) in LSP mode, `lsp-mode` activates automatically for every file. If the language server is installed into a virtual environment, `lsp-mode` will not find it without activated virtualenv.

A possible complication is that while `lsp-mode` starts a language server per workspace, `pyvenv` [applies its configuration globally](https://github.com/jorgenschaefer/pyvenv/issues/47), so that something like a `pipenv-deactivate` + `pipenv-activate` combo is needed every time a buffer is opened or switched to.

mathieupost:
Pipenv was not automatically enabled when entering a Pipenv project, now it is. Not sure why it was disabled explicitly in the first place.

I guess this also resolves #1666

### DIFF
--- a/modules/lang/python/config.el
+++ b/modules/lang/python/config.el
@@ -209,7 +209,7 @@ called.")
       (:description . "Run Python script")))
   (map! :map python-mode-map
         :localleader
-        :prefix "e"
+        :prefix ("e" . "pipenv")
         :desc "activate"    "a" #'pipenv-activate
         :desc "deactivate"  "d" #'pipenv-deactivate
         :desc "install"     "i" #'pipenv-install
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR! Sorry to say, I cannot accept it. `pipenv-with-projectile` was intentionally disabled because all venv management is manual in Doom. I'd recommend you use direnv if you want it to be automatic, otherwise set `pipenv-with-projectile` yourself, manually. This may change when I get around to rethinking the python module, but for now I'd rather it stay this way.

mathieupost:
Alright, good to know! I'll use direnv instead then :)

PR #2819, opened 2020-04-02
Enable treemacs workspace/persp support

jsravn:
With https://github.com/Alexander-Miller/treemacs/issues/592 merged,
treemacs now fully supports perspective mode. That is, each treemacs
buffer is scoped to a perspective and initializes itself using projectile.

This change re-enables treemacs-persp, enables the `Perspectives` scope,
and adjusts the doom treemacs init behavior to support the new
treemacs-persp behavior.

I haven't been able to reproduce the prior errors with melpa so I think
this is safe to re-enable now. In my testing it seems to work
well.

Only one issue is that users might need to remove their treemacs persist
file (`~/.emacs.d/.local/cache/treemacs-persist`) after this change if
using persp-mode. I'm not sure if it is necessary since I blew away my
own before testing.

### DIFF
--- a/modules/ui/treemacs/autoload.el
+++ b/modules/ui/treemacs/autoload.el
@@ -6,8 +6,9 @@
     (cl-letf (((symbol-function 'treemacs-workspace->is-empty?)
                (symbol-function 'ignore)))
       (treemacs--init))
-    (dolist (project (treemacs-workspace->projects (treemacs-current-workspace)))
-      (treemacs-do-remove-project-from-workspace project))
+    (unless (bound-and-true-p persp-mode)
+      (dolist (project (treemacs-workspace->projects (treemacs-current-workspace)))
+        (treemacs-do-remove-project-from-workspace project)))
     (with-current-buffer origin-buffer
       (let ((project-root (or (doom-project-root) default-directory)))
         (treemacs-do-add-project-to-workspace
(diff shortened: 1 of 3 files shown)


jsravn:
@hlissner the only minor thing I've noticed is that on a fresh treemacs startup (empty `treemacs-persist`) it always gives the popup to select the treemacs project. I narrowed it down to this section of `treemacs--init`:

```lisp
       (when (treemacs-workspace->is-empty?)
         (-> (treemacs--read-first-project-path)
             (treemacs--canonical-path)
             (treemacs-do-add-project-to-workspace))
         (treemacs-log "Created first project."))
```

Which baffles me, since in `+treemacs--init` it expressly disables this:

```lisp
    (cl-letf (((symbol-function 'treemacs-workspace->is-empty?)
               (symbol-function 'ignore)))
      (treemacs--init))
```

This is reaching the limits of my elisp knowledge though. Stepping through edebug I see it definitely seems to execute the above code in treemacs-init, so for some reason the cl-letf is not applying.

jsravn:
^ seems related to byte compilation. If I delete treemacs-core-utils.elc everything works fine.

hlissner:
Thanks for the PR!

> The only minor thing I've noticed is that on a fresh treemacs startup (empty treemacs-persist) it always gives the popup to select the treemacs project. I narrowed it down to this section of treemacs--init:

I'll look into this shortly.

PR #2824, opened 2020-04-04
fix: revert ranger version to fix paste over

JakDar:
With current version of ranger.el if you copy file with ranger and paste it in the same place ranger fails  with erro instead of  appending "~1" as before. 

Related issue:
https://github.com/ralesi/ranger.el/issues/217

If newer features are not worth it, maybe we should pin older commit while waiting for fix.


yuhan0:
I opened a PR upstream to fix the issue: https://github.com/ralesi/ranger.el/pull/220 - sorry about that

hlissner:
I'll keep an eye on that PR and bump once it is merged. Thanks for the PR and for bringing it to my attention.

PR #2829, opened 2020-04-06
Change projectile dirconfig file to .project

jsravn:
So it matches the doom project marker `.project`. Otherwise you need to
create both a `.project` and then a `.projectile` with the project
configuration.

To maintain backwards compatibility, it will prefer a `.projectile` file
for dirconfig if found.

### DIFF
--- a/core/core-projects.el
+++ b/core/core-projects.el
@@ -75,6 +75,17 @@ Emacs.")
 
   (push (abbreviate-file-name doom-local-dir) projectile-globally-ignored-directories)
 
+  ;; Override projectile's dirconfig file '.projectile' with doom's project marker '.project'.
+  (defun projectile-dirconfig-file ()
+    "Return the absolute path to the project's dirconfig file.
+
+By default this will be `.project' in the root of the project.
+If a `.projectile' file exists, it will use that instead."
+    (cond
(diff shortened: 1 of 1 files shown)


jsravn:
This only seems to do anything if using projectile's native indexing. So it's probably not that useful, although may help some people out. If using default projectile with fd installed, you have to create a `.ignore` file to ignore directories (which isn't documented anywhere).

core/core-projects.el
```diff
@@ -75,6 +75,17 @@ Emacs.")
 
   (push (abbreviate-file-name doom-local-dir) projectile-globally-ignored-directories)
 
+  ;; Override projectile's dirconfig file '.projectile' with doom's project marker '.project'.
+  (defun projectile-dirconfig-file ()
```
hlissner, line 79: Please use `defadvice!` to override functions.

```elisp
(defadvice! +doom--recognize-project-el-dirconfig-file-a ()
  "Recognize .project as a dirconfig file, besides .projectile."
  :override #'projectile-dirconfig-file
  ...)
```

hlissner: Also, either remove the comment over the advice, or clarify the docstring, as they are kind of redundant.

hlissner:
Looks good to me! Thanks for the help!

PR #2830, opened 2020-04-06
fix: change display-line-numbers-type to 't

Lenin1917:
I can not change display-line-numbers-type variable for org buffer https://github.com/Lenin1917/dotfiles/blob/master/emacs/.config/doom.d/config.el#L25 before edit t -> 't

### DIFF
--- a/core/templates/config.example.el
+++ b/core/templates/config.example.el
@@ -32,7 +32,7 @@
 
 ;; This determines the style of line numbers in effect. If set to `nil', line
 ;; numbers are disabled. For relative line numbers, set this to `relative'.
-(setq display-line-numbers-type t)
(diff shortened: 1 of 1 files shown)


hlissner:
Hi! Thanks for the PR, but I cannot accept it because this change does nothing. That it fixes your issue is definitely a misunderstanding of the problem, because `t` and `'t` are equivalent.

If you want to change `display-line-numbers-type` in `org-mode`, adding the following to `~/.doom.d/config.el`:

```elisp
(setq-hook! 'org-mode-hook
  display-line-numbers-type NEWVALUE)
```

Lenin1917:
Thanks!

Issue #2827, opened 2020-04-05
"bibliographic" module


bdarcus:
# Description

We need a “bibtex” or “bibliographic” module, similar to the bibtex layer in spacemacs.

There’s an earlier PR #698, which started as a suggested “reference” module, but then evolved into a `+reference` flag on the `org` module, and I guess was never finished.

I would argue it would be better to have a separate module, since the functionality here needs to provide integration into document authoring workflows in `org`, `markdown` (see screeenshot below of what this looks like), and `latex`, along with support for working from bibtex and biblatex files.

![Screenshot from 2020-04-05 10-54-09](https://user-images.githubusercontent.com/1134/78501872-d93a5800-772b-11ea-8353-09f97c72c9b1.png)

# Spacemacs Implementation

What I'm proposing is similar to the spacemacs [bibtex layer](https://github.com/syl20bnr/spacemacs/tree/master/layers/%2Blang/bibtex), which integrates the following packages:

- Auctex
- Org
- Org-ref
- Markdown-mode
- Helm-bibtex
- Biblio
- biblio-core

All of the above are on MELPA.

A doom module should obviously also integrate `ivy-bibtex` (which provides equivalent functionality to `helm-bibtex`) depending on which completion engine feature is selected.

I also wonder about two other packages:

1. `org-noter` (when the `pdf` and `org` features are selected?); #698 contemplates adding this to the pdf module, but I think it belongs here
2. `ebib` (here I'm less sure; maybe others with experience could weigh in?)

# Conclusion

I know a lot about bibliographic tools in general (I created the CSL language that is used by a lot of bibliographic tools and services), less about emacs and doom, and elisp programming. But I am doing what I can to cobble this together [here](https://github.com/bdarcus/.doom.d/blob/master/org-academic.org#bibliographic-management-and-integration), where the doom org module and org-roam support is the core.

I will continue to work on this and do what I can, but hoping someone with greater expertise than I will ultimately submit a proper doom module.

If this is an appropriate request, I can post additional details (like commands) as a needed.

brianmcgillion:
modules/lang/latex has most of the setup. The notable exceptions are org-ref and org-noter.

bdarcus:
Now that I'm thinking about this more, am not exactly sure how this should all fit together in the context of doom. We basically have a few areas of functionality:
1. editing and working with bib(la)tex files
2. finding and importing bibliographic data and adding to those bib(la)tex files (biblio and org-ref have overlapping functionality here)
[...]

PR #2838, opened 2020-04-07
Remove Ivy/helm-bibtex

patrl:
Ivy/helm-bibtex is unused throughout the latex module, aside from implicitly by setting the `bibtex-completion-bibliography` variable. This is confusing, and unnecessary.

Ivy/helm-bibtex is a great tool, but it is also language agnostic (it can be used to insert citations in org and markdown modes), and latex mode already has excellent citation autocompletion from `company-reftex`. In my opinion, ivy/helm-bibtex would be a good candidate for inclusion in a distinct bib module, but should be removed from the latex module.

### DIFF
--- a/modules/lang/latex/+ref.el
+++ b/modules/lang/latex/+ref.el
@@ -1,10 +1,5 @@
 ;;; lang/latex/+ref.el -*- lexical-binding: t; -*-
 
-(when (stringp +latex-bibtex-file)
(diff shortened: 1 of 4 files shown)


hlissner:
Sounds good to me (as does a separate biblio module, but I'll follow up on that in #2827). Thanks for the help!

PR #2842, opened 2020-04-08
tools/lookup: fix synosaurus -> synosaurus-wordnet

paulapatience:
The synosaurus-wordnet--command variable is defined in synosaurus-wordnet rather than in synosaurus.


hlissner:
Ah, good catch. Thanks!

PR #2845, opened 2020-04-08
Disable deft's buffer auto-save

jsravn:
By default deft will save a changed buffer that it opens after 1 second.

This causes problematic interactions with the rest of doom - like with
wsbutler (insert space, try to insert a link, and it will eat the
space before the link is inserted).

I think it's best for auto-save to be managed by a separate package,
rather than this ad-hoc save that occurs only in buffers that deft opens.


hlissner:
Makes sense to me. Thanks!

Issue #2846, opened 2020-04-08
Should org-roam start up w/ org-mode?


jsravn:
**What did you expect to happen?**
I restore a session with several org buffers in the org-roam directory. Or I open up several org buffers directly with projectile, deft, or find-file. Then I run `org-roam` and expect to see the backlinks, etc. working in the org-roam buffer. 

**What actually happened?**
The `org-roam` buffer is empty for all of the org buffers. This is because org-roam-mode needs to be started _before_ any org-roam files are opened, to insert its `post-command-hook` when the buffer is opened.

As a workaround I have to close all org buffers, then run `org-roam`, then reopen them.

Currently org-roam-mode is configured to lazily load on an org-roam command only. Wouldn't it be better to do `:after org` to ensure it is always loaded on time? That way we ensure the org-roam hooks are always inserted for org buffers in the org-roam directory.

PR #2847, opened 2020-04-08
Load org-roam after org-mode

jsravn:
Fix https://github.com/hlissner/doom-emacs/issues/2846

### DIFF
--- a/modules/lang/org/contrib/roam.el
+++ b/modules/lang/org/contrib/roam.el
@@ -2,16 +2,7 @@
 ;;;###if (featurep! +roam)
 
 (use-package! org-roam
-  :commands (org-roam
-             org-roam-capture
-             org-roam-date
-             org-roam-find-file
-             org-roam-graph-show
-             org-roam-insert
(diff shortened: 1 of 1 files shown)


modules/lang/org/contrib/roam.el
```diff
@@ -2,16 +2,7 @@
...
-             org-roam-capture
-             org-roam-date
-             org-roam-find-file
-             org-roam-graph-show
-             org-roam-insert
-             org-roam-switch-to-buffer
-             org-roam-today
-             org-roam-tomorrow
-             org-roam-yesterday)
+  :after org
```
hlissner, line 5: Please restore the `:commands` statement, otherwise none of the keybinds will work before org-roam has loaded.

hlissner: Also, lets use `:hook (org-load . org-roam-mode)` and remove `(org-roam-mode +1)` from its `:config`'s body.

hlissner:

PR #2851, opened 2020-04-08
Fix install link

tbille:
The install link in the docs is incorrect. This will link to the install section at the top of the file.


hlissner:
Good catch. Thanks!

Issue #2852, opened 2020-04-08
lsp-mode set-lookup-handlers! doesn't seem to override the major mode handlers


jsravn:
**What did you expect to happen?**
`+lookup/references` inside of a golang file with lsp mode, I expect it to use the LSP functions.

**What actually happened?**
It ignores `lsp-find-references` and runs the default go-guru command.

**Additional details:**
- https://github.com/jsravn/dotfiles/tree/master/dot_doom.d
- `SPC c D`

If I comment out the default lookup handlers, everything works as expected. This seems to point to a problem with `set-lookup-handlers!`: 

https://github.com/hlissner/doom-emacs/blob/develop/modules/lang/go/config.el#L9-L12

**Steps to reproduce:**
1. Enable `(go +lsp)`
2. Check out a golang project, 
3. Observe none of the lsp lookup handlers are invoked

**System information:**
<details><pre>
emacs   version    27.0.90
        features   XPM JPEG TIFF GIF PNG RSVG CAIRO SOUND GPM DBUS GLIB NOTIFY INOTIFY ACL GNUTLS LIBXML2 FREETYPE HARFBUZZ LIBOTF ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS LIBSYSTEMD JSON PDUMPER LCMS2 GMP
        build      Apr 07, 2020
        buildopts  (--prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib --localstatedir=/var --mandir=/usr/share/man --with-gameuser=:games --with-sound=alsa --with-modules --without-gconf --without-gsettings --with-x-toolkit=gtk3 --without-xaw3d --without-m17n-flt --with-cairo --without-compress-install 'CFLAGS=-march=x86-64 -mtune=generic -O2 -pipe -fno-plt -g -flto -s -fuse-ld=gold' CPPFLAGS=-D_FORTIFY_SOURCE=2 LDFLAGS=-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now)
        windowsys  x
        daemonp    server-running
doom    version    2.0.9
        build      HEAD -> develop, origin/develop, origin/HEAD 8d7751871 2020-04-08 04:15:16 -0400
        dir        ~/.doom.d/
system  type       gnu/linux
        config     x86_64-pc-linux-gnu
        shell      /usr/bin/zsh
[...]

PR #2853, opened 2020-04-08
Fix helm-xref when using Emacs 27+

jsravn:
Fixes https://github.com/hlissner/doom-emacs/issues/2852


hlissner:

PR #2854, opened 2020-04-08
web: add +web-django-mode

seanfarley:
This adds support for django templates.

I'm not sure the mode line is correct but at least I got this working in my
setup.


hlissner:
Thanks!

PR #2856, opened 2020-04-08
Add +lsp to Julia and override some defaults in julia-repl

AlexLewandowski:
This pull request adds a README, additional functionality for `julia-repl` and a `+lsp` flag to the Julia module.

The language server must be manually installed, as per guidelines.

I have overriden some defaults in `julia-repl`. In particular, I override:

- `julia-repl--bufer-name` to name the Julia REPL as `*julia:[workspace-name]*`. 
- The escape sequence from `C-c C-x` to `C-c C-c`.


hlissner:
This looks good to me, though I don't like external (interactive) tools -- like REPLs -- opening themselves when you invoke a major mode, so I've deferred julia-repl. The canonical way to invoke a REPL is `SPC o r` (in a popup) or `SPC o R` (replace the current window). Or `C-c o r` / `C-c o R` if you don't use evil.

I've also made the namespacing-by-workspace advice conditional, as it would throw an error if `:ui workspaces` were disabled.

PR #2857, opened 2020-04-09
Fix company-coq temporarily

ccyip:
This patch is slightly more general than the one I posted in discord, so I think other coq folks may be interested in having it too. I actually think it is cleaner to just have a blacklist variable for `+company-init-backends-h`, as it has already done something similar to `fundamental-mode` and `special-mode`. But you may have a better plan for this. Anyway, feel free to close it if you think this hack is too dirty to merge :)

### DIFF
--- a/modules/lang/coq/config.el
+++ b/modules/lang/coq/config.el
@@ -89,3 +89,18 @@
           "e" #'company-coq-document-error
           "E" #'company-coq-browse-error-messages
           "h" #'company-coq-doc)))
+
+;; `+company-init-backends-h' in `after-change-major-mode-hook' overrides
+;; `company-backends' set by `company-coq' package. This dirty hack fixes
+;; completion in coq-mode. TODO: remove when company backends builder is
+;; reworked.
+(defvar-local +coq--company-backends nil)
+(after! company-coq
+  (defun +coq--record-company-backends-h ()
+    (setq +coq--company-backends company-backends))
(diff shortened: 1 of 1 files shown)


hlissner:
Sorry for the late response. This issue has to do with a larger, more systemic issue with `set-company-backend!` and the `:completion company` module's opinions. A proper fix will have to happen there, rather than pepper our modules with hacks to address it, so I will have to refuse this module. I've created [a project](https://github.com/hlissner/doom-emacs/projects/8) to track it. Anyhow, thanks for bringing this to my attention!

quinn-dougherty:
for the record / in case anyone's reading this later: I tried @ccyip's code in my `config.el` and it did not solve the problem. 

In the past other doom builds would have company running on any coq buffer as long as a second coq file was booted somewhere in the session at some point, and I haven't even had that for years (tho that may be to do with nixos?)

ZHANGTIANYAO1:
> for the record / in case anyone's reading this later: I tried @ccyip's code in my `config.el` and it did not solve the problem.
> 
> In the past other doom builds would have company running on any coq buffer as long as a second coq file was booted somewhere in the session at some point, and I haven't even had that for years (tho that may be to do with nixos?). I can still make greek letters with company but it does not autofill.
> 
> Has anyone tried helm, ido, or ivy for coq?

Same issue

PR #2861, opened 2020-04-09
adding new biblio module

bdarcus:
Addressing #2827, this is a minimal biblio module that provides `ivy-bibtex` or `helm-bibtex` (which were recently removed from the latex module in #2838) depending on completion feature, and provides a single global keybinding.

The idea is this provides the core functionalty, and the community can iteratively build it up as appropriate.

### DIFF
--- a/modules/tools/biblio/config.el
+++ b/modules/tools/biblio/config.el
@@ -1,0 +1,7 @@
+;;; tools/biblio/config.el -*- lexical-binding: t; -*-
+
+(use-package! ivy-bibtex
+  :when (featurep! :completion ivy)
+  :defer t
+  :config
+  (add-to-list 'ivy-re-builders-alist '(ivy-bibtex . ivy--regex-plus)))
--- a/modules/tools/biblio/packages.el
+++ b/modules/tools/biblio/packages.el
@@ -1,0 +1,7 @@
+;; -*- no-byte-compile: t; -*-
+;;; tools/biblio/packages.el
+
+(when (featurep! :completion ivy)
+  (package! ivy-bibtex :pin "3cff6bd702"))
+(when (featurep! :completion helm)
+  (package! helm-bibtex :pin "3cff6bd702"))
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -123,6 +123,11 @@
       (:prefix-map ("n" . "notes")
         :desc "Search notes for symbol"        "." #'+default/search-notes-for-symbol-at-point
         :desc "Org agenda"                     "a" #'org-agenda
+        (:when (featurep! :tools biblio)
+          :desc "Bibliographic entries"        "b"
+          (cond ((featurep! :completion ivy)   #'ivy-bibtex)
+                ((featurep! :completion helm)  #'helm-bibtex)))
+
         :desc "Find file in notes"             "f" #'+default/find-in-notes
         :desc "Browse notes"                   "F" #'+default/browse-notes
         :desc "Org store link"                 "l" #'org-store-link
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -454,6 +454,11 @@
       (:prefix-map ("n" . "notes")
         :desc "Search notes for symbol"      "*" #'+default/search-notes-for-symbol-at-point
         :desc "Org agenda"                   "a" #'org-agenda
+        (:when (featurep! :tools biblio)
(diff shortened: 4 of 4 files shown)


patrl:
This is probably a matter of taste, but I would personally favor this being a global binding, rather than conventionalized on particular modes. Ivy/helm-bibtex essentially provides a UI for, e.g., opening pdfs, opening note files, etc. Doom already includes bindings under `SPC n` for, e.g., navigating to org-roam files, so I think this is defensible.

bdarcus:
No objection from me.

bdarcus:
I updated to make the binding global. But it only works if ivy-bibtex or helm-bibtex is first loaded. I assume I need a hook, but not sure how best to do that.

bdarcus:
I added global keybindings for evil and emacs and rebased. Hope this is OK.

bdarcus:
I just fixed a dumb typo.

bdarcus:
Good question.

What should happen?

I'm getting to the edge of my depth here with elisp/emacs/doom :-)

bdarcus:
Can/should we include one as a requirement in the docs?

patrl:
That seems like a reasonable solution. Once/if we add additional tools such as ebib, this will be redundant in any case.

bdarcus:
Note that #2888 adds an additional package (org-ref) and the README.org file.

modules/tools/biblio/config.el
```diff
@@ -0,0 +1,13 @@
+;;; tools/biblio/config.el -*- lexical-binding: t; -*-
+
+(use-package! ivy-bibtex
+  :when (featurep! :completion ivy)
+  :commands (ivy-bibtex)
```
hlissner, line 5: Replace `:commands (ivy-bibtex)` with `:defer t`. `ivy-bibtex` [autoloads this command for us](https://github.com/tmalsburg/helm-bibtex/blob/master/ivy-bibtex.el#L124), so we don't have to.

modules/tools/biblio/config.el
```diff
@@ -0,0 +1,13 @@
...
+  :config
+  (add-to-list 'ivy-re-builders-alist '(ivy-bibtex . ivy--regex-plus)))
+
+(use-package! helm-bibtex
+  :when (featurep! :completion helm)
+  :commands (helm-bibtex))
```
hlissner, line 11: Remove this block. `helm-bibtex` [autoloads this command for us](https://github.com/tmalsburg/helm-bibtex/blob/master/helm-bibtex.el#L205), so we don't have to.

modules/tools/biblio/config.el
```diff
@@ -0,0 +1,13 @@
...
+
+;;; config.el ends here
```
hlissner, line 13: Please remove this line. This is not a convention Doom follows for its modules' files.

modules/tools/biblio/packages.el
```diff
@@ -0,0 +1,7 @@
+;; -*- no-byte-compile: t; -*-
+;;; tools/biblio/packages.el
+
+(when (featurep! :completion ivy)
+  (package! ivy-bibtex))
+(when (featurep! :completion helm)
+  (package! helm-bibtex))
```
hlissner, line 7: Please pin both these packages. Put your cursor in their `package!` form and run `M-x doom/update-pinned-package-form`.

bdarcus:
I just pushed the requested changes.

modules/tools/biblio/config.el
```diff
@@ -0,0 +1,10 @@
+;;; tools/biblio/config.el -*- lexical-binding: t; -*-
+
+(use-package! ivy-bibtex
+  :when (featurep! :completion ivy)
+  :defer t
+  :config
+  (add-to-list 'ivy-re-builders-alist '(ivy-bibtex . ivy--regex-plus)))
+
+(use-package! helm-bibtex
+  :when (featurep! :completion helm))
```
hlissner, line 10: This has to be removed too, otherwise it is eagerly loading the package at startup, if you have helm enabled. No `use-package!` block is needed for this package, since it autoloads all its entry points for us.

bdarcus: So you're saying we don't need anything in config.el for helm-bibtex at this point?

But we do for ivy-bibtex, because of the config details?

bdarcus:
I'm not sure how notification works on these tickets, but I think I'm done.

hlissner:
Looks good to me, thanks!

bdarcus:
Thank you! I appreciate your patience.

>

Issue #2869, opened 2020-04-10
magit error when switching to a new projectile project


jsravn:
**What did you expect to happen?**
Switch to a new project without errors.

**What actually happened?**
Get the below error the first time I try to switch projects.

<details><pre>
Debugger entered--Lisp error: (error "Buffer does not seem to be associated with any fil...")
  signal(error ("Buffer does not seem to be associated with any fil..."))
  error("Buffer does not seem to be associated with any fil...")
  revert-buffer--default(t t)
  revert-buffer(t t)
  (if (or (buffer-modified-p buffer) (and file (file-exists-p file))) (if (and (boundp 'vc-mode) vc-mode) (progn (vc-refresh-state))) (revert-buffer t t))
  (let* ((buffer (or (buffer-base-buffer) (current-buffer))) (file (buffer-file-name buffer))) (if (or (buffer-modified-p buffer) (and file (file-exists-p file))) (if (and (boundp 'vc-mode) vc-mode) (progn (vc-refresh-state))) (revert-buffer t t)))
  (save-current-buffer (set-buffer buffer) (kill-local-variable '+magit--stale-p) (let* ((buffer (or (buffer-base-buffer) (current-buffer))) (file (buffer-file-name buffer))) (if (or (buffer-modified-p buffer) (and file (file-exists-p file))) (if (and (boundp 'vc-mode) vc-mode) (progn (vc-refresh-state))) (revert-buffer t t))))
  +magit--revert-buffer(#<buffer *doom*>)
  (progn (+magit--revert-buffer (current-buffer)))
  (if +magit--stale-p (progn (+magit--revert-buffer (current-buffer))))
  +magit-revert-buffer-maybe-h()
  run-hooks(doom-switch-buffer-hook)
  (save-current-buffer (set-buffer (if (windowp buffer) (window-buffer buffer) buffer)) (run-hooks 'doom-switch-buffer-hook))
  (progn (save-current-buffer (set-buffer (if (windowp buffer) (window-buffer buffer) buffer)) (run-hooks 'doom-switch-buffer-hook)) buffer)
  (if buffer (progn (save-current-buffer (set-buffer (if (windowp buffer) (window-buffer buffer) buffer)) (run-hooks 'doom-switch-buffer-hook)) buffer) nil)
[...]

PR #2870, opened 2020-04-10
Fix magit hook trying to revert buffer without a file

jsravn:
Fix https://github.com/hlissner/doom-emacs/issues/2869


jsravn:
Not sure this is the right approach as I didn't have time to dig too much into what the magit-revert stuff is supposed to be doing.

hlissner:
Thanks. I'd already been working on a solution to this, but hadn't pushed it yet. As of 3622e36 it should be fixed.

PR #2872, opened 2020-04-10
Clarify LaTeX prerequisites

yoavm448:
Ghotscript <=9.27 is [reported buggy in general](https://www.gnu.org/software/auctex/manual/preview-latex/No-images-are-displayed-with-gs-9_002e27-and-earlier.html), and auctex's preview doesn't need `dvipng`/`dvips`. It can use them if you ask it to, but the default pipeline is only pdflatex and ghotscript. 

I couldn't be sure if `preview` actually used `dvi(png|ps)`, so I manually moved them out of `PATH` and re-ran the previews. (It worked.)


hlissner:
Thanks!

PR #2874, opened 2020-04-10
[WIP] - RFQ on adding or-ref and org-noter

brianmcgillion:
https://github.com/hlissner/doom-emacs/pull/2861

With the above change in for review would it make more sense to put org-ref and org-noter to the biblio feature set rather than including it in org?


HaoZeke:
I'm in favor of leaving it in the `org` module, since it might be confusing otherwise.

PR #2883, opened 2020-04-11
lang/org: add support for pivotaltracker, trello and sidebar [WIP]

peterhoeg:
It's WIP because I would like confirmation on the direction before doing all the documentation.

Couple of things:

1. take up "localleader S" for Sync (trello and pt)
2. take up "localleader u" for "UI toggle" (sidebar)
3. instead of sticking the features under contrib, they are in the root prefixed with a + like the manual says. I'm guessing the contrib bit is a holdover from the early days.
4. versions not `:pin`ned yet

Questions:
1. Should I move the other contrib things into +name instead?

### DIFF
--- a/modules/lang/org/+jira.el
+++ b/modules/lang/org/+jira.el
@@ -1,0 +1,5 @@
+;;; lang/org/+jira.el -*- lexical-binding: t; -*-
+;;;###if (featurep! +jira)
+
+(use-package! ox-jira
+  :defer-incrementally org)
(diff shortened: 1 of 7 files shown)


hlissner:
Hi! Thanks for the PR, but I cannot accept these submodules because their configs are too trivial to warrant inclusion (particularly +jira).

That said, for posterity, these should have been in the contrib/ folder. It's not a holdover, I simply suspect there will be a lot of them, considering the nature of org-mode, and I'd rather use it than pollute the module's root.

PR #2884, opened 2020-04-11
Add org-noter package

brianmcgillion:
This package allows the creation of notes on a document (pdf, EPub,
etc.) which will be kept in sync with the document. Providing context
sensitive annotations that are managed in plaintext.

### DIFF
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -128,7 +128,9 @@
           :desc "Bibliographic entries"        "b"
           (cond ((featurep! :completion ivy)   #'ivy-bibtex)
                 ((featurep! :completion helm)  #'helm-bibtex)))
-
+        (:when (featurep! :lang org +noter)
+          :desc "Org noter"                    "e" #'org-noter)
+      
         :desc "Find file in notes"             "f" #'+default/find-in-notes
         :desc "Browse notes"                   "F" #'+default/browse-notes
         :desc "Org store link"                 "l" #'org-store-link
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -128,6 +128,10 @@
           :desc "Bibliographic entries"        "b"
           (cond ((featurep! :completion ivy)   #'ivy-bibtex)
                 ((featurep! :completion helm)  #'helm-bibtex)))
+
+        :desc "Toggle org-clock"               "c" #'+org/toggle-clock
+        :desc "Cancel org-clock"               "C" #'org-clock-cancel
+        :desc "Open deft"                      "d" #'deft
         (:when (featurep! :lang org +noter)
           :desc "Org noter"                    "e" #'org-noter)
       
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -464,6 +464,9 @@
         :desc "Toggle org-clock"             "c" #'+org/toggle-clock
         :desc "Cancel org-clock"             "C" #'org-clock-cancel
         :desc "Open deft"                    "d" #'deft
+        (:when (featurep! :lang org +noter)
+          :desc "Org noter"                  "e" #'org-noter)
+
         :desc "Find file in notes"           "f" #'+default/find-in-notes
         :desc "Browse notes"                 "F" #'+default/browse-notes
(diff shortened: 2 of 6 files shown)


modules/lang/org/contrib/noter.el
```diff
@@ -0,0 +1,22 @@
...
+  :commands org-noter
+  :preface
+  ;; Allow the user to preempt this and set the document search path
+  ;; If not set then use `org-directory'
+  (defvar org-noter-notes-search-path nil)
+  :init
+  (map! :leader
+        ;;; <leader> n --- notes
+        (:prefix-map ("n" . "notes")
+          :desc "Org Noter"          "r" #'org-noter
```
bdarcus, line 16: Minor thing, but isn't it called "Org-noter"; with a hyphen?

hlissner: Modules shouldn't be defining their own leader keys (localleader keys are fine though).

Please move this leader key to the `:config default` module.

hlissner: That said, `SPC n r` is already being used by org-roam, so I'll have to consider a new key for it...

modules/lang/org/contrib/noter.el
```diff
@@ -0,0 +1,22 @@
+;;; lang/org/contrib/noter.el -*- lexical-binding: t; -*-
+;;;###if (featurep! +noter)
+
+
+;; org-noter
```
hlissner, line 5: Please remove the extraneous space and redundant comment.

modules/lang/org/contrib/noter.el
```diff
@@ -0,0 +1,22 @@
...
+(use-package! org-noter
+  :commands org-noter
```
hlissner, line 7: It appears [org-noter already autoloads this command](https://github.com/weirdNox/org-noter/blob/master/org-noter.el#L2081) so we don't have to. Best to replace `:commands org-noter` with `:defer t`.

modules/lang/org/contrib/noter.el
brianmcgillion: When I originally created the patch Org-roam was behind m in emacs-bindings and as I noticed later r in evil-bindings. So I will address these.

But it brings me to a point that I raised in a private chat @hlissner in discord. There are many similarities between evil-bindings and emacs-bindings. The leader key is different but all the rest of the combos are the same. To save on having to make 2 versions of every binding, could we create a +common-global-bindings.el that is loaded by +evil-bindings and +emacs-bindings. 1 it will reduce the differences of C-c n m being roam in one set and SPC n r being roam in another, in addition to reducing the over all maintenance as I have noticed a few cases where either emacs has been updated or evil, depending on the committer.

modules/config/default/+emacs-bindings.el
```diff
@@ -128,7 +128,9 @@
           :desc "Bibliographic entries"        "b"
           (cond ((featurep! :completion ivy)   #'ivy-bibtex)
                 ((featurep! :completion helm)  #'helm-bibtex)))
-
+        (:when (featurep! :lang org +noter)
+          :desc "Org noter"                    "e" #'org-noter)
```
bdarcus, line 136: "Org-noter" or remove description?

modules/config/default/+evil-bindings.el
```diff
@@ -464,6 +464,9 @@
         :desc "Toggle org-clock"             "c" #'+org/toggle-clock
         :desc "Cancel org-clock"             "C" #'org-clock-cancel
         :desc "Open deft"                    "d" #'deft
+        (:when (featurep! :lang org +noter)
+          :desc "Org noter"                  "e" #'org-noter)
```
bdarcus, line 468: "Org-noter" or remove description?

modules/config/default/+emacs-bindings.el
brianmcgillion: not sure. "Org agenda" above, "Org store link" below.

bdarcus:
Ok; was just a question, based on looking at the noter repo.

If no comment from anyone else, just keep as is?

hlissner:
> "Org Noter"

This is fine for now. We can always changed it if we come up with a better one.

> +comming-global-bindings.el

I hear you, I'm still considering my options.

----

Anyhow, thanks for the PR!

PR #2885, opened 2020-04-11
Align org-roam emacs-bindings with evil-bindings 

brianmcgillion:


brianmcgillion:
@hlissner I just noticed that these key bindings are listed directly in default/+evil-bindings, so should I move the emacs one to +emacs-bindings also?

hlissner:
> should I move the emacs one to +emacs-bindings also?

Sounds like a good idea!

brianmcgillion:
Sorry, I misunderstood completely! will do!

hlissner:
No worries! Thanks for the quick response and the PR!

PR #2888, opened 2020-04-11
Add org-ref and org-roam-bibtex to :tools/biblio

brianmcgillion:
org-ref allows for easy citation management and provides numerous interfaces to online publishing and reference management tools.

org-roam-bibtex provides deep integration of bibtex-completion and {ivy,helm}-bibtex into the org-roam workflow.

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -85,7 +85,7 @@
 
        :tools
        ;;ansible
-       ;;biblio            ; Writes a PhD for you (citation needed)
+       ;;biblio            ; A collection of bibliographic reference tools
        ;;debugger          ; FIXME stepping through code, to help you add bugs
        ;;direnv
        ;;docker
--- a/modules/tools/biblio/README.org
+++ b/modules/tools/biblio/README.org
@@ -8,6 +8,7 @@
   - [[#maintainers][Maintainers]]
   - [[#module-flags][Module Flags]]
   - [[#plugins][Plugins]]
+  - [[#note-about-org-ref][Note about org-ref]]
 - [[#prerequisites][Prerequisites]]
   - [[#pdf-viewing][PDF viewing]]
   - [[#bibtex-completion][Bibtex completion]]
@@ -17,7 +18,11 @@
     - [[#processor-configuration][Processor configuration]]
     - [[#other-configuration-options][Other configuration options]]
   - [[#path-configuration][Path configuration]]
+    - [[#bibtex-completion-helm-bibtex-ivy-bibtex][bibtex-completion (helm-bibtex, ivy-bibtex)]]
+    - [[#org-ref][org-ref]]
   - [[#templates][Templates]]
+    - [[#bibtex-completion-1][Bibtex-completion]]
+    - [[#org-roam-bibtex-1][org-roam-bibtex]]
 - [[#troubleshooting][Troubleshooting]]
 
 * Description
@@ -41,6 +46,14 @@ This module provides no flags.
   + [[https://github.com/tmalsburg/helm-bibtex][helm-bibtex]]
 + ~:completion ivy~
   + [[https://github.com/tmalsburg/helm-bibtex][ivy-bibtex]]
++ ~:lang org~
+  + [[https://github.com/jkitchin/org-ref][org-ref]]
++ ~:lang org +roam~
+  + [[https://github.com/Zaeph/org-roam-bibtex][org-roam-bibtex]]
+** Note about org-ref
+Although org-ref supports both ivy and helm, certain functionality e.g. the ability to drag-n-drop a pdf into a bib file in order to have org-ref automatically generate a new bibtex entry will pull in helm, as it is hard coded to support that.
+
+For this reason many of the options are not enabled by default, if you wish to enable the other functions see: [[https://github.com/jkitchin/org-ref#some-other-useful-libraries-in-org-ref][org-ref additional libraries]]
 
 * Prerequisites
 There are no hard dependencies for this module.
@@ -56,8 +69,8 @@ For vertico, helm, or ivy bibtex completion you should enable =:completion verti
 
 * Features
 Both [[https://github.com/tmalsburg/helm-bibtex][helm-bibtex]] (includes helm-bibtex, ivy-bibtex and bibtex-completion code)
-and [[https://github.com/bdarcus/bibtex-actions][citar]] provide an extensive range of features so it is best to check their
-respective sites for a full list of features.
+[[org-ref]], and [[https://github.com/bdarcus/bibtex-actions][citar]] provide an extensive range of features so it is best to check
+their respective sites for a full list of features.
 
 On a high-level you can expect:
 + bibliography management
@@ -112,6 +125,9 @@ directory for your CSL styles:
 (setq org-cite-csl-styles-dir "~/Zotero/styles")
 #+END_SRC
 
+** org-roam-bibtex
+This package integrates org-ref and bibtex-completion into [[https://github.com/jethrokuan/org-roam][org-roam]] mode, which can be enabled via =lang org +roam=. For academics and those working extensively with documents which they need to keep detailed notes on for cross-referencing.
+
 ** Path configuration
 
 You must set the path variable for either =citar= (if using =vertico=
@@ -139,13 +155,66 @@ enhanced functionality:
        citar-notes-paths '("/path/to/your/notes/"))
 #+END_src
 
+
+It is of course possible to set the variables directly through their respective packages as shown below, in which case you can ignore the ~+biblio...~ versions. See [[*bibtex-completion (helm-bibtex, ivy-bibtex)][bibtex-completion]] and [[*org-ref][org-ref]] configurations below.
+
+#+BEGIN_src emacs-lisp
+(setq! +biblio-pdf-library-dir "/path/to/bibrary/"
+       +biblio-default-bibliography-files '("/path/to/bibliography.bib")
+       +biblio-notes-path "/path/to/notes/")
+#+END_src
+
+Unless ~+biblio-notes-path~ ends in a ~/~ it is assumed to be a single file and not a directory.
+
+Each of the variables overlays the corresponding variables in the following way
++ ~+biblio-default-bibliography-files~
+  - ~bibtex-completion-library-path~
+  - ~org-ref-pdf-directory~
++ ~+biblio-default-bibliography-files~
+  - ~reftex-default-bibliography~
+  - ~bibtex-completion-bibliography~
+  - ~org-ref-default-bibliography~
++ ~+biblio-notes-path~
+  - ~+bibibtex-completion-notes-path~
+  - ~org-ref-bibliography-notes~
+  - ~org-ref-notes-directory~
+
+*** bibtex-completion (helm-bibtex, ivy-bibtex)
+These packages share the same common backend ~bibtex-completion~
+#+BEGIN_src emacs-lisp
+(after! bibtex-completion
+  (setq! bibtex-completion-bibliography "/path/to/bib/"
+        bibtex-completion-library-path "/path/to/Papers/"
+        bibtex-completion-notes-path "/path/to/your/notes/"))
+#+END_src
+*** org-ref
+#+BEGIN_src emacs-lisp
+(after! org-ref
+  (setq! org-ref-bibliography-notes "/path/to/notes/"
+        org-ref-default-bibliography "/path/to/bib"
+        org-ref-pdf-directory "/path/to/papers/"
+        reftex-default-bibliography org-ref-default-bibliography))
+
 ** Templates
 
 This module provides reasonable default templates for the packages. However, if
 you wish to change them, refer to the respective packages' documentation for in-depth
 instructions.
 
-* Troubleshooting
-# Common issues and their solution, or places to look for help.
+*** Bibtex-completion
+#+BEGIN_src emacs-lisp
+(after! bibtex-completion
+  (setq!  bibtex-completion-notes-template-multiple-files
+         "${title} : (${=key=})\n Some more format options"))
+#+END_src
 
-Refer to the respective package repositories.
+*** org-roam-bibtex
+To change the default template you can change it according to [[https://github.com/Zaeph/org-roam-bibtex][org-roam-bibtex]]
+#+BEGIN_src emacs-lisp
+(after! org-roam-bibtex
+  (setq org-roam-bibtex-preformat-keywords '("some" "key" "words"))
+  (setq org-roam-bibtex-template '(("Some template"))))
+#+END_src
+
+* Troubleshooting
+# Common issues and their solution, or places to look for help
--- a/modules/tools/biblio/config.el
+++ b/modules/tools/biblio/config.el
@@ -1,6 +1,36 @@
 ;;; tools/biblio/config.el -*- lexical-binding: t; -*-
 
 ;;
+
+;; Internal function to set the various paths used in the
+;; reference packages.
+(defun +biblio-set-paths-fn (&optional symbol value)
+  (when symbol
+    (set-default symbol value))
+  (when value
+    (cond ((eq symbol '+biblio-pdf-library-dir)
+           (when (featurep! :lang org)
+             (setq org-ref-pdf-directory value))
+           (setq bibtex-completion-library-path value))
+          ((eq symbol '+biblio-default-bibliography-files)
+           (when (featurep! :lang org)
+             (setq reftex-default-bibliography value
+                   org-ref-default-bibliography value))
+           (setq bibtex-completion-bibliography value)))))
+
+(defcustom +biblio-pdf-library-dir nil
+  "Directory where pdf files are stored. Must end with a slash."
+  :type 'string
+  :set #'+biblio-set-paths-fn)
+
+(defcustom +biblio-default-bibliography-files nil
+  "A list of default bibtex files to use."
+  :type '(repeat :tag "List of bibtex files" file)
+  :set #'+biblio-set-paths-fn)
+
+(defvar +biblio-notes-path nil)
+
+
 ;;; `org-cite'
 
 (use-package! oc
@@ -41,15 +71,102 @@
   :when (or (featurep! :completion ivy)
             (featurep! :completion helm))
   :defer t
-  :config
+  :preface
+  ;; Allow the user to set a template of their own via (setq). if the user does
+  ;; not set one fall back to the +biblio variants which have a reasonable
+  ;; fallback.
+  (defvar bibtex-completion-notes-template-multiple-files nil)
+   :config
+
+  (when (featurep! :completion ivy)
+    (add-to-list 'ivy-re-builders-alist '(ivy-bibtex . ivy--regex-plus)))
+
   (setq bibtex-completion-additional-search-fields '(keywords)
-        ;; Tell bibtex-completion to look at the File field of the bibtex to
-        ;; figure out which pdf to open:
-        bibtex-completion-pdf-field "file"))
+        ;; This tell bibtex-completion to look at the File field of the bibtex
+        ;; to figure out which pdf to open
+        bibtex-completion-pdf-field "file")
+  ;; orb will define handlers for note taking so not needed to use the
+  ;; ones set for bibtex-completion
+  (unless (featurep! :lang org +roam2)
 
+    (setq bibtex-completion-notes-path +biblio-notes-path)
+    (unless bibtex-completion-notes-template-multiple-files
+      (setq bibtex-completion-notes-template-multiple-files
+            "${title} : (${=key=})
 
-(use-package! ivy-bibtex
-  :when (featurep! :completion ivy)
-  :defer t
+- tags ::
+- keywords :: ${keywords}
+
+\n* ${title}\n  :PROPERTIES:\n  :Custom_ID: ${=key=}\n  :URL: ${url}\n  :AUTHOR: ${author-or-editor}\n  :NOTER_DOCUMENT: /${file}\n  :NOTER_PAGE: \n  :END:\n\n"))))
+
+;; TODO which set of keys that should be bound for commonly used functions
+;; see https://github.com/jkitchin/org-ref/blob/master/org-ref-core.el#L3998
+(use-package! org-ref
+  :when (featurep! :lang org +ref)
+  :after org
+  :preface
+  ;; This need to be set before the package is loaded, because org-ref will
+  ;; automatically `require' an associated package during its loading.
+  (setq org-ref-completion-library (cond ((featurep! :completion ivy)  #'org-ref-ivy-cite)
+                                         ((featurep! :completion helm) #'org-ref-helm-bibtex)
+                                         (t                            #'org-ref-reftex)))
   :config
-  (add-to-list 'ivy-re-builders-alist '(ivy-bibtex . ivy--regex-plus)))
+;;  (add-to-list 'ivy-re-builders-alist '(ivy-bibtex . ivy--regex-plus))
+
+  ;; Although the name is helm-bibtex, it is actually a bibtex-completion function
+  ;; it is the legacy naming of the project helm-bibtex that causes confusion.
+  (setq org-ref-open-pdf-function 'org-ref-get-pdf-filename-helm-bibtex)
+  ;; orb will define handlers for note taking so not needed to use the
+  ;; ones set for bibtex-completion
+  (unless (featurep! :lang org +roam2)
+    ;; determine how org ref should handle the users notes path (dir, or file)
+    (if (directory-name-p +biblio-notes-path)
+        (setq org-ref-notes-directory +biblio-notes-path)
+      (setq org-ref-bibliography-notes +biblio-notes-path))
+    ;; Allow org-ref to use the same template mechanism as {helm,ivy}-bibtex for
+    ;; multiple files if the user has chosen to spread their notes.
+    (setq org-ref-notes-function (if (and org-ref-notes-directory (directory-name-p org-ref-notes-directory))
+                                     #'org-ref-notes-function-many-files
+                                   #'org-ref-notes-function-one-file))))
+
+
+(use-package! org-roam-bibtex
+  :when (featurep! :lang org +roam2)
+  :after org-roam
+  :preface
+  ;; if the user has not set a template mechanism set a reasonable one of them
+  ;; The package already tests for nil itself so we define a dummy tester
+  (defvar orb-preformat-keywords
+    '("title" "url" "file" "author-or-editor" "keywords" "citekey" "pdf"))
+  ;;:hook (org-roam-mode . org-roam-bibtex-mode)
+  :custom
+  (orb-note-actions-interface (cond ((featurep! :completion ivy)  'ivy)
+                                    ((featurep! :completion helm) 'helm)
+                                    ((t                           'default))))
+  :config
+  (setq orb-insert-interface (cond ((featurep! :completion ivy)  'ivy-bibtex)
+                                   ((featurep! :completion helm) 'helm-bibtex)
+                                   ((t                           'generic))))
+  (setq orb-process-file-keyword t
+        orb-file-field-extensions '("pdf"))
+
+  (add-to-list 'org-roam-capture-templates
+               '("b" "Bibliography note" plain
+                 "%?
+- keywords :: %^{keywords}
+- related ::
+
+* %^{title}
+:PROPERTIES:
+:Custom_ID: %^{citekey}
+:URL: %^{url}
+:AUTHOR: %^{author-or-editor}
+:NOTER_DOCUMENT: %^{file}
+:NOTER_PAGE:
+:END:\n\n"
+                 :if-new (file+head "${citekey}.org" ":PROPERTIES:
+:ROAM_REFS: cite:${citekey}
+:END:
+#+TITLE: ${title}\n")
+                 :unnarrowed t))
+  (require 'org-ref))
--- a/modules/tools/biblio/packages.el
+++ b/modules/tools/biblio/packages.el
@@ -9,5 +9,8 @@
   (package! helm-bibtex :pin "aa775340ba691d2322948bfdc6a88158568a1399"))
 (when (featurep! :completion vertico)
   (package! citar :pin "51b30f2e4091a41243ae62cfbac53e7a579f3168"))
-
 (package! citeproc :pin "538fed794c29acf81efee8a2674268bd3d7cc471")
+(when (featurep! :lang org +ref)
+  (package! org-ref :pin "429733150548a847966685680bca0a20ec3b1ad9"))
+(when (featurep! :lang org +roam2)
+  (package! org-roam-bibtex :pin "cf811abf273ad28d32caad3a93318f92da034556"))


bdarcus:
While you wait on a review, two things:

1. @patrl asked on #2861 what happens if one just have either the `ivy` or `helm` feature selected. I suggested we include in the README that the module depends on one.
2. depending on time, we should add the new [org-roam-bibtex](https://github.com/Zaeph/org-roam-bibtex) extension, which tightly integrates org-roam and ivy/helm-bibtex and org-ref; would be dependent on the `+roam` feature

bdarcus:
Also, just saw [this](https://github.com/melpa/melpa/pull/6840); bibtex-completion is being split from ivy/helm-bibtex. Does it impact us?

Guessing not, but just thought I'd ask.

brianmcgillion:
> Also, just saw [this](https://github.com/melpa/melpa/pull/6840); bibtex-completion is being split from ivy/helm-bibtex. Does it impact us?
> 
> Guessing not, but just thought I'd ask.es it did as doom/upgrade has been failing for the last hours. I am converted to being a :pin advocate

brianmcgillion:
it did as doom/upgrade has been failing for the last hours. I am converted to being a :pin advocate

brianmcgillion:
> While you wait on a review, two things:
> 
> 1. @patrl asked on #2861 what happens if one has neither the `ivy` or `helm` feature selected. I suggested we include in the README that the module depends on one.
> 2. Depending on your time (though it should be easy; doesn't appear to require much or any config), we should add the new [org-roam-bibtex](https://github.com/Zaeph/org-roam-bibtex) extension, which tightly integrates org-roam and ivy/helm-bibtex and org-ref, and has been submitted to MELPA. This would need to be dependent on the org `+roam` feature.

Yes I will do this, once the package in in melpa

modules/tools/biblio/config.el
```diff
@@ -5,3 +5,44 @@
...
+
+;; TODO bibtex-complete is being split into its own package
+;; https://github.com/melpa/melpa/pull/6840
+;; so we can remove the dependency on helm/ivy here and just depend bibtex-complete
+(when (or (fboundp 'ivy-bibtex) (fboundp 'helm-bibtex))
+  (after! bibtex-completion
+    ;; TODO do we need to test if the user set these first
+    (setq bibtex-completion-additional-search-fields '(keywords)
+          bibtex-completion-pdf-field "file")
+    (setq bibtex-completion-bibliography (list (concat org-directory "library.bib"))
```
bdarcus, line 18: By default, are these variables here set to nil, so that we need to set something reasonable?

brianmcgillion: yes, 

https://github.com/tmalsburg/helm-bibtex/blob/master/bibtex-completion.el#L262
- not sure why keywords is not a default.

https://github.com/tmalsburg/helm-bibtex/blob/master/bibtex-completion.el#L309

modules/tools/biblio/packages.el
```diff
@@ -1,7 +1,11 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; tools/biblio/packages.el
 
+;;(package! bibtex-completion)
```
brianmcgillion, line 4: Pulled in in this commit set (https://github.com/hlissner/doom-emacs/pull/2907)

bdarcus:
This looks good. It adds documentation, and significant new functionality, to the new biblio package.

I've tested it, and it works as I expect.

modules/tools/biblio/config.el
```diff
@@ -2,13 +2,72 @@
...
   :when (featurep! :completion ivy)
   :defer t
   :config
   (add-to-list 'ivy-re-builders-alist '(ivy-bibtex . ivy--regex-plus)))
+
+
+;; TODO which set of keys that should be bound for commonly used functions
+;; see https://github.com/jkitchin/org-ref/blob/master/org-ref-core.el#L3998
+(use-package! org-ref
+  :when (featurep! :lang org)
```
Wetlize, line 33: Is it necessary to future gate `org-ref`? Despite the name it also provides features outside of the org-mode.

modules/tools/biblio/packages.el
```diff
@@ -6,3 +6,8 @@
   (package! ivy-bibtex :pin "b14b628261"))
 (when (featurep! :completion helm)
   (package! helm-bibtex :pin "b14b628261"))
+(when (featurep! :lang org)
```
Wetlize, line 9: Same thing, this feature gate is probably redundant.

modules/tools/biblio/config.el
```diff
@@ -2,13 +2,72 @@
...
+  :after (org bibtex-completion)
+  :config
+  ;; Although the name is helm-bibtex, it is actually a bibtex-completion function
+  ;; it is the legacy naming of the project helm-bibtex that causes confusion.
+  (setq org-ref-open-pdf-function 'org-ref-get-pdf-filename-helm-bibtex)
+  ;; org-roam-bibtex will define handlers for note taking so not needed to use the
+  ;; ones set for bibtex.completion
+  (unless (featurep! :lang org +roam)
+    ;; Allow org-ref to use the same template mechanism as {helm,ivy}-bibtex
+    (setq org-ref-notes-(function )
```
Wetlize, line 43: Should be `org-ref-notes-function`.

modules/tools/biblio/config.el
```diff
@@ -2,13 +2,72 @@
...
+          (lambda (thekey)
+            (let ((bibtex-completion-bibliography (org-ref-find-bibliography)))
+              (bibtex-completion-edit-notes
+               (list (car (org-ref-get-bibtex-key-and-file thekey)))))))))
```
Wetlize, line 47: This is the same as `org-ref-notes-function-many-files` function, any reason to extract lambda?
Furthermore, we can do the next thing:
```lisp
(setq 
  ...
  org-ref-notes-function (if (directory-name-p org-ref-notes-directory)
                              #'org-ref-notes-function-many-files
                            #'org-ref-notes-function-one-file))

File: modules/tools/biblio/config.el [hunk 7]
```diff
@@ -2,13 +2,72 @@
...
+￼- keywords :: ${keywords}
+￼
+￼\n* ${title}\n  :PROPERTIES:\n  :Custom_ID: ${=key=}\n  :URL: ${url}\n  :AUTHOR: ${author-or-editor}\n  :NOTER_DOCUMENT: /${file}\n  :NOTER_PAGE: \n  :END:\n\n")))
 
 
 (use-package! ivy-bibtex
   :when (featurep! :completion ivy)
   :defer t
   :config
   (add-to-list 'ivy-re-builders-alist '(ivy-bibtex . ivy--regex-plus)))
```
Wetlize, line 27: We can refactor this into:
```lisp
    (add-to-list 'ivy-re-builders-alist '(ivy-bibtex . ivy--regex-plus)))
```
under the `:config` section of `bibtex-completion` package.

Wetlize:
I've noticed that a a lot of bibliographical packages want to specify the same patterns as user options:
- pdf directory path
- path to default bibliographical files
- path to bibliographical notes

So I propose to expose the next user options on top of config.el:
```lisp
(defvar +biblio-pdf-library-dir "~/Documents/Library/"
  "Directory where pdf files are stored. Must end with a slash.")

(defvar +biblio-default-bibliography-files '("~/Documents/Biblio/biblio.bib")
  "A list of default bibtex files to use.")

(defvar +biblio-notes-path "~/Documents/Notes/"
  "The place where you will store your notes for bibliography files.

This can be either a single file or directory of files.
In case of directory the path must end with a slash.")
```
They can affect next variables:
- `+biblio-default-bibliography-files`:
  - `bibtex-completion-library-path`
  - `org-ref-pdf-directory`
- `+biblio-default-bibliography-files`:
  - `reftex-default-bibliography`
  - `bibtex-completion-bibliography`
  - `org-ref-default-bibliography`
- `+biblio-notes-path`:
  - `bibtex-completion-notes-path`
  - `org-ref-bibliography-notes`
  - `org-ref-notes-directory`

This would simplify and unify the whole interface better.

modules/tools/biblio/config.el
bdarcus: I think we were originally assuming one wouldn't use org-ref without having the org feature selected, but you raise a good point. No objections from me to removing it.

bdarcus:
> I've noticed that a a lot of bibliographical packages want to specify the same patterns as user options:

Yes, I've noticed (and grumbled about) that too.

...

> So I propose to expose the next user options on top of config.el:
...
> This would simplify and unify the whole interface better for the end user.

+1

Wetlize:
I also remember that back in time I had difficulties understand how all those packages suppose to work together. We can briefly describe them in the documentation or just link to (or include as a quote block) [this](https://old.reddit.com/r/emacs/comments/cxu0qg/emacs_org_and_bibtex_as_alternative_to_zotero_and/eyqe4vq/) reddit thread.

modules/tools/biblio/config.el
brianmcgillion: Many of the modules under org-ref (require 'org) or org-ref-core which requires org so it will pull in org. hence you should have it, although I agree that it is not org olny.

modules/tools/biblio/config.el
brianmcgillion: TAB, seems to have a nasty habit of turning (keywords) into code, so TAB at the beginning of the line to align or else ...

modules/tools/biblio/config.el
brianmcgillion: https://github.com/jkitchin/org-ref/blob/master/org-ref.org#setting-up-notes-to-work-with-multiple-notesorg-files

modules/tools/biblio/packages.el
brianmcgillion: same comment as above

brianmcgillion:
> This would simplify and unify the whole interface better for the end user.

I totally agree. I had thought of something similar when doing it, but then discounted the idea because many folks are coming from own configs, so having different variants helm-bibtex, bibtex complete, org-ref .... So I was figuring that the permutation checking might be a mess. But having said that they can always use the standard (after! ...) if they wish and ignore the +biblio variables and just the packages directly.

I will send an update in a few.

brianmcgillion:
> I also remember that back in time I had difficulties to understand how all those packages suppose to work together. We can very briefly describe their roles in the documentation.
> 
> As a reference, there's a [reddit thread](https://reddit.com/r/emacs/comments/cxu0qg/emacs_org_and_bibtex_as_alternative_to_zotero_and/eyqe4vq/) that already does so, but not briefly.

Thanks for the link. That will surely help!

modules/tools/biblio/config.el
Wetlize: The case of outdated documentation :(
https://github.com/jkitchin/org-ref/commit/dc3d259a876e53b939bb885d915ff013e122acc0

modules/tools/biblio/config.el
```diff
@@ -2,13 +2,72 @@
...
+
+
+;; TODO which set of keys that should be bound for commonly used functions
+;; see https://github.com/jkitchin/org-ref/blob/master/org-ref-core.el#L3998
+(use-package! org-ref
```
Wetlize, line 32: We should also preset `org-ref-completion-library` under `:init` section:
```lisp
;; This need to be set before the package is loaded, because org-ref will
;; automatically `require' an associated package during its loading.
(setq org-ref-completion-library (cond ((featurep! :completion ivy)  #'org-ref-ivy-cite)
                                       ((featurep! :completion helm) #'org-ref-helm-bibtex)
                                       (t                            #'org-ref-reftex)))
```

brianmcgillion:
@Wetlize, you raise another issue that @bdarcus and I debated a bit. Should org-noter be under the biblio module or as +noter under org/contrib? Technically you do not need the biblo packages to take notes, however, who takes notes that does not in some way use the biblio packages?

Wetlize:
@brianmcgillion I think we should keep them separated. One can think about annotations and bibliography as 2 different things. The bibliography module can just reference `:lang org +noter` somehow in the documentation, as a potential object of interest.

brianmcgillion:
would crossreferencing the setting of the notes path be too much confusion? so from noter, 

```
(if (fboundp +biblio-notes-path)
    (setq org-noter-notes-search-path  +biblio-notes-path))
```
or something to that effect?

modules/tools/biblio/config.el
brianmcgillion: thanks again.

Wetlize:
> would cross-referencing the setting of the notes path be too much confusion?

I think we should avoid doing so and keep it dumb as for now (i.e. just default `org-noter-notes-search-path` to `org-directory`). In future, if users will find this confusing and not intuitive we can change how org-noter (and maybe other org related bibliography packages) will integrate with the rest of Doom.

modules/tools/biblio/config.el
```diff
@@ -1,14 +1,114 @@
...
-  :when (featurep! :completion ivy)
-  :defer t
+\n* ${title}\n  :PROPERTIES:\n  :Custom_ID: ${=key=}\n  :URL: ${url}\n  :AUTHOR: ${author-or-editor}\n  :NOTER_DOCUMENT: /${file}\n  :NOTER_PAGE: \n  :END:\n\n")))
+
+
+;; TODO which set of keys that should be bound for commonly used functions
+;; see https://github.com/jkitchin/org-ref/blob/master/org-ref-core.el#L3998
+(use-package! org-ref
+  :when (featurep! :lang org)
+  :after (org bibtex-completion)
```
Wetlize, line 54: Can we just `:defet t` this? I cannot wrap my head why would we need to wait for all these packages to apply configuration for this package.

brianmcgillion:
rebase and bump the biblio packages

modules/tools/biblio/config.el
```diff
@@ -1,14 +1,119 @@
...
+  :init
```
Wetlize, line 76:
```suggestion
  :preface
```
Whoops, when `:after` is present `:init` runs after these all packages are loaded. This needs to be set earlier, otherwise it will fail to load a proper for user completion library, defaulting him to helm.

Wetlize, line 76:
```suggestion
  :preface
```
We need to `:preface` this for a better robustness. Something can load `org-ref` before this block will be evaluated. If this will happen, `org-ref-completion-library` will be set only after the package is loaded, failing `require` a needed for user completion engine and defaulting him to helm.

This just actually happened in my config with this module and took me some time to figure out.

brianmcgillion: thank you. Change pushed

brianmcgillion:
rebased, and updated the time stamp

bdarcus:
Any chance we can get this merged soon @hlissner? Anything we're missing?

brianmcgillion:
Rebased and bump the pins

hlissner:
Sorry for the delay. As I'm very new to bibtex, org-roam and org-ref, reviewing this is taking me a while; digging through these packages. There are few pain points I'm looking for alternatives around:

1. The three `defcustom`s. I'd _like_ not to introduce new variables, and am investigating a way around these; especially because we do not use `defcustom` anywhere else. I want to avoid them.

2. We have a lot of `(unless X (setq X DEFAULTS))`, and I'm going through each of their upstream defaults to see if these checks are warranted (or if we could simply define them _in_ their respective `defvar`s.

3. I don't fully understand the relationship between org-ref and bibtex-completion, and wonder why `:after (org bibtex-completion)` is necessary. This complicates the customization path for the user, because neither `(after! org ...)` nor `(after! bibtex-completion ...)` will be enough to configure the package (and due to `use-package`'s [weird :after semantics](https://github.com/jwiegley/use-package/issues/829), any config in `(after! org-ref ...)` will be overruled by what's in this module's `:config` block for it.

I'll keep you posted as I truck along.

bdarcus:
Thanks for the update @hlissner. 

Given how common it is to integrate these tools (and related, how commonly people get confused how they're supposed to fit together), perhaps your investigations can lead to suggestions upstream.

bdarcus:
@brianmcgillion - you might bump `org-roam-bibtex` again, as they just issued a new release, with new features.

brianmcgillion:
Thank you @hlissner for the thorough review.

> * The three `defcustom`s. If possible, I'd _like_ to avoid introducing new variables, and am investigating a way around these; especially because we do not use `defcustom` anywhere else.
> * We have a lot of `(unless X (setq X DEFAULTS))`, and I'm going through each of their upstream defaults to see if these checks are warranted (or if we could simply define them _in_ their respective `defvar`s).

The first drafts used the (unless ... set...) form as seen in other modules. It was suggested by @Wetlize https://github.com/hlissner/doom-emacs/pull/2888#issuecomment-616712460 that it would greatly cleanup the interface if we reduce the multiple packages "file path" variables down to just 3 which would then set them in the respective packages.

The problem stems from the packages providing some level of functionality for bibliographic management but there is not one overarching package tying them together, e.g. reftex-default-bibliography, org-ref-default-bibliography,  bibtex-completion-bibliography all point to the same place and all have to be set, (to the confusion of most users). In the documentation, we indicate how the user can set the variables directly if they so wish rather than using the defcustom.

brianmcgillion:
> 3\. I don't fully understand the relationship between org-ref and bibtex-completion, and wonder why `:after (org bibtex-completion)` is necessary. This complicates the customization path for the user, because neither `(after! org ...)` nor `(after! bibtex-completion ...)` will be enough to configure the package (and due to `use-package`'s [:after semantics](https://github.com/jwiegley/use-package/issues/829), any config in a user's `(after! org-ref ...)` will be overruled).

Will be fixed in the next push. I believe that `after! org` will be sufficient, as the bibtex-completion will be loaded before this.

das-s:
If I might chime in here as someone who is already using this merge request:
IMHO the module is not documented enough for people who are new to reference managment or consider switching from more mainstream programs like zotero or citavi. For example, the linked reddit post in the current readme only considers papers but books or online sources are not touched upon. Especially org-ref and its corresponding packages are neither documented very well nor easily discoverable for the new user. Maybe implementing a dedicated keybinding layer would be sufficient to make discoverability easier. This might not be up to the task for the current PR but should be considered. 

I'd help with documenting if my skills suffice. 
Some ideas about approaching this documentation:
* What to do if drag-n-drop fails? 
* What if there is no DOI?
* How to add entries manually

bdarcus:
> If I might chime in here as someone who is already using this merge request:
> IMHO the module is not documented enough for people who are new to reference managment or consider switching from more mainstream programs like zotero or citavi. For example, the linked reddit post in the current readme only considers papers but books or online sources are not touched upon. Especially org-ref and its corresponding packages are neither documented very well nor easily discoverable for the new user. 

I agree with all of the above.

It's related to the problem @brianmcgillion noted just above that the functionality is spread across multiple packages, which we are trying to integrate as nicely as we can.

This is pretty common in the emacs world, and while it has its annoyances, there is value to it as well.

The practical question for this module documentation is what we should be responsible for documenting, and what should we defer to the packages we're using?

It's not really our fault if upstream package documentation is lacking. OTOH, we should certainly a) report gaps upstream, and b) look for places where we can provide a bridge.

> Maybe implementing a dedicated keybinding layer would be sufficient to make discoverability easier. 

You mean "dedicated" to Doom and this module, that supersede those in the packages we're using?

Can you give an example where this is currently a problem that we don't do this?

> This might not be up to the task for the current PR but should be considered for the future.
> 
> I'd help with documenting if my skills suffice.
> Some ideas about approaching this documentation:
> 
> * What to do if drag-n-drop fails?
> * What if there is no DOI?
> * How to add entries manually
> * Whats a general workflow?
> * How to add notes to file?
> * How can org-noter be incorporated?
> * How to create an org-roam entry with cite links?

All are good and reasonable issues.

But per my points above, which of those are realistically our responsibility to document, and which should be reported upstream?

das-s:
> You mean "dedicated" to Doom and this module, that supersede those in the packages we're using?

Sorry, I worded this incorrectly. I meant a dedicated default keybinding prefix when using `biblio`, for example `SPC-r`(could also only be active on selected mode-maps like `org` or `bibtex`). This would then pack some useful commands behind it for getting started. In case someone knows their way around it can be overwritten by custom bindings. Admittedly, this approach might be beyond the philosophy of doom-emacs.

> Can you give an example where this is currently a problem that we don't do this?

Does this get answered with my previous paragraph? If not, could you please rephrase this? Do you mean doom-emacs in general or specific changes in the current PR? 

> But per my points above, which of those are realistically our responsibility to document, and which should be reported upstream?

I admit that this a weak point and those questions might be more fitting on a blog or in `doom-emacs/docs/` documentation. Most of it should be documented upstream. Still, doom provides a features section for some [modules](https://github.com/hlissner/doom-emacs/blob/develop/modules/tools/lookup/README.org). Maybe one could think of adding some helpful tips there, but this is most likely outside the range of this PR.

bdarcus:
> I meant a dedicated default keybinding prefix when using `biblio`, for example `SPC-r`(could also only be active on selected mode-maps like `org` or `bibtex`). This would then pack some useful commands behind it for getting started. In case someone knows their way around it can be overwritten by custom bindings. Admittedly, this approach might be beyond the philosophy of doom-emacs.

It probably would, but @hlissner would be arbiter of that question.

You can always create your own keybindings to do this too.

> Maybe one could think of adding some helpful tips there, but this is most likely outside the range of this PR.

Sure. Hopefully we can get this PR merged, and maybe that can be a latter step.

Feel free to collect of that info for possible inclusion.

brianmcgillion:
rebased and bumped the pins

LauraBMo:
Hey, thanks a lot for extending the module tools: biblio. As LaTeX user it will be really handy.
Nevertheless, I really think that `org-ref` should be offered as an option with a flag.
It is a great package, but I think it does not integrate well with Doom philosophy.
- It slowdowns the init time a lot.

bdarcus:
The assumption currently is that if someone is using org, then we can
assume you want org-ref.

So you're saying that's not the case with you?

FWIW, I have no objection to adding that flag; just clarifying.

LauraBMo:
Yes, I am using `ox-bibtex` [https://github.com/yyr/org-mode/blob/master/contrib/lisp/ox-bibtex.el](url)

brianmcgillion:
it is deferred so do not load on init, only when referenced. Also it uses ivy-bibtex if you have ivy enabled and not helm (https://github.com/hlissner/doom-emacs/pull/2888/commits/5f0edbca435e71dfbb3484ed7f49a1717c36b74a#diff-9535f000ae2e09e5c9700b6136569cc6R79), the overall package used to be  called helm-bibtex, but that has now been split into 3 packages (https://github.com/tmalsburg/helm-bibtex/commit/473761dffe956b5aa7d60d399d037ec7d80c69b8). ivy-bibtex, helm-bibtex and bibtex-completion. bibtex-completion is used by helm/ivy as the backend, but used to be included by requiring helm-bibtex. I can check if the reference has been changed in orf-ref.

AFAIK the "ox-" series of packages are used for exporting form org to some other format (https://orgmode.org/worg/exporters/ox-overview.html).

As regards to speed, I see no additional delays with init. I have these packages enabled and still get between 0.5 and 0.6 second times on init.

For pure latex enabled use cases reftex is included in the latex module under the +ref flag.

As for the doom philosophy, I can not really comment there. But what I can see is that it seems to about giving a reasonable set of defaults for the average user of a package, with the flexibility to modify those to your needs. So if org-ref is not the common citation mechanism used by org users I have no issue to flag it out. If, however, it is one or 2 folks, I would probably suggest to add the following to the documentation `(package! org-ref :disable)` for people to opt-out.

LauraBMo:
`(package! org-ref :disable)` seems a good option.
Just for clarifying, it is `org-ref` who loads `helm`.
The line  `(setq org-ref-completion-library #'org-ref-ivy-cite)` does not solve the problem.
I am referring to [for now it is stuck in helm](https://github.com/jkitchin/org-ref/issues/552#issuecomment-537116098) (actually, that is the main reason I do not use `org-ref`).

bdarcus:
Org-roam just added tagging support, which can be used to type bibliographic notes, including (optionally) by placing them in a subdir.

https://org-roam.github.io/org-roam/manual/Tags.html#Tags

bdarcus:
> Sorry for the delay. As I'm very new to bibtex, org-roam and org-ref, reviewing this is taking me a while; digging through these packages. There are few pain points I'm trying to deal with:
> 
> 1. The three `defcustom`s. If possible, I'd _like_ to avoid introducing new variables, and am investigating a way around these; especially because we do not use `defcustom` anywhere else.
> 2. We have a lot of `(unless X (setq X DEFAULTS))`, and I'm going through each of their upstream defaults to see if these checks are warranted (or if we could simply define them _in_ their respective `defvar`s).
> 3. I don't fully understand the relationship between org-ref and bibtex-completion, and wonder why `:after (org bibtex-completion)` is necessary. This complicates the customization path for the user, because neither `(after! org ...)` nor `(after! bibtex-completion ...)` will be enough to configure the package (and due to `use-package`'s [:after semantics](https://github.com/jwiegley/use-package/issues/829), any config in a user's `(after! org-ref ...)` will be overruled).
> 
> I'll keep you posted as I truck along.

Can we assess where we're at?

My understanding is that @brianmcgillion addressed 3 in a subsequent commit, and provided context for 1 and 2 in [this comment](https://github.com/hlissner/doom-emacs/pull/2888#issuecomment-623129214). 

So the question of those variables is the key thing we're stuck on ATM?

jlmaccal:
I ran into a problem with this module and `(format +onsave)`.

Specifically, when `(format +onsave)` is set, dragging a pdf file onto an open bib file does not work correctly. Org-ref correctly looks up the reference information based on the doi and inserts it into the bib file. The file is then saved and autoformatted. This leaves the point in the wrong place for the remainder of org-ref's drag and drop handling, which leads to strange filenames, like `nil.pdf` and other weird problems. This is fixed by disabling `(format +onsave)`.

I would suggest that either autoformating of bib files should be disabled if both the biblio module is loaded and `+onsave` is set, or that appropriate advice is added around org-ref functions to temporarily disable autosave.

bdarcus:
Just a gentle bump, @hlissner, while I'm thinking about :-)

https://github.com/hlissner/doom-emacs/pull/2888#issuecomment-629780792

JonathanReeve:
I've been testing this branch, and I'm no expert, so take this as you will, but I find that enabling this module causes everything org-related in Doom not to work, on my setup, at least. It complains: `(wrong-type-argument stringp nil)`. I did `toggle-debug-on-error` and I got this traceback (truncated, but called from require(org) and so on. 

```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  directory-name-p(nil)
  (if (directory-name-p org-ref-notes-directory) (function org-ref-notes-function-many-files) (function org-ref-notes-function-one-file))
  (setq org-ref-notes-function (if (directory-name-p org-ref-notes-directory) (function org-ref-notes-function-many-files) (function org-ref-notes-function-one-file)))
  (if nil nil (setq org-ref-notes-function (if (directory-name-p org-ref-notes-directory) (function org-ref-notes-function-many-files) (function org-ref-notes-function-one-file))))
  (progn (require (quote org-ref) nil nil) (setq org-ref-open-pdf-function (quote org-ref-get-pdf-filename-helm-bibtex)) (if nil nil (setq org-ref-notes-function (if (directory-name-p org-ref-notes-directory) (function org-ref-notes-function-many-files) (function org-ref-notes-function-one-file)))) t)
  (closure (t) nil (progn (require (quote org-ref) nil nil) (setq org-ref-open-pdf-function (quote org-ref-get-pdf-filename-helm-bibtex)) (if nil nil (setq org-ref-notes-function (if (directory-name-p org-ref-notes-directory) (function org-ref-notes-function-many-files) (function org-ref-notes-function-one-file)))) t))()
  eval-after-load-helper("/home/jon/.emacs.doom.d/.local/straight/build/org-mode/org.elc")
  run-hook-with-args(eval-after-load-helper "/home/jon/.emacs.doom.d/.local/straight/build/org-mode/org.elc")
  do-after-load-evaluation("/home/jon/.emacs.doom.d/.local/straight/build/org-mode/org.elc")
```

What I'm guessing is happening is that [this line](https://github.com/brianmcgillion/doom-emacs/blob/bibtex/modules/tools/biblio/config.el#L91) calls `(if (directory-name-p org-ref-notes-directory))`, which on my system is `nil`, since I don't have one of these directories—I just use a single org-ref notes file. 

My setup is pretty much just like the one in [your config.el](https://github.com/brianmcgillion/doomd/blob/master/config.org), with these lines: 

```
(setq! +biblio-pdf-library-dir "~/Dropbox/Papers/"
       +biblio-default-bibliography-files '("~/Dropbox/Papers/library.bib")
       +biblio-notes-path "~/Dropbox/Org/Projects/books.org")
```

So I don't know what's going wrong. 

But I do know that I can work around this by adding the line: 

```
(setq! org-ref-notes-directory "")
```

right above my config lines from earlier. If my elisp-fu were better, I'd try to fix this, but alas, that's all I got. Hope this helps.

theianjones:
I just tested this out and was able to get everything working. I did end up with `Warning (:warning): Bibtex-completion couldn't find entry with key "EvilreadthedocsioenlatestPdf".` but I think this may be a user warning?

modules/tools/biblio/config.el
```diff
@@ -1,14 +1,119 @@
...
+           (when (featurep! :lang org)
+             (setq reftex-default-bibliography value
+                   org-ref-default-bibliography value))
+           (setq bibtex-completion-bibliography value))
+          ((eq symbol '+biblio-notes-path)
+           (when (featurep! :lang org)
+             (if (directory-name-p value)
+                 (setq org-ref-notes-directory value)
+               (setq org-ref-bibliography-notes value)))
+           (setq bibtex-completion-notes-path value)))))
```
Emiller88, line 23: This is really clever, but it might be more readable and easy to understand if you followed how the org and capture declares the variables 

https://github.com/hlissner/doom-emacs/blob/0f3bd571011f0f13a77b2f22e2c25dd5bfe03c2a/modules/lang/org/config.el#L24-L61

https://github.com/hlissner/doom-emacs/blob/0f3bd571011f0f13a77b2f22e2c25dd5bfe03c2a/modules/lang/org/autoload/org-capture.el#L73-#160

And just set them like the modules readme covers, making it easier for users to hack on rather than follow the logic of the function. For example
```lisp
(after! bibtex-completion
  (setq! bibtex-completion-bibliography +biblio-default-bibliography-files
        bibtex-completion-library-path +biblio-pdf-library-dir
        bibtex-completion-notes-path +biblio-notes-path))
```

Emiller88: I think this module should also consider setting some "opinionated" defaults of where files should be like other doom modules, see the org-mode stuff again.

Maybe 
```lisp
(setq! +biblio-pdf-library-dir (concat org-directory "pdfs")
       +biblio-default-bibliography-files '((concat org-directory "bib"))
       +biblio-notes-path (concat org-directory "notes"))
```
My example makes it a hard dependency on org, but there could be a `+bib-directory` variable created. This would make fewer changes for the end user if anything changes in the future.

bdarcus: Unfortunately, this PR seems stalled ATM.

It seems that @brianmcgillion must be busy with other things.

And while I updated [my own branch](https://github.com/bdarcus/doom-emacs/tree/bd-biblio) with some of the suggestions above, I'm an elisp novice, so not always super clear how to proceed on some of the details.

Finally, we've not had any feedback from @hlissner since end-of-April, so I'm not sure what, if anything, remains to be done. 

I do know at least some of his concerns have been addressed by @brianmcgillion, and by the suggestions of @myshevchuk and upstream changes in `org-roam`.

modules/tools/biblio/config.el
```diff
@@ -1,14 +1,119 @@
...
+  ;; The package already tests for nil itself so we define a dummy tester
+  (defvar org-roam-bibtex-preformat-keywords nil)
+  (defvar org-roam-bibtex-templates nil)
+  :hook (org-roam-mode . org-roam-bibtex-mode)
+  :config
+  (unless org-roam-bibtex-preformat-keywords
+    (setq org-roam-bibtex-preformat-keywords
+          '("=key=" "title" "url" "file" "author-or-editor" "keywords")))
+  (unless org-roam-bibtex-templates
+    (setq org-roam-bibtex-templates
```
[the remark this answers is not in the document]
brianmcgillion: Sorry for the delay, missed the updated thread here. Changes pushed.

modules/tools/biblio/config.el
brianmcgillion: Thanks @Emiller88, I will take a look

brianmcgillion:
> But I do know that I can work around this by adding the line:
> 
> ```
> (setq! org-ref-notes-directory "")
> ```
> 
> right above my config lines from earlier. If my elisp-fu were better, I'd try to fix this, but alas, that's all I got. Hope this helps.

Thanks @JonathanReeve, I think your deduction is spot on. `directory-name-p` does not seem to behave logically, i.e. nil in my understanding is not a directory name so should take the alternative path just like " " or "file", however, it is instead treated as an invalid parameter. I will look to fix that.

modules/tools/biblio/README.org
```diff
@@ -0,0 +1,146 @@
+#+TITLE:   tools/biblio
+#+DATE:    April 11, 2020
```
bdarcus, line 2:
```suggestion
#+DATE:    June 26, 2020
```

bdarcus: So how do we get this unstuck @hlissner?

tecosaur:
@brianmcgillion how's it going? Have you had any further time to look into this? I'm just a casual observes but it looks like this is 95% of the way there

modules/tools/biblio/README.org
hlissner: May as well move it up. The date should represent when it was added to Doom.

AnikaSU:
> An user of `org-roam` may use a one long file note using `org-ref`, so it would be a good idea that the module provides an independent flag (e.g. +roam-bibtex) for `org-roam-bibtex`.

I agree with that

modules/tools/biblio/README.org
bdarcus: "It" being the module; so when it was first merged?

Or "it" meaning this README, so when it is merged?

hlissner: It being the module.

helpotters:
Any updates about the org-ref and org-roam-bibtex integrations?

Emiller88:
@helpotters I can't speak for the plan on a timeline for integration of it into doom, but it's working very well for me by just dropping the changes here into my private config. I think if more people can test the stability of the module in their work we can find any nagging issues.

Emiller88:
I'm having an issue that I believe is related to the configuration. Occasionally I get an error when creating a new notes file.

`if: Symbol’s function definition is void: pushnew`

modules/tools/biblio/config.el
```diff
@@ -1,14 +1,117 @@
...
+  :hook (org-roam-mode . org-roam-bibtex-mode)
+  :config
+  (unless orb-templates
+    (setq orb-templates
+          '(("r" "ref" plain (function org-roam-capture--get-point)
+             ""
+             :file-name "${slug}"
+             :head "#+TITLE: ${=key=}: ${title}\n#+ROAM_KEY: ${ref}
+
+- tags ::
```
Emiller88, line 113:
```suggestion
             :head "#+title: ${=key=}: ${title}\n#+roam_key: ${ref}\n#+roam_tags:
             
```
`org-roam` recommends using lowercase keywords now, and they also have added `roam_tags` which can replace the tags section for better functionality.

Emiller88 (changes requested):

helpotters:
If it's solid enough to run for general testing, I'd be enthusiastic to
participate. Thank you for the suggestion and encouragement!

bdarcus:
It's definitely solid enough for testing (and even merging probably).

>

modules/tools/biblio/config.el
brianmcgillion: Thanks, sorry for the late reply. I changed to lower case, however, I have not tested without the tags :: yet. But I will do that soon and delete as needed

brianmcgillion:
> I've been testing this branch, and I'm no expert, so take this as you will, but I find that enabling this module causes everything org-related in Doom not to work, on my setup, at least. It complains: `(wrong-type-argument stringp nil)`. I did `toggle-debug-on-error` and I got this traceback (truncated, but called from require(org) and so on.

@JonathanReeve could you test, with some guidance from @Lejon the single file option should work. I tested a single org file without Roam enabled and it created the notes. If not please let me know!

brianmcgillion:
> If it's solid enough to run for general testing, I'd be enthusiastic to participate. Thank you for the suggestion and encouragement!

I use it on a daily basis. I just rebase ontop of doom/develop occasionally and it is working fine. Of course, I have my workflow and these tools allow many ways to achieve similar things, so I have not exhausted all config possibilities.

bdarcus (commented):
> @JonathanReeve could you test, with some guidance from @lejon the single file option should work. I tested a single org file without Roam enabled and it created the notes. If not please let me know!

After we get confirmation on this piece, LGTM!

JonathanReeve:
I've been using this successfully for some time now. Even started using org-roam, and org-roam-bibtex, as well. I haven't tested it with minimal or little config, but it works with my setup. :shipit:

A6GibKm:
Looking forward for this :+1:

modules/tools/biblio/config.el
```diff
@@ -1,14 +1,116 @@
...
+  ;; This need to be set before the package is loaded, because org-ref will
+  ;; automatically `require' an associated package during its loading.
+  (setq org-ref-completion-library (cond ((featurep! :completion ivy)  #'org-ref-ivy-cite)
+                                         ((featurep! :completion helm) #'org-ref-helm-bibtex)
+                                         (t                            #'org-ref-reftex)))
   :config
-  (add-to-list 'ivy-re-builders-alist '(ivy-bibtex . ivy--regex-plus)))
+  ;; Although the name is helm-bibtex, it is actually a bibtex-completion function
+  ;; it is the legacy naming of the project helm-bibtex that causes confusion.
+  (setq org-ref-open-pdf-function 'org-ref-get-pdf-filename-helm-bibtex)
```
lzmartinico, line 85: Hi, I just started testing your PR and it mostly works great! One questions I had though is why is this set to override org-ref-open-pdf-at-point? This change breaks opening a PDF file from a cite link (through pressing Enter for the org-ref modal, and Open PDF)

junwei-wang:
I am looking forward to merging this pull request.

modules/tools/biblio/README.org
```diff
@@ -0,0 +1,149 @@
...
+#+BEGIN_src emacs-lisp
+(setq! +biblio-pdf-library-dir "/path/to/bibrary/"
+       +biblio-default-bibliography-files '("/path/to/bibliography.bib")
+       +biblio-notes-path "/path/to/notes/")
+#+END_src
+
+Unless ~+biblio-notes-path~ ends in a ~/~ it is assumed to be a single file and not a directory.
+
+Each of the variables overlays the corresponding variables in the following way
++ ~+biblio-default-bibliography-files~
```
aaaawwWWWwwwwWWW, line 102:
```suggestion
+ ~+biblio-pdf-library-dir~
```

aaaawwWWWwwwwWWW:
since hugo and ox-hugo are already integrated into doom, would be nice the addition of https://github.com/andras-simonyi/citeproc-org into biblio

bdarcus:
> since hugo and ox-hugo are already integrated into doom, and now with this org-ref, would be nice the addition of https://github.com/andras-simonyi/citeproc-org into biblio

That would be a good next step, but we first have to get this merged. It's now been more than eight months.

aaaawwWWWwwwwWWW:
> That would be a good next step, but we first have to get this merged. It's now been more than eight months.

Sure thing! I've been using this PR for almost 11 days now and works great!

modules/tools/biblio/config.el
```diff
@@ -1,14 +1,116 @@
...
+  :hook (org-roam-mode . org-roam-bibtex-mode)
+  :config
+  (unless orb-templates
+    (setq orb-templates
+          '(("r" "ref" plain (function org-roam-capture--get-point)
+             ""
+             :file-name "${slug}"
+             :head "#+TITLE: ${=key=}: ${title}\n#+roam_key: ${ref}\n#+roam_tags: lit
+
+- tags ::
```
lzmartinico, line 112: What's the point of having this when you also have roam_tags? As far as I can tell, this and the keywords below aren't easily searchable. In fact, it would be really nice if the entries in keywords automatically linked to a note!

bdarcus: What's "this"? The tags?

If yes, those aren't functionally the same as roam_tags. It's like the difference between keywords/tags and categories.

lzmartinico: Yes sorry, I meant the tags. I was looking at the org-roam forum and it seems like they are half-deprecated, or at least there is no built-in way to search through them, so it's not clear to me how you would use them - insert a backlink to a "category" note?

junwei-wang: `SPC n r f` -> `org-roam-find-find`

lzmartinico: As far as I can tell, org-roam-find-file (I assume find-find was a typo?) only filters through the roam_tags, and not the tags:: ? Or am I missing some custom search filter?

lzmartinico: I went back and looked at Jethro's [How to take smart notes](https://blog.jethro.dev/posts/how_to_take_smart_notes_org/), which was helpful to understand how they used tags at the time; as I was guessing, it's for backlinks. I think then it makes sense to keep that in the template. On the other hand, it would be nice if a backlink was automatically added for backlinks!

bdarcus:
Is anyone else seeing this?

Using `ivy-bibtex`, when I select a candidate, and do `M-o` to get the action menu, I just get a generic ivy menu; not the `ivy-bibtex` actions I'm expecting.

natejlong:
@bdarcus I'm also seeing that generic ivy menu

bdarcus:
Thanks @natejlong.

I don't ever recall seeing that before. Is that something new in ivy?

bdarcus:
As we're still waiting to get this merged, I just thought I'd update on the discussions I've been involved in on this PR.

https://github.com/tmalsburg/helm-bibtex/pull/355

What's interesting about this is it proposes to adapt to bibtex-completion to make ivy/helm-bibtex-like functionality available in default Emacs completion, and hence also selectrum.

I think it would be straightforward to adapt here, for when there is a selectrum module merged.

natejlong:
Honestly it's the first time I've noticed it too! I was using helm-bibtex previously, not a big expert on helm or ivy to be honest. 

I should've also mentioned I'm not using this PR exactly, so it might be a config/package version difference, but if there's anything I can do to help you out let me know

bdarcus:
> Honestly it's the first time I've noticed it too! I was using helm-bibtex previously, not a big expert on helm or ivy to be honest.

I think we want that keybinding to bypass that interim menu and go straight to the actions, but I'm not an expert either, so not sure.

bdarcus:
As we wait on input from henrik, a bit of an update on the discussion I linked to above on adding support for completing-read/selectrum to bibtex-completion, as it has bearing on an initial decision on a module keybinding.

Both `ivy-bibtex` and `helm-bibtex` have their respective entry points, which brings up a search interface for a bibliography collection, and then one choose actions for the selected candidate(s).

But @minad made the, I think very good, suggestion to instead expose those actions as full interactive commands, so that the user can more directly do, for example `M-x ivy-bibtex-insert-key`.

Which brings me to this:

``` elisp
:desc "Bibliographic entries"        "b"
        (cond ((featurep! :completion ivy)   #'ivy-bibtex)
              ((featurep! :completion helm)  #'helm-bibtex)))
```
              
It strikes me that those commands really means "search bibliographic entries" (which is basically the docstring for these functions), and so I think ideally the binding should be "SPC n b s", and that perhaps later we can add additional commands to "SPC n b."

bdarcus:
As for @hlissner, here's what he said on discord awhile back in response to a query from @bmg:

> I have a couple design issues with the PR that I can't solve without more insight into how to use org-ref and org-roam-bibtex.

If this is the holdup ATM, it might give someone (@myshevchuk?) an idea of where to look, if they have the time and inclination.

lejon:
Please take me off this thread since I’m not the right Leif Jonsson.

BR
-Leif Jonsson

bdarcus:
Can we do that?

You might just click the unsubscribe link at the bottom of the
notification email?

lejon:
You’re right! I thought that was GitHub in general, but it was this specific thread!
Thanks!

Cheers!
-Leif

bdarcus:
> Is anyone else seeing this?
> 
> Using `ivy-bibtex`, when I select a candidate, and do `M-o` to get the action menu, I just get a generic ivy menu; not the `ivy-bibtex` actions I'm expecting.

Seems a bug:

https://github.com/abo-abo/swiper/issues/2397

myshevchuk:
@bdarcus Hi, is there anything specific I could help with? I looked at the configuration code and most of it is now outdated. Also, ORB is still unstable and a few possibly breaking changes are expected.

bdarcus:
> @bdarcus Hi, is there anything specific I could help with? I looked at the configuration code and most of it is now outdated. Also, ORB is still unstable and a few possibly breaking changes are expected.

No, I was just wondering if you saw anything off that you thought might have caused the concern that @hlissner noted, but didn't really explain.

But what you say itself might itself suggest removing this for now.

Should we just do that and merge the other stuff, @brianmcgillion?

Jonghyun-Yun:
The module causes errors with vanilla `config.el`, which is something people don't expect to see. How about revising 
https://github.com/brianmcgillion/doom-emacs/blob/6d61f98ef11c4fcdfd30135564a9290d7337ca27/modules/tools/biblio/config.el#L88-L93 as follows?

``` emacs-lisp
    (setq org-ref-notes-function #'org-ref-notes-function-one-file)
    (unless (null org-ref-notes-directory)
      (setq org-ref-notes-function (if (directory-name-p org-ref-notes-directory)
                                       #'org-ref-notes-function-many-files
                                     #'org-ref-notes-function-one-file)))
```

bdarcus:
In case anyone is testing the new [WIP selectrum PR](https://github.com/hlissner/doom-emacs/pull/4664), I've put together a PR for bibtex-completion that makes it compatible.

So if you want to test the two together, I now finally got it all working correctly ....

https://github.com/tmalsburg/helm-bibtex/pull/361

Screenshot from 2021-03-01 16-20-49

The screenshot demonstates my point above about it possibly making sense to change the keybindings for the bibliography module; that `ivy-bibtex` and `helm-bibtex` should probably be bound to `n b s` for "search".

bdarcus:
> In case anyone is testing the new [WIP selectrum PR](https://github.com/hlissner/doom-emacs/pull/4664), I've put together a PR for bibtex-completion that makes it compatible.

Update; now a separate package.

https://github.com/bdarcus/bibtex-actions

helpotters:
I'm a bit lost. Is there a new package just for ivy-bibtex/helm-bibtex compatability? kkkkkkkk

bdarcus:
> I'm a bit lost. Is there a new package just for ivy-bibtex/helm-bibtex compatability?

What are you referring to?

If the package I posted just above, the easiest way to think of it is it's like `selectrum-bibtex` (though will work with other completing-read-based completion systems).

If this PR, no; the module already has that; this is expanding the features.

helpotters:
The org-file for biblio is not very clear in how to setup or utilize the package. 

For example, something as simple as inserting a citation from the default bibtex file.

bdarcus:
> The org-file for biblio is not very clear in how to setup or utilize the package.
> 
> For example, something as simple as inserting a citation from the default bibtex file.

It's been awhile (this PR was submitted almost a year ago at this point), but I think that's because these packages, and this module, provide a range of ways that could work.

Are you authoring in latex-mode, markdown-mode, or org-mode, for example?

Using primarily ivy-bibtex or helm-bibtex to insert the citations, or something else?

Aside: if the former, you open those commands, and insert the citation from there. But that's extensively documented at [that project](https://github.com/tmalsburg/helm-bibtex).

Do others have any views on this question; about the depth of what should be documented here?

init.example.el
```diff
@@ -85,6 +85,7 @@
 
        :tools
        ;;ansible
+       ;;biblio            ; A collection of bibliographic reference tools
```
tecosaur, line 88: Can this be sassier? :grin:

bdarcus: Best comment ever!

Any suggestions?

tecosaur: Courtesy of the Doom discord:
> Writes a PhD for you[citation needed]

brianmcgillion:
> I believe these should be, and are, set with the +biblio variables @brianmcgillion, so you can remove this part?

At the top of the section it states "It is of course possible to set the variables directly through their respective packages as shown below, in which case you can ignore the ~+biblio...~ versions.". So the instructions are there more for completeness if someone chooses that option.

init.example.el
payasrelekar: One question, is this going to be a separate module, as seen in above line, org should it be a flag for org module, like +roam, +deft etc.
Considering this is heavily dependent on org-mode and org-roam, shouldn't be a flag?

bdarcus: > One question, is this going to be a separate module, as seen in above line, org should it be a flag for org module, like +roam, +deft etc.
> Considering this is heavily dependent on org-mode and org-roam, shouldn't be a flag?

I'm confused. It's already a separate module.

And a user of this module may heavily use latex, for example, but not org at all.

It won't load those org-related packages unless you use the `org` module.

bdarcus:
For those of you interested in this, but not on the org list, after many years of stops and starts, `org-cite` (native org citation support) is set to be merged soon, and Bastien put out a call for testing today.

https://lists.gnu.org/archive/html/emacs-orgmode/2021-05/msg00003.html

So if you contribute to that, it will get it merged sooner rather than later.

If all goes smoothly, this should be merged for org 9.5.

natejlong:
> For those of you interested in this, but not on the org list, after many years of stops and starts, org-cite (native org citation support) is set to be merged soon, and Bastien put out a call for testing today.

Thank you so much for linking this, I had no idea this was in the works. More importantly thank you @bdarcus for your work on the feature itself! I read through the thread there and the citation syntax looks like _exactly_ what I've wanted in org. So excited to try this out

bdarcus:
Does `org-cite` have any implications for the details of this package?

I'm wondering in particular about `org-ref`.

I do know it's author (John Kitchen) is hoping to rewrite that package to support it, but also that he doesn't have much time to do this, and has estimated it might take a year.

So not sure; maybe we should put that package behind a flag?

bdarcus:
> > In case anyone is testing the new [WIP selectrum PR](https://github.com/hlissner/doom-emacs/pull/4664), I've put together a PR for bibtex-completion that makes it compatible.
> 
> Update; now a separate package.
> 
> https://github.com/bdarcus/bibtex-actions

Another update:

Now on MELPA, and now incorporated into the above monster selectrum PR.

So if you have `selectrum` and `biblio` selected in your init.el, you will get `bibtex-actions`.

tecosaur:
I'm definitely of the mind that adding citeproc would be a good idea :+1: 
Also, I'd estimate the merge timeline to be of the order of weeks based on current progress.

bdarcus:
> I'm definitely of the mind that adding citeproc would be a good idea.

That, plus putting `org-ref` behind a flag, would seem the best way forward, giving users the most flexibility.

I don't, for example, use `org-ref`, and likely won't with `org-cite`. But I suspect others will use both.

brianmcgillion:
based on a discussion with @bdarcus I have added +ref for inclusion of org-ref. Please add (org +ref) to your .doom.d/init.el

bdarcus:
> based on a discussion with @bdarcus I have added +ref for inclusion of org-ref. Please add (org +ref) to your .doom.d/init.el

Thanks @brianmcgillion!

FYI, I've learned a bit more about `org-cite` "follow" processors in the last 48 hours, after earlier not paying much attention to them.

The `org-cite` author describes them succinctly as "Follow is what happens when you use C-o on a citation." 

In other words, it allows you to bind a function to `org-open-at-point`, to be able to run actions on either org citations or citation-references.

My plans for `bibtex-actions`, which will allow a kind of lightweight embarkified micro-org-ref:

https://github.com/bdarcus/bibtex-actions/issues/141

If and when John rewrites `org-ref`, I'd expect it to be just such a "follow" processor.

chuxubank (dismissed):

iyefrat:
@brianmcgillion Hey! I have a few remarks on the PR:

- I think the `+ref` flag makes more sense on the `biblio` module, since all the code that deals with it is in the `biblio` module, maybe as a `+org-ref` flag.
- You might want to state more clearly in the readme that  `org-ref` installs **both** `ivy` and `helm` as dependencies, as users might wonder where the helm installation came from.
- I think you forgot to update the existence of the `+ref` flag in the readme.
-  You might also want to introduce a doctor check that warns users if they have `:tools biblio +org-ref` on but not `:lang org` on, and maybe one that warns them against using `org-ref` if they are using a completion system that isn't `ivy` or `helm` since it seems that'll cause issues.

modules/tools/biblio/config.el
```diff
@@ -1,14 +1,116 @@
...
-  (add-to-list 'ivy-re-builders-alist '(ivy-bibtex . ivy--regex-plus)))
+  ;; Although the name is helm-bibtex, it is actually a bibtex-completion function
+  ;; it is the legacy naming of the project helm-bibtex that causes confusion.
+  (setq org-ref-open-pdf-function 'org-ref-get-pdf-filename-helm-bibtex)
+  ;; orb will define handlers for note taking so not needed to use the
+  ;; ones set for bibtex-completion
+  (unless (featurep! :lang org +roam)
+    ;; Allow org-ref to use the same template mechanism as {helm,ivy}-bibtex for
+    ;; multiple files if the user has chosen to spread their notes.
+    (setq org-ref-notes-function (if (and org-ref-notes-directory (directory-name-p org-ref-notes-directory))
```
bdarcus, line 91: This will need to be changed to incorporate the org-ref flag?

Default behavior with org should be to use org-cite (when it's merged, though that will be before this module is merged here).

bdarcus:
I just added the draft PR to add org-cite support https://github.com/hlissner/doom-emacs/pull/5212, to hopefully make it easier to align this and (maybe) #4664.

bdarcus:
>     * You might want to state more clearly in the readme that  `org-ref` installs **both** `ivy` and `helm` as dependencies, as users might wonder where the helm installation came from.

A related issue: `org-roam-bibtex` currently requires `org-ref`, so has the same issue, though they intend to move to `org-cite`.

BTW, both `org-roam` and `org-roam-bibtex` are set for a major breaking change in a few days, to correspond with the release of `org-roam v2`.

bdarcus:
Just by way of update, org-cite is now in org-mode master.

bdarcus:
> Just by way of update, org-cite is now in org-mode master.

I think this is ready for review.

https://github.com/hlissner/doom-emacs/pull/5212

@hlissner will ultimately have to decide the path forward.

I think it's no-brainer that org-cite becomes default when the org + biblio modules.

Less clear is what to do about this PR.

I have added a section to #5212 that addresses that, but to summarize here I see two options:

1. close this PR in favor of #5212, but add some of this there
2. close #5212 in favor of this, but merge that code here

Or I guess a third option is #5212 could add org-cite on top of this. This just might need to be adjusted to ensure that's easy to do.

1 is a cleaner path going forward, particularly with the new vertico module (org-ref, as discussed above, will install both ivy and helm even if you use vertico), while 2 provides more options.

bdarcus:
Screencast demo of WIP org-cite version of org-ref.

https://www.youtube.com/watch?v=4ta4J20kpmM

I was already leaning towards removing org-ref from here, but this makes the case stronger.

bdarcus:
Update, after talking to @brianmcgillion today.

We're going to drop `org-ref` from here, but I plan to (already have actually) add it's successor, `org-ref-cite`, to #5290.

The goal will be to merge that first, for Brian to rebase on that, and then this PR will just be about adding ORB.

modules/tools/biblio/config.el
chuxubank:
```suggestion
  (setq org-ref-get-pdf-filename-function #'org-ref-get-pdf-filename-helm-bibtex)
```

iyefrat:
@brianmcgillion so since #5290 got merged this can be reworked on top on that (and be merged finally :grimacing:)

brianmcgillion:
will close this merge request. Will add org-roam-bibtex in https://github.com/hlissner/doom-emacs/pull/6077 and org-ref can be discussed for inclusion separately.

PR #2889, opened 2020-04-11
[RFC] Allow helm-projecile to remap projectile

brianmcgillion:
Helm projectile provides the function helm-projectile-on that performs
the remapping of projectile functions to helm-projectile equivalents.
Use this functionality so that it always stays in sync between versions.


hlissner:
Thanks for the PR, but I can't accept this one. I remap them manually so that the keybinds are available before the package is loaded. If done with `helm-projectile-on` then the user would have to wait until the package is loaded to use those keys. Thanks anyhow!

Issue #2802, opened 2020-03-31
Could not find package X in recipe repositories


flyingns:
After updating doom with `doom upgrade`, or its packages with `doom update`, I get errors like these:

```
Package error: (error Could not find package org-roam in recipe repositories: (org-elpa melpa gnu-elpa-mirror emacsmirror-mirror))
Package error: (error Could not find package company-org-roam in recipe repositories: (org-elpa melpa gnu-elpa-mirror emacsmirror-mirror))
```

And/or errors like these:

```
Cannot open load file: No such file or directory, org-roam
Cannot open load file: No such file or directory, org-superstar
```

jadestrong:
Me too.

hlissner:
It's a known issue where your local melpa repo isn't being updated. I'm working on a fix, but in the mean time you'll need to do:

```sh
# Update MELPA
cd ~/.emacs.d/.local/straight/repos/melpa
git pull

# Clear straight's cache
rm -f ~/.emacs.d/.local/straight/build-cache.el
doom sync
```

jcf:
I've encountered this same issue, and the recommended workaround fails like so:

``` sh
$ doom sync
Building straight...
Building straight...done
Building use-package...
Building use-package → Building bind-key...
Building use-package → Building bind-key...done
Building use-package...
Building use-package...done
> Synchronizing your config with Doom Emacs...
  > Regenerating envvars file at "~/.emacs.d/.local/env"
  - Scraping shell environment
    - Ignoring HOME=/Users/jcf
    - Ignoring GPG_TTY=/dev/ttys006
    - Ignoring PWD=/Users/jcf/.emacs.d/.local/straight/repos/melpa
    - Ignoring __CF_USER_TEXT_ENCODING=0x1F5:0x0:0x2
    - Ignoring SSH_AUTH_SOCK=/Users/jcf/.gnupg/S.gpg-agent.ssh
    - Ignoring TERM=xterm-256color
    ✓ Successfully generated "~/.emacs.d/.local/env"
  > (Re)generating core autoloads...
    > Generating core autoloads...
    > Byte-compiling core autoloads file...
    ✓ Generated .local/autoloads.elc
  > Installing packages...
    → Cloning auto-minor-mode...
x There was an unexpected error:
  Type: doom-package-error
  Message:
    Error with packages
  Data:
    "auto-minor-mode"
    (error "Failed to run \"git\"; see buffer *straight-process*")
[...]

PR #2893, opened 2020-04-12
Add recipe for company-org-roam

daveduthie:


hlissner:
company-org-roam is [on MELPA](https://melpa.org/#/company-org-roam), so the recipe isn't necessary.

If you're experiencing the package errors like this one:

```
Package error: (error Could not find package org-roam in recipe repositories: (org-elpa melpa gnu-elpa-mirror emacsmirror-mirror))
```

Then see https://github.com/hlissner/doom-emacs/issues/2802#issuecomment-606654151.

PR #2896, opened 2020-04-13
Enable visual-line-mode in org-roam-backlinks buffer

alarsyo:
`org-roam` has a buffer that shows backlinks to the file currently displayed. If the backlink is in a multiline paragraph, `org-roam` will show it as a single line inside this buffer, making it unreadable for Doom users by default, since `truncate-lines` is `t`.

This adds a hook to enable `visual-line-mode` (soft wrapping) when in the `org-roam-backlinks` buffer.

I've made the change in my private config, but I feel like this is a good default for Doom users.

Before: 
2020-04-13-193359_642x184_scrot

After:
2020-04-13-193104_637x181_scrot

### DIFF
--- a/modules/lang/org/contrib/roam.el
+++ b/modules/lang/org/contrib/roam.el
@@ -3,6 +3,7 @@
 
 (use-package! org-roam
   :hook (org-load . org-roam-mode)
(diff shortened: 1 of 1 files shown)


hlissner:
Makes sense to me. Thanks!

PR #2906, opened 2020-04-15
Add set -euo pipefail to sh-mode file template

rushsteve1:
Adds a line with set -euo pipefail to the sh-mode file template, which improves the erroring of bash scripts.
As described by https://ashishb.net/all/the-first-two-statements-of-your-bash-script-should-be/


hlissner:
Makes sense to me. Thanks!

PR #2916, opened 2020-04-17
Update hook to match rename from refactoring

albertfong:
If you have omnisharp & csharp set up, and you close a cs file, you'll get `Symbol's function definition is void: +csharp-cleanup-omnisharp-server-h`.

This is because 169f9a6121d8e0c75c0b27c2566f050477828dce moved its definition to csharp/autoload.el, but also renamed it to +csharp-kill-omnisharp-server-h.

This PR updates the hook to match the new name.


hlissner:
Ah, good catch! Thanks!

Issue #2837, opened 2020-04-07
gccemacs support?


sebastiansturm:
Several people have claimed that gccemacs is stable enough for daily use; after having the first 3 compilation attempts segfault on me, today's the first time I've been able to build it from source and indeed it seems to work very well so far (though right now it's still mostly busy compiling various packages I have installed). What do you think about having doom's cli tool call `(native-compile-async)` during `.elc` generation if `native-compile-async` is defined?

sebastiansturm:
sorry, this is of course not a bug. Will try to remove this false label if I can

gagbo:
Can you provide a small patch so people who want to opt in a world full of undefined bugs can try ? :)

sebastiansturm:
my gccemacs is still (or rather, again) compiling because I decided to recompile `.local/straight` from scratch. If upfront compilation is not needed (though I'd very much prefer that, also it's much more in line with doom's compilation-centric approach), it is nearly sufficient to add `(setq comp-deferred-compilation t)` to `early-init.el`, then start Emacs and be very patient.
That led to Emacs complaining about missing files within the `evil-collection` directory on next startup though. I thought it might be because the .eln-file resides in a platform-specific subdirectory, so the first thing I'll try once the compilation is done is to symlink non-el files to the eln directory and see if that helps. If it does, I'll write a small function for private use that does this automatically (and maybe post it here), and ask on emacs-devel how to solve this issue properly. If it doesn't help, maybe I'll post the error message here or ask on emacs-devel right away

sebastiansturm:
so, this time I just did `(native-compile-async "~/.emacs.d/.local/straight/build/" t t)` and got the same evil-collection error as last time when I restarted Emacs afterwards. That error went away after symlinking `~/.emacs.d/.local/straight/build/evil-collection/modes` to the `eln-...` directory created within; surprisingly no other package raised any complaints so far.
Better take this information with a huge grain of salt though; I still have lots of `.eln` files from my previous attempt at `~/.emacs.d/.local/straight/repos/...`. I guess these are unused but I won't remove them yet as otherwise the remainder of this evening will again be spent recompiling. Will clear out all eln stuff below `repos` later so I can recompile overnight if this should turn out to break things.
After doom startup, most package-provided functions are described as autoloaded lisp functions by `describe-function` but after invoking them once, they are described as native code so I guess gccemacs is doing its magic. Haven't done benchmarks of any sort so far

sebastiansturm:
[...]

PR #2917, opened 2020-04-17
[WIP] Support native compilation of packages

flatwhatson:
Adds experimental support for native-compilation, built on top of the experimental native-compilation support for straight.el (https://github.com/raxod502/straight.el/pull/507).

To play with this, you'll need an emacs built from the `feature/native-comp` branch, and the version of `straight.el` linked above.

### DIFF
--- a/core/cli/packages.el
+++ b/core/cli/packages.el
@@ -263,10 +263,12 @@ declaration) or dependency thereof that hasn't already been."
                              (file-exists-p (straight--modified-dir (or local-repo package)))
                              (cl-loop with want-byte   = (straight--byte-compile-package-p recipe)
                                       with want-native = (if (require 'comp nil t) (straight--native-compile-package-p recipe))
+                                      with outdated = nil
                                       for file in (doom-files-in build-dir :match "\\.el$" :full t)
                                       if (or (if want-byte   (doom--elc-file-outdated-p file))
                                              (if want-native (doom--eln-file-outdated-p file)))
-                                      return t))
+                                      do (setq outdated t)
+                                      finally return outdated))
                          (puthash package t straight--packages-to-rebuild))))
                 (straight-use-package (intern package))))
          (progn
--- a/core/cli/packages.el
+++ b/core/cli/packages.el
@@ -134,8 +134,7 @@ list remains lean."
       t)))
 
 (defun doom--eln-file-outdated-p (file)
-  (when-let* (((require 'comp nil 'noerror))
-              (eln-file (comp-output-filename file))
+  (when-let* ((eln-file (comp-output-filename file))
               (error-file (concat eln-file ".error")))
     (push eln-file doom--expected-eln-files)
     (cond ((file-exists-p eln-file)
@@ -180,8 +179,8 @@ list remains lean."
            else do
            (let ((inhibit-message t))
              (sleep-for 0.1)))
-  ;; HACK Write .error files for any missing files which still don't
-  ;; exist. We'll just assume there was some kind of error...
+  ;; HACK Write .error files for any missing files which still don't exist.
+  ;;      We'll just assume there was some kind of error...
   (cl-loop for eln-file in doom--expected-eln-files
            for error-file = (concat eln-file ".error")
            unless (or (file-exists-p eln-file)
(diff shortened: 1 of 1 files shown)


hlissner:
This looks great! While I can't accept the PR (because our CLI is undergoing a huge rewrite), I'll merge it into the rewrite as soon as it's ready. Thanks for your help!

Emantor:
Related to #2837

quangIO:
Great work! I am using this with gccemacs (Binary version here for reference https://github.com/quangIO/gccemacs-bin/) and have no issue so far but a huge performance boost.

hlissner:
Sorry for the late follow up! Since there's been an uptick on native-comp users (and my cli rewrite will need a little longer), I've decided to merge this for now. It still generate excessive *.eln.error files in packages' build folder, but this shouldn't have any adverse affect. In any case, a better implementation is in the works. Thanks again for this PR and help testing it @flatwhatson!

AllenDang:
@flatwhatson Really glad to see this PR.
After doom clean & doom sync, I'm stuck as - Waiting for 1 async jobs... for 20 minutes.
Is this normal?

flatwhatson:
Oh right!  Yes, building `evil-collection-vterm.el` always hangs.  I've just been killing the child process to allow `doom sync` to finish, you'll see "evil-collection-vterm" in its name.  Probably should add that one to the compilation blacklist, and report it upstream.

robbert-vdh:
You'll also have to manually compile the vterm module itself because the path detection is broken right now (https://github.com/akermu/emacs-libvterm/issues/363):

```shell
cd ~/.emacs.d/.local/straight/build/vterm/
cmake -Bbuild -DCMAKE_BUILD_TYPE=Release
make -Cbuild
```

AllenDang:
Path detection is broken? Is this the root cause doom cannot start lsp server and projectile related function doesn't work?

robbert-vdh:
@AllenDang The path detection is broken for the vterm module because it uses the location of `vterm.el` to find the source files for the vterm module, and with gccemacs the compiled version of this file now sits in a different subdirectory. LSP and projectile (along with everything else) work just fine for me with gccemacs built yesterday.

AllenDang:
@robbert-vdh After doom upgrade, no errors occurred again. And I also set following two lines in config.el
```
(setenv "LIBRARY_PATH" "/usr/local/opt/gcc/lib/gcc/10")
(setq comp-deferred-compilation t)
```

And I find that emacs is slower than before. Don't know what's wrong...

AllenDang:
I guess the native compile result of all modules are output to ~/.emacs.d/.local/straight with suffix .elc right?
Are they going to be loaded as *.eln automatically?
Because I didn't feel any performance boost.
Should I add any thing to config.el or early-init.el to make it work?
Should I still set comp-deferred-compilation to true?

I compiled gccemacs with NATIVE_FAST_BOOST=1, will this be the root cause why performance is slow?

zeronone:
@AllenDang `*.eln` files are are created in a `eln-<system-architecture>` directory for each package.
You can find them with following:
```
find ~/.emacs.d/.local/straight/build -name "eln-*" -type d
```

To check if a native-compiled version is being used, check `describe-function` (`SPC h f`) for that function (i.e: `org-mode`).

AllenDang:
@zeronone Thanks! I somehow managed to make it work! Seems I don't need to add following two lines to config.el.

```lisp
(setenv "LIBRARY_PATH" "/usr/local/opt/gcc/lib/gcc/10")
(setq comp-deferred-compilation t)
```

AllenDang:
I noticed that doom with native-compile seems really slow. At least much slower then spacemacs.
I can set company-idle-delay to 0 in spacemacs and still feel smooth.
But with company-idle-dely as 0.2 in doom feels lag when typing.

AloisJanicek:
> Seems I don't need to add following two lines to config.el.

I don't know about the first line, but you need to set `comp-deferred-compilation` to non-nil as advised in gccmacs documentation on [Deferred compilation and Fast bootstrap](http://akrl.sdf.org/gccemacs.html#org4be982d) if you compiled you emacs with `NATIVE_FAST_BOOT=1`. The issues you had experienced are most likely related to the fact, that you didn't have write access to `/usr/share/emacs/` directory where are tons of elisp files waiting for native compilation.

I just compiled [emacs-native-comp-git-enhanced](https://aur.archlinux.org/packages/emacs-native-comp-git-enhanced) from AUR, executed `doom sync`, set `(setq comp-deferred-compilation t)` in my `init.el`, did `chmod -R 757 /usr/share/emacs/` in bash and my doom is ridiculously fast.

flatwhatson:
**The ONLY requirement for native-compiled Doom is a native-comp build of Emacs.**

I wouldn't recommend using either `NATIVE_FAST_BOOT` or `comp-deferred-compilation`.

The `NATIVE_FAST_BOOT` option is a hack which was useful when native-comp was very slow and Emacs would take *hours* to build.  That's no longer the case, a native-comp build doesn't take too long and lazily compiling the core Elisp is more trouble than it's worth.  Much simpler to compile everything ahead-of-time as is the default.

The `comp-deferred-compilation` option is necessary if using `NATIVE_FAST_BOOT`, but otherwise isn't too useful when Doom is already compiling your packages ahead-of-time.

Changing permissions on `/usr/share/emacs` might be necessary when using `NATIVE_FAST_BOOT`, but is a **bad idea**.  This directory contains code which is loaded and executed by Emacs.  Making it globally writable might allow an attacker to inject some malicious Elisp and **steal all your cookies**.

If you aren't noticing a performance improvement, first check that your Emacs is actually able to natively-compile things.  Running `find ~/.emacs.d -name '*.eln' | wc -l` after a `doom build -r` should report roughly a thousand items.  If it's zero, native compilation isn't working on your system and needs to be debugged separately.

AllenDang:
@flatwhatson Thanks for your advise! I'm re-compiling gccemacs without native_fast_boot now.

AllenDang:
@flatwhatson I compiled gccemacs without native_fast_boot. And I got 872 with `find ~/.emacs.d -name '*.eln' | wc -l`.
But the performance is still slow. I'm using MacOS catalina.
Do you have any thoughts about the root cause.

flatwhatson:
Perhaps you're just looking for "performance" in the wrong places?  Native compilation doesn't doesn't speed up Doom's startup, nor does it speed up opening a file in a new mode (where a bunch of Elisp is lazily loaded).  Actual *editing* operations and moving through larger ivy menus should be noticeably faster.

AllenDang:
@flatwhatson The performance I mean is editing experience and auto-complete from lsp-mode.

hlissner:
> You'll also have to manually compile the vterm module itself because the path detection is broken right now ([akermu/emacs-libvterm#363](https://github.com/akermu/emacs-libvterm/issues/363)):
> 
> ```shell
> cd ~/.emacs.d/.local/straight/build/vterm/
> cmake -Bbuild -DCMAKE_BUILD_TYPE=Release
> make -Cbuild
> ```

This is now its own issue: #3683. I've posted an in-Emacs workaround for it there as well.

PR #2919, opened 2020-04-17
Autostart LSP also in scss-mode

AloisJanicek:
LSP works fine with scss-mode but currently has to be started manually.


hlissner:
Thanks!

PR #2920, opened 2020-04-17
Autoload all 4 org decrypt/encrypt user commands

AloisJanicek:
currently if someone wants to use them, it has to manually autoload
them first in user `config.el` with `(unless (fboundp ...) (autoload ...))`


hlissner:
Thanks!

PR #2925, opened 2020-04-17
doom/upgrade use -y CLI flag

rushsteve1:
When using the internal doom/upgrade command to upgrade using the CLI
tool, do so with the -y flag.

This is admittedly a temporary solution, but without it the CLI tool
will prompt for confirmation, which cannot easily or obviously be given
while inside Emacs.
A better solution would be to somehow integrate the prompt so that it
appears in the minibuffer as normal Emacs prompts do. But for now this
at least fixes doom/upgrade.


hlissner:
Thanks!

PR #2927, opened 2020-04-18
Make :ls/:buffers invoke ibuffer

rgrinberg:
Only when evil +everywhere and ibuffer are enabled

### DIFF
--- a/modules/editor/evil/+commands.el
+++ b/modules/editor/evil/+commands.el
@@ -99,3 +99,7 @@
 
 ;;; Org-mode
 (evil-ex-define-cmd "cap"         #'org-capture)
+
+;;; ibuffer
+(when (featurep! :emacs ibuffer)
+  (evil-ex-define-cmd "buffers" #'ibuffer))


modules/emacs/ibuffer/config.el
```diff
@@ -68,6 +68,9 @@
                default-directory)
            default-directory)))))
 
+  (when (featurep! :editor evil +everywhere)
+    (evil-ex-define-cmd "buffers" #'ibuffer))
```
hlissner, line 72: Could you move this to modules/editor/evil/+commands.el, in a `(when (featurep! :emacs ibuffer) ...)` check? This rules out any race conflicts caused by ibuffer loading before evil-ex is loaded [due to how we lazy load of Evil's ex commands](https://github.com/hlissner/doom-emacs/blob/develop/modules/editor/evil/config.el#L195-L197).

hlissner: Oh, and I don't think the `+everywhere` check is necessary. Extra ex commands aren't anywhere near as imposing as all the extra evilified keybinds that comes with that flag.

rgrinberg: Done.

hlissner:

PR #2929, opened 2020-04-18
Fix :last and :popup

rgrinberg:
The doom/ prefix for popup commands no longer exists


hlissner:
Good catches. Looks like these have been sitting around for quite a while. Thanks!

PR #2933, opened 2020-04-18
Register org's DONE as fixed pitch in mixed-pitch

yoavm448:
Currently, the TODO is fixed-pitch but DONE is mixed, which I find very odd. If there _is_ a reason for it, I'd be curious to hear it.


hlissner:
Makes sense. I not sure why it was omitted either.

Thanks for the PR!

PR #2948, opened 2020-04-21
Add full name to :cap

rgrinberg:
Makes it available as `:capture` which is a slightly easier to understand
name. The short alias is still available


rgrinberg:
Done!

hlissner:
Thanks!

PR #2950, opened 2020-04-21
Add lsp-command-map to SPC-c-l

jsravn:
The default lsp-mode bindings are bound to `s-l` which is impractical
for anyone using keybindings in their window manager. This makes all the
default lsp bindings available on SPC-c-l.


jsravn:
Sorry, this is broken - I'm not sure how to add a lazily loaded command map.

brianmcgillion:
```
(map! 
      (;;:feature that will be lazy loaded
        :leader
        (:prefix-map ("c" . "Feature")
          :desc "Some name"                "l" #'func1
          :desc "Some name2"              "m" #'func2)))
```
```
(use-package! some-package
  :bind (:map package-map
          ("<left>" . #'do-something1)
          ("<right>" . #'do-something2)))
```

Depending on the case, both should allow lazy loading

jsravn:
Replacement at https://github.com/hlissner/doom-emacs/pull/2952

PR #2952, opened 2020-04-21
Add lsp-command-map to SPC-c-l

jsravn:
The default lsp-mode bindings are bound to `s-l` which is impractical
for anyone using keybindings in their window manager. This change makes
all the default lsp bindings (see
https://github.com/emacs-lsp/lsp-mode#commands) available on SPC-c-l.

So for example, restarting the lsp server can be done with `SPC c l s r`.


modules/tools/lsp/config.el
```diff
@@ -114,6 +116,10 @@ This also logs the resolved project root, if found, so we know where we are."
                               (lambda (it) (format "[%s]" (lsp--workspace-print it)))
                               lsp--buffer-workspaces))))))
 
+  (dolist (leader-key (list doom-leader-key doom-leader-alt-key))
+    (let ((lsp-keymap-prefix (concat leader-key " c l")))
+      (lsp-enable-which-key-integration)))
+
```
hlissner, line 122: Could you wrap this (and the `(setq lsp-keymap-prefix nil)` in

```elisp
(when (featurep! :config default +bindings)
  ...)
```

jsravn: I'm thinking of removing `(setq lsp-keymap-prefix nil)` since that will break the bindings for people currently using it - and it doesn't hurt to keep it in. WDYT?

jsravn: Left it alone for now.

hlissner:
Thanks for the help! Sometime in the not-too-near future, once LSP/eglot is promoted to more of a first-class citizen in Doom, I think I may move the lsp keybinds to their own prefix, e.g. `SPC l`, but until then I think this works just fine.

PR #2956, opened 2020-04-22
Add Poetry support

tokilone:
Poetry is a tool for managing Python projects. There's a very nice wrapper around that tool called [poetry.el](https://github.com/galaunay/poetry.el). I think it's worth adding a module flag for it.


tokilone:
Done

hlissner:
Thanks for your help!

PR #2957, opened 2020-04-22
mu4e: Adjust config to mu4e 1.4

agraul:
### Description
`mu4e` 1.4 brings quite a few changes, including the deprecation of previously used variables. These are now guarded by a version comparison. `org-mu4e` is not needed for linking anymore, `org-mu4e-link-query-in-headers-mode` was renamed. It's new version defaults to `nil` and is therefore not set anymore -- I am not sure if the default was changed or setting it was not strictly needed previously.

### ToDos
- [x] Adjust `set-email-account!`
- [x] Update README

### DIFF
--- a/modules/email/mu4e/config.el
+++ b/modules/email/mu4e/config.el
@@ -14,9 +14,11 @@
   :commands mu4e mu4e-compose-new
   :init
   (provide 'html2text) ; disable obsolete package
-  (setq mu4e-maildir "~/.mail"
-        mu4e-attachment-dir "~/.mail/.attachments"
-        mu4e-user-mail-address-list nil)
+  (require 'mu4e-meta)
+  (when (version< mu4e-mu-version "1.4")
+    (setq mu4e-maildir "~/.mail"
+          mu4e-user-mail-address-list nil))
+  (setq mu4e-attachment-dir "~/.mail/.attachments")
   :config
   (pcase +mu4e-backend
     (`mbsync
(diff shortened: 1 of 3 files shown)


modules/email/mu4e/config.el
```diff
@@ -14,9 +14,10 @@
   :commands mu4e mu4e-compose-new
   :init
   (provide 'html2text) ; disable obsolete package
-  (setq mu4e-maildir "~/.mail"
-        mu4e-attachment-dir "~/.mail/.attachments"
-        mu4e-user-mail-address-list nil)
+  (when (version< mu4e-mu-version "1.4")
```
agraul, line 17: This is not working: `(void-variable mu4e-mu-version)`, I need to look into that again

agraul: Manually requiring `mu4e-meta` works, I am not sure about the performance implications. `mu4e-meta` is pretty small though, this is the one on my system:
```lisp
;; auto-generated
(defconst mu4e-mu-version "1.4.1"
  "Required mu binary version; mu4e's version must agree with this.")

(defconst mu4e-builddir "/home/abuild/rpmbuild/BUILD/mu-1.4.1"
  "Top-level build directory.")

(defconst mu4e-doc-dir "/usr/share/doc/mu"
  "Mu4e's data-dir.")

(provide 'mu4e-meta)
```

hlissner: I would like to avoid this in the long run. Perhaps we can perform that logic inside the :config block, but this is fine for the time being.

hlissner:

PR #2960, opened 2020-04-23
switching org-roam-completion to use 'default

bdarcus:
I do not understand why this is the case, but the conditional setup for the completion system was causing fuzzy completion to fail in ivy. This fixes that.

### DIFF
--- a/modules/lang/org/contrib/roam.el
+++ b/modules/lang/org/contrib/roam.el
@@ -36,11 +36,7 @@
   (setq org-roam-directory (expand-file-name (or org-roam-directory "")
                                              org-directory)
         org-roam-verbose nil  ; https://youtu.be/fn4jIlFwuLU
-        org-roam-completion-system
-        (cond ((featurep! :completion helm) 'helm)
-              ((featurep! :completion ivy) 'ivy)
-              ((featurep! :completion ido) 'ido)
-              ('default)))
+        org-roam-completion-system 'default)
 
   ;; HACK Hide the mode line in the org-roam buffer, since it serves no purpose.
   ;;      This makes it easier to distinguish among other org buffers.
--- a/modules/lang/org/contrib/roam.el
+++ b/modules/lang/org/contrib/roam.el
@@ -36,7 +36,14 @@
   (setq org-roam-directory (expand-file-name (or org-roam-directory "")
                                              org-directory)
         org-roam-verbose nil  ; https://youtu.be/fn4jIlFwuLU
-        org-roam-completion-system 'default)
+        org-roam-completion-system
+        (cond ((featurep! :completion helm) 'helm)
+              ((featurep! :completion ivy) 'ivy)
+              ((featurep! :completion ido) 'ido)
+              ('default))
+        org-roam-completion-fuzzy-match
+        (or (featurep! :completion helm +fuzzy)
+            (featurep! :completion ivy +fuzzy)))
 
   ;; HACK Hide the mode line in the org-roam buffer, since it serves no purpose.
   ;;      This makes it easier to distinguish among other org buffers.


hlissner:
Could you expand on how it fails? Any error messages (or warnings in \*Messages\*)?

hlissner:
Perhaps `(setq org-roam-completion-fuzzy-match t)` is needed?

bdarcus:
I will copy-and-paste this from another user. It doesn't fail; it just reverts to something other than fuzzy completion.

> org-roam-completion-system in Doom is automatically set to either ivy or helm instead of default , with helm apparently taking precedence over ivy if both are installed (at least for me). Fuzzy searching in org-roam appears to work with helm, but fail with ivy (possibly reverting to ido?). If I manually run (setq org-roam-completion-system 'default after everything is loaded (without helm installed), then org-roam-find-file and org-roam-insert both correctly use ivy with fuzzy searching. Everything works as I would expect it to. So far I have to run this manually, as simply adding it to config.el results in org-roam-completion-system using ivy instead of default. I'm assuming there is a way to defer setting it's value until something else is loaded, but I'm not sure how to implement that.
I have also tried (setq org-roam-completion-fuzzy-match t), with org-roam-completion-system set to ivy, but this does not match ivy's behavior. For example, searching 'b t' doesn't match 'bib_test', but searching 'bt' (without the space between letters) does. I'm sure this is otherwise expected behavior, but I'm leaving it here as a note.

hlissner:
> For example, searching 'b t' doesn't match 'bib_test', but searching 'bt' (without the space between letters) does.

I'm confused. This is correct behavior for ivy's fuzzy search algorithm. Spaces are not magical characters. `b t` is also searching for candidates with a space in them (it won't match `bib_test`).

alan-carroll:
I don't know if 'fuzzy' is the correct terminology, but 'b t' narrows the search to include 'bib_test'. I can rapidly narrow the search for a long file name by typing single characters with spaces between them

bdarcus:
The example I see is if I'm wanting to narrow to "org academic" and I did `org-roam-find-file` and then "o r" it would not match.

It does with this change though.

hlissner:
This appears to work for me:

```elisp
(setq org-roam-completion-system
      (cond ((featurep! :completion helm) 'helm)
            ((featurep! :completion ivy) 'ivy)
            ((featurep! :completion ido) 'ido)
            ('default))
      org-roam-completion-fuzzy-match
      (or (featurep! :completion helm +fuzzy)
          (featurep! :completion ivy +fuzzy))
```

hlissner:
Or perhaps it would be better if `org-roam-completion--completing-read` didn't [specify its own `:re-builder`](https://github.com/jethrokuan/org-roam/blob/master/org-roam-completion.el#L86-L87)? This overrides user customizations.

bdarcus:
I committed your suggestion.

Should we instead, or also, offer your suggestion to org-roam?

alan-carroll:
> This appears to work for me:
> 
> ```emacs-lisp
> (setq org-roam-completion-system
>       (cond ((featurep! :completion helm) 'helm)
>             ((featurep! :completion ivy) 'ivy)
>             ((featurep! :completion ido) 'ido)
>             ('default))
>       org-roam-completion-fuzzy-match
>       (or (featurep! :completion helm +fuzzy)
>           (featurep! :completion ivy +fuzzy))
> ```

This should work if you are using helm, but the problem with ivy appears unrelated to the conditional setup. With ivy installed, `(setq org-roam-completion-system 'default)` gives expected search behavior, but `(setq org-roam-completion-system 'ivy)` does not, regardless of whether `org-roam-completion-fuzzy-match` is set. 

I don't understand the internal workings enough to contribute to how the value should be set automatically, simply that I must currently manually set it to default after everything has loaded. The issue with ivy may be in org-roam

bdarcus:
I did confirm it works for me with ivy.

hlissner:
Looks good to me! Thanks for bringing it to my attention. I'll keep an eye on jethrokuan/org-roam#514.

PR #2961, opened 2020-04-24
tmux: Quoting shell arg in listing windows / panes

tylerware:
Specifying a session and / or window to +tmux-list-windowsand +tmux-list-panes just failed, because of the special characters used in the ids for sessions & windows ($ and @). Quoting the ids before building up the tmux command fixes this.

Specifically, the following don't work before this fix (you'll need to start a tmux session first):
```emacs-lisp
(+tmux-list-windows (car (+tmux-list-sessions)))
(+tmux-list-panes (car (+tmux-list-sessions)))
```


hlissner:
Makes sense to me. Thanks!

PR #2962, opened 2020-04-24
yaml: hook -> mode

peterhoeg:
There was a typo in the `use-package!` definition.

Requesting a pull to hlissner:develop from peterhoeg:fix_yaml

Write a message for this pull request. The first block
of text is the title and the rest is the description.


hlissner:
Good catch! Thanks!

PR #2970, opened 2020-04-25
Fix typo in key bindings

ineu:
This fixes a typo for a minibuffer keybinding.

@hlissner `*-char` and `*-word` commands in this section are generally mixed up, please let me know if I should include the fix to them as well.

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -22,7 +22,7 @@
     (define-key! :keymaps +default-minibuffer-maps
       "C-f"    #'forward-word
       "C-b"    #'backward-word
-      "M-f"    #'foward-char
+      "M-f"    #'forward-char
       "M-b"    #'backward-char
       "C-j"    #'next-line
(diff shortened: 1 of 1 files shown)


hlissner:
Good catch! Thanks for the fix! 

> *-char and *-word commands in this section are generally mixed up, please let me know if I should include the fix to them as well.

Actually, those swaps were intentional because most evil motions utilize the ctrl key far more than meta, so having these two stand out was a bit confusing, especially when they get used more than their single-character counterparts. What do you think? Is that too jarring a departure from their original functions?

ineu:
Well, that's a bit unexpected. I never use the arrow keys, so c-f/m-f/c-b/m-b is a kind of a muscle memory. And in case of the minibuffer I don't switch to command mode (because esc simply closes the minibuffer), so using good old emacs keys. And when you first try m-f, that's fine because you can reach any position by walking single char at a time. But when you try c-f and it jumps between words, it's not really clear how to move to a char

PR #2975, opened 2020-04-25
Update README badge

gagbo:
latest tag on emacs-27 is 27.0.91


hlissner:
I had an unpushed redesign of the readme I decided to push today, which includes an update to the badge (or at least generalizes it to say `27.x`). Thanks for the PR in any case!

PR #2977, opened 2020-04-25
Add .jinja2 file extension to supported web files.

amrsoll:
Seeing that the `web` lang module already supports .jinja files, I don't think it's a far cry to add the .jinja2 extension to the list.
It's not super fancy but better than no major mode support.
Working with limited lisp experience to expand the highlighting functionality. :shrug:

### DIFF
--- a/modules/lang/web/+html.el
+++ b/modules/lang/web/+html.el
@@ -13,6 +13,7 @@
   :mode "\\.vue\\'"
   :mode "\\.twig\\'"
   :mode "\\.jinja\\'"
+  :mode "\\.jinja2\\'"
   :mode "wp-content/themes/.+/.+\\.php\\'"
   :mode "templates/.+\\.php\\'"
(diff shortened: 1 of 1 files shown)


modules/lang/web/+html.el
```diff
@@ -13,6 +13,7 @@
   :mode "\\.vue\\'"
   :mode "\\.twig\\'"
   :mode "\\.jinja\\'"
+  :mode "\\.jinja2\\'"
```
hlissner, line 16: It'd be better if you change the first jinja regexp instead of adding a whole new one:

```diff
- :mode "\\.jinja\\'"
+ :mode "\\.jinja2?\\'"
```

amrsoll: Totally correct, and thank you very much for this fine piece of software :) I just started using it myself!

hlissner:
Thanks!

PR #2980, opened 2020-04-26
Oooops, I did a boo-boo

tecosaur:
Abbrev-mode betrayed me :( for that you get two commits for the price of one.

<kbd>SPC h for</kbd> → <kbd>SPC h f</kbd>


hlissner:
Not sure we can ever recover from this, but we must try!

PR #2981, opened 2020-04-26
Set `evil-shift-width` to `standard-indent` instead of `tab-width`

midchildan:
`tab-width` represents the number of spaces that a `Tab` character counts for, while `standard-indent` represents the number of spaces required for each level of indentation. Since `shiftwidth` is solely concerned with indentation, it'd make more sense to assign it to `standard-indent`. This change would benefit people who have different settings for `tab-width` and `standard-indent`.


midchildan:
I've now made a change to set `standard-indent` to `4` to retain previous behavior.

midchildan:
I'm closing this because I figured out that `tab-width` is used extensively across numerous major modes to configure indentation.

PR #2983, opened 2020-04-26
Make rg --path-separator argument work under Windows

guraltsev:
Under windows the formatting of the ```--path-separator``` argument was wrong.


hlissner:
Thanks!

PR #2984, opened 2020-04-27
Fix getting started link in README

ANGkeith:
- Fixed typo
- Replace Screenshot (thumbnail?) with Getting started
- Also removed the Screenshots link since it is broken and the
screenshot is just below it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # Doom Emacs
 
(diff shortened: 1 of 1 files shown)


hlissner:
Hi! Thanks for the PR but I cannot accept it. I'd rather the screenshots link be fixed, rather than removed. Also, linking the whole phrase "read our getting started guide" was intentional. I do not want to make the link a smaller target.

PR #2985, opened 2020-04-27
Update ivy readme

ANGkeith:


hlissner:
Hi! Thanks for the PR, but I cannot accept it. The `+fuzzy` flag is already documented in the "Module Flags" section. Perhaps that can be improved, but in any case, a dedicated section would be redundant.

PR #2987, opened 2020-04-27
org +roam: Prefer org-roam-graph

ocharles:
org-roam-graph-show is deprecated


hlissner:
Thanks for the PR. 019e988 should update these references. Thanks for bringing it to my attention!

Issue #2072, opened 2019-11-18
[BUG] Langtool can produce JNI errors unless classpath is set


chaorace:
**Describe the issue**
In some (all?) configurations, invoking langtool will fail with a JNI error.
(The error is temporarily displayed by the langtool package and does not throw a lisp error.)

I am able to resolve the issue by manually specifying the Java Classpath using the variable langtool-java-classpath variable, [as described here](https://www.reddit.com/r/emacs/comments/2zqetc/languagetool_exited_abnormally_with_code_1/ctapph0/).
It seems that the Java Classpath needs to be explicitly specified in some (possibly all?) cases.

In any case, we should really be explicitly doing so in all cases where we explicitly specify a jarfile, in order to avoid the possibility of this problem.

**Steps to reproduce**
- Install langtool (my particular version is the Arch official languagetool package)
- Configure +write module with langtool
- Invoke `langtool-check`
- Experience error

**System information**
<details><pre>
emacs   version    26.3
        features   XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS GLIB NOTIFY ACL GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS LIBSYSTEMD LCMS2
        build      Aug 29, 2019
        buildopts  (--prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib --localstatedir=/var --with-x-toolkit=gtk3 --with-xft --with-modules 'CFLAGS=-march=x86-64 -mtune=generic -O2 -pipe -fno-plt' CPPFLAGS=-D_FORTIFY_SOURCE=2 LDFLAGS=-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now)
        windowsys  x
        daemonp    daemon
doom    version    2.0.9
        build      HEAD -> develop ab0a90cd2 2019-11-15 02:17:08 -0500
system  type       gnu/linux
        config     x86_64-pc-linux-gnu
        shell      /bin/bash
        uname      Linux 4.14.151-1-MANJARO #1 SMP PREEMPT Tue Oct 29 14:36:16 UTC 2019 x86_64
[...]

PR #2994, opened 2020-04-28
Fix langtool config for Linux

jsravn:
Fixes https://github.com/hlissner/doom-emacs/issues/2072.

I have left the mac portion alone as I don't have a mac handy to test this.


hlissner:
Thanks!

PR #3002, opened 2020-04-29
Remove `lsp-groovy-server-install-dir`

sei40kr:
Remove `lsp-groovy-server-install-dir` setting since it no longer exists.

https://github.com/emacs-lsp/lsp-mode/pull/1319


hlissner:
Thanks!

PR #3003, opened 2020-04-29
Improve ivy-rich with counsel-bookmark

rgrinberg:
This adds a column that displays the path where the bookmark exists.

### DIFF
--- a/modules/completion/ivy/config.el
+++ b/modules/completion/ivy/config.el
@@ -129,7 +129,11 @@ evil-ex-specific constructs, so we disable it solely in evil-ex."
                  (ivy-rich-counsel-function-docstring (:face font-lock-doc-face))))
               ;; Apply switch buffer transformers to `counsel-projectile-switch-to-buffer' as well
               'counsel-projectile-switch-to-buffer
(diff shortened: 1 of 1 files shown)


hlissner:
Sounds good to be, but does it handle remote bookmarks? Or fail gracefully, at least?

rgrinberg:
There's `info` which is a bit more robust:

```emacs-lisp
(defun ivy-rich-bookmark-info (candidate)
  (let ((filename (ivy-rich-bookmark-filename candidate)))
    (cond (filename
           (cond ((null filename)
                  "")
                 ((file-remote-p filename)
                  candidate)
                 ((file-exists-p filename)
                  (file-truename filename))
                 (t filename))))))
```

But it seems like all these sys calls can potentially reduce responsiveness. I feel like if the filename is null, I wouldn't care if it's not displayed. If you prefer, feel free to use this version and I'll customize things for my personal config.

hlissner:
I think `ivy-rich-bookmark-filename` is good enough. I just didn't want it throwing an unrecoverable error if the filename didn't exist. Thanks for the PR!

PR #3006, opened 2020-04-30
Switching workspaces: Display list of open buffers with ivy-rich preview

rgrinberg:
This makes it simpler to immediately pick the desired workspace

The formatting of the buffers is a bit spartan. If the idea is approved in
principle, we can always improve the display.


hlissner:
Sorry for the _tremendously_ late response. I've updated and merged this in 086e9a9. Thanks for the PR!

PR #3011, opened 2020-04-30
Add google-java-format support to Java

sei40kr:
## Description

Add `+google-java-format` flag to `lang/java` module.
https://github.com/google/google-java-format

We can also use Google style in `clang-format` by configuring `.clang-format`, but it doesn't actually obey Google Java Style Guide.

### DIFF
--- a/modules/lang/java/README.org
+++ b/modules/lang/java/README.org
@@ -25,6 +25,7 @@ This module adds [[https://www.java.com][java]] support to Doom Emacs, including
 ** Module Flags
 + =+lsp= Enables integration for the eclipse.jdt.ls LSP server.
 + =+meghanada= Enables the [[https://github.com/mopemope/meghanada-emacs/tree/master][meghanada-mode]]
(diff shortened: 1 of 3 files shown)


sei40kr:
We can also obey Google Java Style Guide by setting `lsp-java-format-settings-url` to `"http://google.github.io/styleguide/eclipse-java-google-style.xml"`.
This probably works faster, but less opinionated.

oscarSantoyo (approved):

hlissner:
Sorry for the tremendously late response. I do want to add google-java-format support to the Java module, but some work is still needed; mainly to accommodate the new :editor format module that #6369 will introduce. However, because I'm prepping to move our modules [to a separate repo](https://github.com/doomemacs/modules) (and because this PR targets the defunct `develop` branch) I'm going to close it for now. That said, I'm not rejecting the PR, but Github gives me no easy way to transfer it to another repo; once the smoke has cleared I will revisit this. Thanks for bringing it to my attention, in any case!

PR #3016, opened 2020-05-01
Add a few missing modes to `hdm` alist

gagbo:
Rule was mostly to add major-modes defined by packages bundled in the module


hlissner:
Thanks!

PR #3020, opened 2020-05-01
Add support for eglot as LSP client implementation

gagbo:
# Choose eglot over lsp-mode in Doom

## Features

- [x] Add a `+eglot` flag to `:tools lsp` which will install [eglot](https://github.com/joaotavora/eglot) over [lsp-mode](https://github.com/emacs-lsp/lsp-mode).
- [x] Automatically switch all lsp client configuration in `:tools lang` modules to use `eglot`
- [x] Provide `(set-eglot-client!)` function which is the thinnest wrapper around `eglot-server-programs` variable (the wrapper only exists to use the `;;;###autodef` feature)
- [x] Add a few language specific settings / functions where available and provided by language server maintainers.
- [X] Provide a `:documentation` lookup-handler that displays the "Hover" information in a pop-up buffer
- [x] Add `doom doctor` warnings if `:tools lsp` has `+eglot` and basically any other lsp flag
- [x] Add `doom doctor` error if `:tools (lsp +eglot)` and `:tools (debugger +lsp)` are active (`dap-mode` has a hard `lsp-mode` dependency)
- [x] Handle `^\\*eglot-help*` buffers properly
- [x] ~~Make `flycheck` and `flymake` live together in peace (eglot only uses Flymake, while for non-lsp buffers users might want flycheck checkers)~~ CANCELLED because we can just make eglot provide a flycheck backend if `:checkers syntax` is enabled.

### Microsoft's Python Language Server
Using eglot means giving up the `lsp-python-ms` package that installs "mspyls" (`Microsoft.Python.LanguageServer`) for you. Users have to build the server themselves from [source](https://github.com/microsoft/python-language-server) or find a location with the binaries that they trust. For reference, `lsp-python-ms` uses `f"https://pvsc.blob.core.windows.net/python-language-server-{'stable' or 'beta' or 'daily'}?restype=container&comp=list&prefix=Python-Language-Server-{'osx' or 'linux' or 'win'}-x64"` to get directly built and updated binaries 

(For the record, eglot being on ELPA and closer to being shipped with emacs, I assume that providing a downloader for microsoft software is a huge nono from a legal standpoint -- they want to distribute only free software after all)

### CCLS
The initialization options specific to MacOS have not been tested at all for eglot for the time being.

### Rust Language Server vs. Rust Analyzer
Rustic handles all eglot-specific vs. lsp-mode-specific configuration for us (thanks a lot !), only thing Doom user should care about is to set `rustic-lsp-client` and `rustic-lsp-server` variables properly

## Differences between the 2 lsp clients ?

*In my opinion*, [eglot](https://github.com/joaotavora/eglot) is :
- more barebones (not the flurry of additional packages like `lsp-ui-*` to check for ; but less packages that wrap configuration around specific lsp servers, no `dap-mode`...)
- closer to historic emacs packages for integration (eldoc, flymake, xref)
- closer to be upstreamed and shipped with emacs (since it's already on ELPA)

This is *only my opinion*, I state it here because I can already imagine everyone asking "Why should I use this over that ?", and I don't want to trigger endless discussions. You can go to [this issue](https://github.com/joaotavora/eglot/issues/180) for more information.


gagbo:
Only issue left is with popup handling, I have no idea for the time being. But this is solvable later I suppose (or with help from others)

gagbo:
The code from `flycheck-eglot.el` is likely to be upstreamed looking at the conversation at flycheck/flycheck#1592 . This is worthwile to monitor until we can just add a `package!` statement for an eglot-backed flycheck checker plugin

Emiller88:
Confirmed this seems to be working with js, in case you still hadn't tested that!

Vue doesn't work out of the box though, this is what I added for lsp-mode, do you have a guess what it would be for eglot?
```lisp
(add-hook 'web-mode-local-vars-hook #'lsp!)
```

gagbo:
It should work the same, `lsp!` has [been modified](https://github.com/hlissner/doom-emacs/pull/3020/files#diff-1885e519d58436eecb7a94984b49efcc) to start eglot instead of lsp-mode if `(lsp +eglot)` has been chosen.

2 possibilities :
- somehow the autoloads didn't correctly update : maybe `rm -f ~/.emacs.d/.local/autoload.el && doom sync` might correct that 
- Or lsp-mode started automagically before and we forgot some config on `web-mode` when `+lsp` is active on `:lang javascript`

Thanks for the feedback, I'll check it !

Emiller88:
Ah this one is my fault for not checking, eglot doesn't have built in support for `vls`

It will need something like 
```lisp
(add-to-list 'eglot-server-programs '(web-mode . ("vls")))
```
But that doesn't work out of the box either, so I'll mess around with it. Thanks thought!

gagbo:
I'm not adding yet support for #3094  in eglot : it may need writing more code, for a lookup-handler without any key binding yet.

Eglot implementations of these extra lookups are already added in `{+emacs,+evil}-bindings.el` under `<leader>c` prefix

modules/config/default/+emacs-bindings.el
```diff
@@ -41,7 +41,7 @@
        :desc "List errors"                           "x"   #'flymake-show-diagnostics-buffer
        (:when (featurep! :checkers syntax)
         :desc "List errors"                         "x"   #'flycheck-list-errors)
-       (:when (featurep! :tools lsp)
+       (:when (and (featurep! :tools lsp) (not (featurep! :tools lsp + eglot)))
```
rushsteve1, line 44: `(not (featurep! :tools lsp + eglot))` is there supposed to be a space or is it meant to be `:tools lsp +eglot`?

modules/lang/python/config.el
```diff
@@ -22,7 +22,10 @@ called.")
         python-indent-guess-indent-offset-verbose nil)
 
   (when (featurep! +lsp)
-    (add-hook 'python-mode-local-vars-hook #'lsp!))
+    (add-hook 'python-mode-local-vars-hook #'lsp!)
+    ;; Use "mspyls" in eglot if in PATH
+    (when (executable-find "Microsoft.Python.LanguageServer")
+      (set-eglot-client! 'python-mode '("Microsoft.Python.LanguageServer"))))
```
rushsteve1, line 28: Should this be wrapped in a `(when (featurep! :tools lsp +eglot) ...)`?

modules/tools/lsp/flycheck-eglot.el
```diff
@@ -0,0 +1,57 @@
...
+             ('eglot-warning 'warning)
+             ('eglot-error 'error)
+             (_ (error "Unknown diagnostic type, %S" diag)))
+           (flymake--diag-text diag)
+           :end-pos (flymake--diag-end diag)
+           :checker checker
+           :buffer (current-buffer)
+           :filename (buffer-file-name)))))
+    ;; NOTE: Setting up eglot to automatically create flycheck errors for the buffer.
+    (eglot-flymake-backend (lambda (flymake-diags &rest _)
```
rushsteve1, line 28: Splitting these two across a few more lines would be easier to read.

modules/config/default/+emacs-bindings.el
gagbo: oh good catch

modules/lang/python/config.el
gagbo: `set-eglot-client!` will no-op automagically if the condition is not there. the `;;;###autodef` cookie on the function and the "file cookie" in the file that defines the function will create that. This way we don't need to litter the code with feature checks

modules/tools/lsp/packages.el
```diff
@@ -1,9 +1,12 @@
...
-  (package! helm-lsp :pin "6b5ce182d7c94c62b55b8f7d0c7e643b2c30e560"))
+(if (featurep! +eglot)
+    (package! eglot :pin "d99a4478a9")
+  (progn
+    (package! lsp-mode :pin "81d62d581b21d847783831e6e5ca9d3c63fe9a4d")
+    (package! lsp-ui :pin "271b47cb33f11915295911f7cf8575f8a82a5e1c")
+    (when (featurep! :completion ivy)
+      (package! lsp-ivy :pin "dce58b5509271bbedb53ba9d0278dcb563a43977"))
+    (when (featurep! :completion helm)
+      (package! helm-lsp :pin "6b5ce182d7c94c62b55b8f7d0c7e643b2c30e560"))))
```
hlissner, line 12: The `progn` is unnecessary. Please remove it.

modules/tools/lsp/config.el
```diff
@@ -7,11 +7,34 @@ workspace buffer is closed.
...
+        eglot-send-changes-idle-time 0.5
+        ;; NOTE: Do NOT set eglot-auto-display-help-buffer to t.
+        ;; With popup-rule! :select t, eglot will steal focus from the source code very often.
+        eglot-auto-display-help-buffer nil)
+  :config
+  (set-popup-rule! "^\\*eglot-help" :size 0.35 :quit t :select t)
+  (when (featurep! :checkers syntax)
+    ;; Eager loading which is okay-ish since we want eglot to feed flycheck as soon as possible.
+    (load! "flycheck-eglot.el")
+    (require 'flycheck-eglot))
```
hlissner, line 30: Two requires of the same file is redundant. Get rid of the `require` statement.

hlissner: Also, this should probably be wrapped in an `(after! flycheck ...)` rather than eagerly loading flycheck together with elgot.

modules/tools/lsp/flycheck-eglot.el
```diff
@@ -0,0 +1,60 @@
+;;; flycheck-eglot --- Hacky eglot support in flycheck -*- lexical-binding: t; -*-
```
hlissner, line 1: Please move this file to autoloads/flycheck-eglot.el, and remove the `(provide 'flycheck-eglot)` and `(require 'flycheck)` lines.

modules/tools/lsp/flycheck-eglot.el
```diff
@@ -0,0 +1,60 @@
...
+
+(flycheck-define-generic-checker 'eglot
+  "Report `eglot' diagnostics using `flycheck'."
+  :start #'flycheck-eglot--start
+  :predicate #'flycheck-eglot--available-p
+  :modes '(prog-mode text-mode))
+
+(push 'eglot flycheck-checkers)
+
+(defun +doom/eglot-prefer-flycheck-h ()
```
hlissner, line 47: This function should be named `+doom-eglot-prefer-flychech-h`. `+doom/...` is a convention for interactive commands, which this is not.

modules/tools/lsp/config.el
```diff
@@ -7,11 +7,34 @@ workspace buffer is closed.
...
+  (set-lookup-handlers! 'eglot--managed-mode
+    :documentation #'+eglot/documentation-lookup-handler
+    :definition '(xref-find-definitions :async t)
+    :references '(xref-find-references :async t)))
```
hlissner, line 34: These two xref handlers shouldn't be necessary. xref is already top of the list by default

modules/lang/cc/config.el
```diff
@@ -236,9 +236,45 @@ If rtags or rdm aren't available, fail silently instead of throwing a breaking e
       (setq-local company-lsp-cache-candidates nil)
       (lsp!))))
 
+(when (and (featurep! +lsp) (featurep! :tools lsp +eglot))
+  ;; TODO : test this value
```
hlissner, line 240: This whole form should be wrapped in an `(after! eglot ...)`

modules/lang/cc/config.el
```diff
@@ -236,9 +236,45 @@ If rtags or rdm aren't available, fail silently instead of throwing a breaking e
...
+  ;; IS-MAC custom configuration
+  (when IS-MAC
+    (add-to-list 'eglot-workspace-configuration
+                 ((:ccls . ((:clang . ,(list :extraArgs ["-isystem/Library/Developer/CommandLineTools/usr/include/c++/v1"
+                                      "-isystem/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include"
+                                      "-isystem/usr/local/include"]
+                          :resourceDir (string-trim (shell-command-to-string "clang -print-resource-dir")))))))))
+  ;; Eglot specific helper, courtesy of MaskRay
+  (defun eglot-ccls-inheritance-hierarchy (&optional derived)
```
hlissner, line 249: Where is `eglot-ccls-inheritance-hierarchy` used? If it is trying to overwrite a defined function this should be advice.

modules/config/default/+emacs-bindings.el
```diff
@@ -52,7 +52,13 @@
...
         (:when (featurep! :completion helm)
          :desc "Jump to symbol in current workspace" "j"   #'helm-lsp-workspace-symbol
-         :desc "Jump to symbol in any workspace"     "J"   #'helm-lsp-global-workspace-symbol)))
+         :desc "Jump to symbol in any workspace"     "J"   #'helm-lsp-global-workspace-symbol))
+       (:when (featurep! :tools lsp +eglot)
+        :desc "LSP Execute code action"              "a" #'eglot-code-actions
+        :desc "LSP Format buffer/region"             "F" #'eglot-format
+        :desc "LSP Rename"                           "r" #'eglot-rename
+        :desc "LSP Find declaration"                 "j" #'eglot-find-declaration
+        :desc "LSP Find implementation"              "J" #' eglot-find-implementation))
```
hlissner, line 61: Remove the space between `#'` and `eglot-find-implementation`

modules/tools/lsp/packages.el
gagbo: That explains why the indentations of `if` is the way it is

modules/tools/lsp/flycheck-eglot.el
gagbo: When / how will the file be loaded then ? `(load! "flycheck-eglot.el")` will still work ?

hlissner: It should be changed to `(load! "autoload/flycheck-eglot")`, but yes, that'll work.

modules/lang/cc/config.el
gagbo: It's just the definition of an interactive function to use eglot to display type-inheritance information (this kind of function is included in `emacs-ccls` but that uses `lsp-mode` API)

modules/tools/lsp/config.el
```diff
@@ -7,11 +7,34 @@ workspace buffer is closed.
 This delay prevents premature server shutdown when a user still intends on
 working on that project after closing the last buffer.")
 
+;; TODO : set eglot-events-buffer-size to nil in doom-debug-mode
 
 ;;
 ;;; Packages
 
+(use-package! eglot
+  :when (featurep! +eglot)
```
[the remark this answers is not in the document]
gagbo: used `:commands (eglot-ensure eglot)`

modules/lang/cc/config.el
gagbo: I'll move the interactive function in autoload somewhere at least.

gagbo: > I'll move the interactive function in autoload somewhere at least.

Oh I remember why I didn't do this : there's no eglot specific autoloads in cc module. So I'm not sure about what to do

gagbo:
Hold, deferring eglot means some (defvar) are needed now when manipulating eglot-server-programs variables in user config

modules/tools/lsp/config.el
```diff
@@ -7,188 +7,7 @@ workspace buffer is closed.
...
-  :commands helm-lsp-workspace-symbol helm-lsp-global-workspace-symbol)
-
-
-(use-package! lsp-ivy
-  :when (featurep! :completion ivy)
-  :commands lsp-ivy-workspace-symbol lsp-ivy-global-workspace-symbol)
+;; TODO : set eglot-events-buffer-size to nil in doom-debug-mode
+(if (featurep! +eglot)
+    (load! "+eglot.el")
+  (load! "+lsp.el"))
```
hlissner, line 13: Please omit the file extension. Otherwise, if we were ever to byte-compile these it'd never load the *.elc files.

modules/tools/lsp/+eglot.el
```diff
@@ -0,0 +1,19 @@
...
+        eglot-autoshutdown t
+        eglot-send-changes-idle-time 0.5
+        ;; NOTE: Do NOT set eglot-auto-display-help-buffer to t.
+        ;; With popup-rule! :select t, eglot will steal focus from the source code very often.
+        eglot-auto-display-help-buffer nil)
+  :config
+  (set-popup-rule! "^\\*eglot-help" :size 0.35 :quit t :select t)
+  (when (featurep! :checkers syntax)
+    (after! flycheck
+      (load! "autoload/flycheck-eglot.el")))
```
hlissner, line 17: Please remove the file extension (see next comment for explanation).

modules/lang/cc/autoload.el
```diff
@@ -122,6 +122,34 @@ simpler."
        #'rtags-imenu
      #'imenu)))
 
+;; Eglot specific helper, courtesy of MaskRay
+;;;###autoload
+(defun eglot-ccls-inheritance-hierarchy (&optional derived)
```
hlissner, line 127: Could you rename this to +cc/eglot-ccls-inheritance-hierarchy? That's Doom's convention for interactive commands.

hlissner:
I think this is good to go! I look forward to using eglot, and would like to make it Doom's default at some point in the future. Thanks for your help and patience!

sei40kr:
@hlissner @gagbo 
I got an error on macOS.

https://github.com/hlissner/doom-emacs/blob/develop/modules/lang/cc/config.el#L251
I think ` is needed in this line.

hlissner:
@sei40kr Fixed in 200fd52. Thanks for letting me know!

bittopaz:
Thank you very much for getting this out, I have been looking forwarding to Eglot integration for a long time. @gagbo

PR #3021, opened 2020-05-01
add support for magit-delta

nitper:
This adds colorful syntax-highlighted diffs in magit-status buffers, thanks to [delta](https://github.com/dandavison/delta) and [magit-delta](https://github.com/dandavison/magit-delta)

This adds a `+delta` flag to the magit module.

### DIFF
--- a/modules/tools/magit/packages.el
+++ b/modules/tools/magit/packages.el
@@ -8,7 +8,7 @@
     (package! xterm-color :pin "d53a39a5af72cd340ebf686e59a37289b4cb6e8c")
     (package! magit-delta
       :recipe (:host github :repo "dandavison/magit-delta" :files ("magit-delta.el") )
-      :pin "a3081b030c1bc21a1c0ec4367a8b08a2a5f031f3"))
+      :pin "f429bf64bb398890b7c584d70660dd202e39824a"))
   (package! magit-gitflow :pin "cc41b561ec6eea947fe9a176349fb4f771ed865b")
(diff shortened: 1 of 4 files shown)


modules/tools/magit/packages.el
```diff
@@ -4,6 +4,11 @@
 (when (package! magit :pin "b1b2683f6012ff3bc29bd9cbe562246477f1523f")
   (when (featurep! +forge)
     (package! forge :pin "09bf8adc9c9afb492632e612f51f39e1cc15fca0"))
+  (when (featurep! +delta)
+    (package! xterm-color :pin "d53a39a5af72cd340ebf686e59a37289b4cb6e8c")
+    (package! magit-delta
+      :recipe (:host github :repo "dandavison/magit-delta" :files ("magit-delta.el") )
+      :pin "f429bf64bb398890b7c584d70660dd202e39824a"))
```
sei40kr, line 11: I found magit-delta on MELPA:
https://melpa.org/#/magit-delta

nitper: Looks like it was added yesterday - I just updated this PR to grab it from there.

hlissner:
Thanks for the PR, but I must refuse it. `magit-delta` is too niche to be included by default with the magit module, and since its config is relatively simple, I'll leave it to users to configure themselves. Thanks for bringing it to my attention in any case!

elmarco:
too bad, I was about to propose the same.

Imho it's not so much niche, and with doom, one cool thing lead to another. Even better would be to enable it by default imho, and have doom doctor check its presence. Anyways, my 2c

PR #3022, opened 2020-05-02
Encrypt and change location of org-gcal-token-file

darth10:
Looks like `org-gcal` stores sensitive data in `~/.emacs.d/org-gcal` (by default) without encryption. 
Here's a few changes to make it a bit more secure. :lock: 

* Move `org-gcal-token-file` to `doom-cache-dir`
* Encrypt `org-gcal-token-file` using GPG
* If old file exists, move it to new location with encryption
* Bump `org-gcal` pin

Any feedback is more than welcome!

### DIFF
--- a/modules/app/calendar/config.el
+++ b/modules/app/calendar/config.el
@@ -49,5 +49,3 @@
   ;; hack to avoid the deferred.el error
   (defun org-gcal--notify (title mes)
     (message "org-gcal::%s - %s" title mes)))
-
-;; (use-package! alert)
--- a/modules/app/calendar/config.el
+++ b/modules/app/calendar/config.el
@@ -42,10 +42,12 @@
              org-gcal-fetch
              org-gcal-post-at-point
              org-gcal-delete-at-point)
+  :init
+  (defvar org-gcal-dir (concat doom-cache-dir "org-gcal/"))
+  (defvar org-gcal-token-file (concat org-gcal-dir "token.gpg"))
   :config
   ;; hack to avoid the deferred.el error
   (defun org-gcal--notify (title mes)
     (message "org-gcal::%s - %s" title mes)))
-
 
 ;; (use-package! alert)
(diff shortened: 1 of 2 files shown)


modules/app/calendar/config.el
```diff
@@ -43,9 +43,15 @@
...
   :config
+  (setq org-gcal--orig-token-file org-gcal-token-file
+        org-gcal-token-file
+        (expand-file-name "org-gcal-token.gpg" doom-cache-dir))
+  ;; Move and encrypt `org-gcal--orig-token-file' to `org-gcal-token-file'.
+  (when (and (file-exists-p org-gcal--orig-token-file)
+             (not (file-exists-p org-gcal-token-file)))
+    (with-temp-file org-gcal-token-file
+      (insert-file-contents org-gcal--orig-token-file))
+    (delete-file org-gcal--orig-token-file))
```
hlissner, line 54: Moving the old file is problematic. 1. If they don't have gpg set up, the move will break. 2. We're breaking our "your system your rules" mantra by changing the user's filesystem beyond `~/.emacs.d/.local` and `~/.doom.d`. While certainly helpful, Doom's responsibility is limited to preventing Emacs and plugins from doing so (as this PR does by changing `org-gcal-token-file`).

In which case, I think it'd be better to simplify this to:

```diff
-  :config
-  (setq org-gcal--orig-token-file org-gcal-token-file
-        org-gcal-token-file
-        (expand-file-name "org-gcal-token.gpg" doom-cache-dir))
-  ;; Move and encrypt `org-gcal--orig-token-file' to `org-gcal-token-file'.
-  (when (and (file-exists-p org-gcal--orig-token-file)
-             (not (file-exists-p org-gcal-token-file)))
-    (with-temp-file org-gcal-token-file
-      (insert-file-contents org-gcal--orig-token-file))
-    (delete-file org-gcal--orig-token-file))
+  :init
+  (defvar org-gcal-dir (concat doom-cache-dir "org-gcal/"))
+  (defvar org-gcal-token-file (concat org-gcal-dir "token.gpg"))
+  :config
```

This also safeguards us against any future files the package stores in `org-gcal-dir` and, like other directory/file variables, sets it early enough for users to change in their config before the package loads.

darth10: Okie dokie! That makes total sense to me.

I've made the changes as requested.

hlissner:
Great work. Thanks for the PR!

PR #3027, opened 2020-05-03
Fix emacs bindings to use :checkers

n-clark:
Replaced :tools flycheck and flyspell with :checkers syntax and
spell.


hlissner:
Ah, good catches. Thanks!

PR #3029, opened 2020-05-04
Fix tide not working with rjsx-mode

qhuyduong:
Tide doesn't work with `rjsx-mode` after changing to `*-mode-local-vars-hook`.
Fixed by adding `+javascript-init-lsp-or-tide-maybe-h` for
`rjsx-mode-local-vars-hook` as well.


hlissner:
Good catch. Thanks!

PR #3030, opened 2020-05-04
Use lsp-format-region-or-buffer if available

jsravn:
Currently the user has to select `SPC c F` for lsp format. But if
lsp-mode is enabled, the user will always want to use the LSP formatter.
So this changes the default `SPC c f` to use lsp formatter if available.
This simplifies the interface a bit so the user doesn't have to worry
about selecting the "right" format, just like how other commands get
auto bound to LSP functions.


hlissner:
Ah, looks like #2952 create a conflict with this PR. Could you rebase it onto `develop`?

hlissner:
Thanks!

PR #3032, opened 2020-05-04
ido: fix up and down keybinds in file dir navigation

belak:
This adds support for up/down/C-n/C-p for file dir navigation. For some reason the keybinds are overridden there and they don't do what I'd expect.


hlissner:
Makes sense to me. Thanks!

PR #3033, opened 2020-05-04
add config/literate readme

Amatrelan:
This is simple readme for config/literate but most of the things and quirks should be clear with this.

Part of #1166 issue.

### DIFF
--- a/modules/config/literate/README.org
+++ b/modules/config/literate/README.org
@@ -1,0 +1,43 @@
+#+TITLE:   config/literate
+#+DATE:    May 4, 2020
+#+SINCE:   v2.0.9
+#+STARTUP: inlineimages nofold
+
+* Table of Contents :TOC_3:noexport:
+- [[#description][Description]]
+  - [[#maintainers][Maintainers]]
+  - [[#module-flags][Module Flags]]
+  - [[#plugins][Plugins]]
+  - [[#hacks][Hacks]]
+- [[#prerequisites][Prerequisites]]
+- [[#features][Features]]
+- [[#configuration][Configuration]]
+- [[#troubleshooting][Troubleshooting]]
+
+* Description
+With this module you can organize your config file inside ~config.org~ file and export them automatically to ~config.el~.
+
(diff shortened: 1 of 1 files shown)


tecosaur:
It's probably worth noting that `$HOME/.config/doom/` is also valid.

modules/config/literate/README.org
```diff
@@ -0,0 +1,43 @@
...
+This module has no dedicated maintainers.
+
+** Module Flags
+This module provides no flags.
+
+** Plugins
++ ~ob-tangle~ (builtin)
+
+** Hacks
+This is only for personal config from ~$HOME/.doom.d/config.org~.
```
tecosaur, line 30: It's probably worth noting that `$HOME/.config/doom/` is also valid.

Amatrelan: That's true! Put it also to that

modules/config/literate/README.org
```diff
@@ -0,0 +1,43 @@
...
+This module has no dedicated maintainers.
+
+** Module Flags
+This module provides no flags.
+
+** Plugins
++ ~ob-tangle~ (builtin)
+
+** Hacks
+This is only for personal config from ~$HOME/.config/doom~ or ~$HOME/.doom.d/config.org~.
```
gagbo, line 30: Either none or both paths should have the `config.org` filename at the end of the path ; otherwise it's confusing, I can think that I might have my org file be just `~/.config/doom` and it'll tangle a file like that

modules/config/literate/README.org
```diff
@@ -0,0 +1,43 @@
...
+
+* Prerequisites
+This module has no prerequisites.
+
+* Features
++ Automatically literate ~config.org~ to ~config.el~ when saving.
```
gagbo, line 36: `s/literate/tangle` : using org-mode vocabulary here helps people familiar with org-mode be sure of what happens

hlissner:
Thanks! Tagging #1166 for posterity.

PR #3039, opened 2020-05-05
Make some improvements to java +lsp

jsravn:
This is a first pass at making `+lsp` more functional:

* Fix lsp not starting automatically when opening java-mode buffers.
Placing it in the lsp-java hook does not work.
* Fix root path for lsp-jt.

### DIFF
--- a/modules/lang/java/+lsp.el
+++ b/modules/lang/java/+lsp.el
@@ -9,17 +9,10 @@
 (add-hook! java-mode-local-vars #'lsp!)
 (map! :map java-mode-map
       :localleader
-      :desc "Update the Eclipse LSP server" "U" #'+java/update-lsp-server
-      :desc "Open the test browser" "T" #'lsp-jt-browser)
-
-;; TODO: This fails on the first java buffer opened - complains about the LSP server not having executeCode capabilities.
-(defun java--lsp-jt-lens-mode-hook ()
-  (when (derived-mode-p 'java-mode) (lsp-jt-lens-mode)))
+      :desc "Update the Eclipse LSP server" "U" #'+java/update-lsp-server)
 
 (use-package! lsp-java
   :after (lsp-clients dap-mode)
-  :hook (lsp-mode . java--lsp-jt-lens-mode-hook)
   :preface
   (setq lsp-java-server-install-dir (concat doom-etc-dir "eclipse.jdt.ls/server/")
-        lsp-java-workspace-dir (concat doom-etc-dir "java-workspace")
-        lsp-jt-root (concat doom-etc-dir "eclipse.jdt.ls/server/java-test/server")))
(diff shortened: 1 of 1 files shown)


jsravn:
I got some feedback from the lsp-java maintainer at https://github.com/emacs-lsp/lsp-java/issues/222#issuecomment-624061914. It looks like lsp-jt isn't recommended, and avy mode should be used for the code lens along w/ directly invoking dap-debug on test classes. I'll have a dig into that later.

modules/lang/java/+lsp.el
```diff
@@ -1,13 +1,18 @@
 ;;; lang/java/+lsp.el -*- lexical-binding: t; -*-
 ;;;###if (featurep! +lsp)
 
+;;;###autoload
+(defun +java/update-lsp-server ()
+  (interactive)
+  (lsp--install-server-internal (gethash 'jdtls lsp-clients) t))
```
hlissner, line 7: This should be in an autoloads file, but...

modules/lang/java/+lsp.el
```diff
@@ -1,13 +1,18 @@
...
+
+(add-hook! java-mode-local-vars #'lsp!)
+(map! :map java-mode-map
+      :localleader
+      :desc "Update the Eclipse LSP server" "U" #'+java/update-lsp-server)
```
hlissner, line 12: Please remove the `map!`. I don't think a update-lsp-server keybind is necessary. At least, it shouldn't be so hardcoded to a single client. A more general `+lsp/update-running-server` command in the `:tools lsp` module would be better.

What's more, I don't think it should be run so often that it warrants a keybind.

modules/lang/java/+lsp.el
```diff
@@ -1,13 +1,18 @@
...
+
 (use-package! lsp-java
-  :after lsp-clients
-  :hook (java-mode-local-vars . lsp!)
```
hlissner, line 6: The reason this did not work was due to how `use-package` processes hooks when an `:after` is involved. Please move the `add-hook` call into the use-package! block's `:preface`. I'd like to keep lsp-java's config inside it.

```elisp
(use-package! lsp-java
  :after lsp-clients
  :preface
  (setq lsp-java-server-install-dir (concat doom-etc-dir "eclipse.jdt.ls/server/")
        lsp-java-workspace-dir (concat doom-etc-dir "java-workspace"))

  (add-hook 'java-mode-local-vars-hook #'lsp!))
```

modules/lang/java/+lsp.el
jsravn: Makes sense.

jsravn:
Addressed the review comments, and also fix `lsp-jt-root` so java-test stuff works (in case someone wants to experiment with it, as it is in a mostly working state).

hlissner:
Thanks again for your help!

PR #3043, opened 2020-05-05
Add Tree Subtree function to bindings to lang org 

snuffop:
add insert and toggle to tables  which  we should probably move from b  as  b would better fit the babel functions.


hlissner:
Hi! Thanks for the PR. Would you mind rebasing your PR to resolve those conflicts?

> which we should probably move from b as b would better fit the babel functions

Where would you suggest? We're running out of mneumonic prefixes.

snuffop:
I "think" i have this correct now?   the PR shows no conflicts.

hlissner:
Thanks!

PR #3049, opened 2020-05-06
Configure java +lsp test runner via dap-mode

jsravn:
This configures dap-mode appropriately so the user can run tests
directly from Doom. It adds two test bindings similar to
other major modes:
* `SPC m t t` runs the test class or method at point.
* `SPC m t a` runs all tests in the current class regardless of point.
* `SPC m t d` debugs the test class or method at point using dap-mode.
* `SPC m t D` debugs all tests in the current class regardless of point.

I also expanded the README with more details about configuring `+lsp`.

### DIFF
--- a/modules/lang/java/+lsp.el
+++ b/modules/lang/java/+lsp.el
@@ -4,10 +4,14 @@
 (use-package! lsp-java
   :after lsp-clients
   :preface
-  (setq lsp-java-server-install-dir (concat doom-etc-dir "eclipse.jdt.ls/server/")
-        lsp-java-workspace-dir (concat doom-etc-dir "java-workspace")
-        lsp-jt-root (concat doom-etc-dir "eclipse.jdt.ls/server/java-test/server/"))
+  (setq lsp-java-workspace-dir (concat doom-etc-dir "java-workspace"))
   (add-hook! java-mode-local-vars #'lsp!)
-  :config
-  ;; TODO keybinds
-  )
+  :init
+  (when (featurep! :tools debugger +lsp)
+    (setq lsp-jt-root (concat lsp-java-server-install-dir "java-test/server/")
+          dap-java-test-runner (concat lsp-java-server-install-dir "test-runner/junit-platform-console-standalone.jar"))
+    (map! :map java-mode-map
+          :localleader
+          (:prefix ("t" . "Test")
+           :desc "Run tests in class" "t" #'dap-java-run-test-class
+           :desc "Run single test method" "s" #'dap-java-run-test-method))))
--- a/modules/lang/java/+lsp.el
+++ b/modules/lang/java/+lsp.el
@@ -18,8 +18,17 @@
           (dap-java-run-test-method)
         (user-error (dap-java-run-test-class))))
 
+    (defun +java/debug-test ()
+      "Runs test at point in a debugger. If in a method, runs the test method, otherwise runs the entire test class."
+      (interactive)
+      (condition-case nil
+          (call-interactively #'dap-java-debug-test-method)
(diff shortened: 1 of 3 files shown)


modules/lang/java/+lsp.el
```diff
@@ -4,10 +4,14 @@
 (use-package! lsp-java
   :after lsp-clients
   :preface
-  (setq lsp-java-server-install-dir (concat doom-etc-dir "eclipse.jdt.ls/server/")
```
jsravn, line 7: I realized this is redundant as doom already sets `lsp-server-install-dir`.

modules/lang/java/README.org
```diff
@@ -111,4 +133,16 @@ According to [[https://github.com/mopemope/meghanada-emacs/]] it adds
...
+** =+lsp=
+Install the eclipse server by executing =M-x lsp-install-server= and selecting
+=jdtls=. After that any newly opened =java= files should start the LSP server
+automatically.
+
+To update the server, perform =SPC u M-x lsp-install-server=.
+
+Note that if you change Java version you may need to remove the LSP server and
+reinstall it. You can do this by removing
+=~/.emacs.d/.local/etc/lsp/eclipse.jdt.ls= then running =lsp-install-server=.
```
hlissner, line 146: Change this to recommend `M-x +lsp/uninstall-server`, because the exact path of the lsp server could change over time. It's less maintenance hassle.

modules/lang/java/autoload.el
```diff
@@ -71,3 +71,9 @@ root)."
     (user-error "This buffer has no filepath; cannot guess its class name"))
   (or (file-name-sans-extension (file-name-base (buffer-file-name)))
       "ClassName"))
+
+;;;###autoload
+(defun +java/run-tests ()
+  "Run tests at point. If in a method, runs the test method, otherwise runs the entire test class."
+
+  )
```
hlissner, line 79: Did you mean to remove this? It doesn't seem to be used.

modules/lang/java/+lsp.el
```diff
@@ -4,10 +4,14 @@
...
-        lsp-jt-root (concat doom-etc-dir "eclipse.jdt.ls/server/java-test/server/"))
+  (setq lsp-java-workspace-dir (concat doom-etc-dir "java-workspace"))
   (add-hook! java-mode-local-vars #'lsp!)
-  :config
-  ;; TODO keybinds
-  )
+  :init
+  (when (featurep! :tools debugger +lsp)
+    (setq lsp-jt-root (concat lsp-java-server-install-dir "java-test/server/")
+          dap-java-test-runner (concat lsp-java-server-install-dir "test-runner/junit-platform-console-standalone.jar"))
```
hlissner, line 12: Did you mean `lsp-server-install-dir`, not `lsp-java-server-install-dir`?

modules/lang/java/autoload.el
jsravn: I meant to implement it, but will remove it for now.

modules/lang/java/+lsp.el
jsravn: It needs to be `lsp-java-server-install-dir` as the test runners are included as part of the LSP server in these sub-folders.

modules/lang/java/autoload.el
jsravn: I implemented it and placed it in `+lsp`, please have a look. The idea is to simulate how other Java IDEs work, where the test run is whatever is at point (either the entire class, or a method).

hlissner:
This shaped up nicely. Thanks again!

PR #3051, opened 2020-05-06
Improve Groovy support

sei40kr:
## Description

* Add Groovy REPL support
* Set the Dash docsets for Groovy


hlissner:
Thanks again!

PR #3052, opened 2020-05-06
Add Dash docsets + Xwidget integration

sei40kr:
## Description

Add Dash docsets + Xwidget integration. We can see Dash docsets in Embedded Webkit Widgets.

Screen Shot 2020-05-07 at 1 05 29

### DIFF
--- a/modules/tools/lookup/config.el
+++ b/modules/tools/lookup/config.el
@@ -181,16 +181,18 @@ See https://github.com/magit/ghub/issues/81"
       (funcall orig-fn url)))
 
   ;; Dash docset + Xwidget integration
-  (when (and (featurep! +xwidget) (display-graphic-p))
+  (when (featurep! +xwidget)
     (defun +lookup/dash-docs-xwidget-webkit-browse-url (url &optional new-session)
-      (setq xwidget-webkit-last-session-buffer +lookup--dash-docs-xwidget-webkit-last-session-buffer)
-      (save-window-excursion
-        (xwidget-webkit-browse-url url new-session))
-      (with-popup-rules!
-        '((set-popup-rule! "^\\*xwidget" :vslot -11 :size 0.35 :select nil))
-        (pop-to-buffer xwidget-webkit-last-session-buffer))
-      (setq +lookup--dash-docs-xwidget-webkit-last-session-buffer xwidget-webkit-last-session-buffer
-            xwidget-webkit-last-session-buffer nil))
+      (if (not (display-graphic-p))
+          (eww url new-session)
+        (setq xwidget-webkit-last-session-buffer +lookup--dash-docs-xwidget-webkit-last-session-buffer)
+        (save-window-excursion
+          (xwidget-webkit-browse-url url new-session))
+        (with-popup-rules!
+          '((set-popup-rule! "^\\*xwidget" :vslot -11 :size 0.35 :select nil))
+          (pop-to-buffer xwidget-webkit-last-session-buffer))
+        (setq +lookup--dash-docs-xwidget-webkit-last-session-buffer xwidget-webkit-last-session-buffer
+              xwidget-webkit-last-session-buffer nil)))
     (setq dash-docs-browser-func #'+lookup/dash-docs-xwidget-webkit-browse-url))
 
   (cond ((featurep! :completion helm)
--- a/modules/tools/lookup/config.el
+++ b/modules/tools/lookup/config.el
@@ -178,6 +178,26 @@ See https://github.com/magit/ghub/issues/81"
     (let ((gnutls-algorithm-priority "NORMAL:-VERS-TLS1.3"))
       (funcall orig-fn url)))
 
+  ;; Dash docset + Xwidget integration
+  (when (and (featurep! :tools lookup +xwidget) (display-graphic-p))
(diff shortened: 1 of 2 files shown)


modules/tools/lookup/config.el
```diff
@@ -178,6 +178,26 @@ See https://github.com/magit/ghub/issues/81"
     (let ((gnutls-algorithm-priority "NORMAL:-VERS-TLS1.3"))
       (funcall orig-fn url)))
 
+  ;; Dash docset + Xwidget integration
+  (when (and (featurep! :tools lookup +xwidget) (display-graphic-p))
```
hlissner, line 182: Use the `(featurep! +xwidget)` short form when you're in the module you're testing against.

modules/tools/lookup/config.el
```diff
@@ -178,6 +178,26 @@ See https://github.com/magit/ghub/issues/81"
...
+
+    (defun +xwidget--webkit-goto-url-a (&rest _)
+      (pop-to-buffer xwidget-webkit-last-session-buffer))
+    (advice-add #'xwidget-webkit-goto-url :after #'+xwidget--webkit-goto-url-a)
+
+    (defun +xwidget--webkit-new-session-a (orig-fun &rest args)
+      (save-window-excursion
+        (apply orig-fun args))
+      (pop-to-buffer xwidget-webkit-last-session-buffer))
+    (advice-add #'xwidget-webkit-new-session :around #'+xwidget--webkit-new-session-a)
```
hlissner, line 195: Can the rule and advise be made less general? I don't want the popup manager to capture _all_ xwidget buffers, only the ones displaying dash docsets. Perhaps a `dash-docs-browser-func` like this could replace all of them:

```lisp
(setq dash-docs-browser-func
      (lambda (url)
        (save-window-excursion
          (xwidget-webkit-browse-url url))
        (with-current-buffer xwidget-webkit-last-session-buffer
          (rename-buffer "*xwidget-webkit dash-docs*"))
        (pop-to-buffer xwidget-webkit-last-session-buffer)))

(set-popup-rule! "^\\*xwidget-webkit dash-docs" :vslot -11 :size 0.35 :select nil)
```

modules/tools/lookup/config.el
```diff
@@ -178,6 +178,26 @@ See https://github.com/magit/ghub/issues/81"
...
+
+    (when (featurep! :editor evil +everywhere)
+      (add-transient-hook! 'xwidget-webkit-mode-hook
+        (+evil-collection-init 'xwidget))))
```
hlissner, line 199: [modules/evil/init.el should already handle this](https://github.com/hlissner/doom-emacs/blob/develop/modules/editor/evil/init.el#L211). Is this necessary? If so, this should be moved there instead.

modules/tools/lookup/config.el
hlissner: The `display-graphic-p` test here also isn't enough. It'll break xwidget support for daemon GUI users. It'd be better to incorporate that check into the `dash-docs-browser-func` (and if it fails, fall back to its original function.

modules/tools/lookup/config.el
```diff
@@ -178,6 +180,19 @@ See https://github.com/magit/ghub/issues/81"
...
       (funcall orig-fn url)))
 
+  ;; Dash docset + Xwidget integration
+  (when (and (featurep! +xwidget) (display-graphic-p))
+    (defun +lookup/dash-docs-xwidget-webkit-browse-url (url &optional new-session)
+      (setq xwidget-webkit-last-session-buffer +lookup--dash-docs-xwidget-webkit-last-session-buffer)
+      (save-window-excursion
+        (xwidget-webkit-browse-url url new-session))
+      (with-popup-rules!
+        '((set-popup-rule! "^\\*xwidget" :vslot -11 :size 0.35 :select nil))
```
hlissner, line 190: The trouble with this is it makes it impossible to customize the popup behavior of the xwidget popup. Did the `rename-buffer` route not work?

sei40kr: @hlissner
It should work. But Xwidget renames the buffer after page loaded in `xwidget-webkit-callback`.

sei40kr: @hlissner 
In addition, I set `xwidget-webkit-last-session-buffer` to nil, to prevent another Xwidget session using dash-docs Xwidget session.

modules/tools/lookup/config.el
```diff
@@ -178,6 +180,21 @@ See https://github.com/magit/ghub/issues/81"
     (let ((gnutls-algorithm-priority "NORMAL:-VERS-TLS1.3"))
       (funcall orig-fn url)))
 
+  ;; Dash docset + Xwidget integration
+  (when (featurep! +xwidget)
+    (defun +lookup/dash-docs-xwidget-webkit-browse-url (url &optional new-session)
```
hlissner, line 185: Please rename the function `+lookup-dash-docs-xwidget-webkit-browse-url-fn`, to conform it to [Doom's naming conventions](https://github.com/hlissner/doom-emacs/blob/develop/docs/contributing.org#naming-conventions). `+lookup/X` indicates an interaction function.

hlissner:
Looks good to me. Thanks!

PR #3060, opened 2020-05-07
README for tools/rgb, also minor typo fix

itai33:
pretty short, but the module is small and I think I covered everything


hlissner:
Looks good to me. Thanks!

PR #3064, opened 2020-05-08
[:ui pretty-code doc] Add link to emacs-mac-port source code

gagbo:
Hopefully this will make it clear that emacs-mac does something special to get ligatures working with the composition-function-table method ; and it serves as an entry point for people who want to get into this kind of detail


hlissner:
Thanks!

PR #3065, opened 2020-05-08
Add org-roam doctor check for the dot executable

yoavm448:


yoavm448:
Actually, let me change the phrasing

Emiller88:
What about if users are using neato? Just a thought, but I think the doctor warning that the default won't work is enough, and we don't need to cover users who read the docs and figured out the switch.

hlissner:
Thanks!

> What about if users are using neato?

I could change it later to check `org-roam-graph-executable` instead of `"dot"`.

yoavm448:
I don't think this is a huge issue, since neato is usually packaged with dot in graphviz, but covering all cases is probably best.

PR #3073, opened 2020-05-09
Allow +vterm/toggle to open remotely

yangsheng6810:
When `default-directory` is remote, prepare the toggled vterm at remote
location. The added code is taken from `multi-term-switch-buffer` in `multi-term.el`.

### DIFF
--- a/modules/term/vterm/autoload.el
+++ b/modules/term/vterm/autoload.el
@@ -34,19 +34,7 @@ If prefix ARG is non-nil, recreate vterm buffer in the current project's root."
         (with-current-buffer buffer
           (unless (eq major-mode 'vterm-mode)
             (vterm-mode))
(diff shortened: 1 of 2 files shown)


yangsheng6810:
Sure. Where should I put the function in? Assume it should not reside in the `autoload.el`. Maybe corresponding `config.el`?

hlissner:
No, it _should_ be in the autoload.el, perhaps in a general `+vterm--change-directory` function that will change the current directory of the current vterm session (which will additionally handle tramp paths).

yangsheng6810:
It seems `vterm` already handles directory change according to `default-directory` when it is not remote. So I named it `+vterm--change-directory-if-remote.

hlissner:
Thanks!

PR #3075, opened 2020-05-09
Fix Rust LSP workspace

sei40kr:
## Description

RLS and rust-analyzer doesn't search `Cargo.toml` through child projects.
It means they won't work with mixed workspace.
So we need to make sure the LSP workspace contains a single project. (but this is hacky way)

Please also refer:
[rls] https://github.com/rust-lang/rls/issues/1202
[rust-analyzer] https://github.com/rust-analyzer/rust-analyzer/issues/3897

Maybe related:
https://github.com/hlissner/doom-emacs/pull/3212

### DIFF
--- a/modules/lang/rust/config.el
+++ b/modules/lang/rust/config.el
@@ -22,7 +22,23 @@
         rustic-format-trigger nil)
 
   (if (featurep! +lsp)
-      (add-hook 'rustic-mode-local-vars-hook #'lsp!)
+      (progn
+        (add-hook 'rustic-mode-local-vars-hook #'lsp!)
+
+        ;; HACK RLS and rust-analyzer doesn't search `Cargo.toml' through child
+        ;;      projects. It means they won't work with mixed workspace. So we
+        ;;      make sure the LSP workspace contains a single project.
+        ;;
+        ;;      [rls] https://github.com/rust-lang/rls/issues/1202
+        ;;      [rust-analyzer] https://github.com/rust-analyzer/rust-analyzer/issues/3897
+        (defun +rust--lsp-suggest-project-root-a ()
+          (when (and (eq major-mode 'rustic-mode) (featurep 'projectile))
+            (condition-case nil
+                (projectile-root-bottom-up (file-truename default-directory)
+                                           `("Cargo.toml"))
+              (error nil))))
+        (advice-add 'lsp--suggest-project-root
+                    :before-until #'+rust--lsp-suggest-project-root-a))
     (setq rustic-lsp-server nil)
     (after! rustic-flycheck
       (add-to-list 'flycheck-checkers 'rustic-clippy)))


gagbo:
The RLS issue seems to be with projects with subprojects but without a workspace directive, and I remember that I didn't have any issue with proper subprojects with RA. Can you provide an example where it doesn't work for you ? I'll try to repro

gagbo:
As far as I just tested : 
 (using `rustic-mode` with `eglot` and `rust-analyzer`), 
 having a proper top level `Cargo.toml` that defines workspace

starts `eglot` without any issue, and I can go to definition across members of the workspace without problem. I remember that I had issues with rls before, and I switched implementation because of it

sei40kr:
@gagbo 
I'm having the issue with this repo:
https://github.com/sei40kr/hello-rust
`Cargo.toml` exists but `lsp-mode` suggests the Git root as LSP workspace.

sei40kr:
@gagbo 

```
+ project/    <- lsp-mode uses this as the workspace
  - .git/
  + rust-project/
    + src/
      - main.rs    <- open this file
    - Cargo.toml
```

gagbo:
So it's not just an issue with Rust I suppose. Replace `rust-project` with `js-project`, and `Cargo.toml` with `package.json` and you'll have the same thing.

 I think it may be an issue where we want projectile/project.el to have priorities for this kind of meta-project without any logical link between the projects

sei40kr:
@gagbo 
We don't have same issue with NPM projects, because the LSP server automatically finds the nearest `package.json`.
So `lsp--suggest-project-root` returns the path to the Git repository root, but LSP features such as code completion still works.

hlissner:
How about if we just move `.git`, `.hg` and `.bzr` from `projectile-project-root-files-bottom-up` to the end of `projectile-project-root-files`? This issue seems to come up a lot with multi-project projects, and this seems like an altogether more portable solution. Perhaps it's unreasonable to always assume your repo root is your project's root, and instead use that as a last resort?

gagbo:
But if I gathered all those "smaller projects" in a single git repo, I'd like to be able to `projectile-find-file` all around the multiple projects, so fixing projectile root to the smaller project is not necessarily going to work.

I don't know how people work to be honest, but it looks like making this ^ decision would end up frustrating other users : 
- LSP wants a project root it can work with from a code point of view
- `projectile-find-file` wants the root to be `.git` because otherwise the files in your project "aren't appearing anymore". And `projectile-find-file` is only an example. If your "meta-project" has a Makefile to orchestrate everything, and you want to run `make` from the projectile-root, what are we supposed to do ?

```
+ project/    <- if projectile-root, then lsp will also use that and break
  - .git/
  - Makefile
  - js-project/
    - package.json
    + src/
  - rust-project/   <- if projectile-root, no easy way to run "make" from "project/"
    - src/
      - main.rs    <- open this file ; want LSP to work with Cargo.toml ; want projectile-run-command "make" to run from the top-level Makefile 
    - Cargo.toml
```

hlissner:
As of 8c3f24f LSP now prompts for the project root, which should effectively resolve this issue. Thanks for bringing it to my attention!

PR #3077, opened 2020-05-10
Add a few Email docs

Emiller88:
Made the links to my mail config permalinks since I moved to `mu4e` and add an initial `notmuch` README. I couldn't remember the git magic to find where a module was created, so the date and version are wrong.


Emiller88:
Update I made an attempt at checking when the notmuch module first came on the scene, but @hlissner should definitely double check.

hlissner:
Thanks!

hlissner:
Linking #1166 for posterity.

PR #3078, opened 2020-05-10
Add initial upload README

Emiller88:
I tried to keep it minimal since it mostly wraps `ssh-deploy`.


hlissner:
Thanks!

Linking #1166 for posterity.

PR #3082, opened 2020-05-10
Unify org-agenda-clock bindings with org-clock

alarsyo:
Following b40d85e9b, make the bindings for the agenda the same


hlissner:
Thanks!

PR #3083, opened 2020-05-10
Add keymap descriptions for Ruby module

lazyatom:
We can make the whichkey menu for Ruby files a bit more user friendly by showing simple names for each section.

I had hoped to set the description for the projectile commands too, but I can't figure out how to set it up since it doesn't actually use a "prefix" in the normal way. If anyone knows how, I'll gladly implement it 😄


modules/lang/ruby/config.el
```diff
@@ -136,7 +136,7 @@
   :config
   (setq rspec-use-rvm (executable-find "rvm"))
   (map! :localleader
-        :prefix "t"
+        :prefix ("t" . "test")
```
hlissner, line 139: Which-key labels aren't supported on minor mode keymaps. This is actually create a global label for all `SPC m t` keybinds. Please remove this.

modules/lang/ruby/config.el
```diff
@@ -164,7 +164,7 @@
     (add-hook 'minitest-mode-hook #'evil-normalize-keymaps))
   (map! :localleader
         :map minitest-mode-map
-        :prefix "t"
+        :prefix ("t" . "test")
```
hlissner, line 167: Which-key labels aren't supported on minor mode keymaps. This is actually create a global label for all `SPC m t` keybinds. Please remove this.

modules/lang/ruby/config.el
lazyatom: Removed in the latest push, apologies, I didn't realise.

hlissner:
Thanks!

PR #3084, opened 2020-05-10
lookup: Update docs

HaoZeke:
As the branch name suggests, using `sudo` with `yay` is an anti-pattern and so this fixes that.


hlissner:
Good catch. Thanks again!

PR #3088, opened 2020-05-11
bindings: update bindings / add remote bindings

UndeadKernel:
Update emacs bindings to match recent evil bindings.
Remove the `g` (lookup) prefix and merge those bindings onto the `s` prefix (just like in the evil bindings).

Add more `remote` bindings for emacs and evil users.

### DIFF
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -56,32 +56,42 @@
 
       ;;; <leader> f --- file
       (:prefix-map ("f" . "file")
+       (:when (featurep! :tools editorconfig)
+        :desc "Open project editorconfig"  "c"   #'editorconfig-find-current-editorconfig)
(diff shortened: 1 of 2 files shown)


hlissner:
How do vanilla users invoke the `+lookup/definition` or `+lookup/references` commands? Is that an oversight or are there other keybinds? I know `+lookup/definition` has `M-.` (remapped from `xref-find-definitions`), but not sure about `+lookup/references`.

UndeadKernel:
The standard key binding for `+lookup/references` in vanilla emacs is `M-?`, which is normally bound to `xref-find-references`.

In current Doom emacs, `+lookup/references` is correctly bound to `M-?`.

hlissner:
Understood. Oh, seems another PR caused a conflict. Could you rebase this PR to resolve it?

UndeadKernel:
Rebased and ready :D

hlissner:
Nice work. Thanks again!

Issue #2579, opened 2020-02-22
[REQUEST] Implementation lookup handlers


sei40kr:
**Describe the feature**
In programming languages like Java, `definition` means' **interface definition**.

So lsp-mode has `lsp-find-implementation` though it already has `lsp-find-definition`.
I want Doom Emacs to support lookup handlers for implementations and a key binding for it.

**System information**
<details><pre>
((emacs
  (version . "27.0.60")
  (features . "NOTIFY KQUEUE ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES THREADS PDUMPER LCMS2 GMP")
  (build . "Feb 20, 2020")
  (buildopts "'CPPFLAGS=-I/usr/local/opt/libxml2/include ' 'LDFLAGS=-L/usr/local/opt/libxml2/lib '")
  (windowsys . batch)
  (daemonp . server-running))
 (doom
  (version . "2.0.9")
  (build . "HEAD -> develop, upstream/develop c62c3bc4c 2020-02-20 20:30:43 -0500")
  (dir . "~/.doom.d/"))
 (system
  (type . darwin)
  (config . "x86_64-apple-darwin19.3.0")
  (shell . "/usr/local/bin/zsh")
  (uname . "Darwin 19.3.0 Darwin Kernel Version 19.3.0: Thu Jan  9 20:58:23 PST 2020; root:xnu-6153.81.5~1/RELEASE_X86_64 x86_64")
[...]

PR #3094, opened 2020-05-11
tools/lookup: Support lookup handlers for implementations and type definition

sei40kr:
## Description

Support lookup handlers for implementations and type definition.
I didn't still add key bindings for them.

https://github.com/hlissner/doom-emacs/issues/2579


hlissner:
Should these two commands fall back to `+lookup/{definition,references}` if no type-definition/implementations handler (respectively) is specified for the current mode?

hlissner:
Thanks!

PR #3100, opened 2020-05-12
Change path separator on Windows when using fd

chuxubank:
Show slash instead of backslash


chuxubank:
Sorry about my poor English, I just force pushed to change the incorrect commit message, no file changes.

hlissner:
Thanks!

PR #3101, opened 2020-05-12
Bind evil-jump-forward for <C-i> on Emacs GUI

simendsjo:
After translating C-i to <C-i> in commit 63ab88 to distinguish C-i from
TAB, the binding for evil-jump-forward no longer works.

### DIFF
--- a/modules/editor/evil/config.el
+++ b/modules/editor/evil/config.el
@@ -59,6 +59,12 @@ directives. By default, this only recognizes C directives.")
 
   (put 'evil-define-key* 'lisp-indent-function 'defun)
 
+  ;; evil binds C-i to evil-jump-forward, but we translate C-i
+  ;; to <C-i> to distinguish TAB from C-i in GUI Emacs.
+  ;; Ref commit 63ab88105f7c64b9b3e59546906ec5fe5857303a
+  (when (display-graphic-p)
+    (define-key evil-motion-state-map (kbd "<C-i>") 'evil-jump-forward))
+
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but I have to refuse it because I've since reverted the `C-i` "fix" in f83499c7a, after realizing we'd have to rebind `<C-i>` *everywhere* a C-i keybind exists (and there are so many). A more institutional fix would be better; this would be a endless hunt for edge cases I don't think is worth the trouble.

simendsjo:
> Thanks for the PR, but I have to refuse it because I've since reverted the `C-i` "fix" in [f83499c](https://github.com/hlissner/doom-emacs/commit/f83499c7a76ff822b3364784d89d63b91cd08fae), after realizing we'd have to rebind `<C-i>` _everywhere_ a C-i keybind exists (and there are so many, across evil-collection, org, magit...). A more institutional fix would be better; this would be a endless hunt for edge cases I don't think is worth the trouble.

This is the reason I thought modifying keybindings to translate "C-i .." to "<C-i> .." before binding might be a suitable part of the fix. As I use C-i to mean jump-forward everywhere, I'm not even sure about any other use for this binding. So I can just have these hacks in my own config instead. But I'm wondering how Spacemacs solved this.

hlissner:
I think they solved it by making it optional (and opt-in). `dotspacemacs-distinguish-gui-tab` is `nil` by default, and most people probably don't notice to report it, or rebind the keys themselves after enabling it.

Issue #3103, opened 2020-05-12
dap-mode web debugging packages are not loaded in typescript-mode


felix-clark:
**What did you expect to happen?**
dap-mode and dap-chrome be loaded upon opening a typescript .ts file when the doom (javascript +lsp) feature is enabled, like it is in js2-mode.

**What actually happened?**
I had to manually (require 'dap-mode) and (require 'dap-chrome) before launching a chrome debug session.

**Comments**
This may be a one-line fix here:

https://github.com/hlissner/doom-emacs/blob/bc5bbb177065fa9ea6968c0f6200295b5660f976/modules/tools/debugger/config.el#L113

The block that follows should perhaps also be loaded in typescript-mode instead of js2-mode alone.

felix-clark:
I submitted a PR but I am not proficient in elisp and there may be a better way to do this.

It might also be worth considering other major modes that should also load dap-mode and its web modules. js-mode? rjsx-mode? js3-mode? I don't have a good sense of any compatibility issues with these.

felix-clark:
This appears to now be working as expected for typescript-mode. I don't have a quick way of testing other possible modes to activate this in so I'll close the issue, but it should now be even more obvious to see how to add them.

PR #3104, opened 2020-05-12
activating dap web modules in TS mode

felix-clark:
Attempt at addressing #3103

### DIFF
--- a/modules/tools/debugger/config.el
+++ b/modules/tools/debugger/config.el
@@ -16,6 +16,14 @@
     (realgud:trepanpl  :modes (perl-mode perl6-mode))
     (realgud:zshdb     :modes (sh-mode))))
 
+(defun +debugger--load-dap-web ()
+  "Load the DAP extensions for web development."
+  (require 'dap-node)
+  (require 'dap-chrome)
+  (require 'dap-firefox)
+  (when IS-WINDOWS
+    (require 'dap-edge))
+  )
 
 ;;
 ;;; Packages
@@ -110,12 +118,9 @@
         (mapc #'require (cddr module)))))
 
   (when (featurep! :lang javascript +lsp)
-    (with-eval-after-load 'js2-mode
-      (require 'dap-node)
-      (require 'dap-chrome)
-      (require 'dap-firefox)
-      (when IS-WINDOWS
-        (require 'dap-edge))))
+    (with-eval-after-load 'js2-mode #'+debugger--load-dap-web)
+    (with-eval-after-load 'typescript-mode #'+debugger--load-dap-web)
+    )
 
   (dap-mode 1))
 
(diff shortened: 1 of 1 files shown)


modules/tools/debugger/config.el
```diff
@@ -110,12 +118,9 @@
...
   (when (featurep! :lang javascript +lsp)
-    (with-eval-after-load 'js2-mode
-      (require 'dap-node)
-      (require 'dap-chrome)
-      (require 'dap-firefox)
-      (when IS-WINDOWS
-        (require 'dap-edge))))
+    (with-eval-after-load 'js2-mode #'+debugger--load-dap-web)
+    (with-eval-after-load 'typescript-mode #'+debugger--load-dap-web)
+    )
```
gagbo, line 123: Can you put closing parens on the same line as the last statement please ? (Point 2 [here](https://github.com/hlissner/doom-emacs/blob/develop/docs/contributing.org#code-style) for reference)

modules/tools/debugger/config.el
```diff
@@ -16,6 +16,14 @@
...
     (realgud:zshdb     :modes (sh-mode))))
 
+(defun +debugger--load-dap-web ()
+  "Load the DAP extensions for web development."
+  (require 'dap-node)
+  (require 'dap-chrome)
+  (require 'dap-firefox)
+  (when IS-WINDOWS
+    (require 'dap-edge))
+  )
```
gagbo, line 26: Can you put closing parens on the same line as the last statement please ? (Point 2 [here](https://github.com/hlissner/doom-emacs/blob/develop/docs/contributing.org#code-style) for reference)

gagbo: Also, can you suffix the name of the function with `-h` since it's only used as a hook ?

modules/tools/debugger/config.el
felix-clark: Thanks for the quick review! fixed in 7f201fc

modules/tools/debugger/config.el
felix-clark: fixed in 7f201fc

modules/tools/debugger/config.el
```diff
@@ -110,12 +117,8 @@
...
 
   (when (featurep! :lang javascript +lsp)
-    (with-eval-after-load 'js2-mode
-      (require 'dap-node)
-      (require 'dap-chrome)
-      (require 'dap-firefox)
-      (when IS-WINDOWS
-        (require 'dap-edge))))
+    (with-eval-after-load 'js2-mode #'+debugger--load-dap-web-h)
+    (with-eval-after-load 'typescript-mode #'+debugger--load-dap-web-h))
```
hlissner, line 121: I think this could better expressed with:

```elisp
(after! (:or js2-mode typescript-mode)
  (require 'dap-node)
  (require 'dap-chrome)
  (require 'dap-firefox)
  (when IS-WINDOWS
    (require 'dap-edge)))
```

Without the need of a hook function. Not to mention, (with-eval-after-load PACKAGE #'function)` does not do evaluate `function`, because it's not being called, but that isn't relevant with the above solution.

felix-clark: Thanks, this is better. 95824c5

hlissner:
Thanks!

PR #3106, opened 2020-05-12
add app/rss readme

Amatrelan:
Simple readme for rss module. 

Should contain all important things to begin use app/rss


hlissner:
Thanks! Linking to #1166 for posterity.

PR #3109, opened 2020-05-12
Documentation fix - use-packages! typo in getting started

jsmestad:
I noticed this when getting my config ported over.


hlissner:
Good catch. Thanks!

PR #3116, opened 2020-05-14
Enable projectile integration for pipenv.el

7omb:
Previously it was disabled by setting pipenv-with-projectile
to nil as described here:

https://github.com/pwalsh/pipenv.el#projectile

Relevant to #1666

### DIFF
--- a/modules/lang/python/config.el
+++ b/modules/lang/python/config.el
@@ -198,7 +198,7 @@ called.")
 (use-package! pipenv
   :commands pipenv-project-p
   :hook (python-mode . pipenv-mode)
-  :init (setq pipenv-with-projectile nil)
+  :init
   :config
   (set-eval-handler! 'python-mode
     '((:command . (lambda () python-shell-interpreter))


hlissner:
Hi! Thanks for the PR, but I cannot accept it. While it does solve the problem, it interferes with the conda, pyvenv, and poetry integrations. As of right now, Doom takes a hands off approach with environment management for python, because it's a terrible mess, and getting all four (and pyenv) to play nicely is a sisyphean. A better solution is needed.

What's more, enabling `pipenv-with-projectile` introduces a few issues: it won't take effect until `python` is loaded and it doesn't solve the case of switching workspaces or buffers to new pipenv projects (it pidgeon holes the user into the projectile-switch-project workflow)

I would recommend using a tool like direnv (doom has a `:tools direnv` module), which serves as a single entry point for "changing environments" between buffers, and is already well integrated into Doom.

7omb:
Hi, thanks for the explanation. You're right, environment management with python is a mess. Thanks for the good work on doom-emacs. I just switched from Spacemacs and really enjoy it.

PR #3124, opened 2020-05-14
Fix (symbolp (quote org-directory)) from 06b3439

gagbo:


hlissner:
Ah, my bad. I already snuck through a fix for this in 85fb59acc, earlier today, before checking PRs. Thanks for the help, in any case!

PR #3133, opened 2020-05-15
Do not set `mu4e-attachment-dir` by default

vqhuy:
as it will be extended to "~/.mail/~/.mail/.attachments".

Furthermore, I doubt that everyone has a specific, central directory for attachments.

### DIFF
--- a/modules/email/mu4e/config.el
+++ b/modules/email/mu4e/config.el
@@ -18,7 +18,6 @@
             (version< mu4e-mu-version "1.4"))
     (setq mu4e-maildir "~/.mail"
           mu4e-user-mail-address-list nil))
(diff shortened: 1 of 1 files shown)


hlissner:
This sounds like an upstream bug; not expanding `~` correctly. I'd rather correct it than remove it, because the default value of `mu4e-attachment-dir` is `~/`. I don't think littering the user's `HOME` is a good default.

How about changing it to:

```lisp
(setq mu4e-attachment-dir
      (lambda () (expand-file-name ".attachments" (mu4e-root-maildir))))
```

hlissner:
As if 7ca20f1 this is done, so I'll consider this resolved. Thanks for bringing it to my attention!

PR #3137, opened 2020-05-16
Wrap flyspell-correct-at-point in feature check

CeleritasCelery:
`flyspell-correct-at-point` is not defined unless the the `spell` module is enabled.


CeleritasCelery:
I moved it to the spell module, but I didn't put it where you suggested because `flyspell-correct-at-point` is part of `flyspell-correct` not built-in `flyspell`.

hlissner:
Thanks!

PR #3141, opened 2020-05-16
fix rails server lag

otavioschwanck:


modules/lang/ruby/config.el
```diff
@@ -183,6 +183,7 @@
   :hook (projectile-rails-mode . auto-insert-mode)
   :init
   (setq auto-insert-query nil)
+  (setq projectile-rails-custom-server-command "bundle exec spring rails server --no-log-to-stdout")
```
hlissner, line 186: Shouldn't we modify `projectile-rails-vanilla-command`, `projectile-rails-spring-command`, and `projectile-rails-zeus-command` instead? If `projectile-rails-custom-server-command` is set, it appears to overwrite the others. Won't think cause issues for zeus/spring users?

```elisp
(setq projectile-rails-vanilla-command "bundle exec rails --no-log-to-stdout"
      projectile-rails-spring-command "bundle exec spring --no-log-to-stdout"
      projectile-rails-zeus-command "zeus ...?"

Comment (otavioschwanck) [reply modules/lang/ruby/config.el:186]: Good idea

Comment (otavioschwanck) [reply modules/lang/ruby/config.el:186]: This variables is the base  commands, it wont modify the server.  The server function is that:

```
(defun projectile-rails-server ()
  "Run rails server command."
  (interactive)
  (when (not (projectile-rails--file-exists-p "config/environment.rb"))
    (user-error "You're not running it from a rails application"))
  (if (member projectile-rails-server-buffer-name (mapcar 'buffer-name (buffer-list)))
      (switch-to-buffer projectile-rails-server-buffer-name)
    (projectile-rails-with-root
     (compile (projectile-rails--command :spring (concat projectile-rails-spring-command " rails server")
                                         :zeus (concat projectile-rails-zeus-command " server")
                                         :vanilla (concat projectile-rails-vanilla-command " server")
                                         :custom projectile-rails-custom-server-command)
              'projectile-rails-server-mode))))
```

bundle exec rails !== bundle exec rails server.  Have any idea of how we can add the --no-log-to-stdout and the end of every command ?

Comment (otavioschwanck) [reply modules/lang/ruby/config.el:186]: Like we talk on discord, lets leave it as is

File: modules/lang/ruby/README.org [hunk 2]
```diff
@@ -88,4 +89,14 @@ The rspec-mode prefix is =SPC m t=.  Here is some examples:
...
+
+** Rails Server
+By default, doom has disabled stdout from rails server to prevent emacs lag and performance issues.  You can enable by adding:
+
+#+BEGIN_SRC elisp
+;; On ~/.doom.d/config.el
+(setq projectile-rails-custom-server-command nil)
+#+END_SRC
+
+Also, you can change =bundle exec spring rails server= to another command, like docker.
```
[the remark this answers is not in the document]
otavioschwanck: Done

hlissner:
Thanks!

PR #3142, opened 2020-05-16
Minimap Module

rushsteve1:
Created a new Minimap module based on `minimap.el` from ELPA. The module sets some DOOM-specific config, and better defaults. Also added keybindings for it.

Some things that can be worked on/improved:
- Fix positioning when window split
- Sometimes it doesn't close when `minimap-mode` is disabled
- Adding more major-modes that the minimap will show in, or showing it in all modes (it is currently at the default which is showing in any mode derived from `prog-mode`)

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -33,6 +33,7 @@
        hl-todo           ; highlight TODO/FIXME/NOTE/DEPRECATED/HACK/REVIEW
        ;;hydra
        ;;indent-guides     ; highlighted indent columns
+       ;;minimap           ; show a map of the code on the side
        modeline          ; snazzy, Atom-inspired modeline, plus API
        ;;nav-flash         ; blink cursor line after big motions
        ;;neotree           ; a project drawer, like NERDTree for vim
@@ -47,7 +48,6 @@
        ;;window-select     ; visually switch windows
        workspaces        ; tab emulation, persistence & separate workspaces
        ;;zen               ; distraction-free coding or writing
-       ;;minimap           ; show a map of the code on the side
 
        :editor
        (evil +everywhere); come to the dark side, we have cookies
--- a/init.example.el
+++ b/init.example.el
@@ -47,6 +47,7 @@
        ;;window-select     ; visually switch windows
        workspaces        ; tab emulation, persistence & separate workspaces
        ;;zen               ; distraction-free coding or writing
+       ;;minimap           ; show a map of the code on the side
 
        :editor
        (evil +everywhere); come to the dark side, we have cookies
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -655,7 +655,8 @@
        :desc "Soft line wrapping"           "w" #'visual-line-mode
        (:when (featurep! :editor word-wrap)
         :desc "Soft line wrapping"         "w" #'+word-wrap-mode)
-       :desc "Zen mode"                     "z" #'writeroom-mode))
+       :desc "Zen mode"                     "z" #'writeroom-mode
+       :desc "Minimap"                      "m" #'minimap-mode))
 
 (after! which-key
   (let ((prefix-re (regexp-opt (list doom-leader-key doom-leader-alt-key))))
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -645,6 +645,8 @@
         :desc "Indent guides"              "i" #'highlight-indent-guides-mode)
        :desc "Indent style"                 "I" #'doom/toggle-indent-style
        :desc "Line numbers"                 "l" #'doom/toggle-line-numbers
+       (:when (featurep! :ui zen)
+        :desc "Minimap"                      "m" #'minimap-mode)
        (:when (featurep! :lang org +present)
         :desc "org-tree-slide mode"        "p" #'org-tree-slide-mode)
        :desc "Read-only mode"               "r" #'read-only-mode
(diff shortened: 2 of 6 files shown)


modules/ui/minimap/config.el
```diff
@@ -0,0 +1,14 @@
...
+  :config
+  (setq minimap-window-location 'right
+        minimap-update-delay 0
+        minimap-width-fraction 0.09
+        minimap-minimum-width 15)
+  (custom-set-faces!
+    `(minimap-current-line-face
+      :background ,(doom-color 'selection))
+    `(minimap-active-region-background
+      :background ,(doom-color 'vertical-bar))))
```
gagbo, line 14: This is incompatible with non-doom themes users, it's better to move these customization to a PR to emacs-doom-themes (Add the minimap-current-line-face and minimap-active-region-background faces to [this file](https://github.com/hlissner/emacs-doom-themes/blob/master/doom-themes-base.el) )

modules/ui/minimap/packages.el
```diff
@@ -0,0 +1,4 @@
+;; -*- no-byte-compile: t; -*-
+;;; ~/Nextcloud/Dotfiles/doom/minimap/packages.el
+
+(package! minimap)
```
gagbo, line 4: Can you add a pin (Usually I just take the master branch at the time I made the PR)

modules/ui/minimap/packages.el
```diff
@@ -0,0 +1,4 @@
+;; -*- no-byte-compile: t; -*-
+;;; ~/Nextcloud/Dotfiles/doom/minimap/packages.el
```
gagbo, line 2: This path should be changed

modules/ui/minimap/config.el
```diff
@@ -0,0 +1,14 @@
+;;; ~/Nextcloud/Dotfiles/doom/minimap/config.el -*- lexical-binding: t; -*-
```
gagbo, line 1: This path should be changed

modules/config/default/+evil-bindings.el
```diff
@@ -655,7 +655,8 @@
        :desc "Soft line wrapping"           "w" #'visual-line-mode
        (:when (featurep! :editor word-wrap)
         :desc "Soft line wrapping"         "w" #'+word-wrap-mode)
-       :desc "Zen mode"                     "z" #'writeroom-mode))
+       :desc "Zen mode"                     "z" #'writeroom-mode
+       :desc "Minimap"                      "m" #'minimap-mode))
```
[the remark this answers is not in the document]
rushsteve1: > Also, this binding is not wrapped in `when featurep! ui minimap` while the emacs one is. I don't know which one we want, but they should be the same

I was just following what Zen mode was doing in the respective files. But I agree, they should be wrapped in `featurep!` flags in both.

modules/ui/minimap/config.el
rushsteve1: Removed them from the `config.el` and opened a PR onto the file you mentioned here https://github.com/hlissner/emacs-doom-themes/pull/467

modules/ui/minimap/config.el
```diff
@@ -0,0 +1,9 @@
+;;; ui/minimap/config.el -*- lexical-binding: t; -*-
+
+(use-package! minimap
+  :hook doom-load-theme-hook
```
hlissner, line 4: This implicit hook won't work, for two reasons. It's equivalent to `(add-hook 'doom-load-theme-hook-hook #'minimap)`. Neither `doom-load-theme-hook-hook` nor `minimap` exist.

This should just be `:defer t`, since we have a keybind to toggle it, or:

```elisp
:hook (doom-first-buffer . minimap-mode)
```

modules/ui/minimap/README.org
```diff
@@ -0,0 +1,48 @@
+#+TITLE:   Minimap
+#+DATE:    May 8, 2020
+#+SINCE:   v2.0
```
hlissner, line 3: Replace v2.0 with v3.0.0, since that is the release it is introduced in.

modules/ui/minimap/README.org
```diff
@@ -0,0 +1,48 @@
+#+TITLE:   Minimap
```
hlissner, line 1: Please change the title to `#+TITLE:   :ui minimap`

modules/ui/minimap/README.org
```diff
@@ -0,0 +1,48 @@
...
+** Module Flags
+This module provides no flags.
+
+** Plugins
++ [[https://elpa.gnu.org/packages/minimap.html][minimap.el]]
+
+** Hacks
++ The colors used by the minimap are terrible so are overridden with
+  =custom-set-faces!= using colors out of the current DOOM theme.
+
```
hlissner, line 34: This is no longer the case and can be removed.

init.example.el
```diff
@@ -47,6 +47,7 @@
        ;;window-select     ; visually switch windows
        workspaces        ; tab emulation, persistence & separate workspaces
        ;;zen               ; distraction-free coding or writing
+       ;;minimap           ; show a map of the code on the side
```
hlissner, line 50: Modules are alphabetized. Please move this above `modeline`, below `indent-guides`.

rushsteve1:
Updated with the fixes you suggested.

If anyone has suggestions on how to fix the issues I listed in my original PR message that'd be appreciated. Specifically the split one is a bit of a pain. The issue with the minimap not closing is odd and I cannot consistently reproduce it.

Also do you think adding `org-mode` to the list of major modes the Minimap will open in is a good idea?

hlissner:
> Also do you think adding org-mode to the list of major modes the Minimap will open in is a good idea?

Until we've worked out the split issue, I think it should be manual. `minimap-create-window` needs to be replaced to utilize `display-buffer-in-side-window`. Side windows in Emacs have properties that should address the split issue.

In the meantime, I'll accept it as it is. Thanks again for your help!

PR #3143, opened 2020-05-16
Add --with-modules to emacs-mac documentation

dani84bs:
Default options for brew install emacs-mac compile wihtout
modules preventing to use vterm.


hlissner:
Makes sense to me. Thanks!

PR #3144, opened 2020-05-17
Upgrade rustic to fix performance regression

brownjohnf:
The `rustic-syntax-propertize` function in rustic (set as
`syntax-propertize-function` in emacs) had a performance regression
(reported in https://github.com/brotzeit/rustic/issues/107) that
caused emacs to effectively lock up every time the viewport changed.
This was fixed upstream in rust-mode by @phillord in
https://github.com/rust-lang/rust-mode/commit/bfe40565753295a4cf8403f4124710acd2827d21,
and ported to rustic by @brotzeit in
https://github.com/brotzeit/rustic/pull/108.

I've confirmed that this version of rustic seems to resolve the issue.

### DIFF
--- a/modules/lang/rust/packages.el
+++ b/modules/lang/rust/packages.el
@@ -1,6 +1,6 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; lang/rust/packages.el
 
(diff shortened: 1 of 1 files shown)


brownjohnf:
Sorry about the delay, but thanks for getting this in. I didn't have time to get to it over the weekend.

PR #3146, opened 2020-05-17
fix compatibility of centaur-bar to TUI

laishulu:
I figured out that for TUI, centaur-bar would not display with
```
  (setq centaur-tabs-set-bar 'left)
```
But it shows up with `nil`, `'under` and `'over`

### DIFF
--- a/modules/ui/tabs/config.el
+++ b/modules/ui/tabs/config.el
@@ -3,9 +3,10 @@
 (use-package! centaur-tabs
   :after-call after-find-file dired-initial-position-hook
   :init
+  (when (window-system)
+    (setq centaur-tabs-set-bar 'left))
   (setq centaur-tabs-set-icons t
         centaur-tabs-gray-out-icons 'buffer
-        centaur-tabs-set-bar 'left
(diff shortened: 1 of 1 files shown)


laishulu:
hold on.

I am asking the upstream to fix the incompatiblity problem of the `'left` value
https://github.com/ema2159/centaur-tabs/issues/90

hlissner:
I look forward to an upstream fix, but I cannot accept this PR. A one-time `window-system` check is not adequate for daemon users, and will need to be done on a per-frame basis. A much more complex solution is needed. I'll leave this open to track upstream progress.

laishulu:
I am tracking the upstream, and will update this page for any progress.

laishulu:
the issue is fixed by upstream (https://github.com/ema2159/centaur-tabs/issues/90).
So please update the pin of the package of `centaur-tabs` :-)

hlissner:
centaur-tabs was bumped in 15f6a6cdd, so I'll consider this resolved. Thanks for keeping me posted!

PR #3147, opened 2020-05-17
works better with projectile

laishulu:
just enable the group by default.

### DIFF
--- a/modules/ui/tabs/config.el
+++ b/modules/ui/tabs/config.el
@@ -13,13 +13,13 @@
         ;; replaces the tab list with that of another Doom workspace. This
         ;; prevents that.
         centaur-tabs-cycle-scope 'tabs)
-  (centaur-tabs-group-by-projectile-project)
 
   :config
   (add-hook '+doom-dashboard-mode-hook #'centaur-tabs-local-mode)
   (add-hook '+popup-buffer-mode-hook #'centaur-tabs-local-mode)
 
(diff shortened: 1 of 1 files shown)


modules/ui/tabs/config.el
```diff
@@ -13,6 +13,7 @@
         ;; replaces the tab list with that of another Doom workspace. This
         ;; prevents that.
         centaur-tabs-cycle-scope 'tabs)
+  (centaur-tabs-group-by-projectile-project)
```
hlissner, line 16: Please move this after the `:config`, otherwise this breaks lazy loading of the package.

laishulu: pushed.

hlissner:
Actually, on second thought, I'd rather we defer to centaur-tabs default on this one. Grouping tabs by project is an opinionated, rather than sensible default. At most, tabs should be grouped by perspectives (`persp-mode`) if `:ui workspaces` is enabled. Otherwise, I think this should be left to the user to add this to their own config.

```elisp
(after! centaur-tabs
  (centaur-tabs-group-by-projectle-project))
;; or
(add-hook 'centaur-tabs-mode-hook #'centaur-tabs-group-by-projectle-project)
```

In any case, thanks for the PR, but I must turn it down.

laishulu:
Got it.
Totally understand your explaination.

PR #3148, opened 2020-05-17
Allow the user to override lispyville themes

CeleritasCelery:
Currently the list of key themes for lispyville is hardcoded in the package config. The user can add more, but there is no way to unload a theme once it has been loaded.This simple change lets users set their own keys themes by setting `lispyville-key-theme` in their private config.


hlissner:
Thanks!

PR #3150, opened 2020-05-17
Add the needed migemo variants to the requirements for Japanese input.

tmtvl:
I have found that openSUSE's repos don't contain cmigemo, so may as well add the link to the README for Japanese input.

Aside from that everything seems to work out of the box.


hlissner:
Thanks!

PR #3152, opened 2020-05-18
add openSUSE installation section to getting started guide

strawberryjello:
I recently installed Doom Emacs on openSUSE, there were a few more steps to get it done than I'd expected. Thought it would save other users some googling if I noted the steps down here.


strawberryjello:
I haven't found any official releases for Emacs 27 in the repositories so far, I'm not sure yet if it's supported. openSUSE lags a bit when it comes to updated packages; I expect that Emacs 27 would have already been released officially elsewhere before it comes to a relatively up-to-date distro like Tumbleweed.

hlissner:

PR #3153, opened 2020-05-18
add missing +features to modules.org

runenesheim:
So I would know when to look for an extra feature for a module I did a little regex 🙈:
`find  .emacs.d -type f -name *.el -exec grep -Po "\(featurep! .*?\)" {} + | uniq | grep --color=always -P "\s\+\w+"`


hlissner:
Nice work! Thanks for the help.

Issue #3154, opened 2020-05-18
Minimap evil toggle keybinding hidden behind "zen"


felix-clark:
The toggle keybinding "\<leader\> t m" for the minimap requires (:ui zen), but now that "minimap" is its own feature it would be nice to have this defined by default so long as the "minimap" feature is enabled.

PR #3155, opened 2020-05-18
using dedicated minimap flag

felix-clark:
gating "toggle minimap" keybinding behind `minimap` rather than `zen`.

#3154


hlissner:
Good catch. Thanks!

PR #3160, opened 2020-05-18
[:ui doom] Add -solaire flag to disable solaire-mode

gagbo:
The question comes often enough that providing a flag to disable `solaire-mode` altogether might be a good solution


hlissner:
Thanks for the PR, but I cannot accept it. I'd rather find a way to fix [these lines](https://github.com/hlissner/doom-emacs/blob/develop/modules/ui/doom/config.el#L29-L36), than add more flags, and the canonical way to disable individual packages is `(package! solaire-mode :disable t)`.

PR #3161, opened 2020-05-18
Use ol-notmuch

johnogle222:
Ensures ol-notmuch is loaded so that storing links in notmuch buffers
will work as expected.


hlissner:
Hi! Thanks for the PR, but a different solution was needed. The canonical way to enable org modules is to modify the `org-modules` variable, rather than load it directly, which I've done in c6fadf4. Thanks for bringing it to my attention!

PR #3162, opened 2020-05-18
Fix minimap package declaration

rushsteve1:
The previous pin was based on the ELPA git revision, not the emacs-straight
mirror.
This changes it to explicitly use the emacs-straight mirror (which it should
have been doing anyway) and changes the pin to match.


rushsteve1:
Henrik beat me to it by under a minute :upside_down_face:

PR #3174, opened 2020-05-19
Fix scoop install instructions

oscarbailey:
Just tried installing on Windows, you need to use "scoop bucket add extras" not "scoop enable extras"


hlissner:
Thanks!

PR #3176, opened 2020-05-19
Fix update-fu commit hash

claudiiii:
Somehow the commit [24101b44...](https://gitlab.com/ideasman42/emacs-undo-fu/-/commit/24101b44fe0ea97e880fb14372ef461ac2cec1da) exists in the "Activity" overview on GitLab but not in the actual [git log](https://gitlab.com/ideasman42/emacs-undo-fu/-/commits/master). 
This pr sets the hash to currently most recent commit and fixes the error thrown on `doom upgrade`.

### DIFF
--- a/modules/emacs/undo/packages.el
+++ b/modules/emacs/undo/packages.el
@@ -3,5 +3,5 @@
 
 (if (featurep! +tree)
     (package! undo-tree :pin "5b6df03781495d8a25695d846b0cce496d3d3058")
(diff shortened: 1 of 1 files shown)


hlissner:
Since I want bump commits to be machine-readable (for future CI), their commit messages must follow a certain format (that I haven't documented anywhere yet, so it's not your fault), so I've dealt with this issue in e9292b2. Anyhow, thanks for the PR and bringing it to my attention!

PR #3177, opened 2020-05-20
use default search on rake tasks on :ruby lang

otavioschwanck:


hlissner:
Can you explain what this does and why it's being changed?

otavioschwanck:
> Can you explain what this does and why it's being changed?

with this, it will use the doom default search instead ido

sei40kr:
@hlissner 
FYI, we don't need Rake anymore because taskrunner supports it.

`rake` -> `+taskrunner/project-tasks`
`rake-rerun` -> `(ivy|helm)-taskrunner-rerun-last-command`

hlissner:
@sei40kr That's true. I'll merge this for the time being, but I'll look into substituting rake for taskrunner soon.

PR #3183, opened 2020-05-20
Update org-roam hyperlinks in org/README.org

marcus-crane:
I noticed that the links in `modules/lang/org/README.org` were still pointing to the original `jethrokuan/org-roam` repo so I've updated them to `org-roam/org-roam` as `org-roam` now has it's own *cough* org in Github. That's a lot of orgs.


hlissner:
Thanks!

PR #3189, opened 2020-05-21
make rails server popup automatically selected

otavioschwanck:
When we open the server, we dont want to be looking at it on 99% of times.  If it is automatic selected, it is more easily hide with q.


hlissner:
By default, popups have a `:quit t` property. This means C-g or ESC will close the popup when pressed from another window (i.e. you don't have to select the popup to close it with `q`).

PR #3196, opened 2020-05-22
lang/org: Support option in custom org links

howsiwei:
Refer https://orgmode.org/manual/External-Links.html.


hlissner:
Thanks!

PR #3199, opened 2020-05-22
lang/org: add indentation for evil-open-bellow inside src blocks

ericlarosa:
Replicates `+org-fix-newline-and-indent-in-src-blocks-a` enforcing tab indentation after evil-insert-bellow on source blocks


hlissner:
I've fixed this upstream, in https://github.com/hlissner/evil-org-mode/commit/2e9c4a295ee6aea7c97c5b1f3892b1c6e28a32d9 and bumped the `:lang org` module to include it, so this can be considered resolved. Thanks for bringing it to my attention!

PR #3208, opened 2020-05-23
Add emacs/vc readme

webframp:
Adds a simple readme for emacs/vc


hlissner:
Thanks! Linking #1166 for posterity.

PR #3209, opened 2020-05-24
Add zen-text-scale to ui/zen/README.org

sumitsahrawat:
Allow users to customize the text scaling factor for the `zen` module.

Tested manually by adding `(setq zen-text-scale nil)` to `~/.doom.d/config.el` and verifying that text scaling is disabled.

### DIFF
--- a/modules/ui/zen/README.org
+++ b/modules/ui/zen/README.org
@@ -13,6 +13,7 @@
   - [[#distractions-free-mode][Distractions-free mode]]
 - [[#configuration][Configuration]]
   - [[#enable-fullscreen-on-activation][Enable fullscreen on activation]]
+  - [[#customize-the-text-scaling-factor][Customize the text scaling factor]]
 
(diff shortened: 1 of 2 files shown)


sumitsahrawat:
I'm new to emacs lisp, and have no idea why some variables are prefixed with a `+` symbol. Looked through the manual and couldn't find anything. I'm assuming that it means that the variable is internal, so I've removed it from `zen-text-scale`. In case that is incorrect, please let me know and I'll add it again.

hlissner:
Hi! Thanks for the PR, but I cannot accept it. The `+` is part of a [Doom Emacs naming convention](https://github.com/hlissner/doom-emacs/blob/develop/docs/contributing.org#naming-conventions) indicating that a variable is defined by a Doom module. `+zen-text-scale` already exists so folks can customize the text scaling in writeroom-mode.

PR #3213, opened 2020-05-24
Prevent doom-capture frame from closing when refiling

wedens:
Prevents the frame opened by `+org-capture/open-frame` from immediately closing when refiling captured entry (`C-c C-w`).

The cause for the problem is that `org-capture-refile` performs capture finalization which triggers `org-capture-after-finalize-hook` and it causes early frame deletion.


hlissner:
Thanks

PR #3222, opened 2020-05-26
Add dhall module

TimWSpence:


hlissner:
Hi! Thanks for the PR, but I'm afraid I must refuse it. Mainly because its configuration is much too trivial to warrant its own module, but also because 1) package isn't pinned, 2) the documentation doesn't follow the template (from #1166), 3) the config.el and packages.el are missing their file headings, and 4) `use-package` is being used, when modules should use `use-package!`. In any case, thanks for taking the time for submitting it!

TimWSpence:
Thanks @hlissner Sorry, the `use-package!` was a typo. Would you like me to fix these things? It's so simple as the dhall mode package is pretty well setup and this means people can just uncomment the dhall module in `init.el` and they're all set

hlissner:
No, not unless dhall-mode needs more configuration. Those issues I listed were secondary to the real problem, and were mentioned simply for posterity. The real problem: the module is too small. i.e. It amounts to one line: `(package! dhall-mode)` to `~/.doom.d/packages.el`. The `use-package` block is unnecessary, since the package [autoloads its `auto-mode-alist` entry](https://github.com/psibi/dhall-mode/blob/ef4d33debe224c6ba37e51a29b9dc8b74f20f1c2/dhall-mode.el#L300).

I can't accept modules that are so trivial that users could set them up themselves in one or two lines, because that encourages others to PR other, similarly small modules for each package on MELPA/ELPA and increases my maintenance burden (more packages to test for and bump in future CI).

TimWSpence:
OK that's no problem at all. I just thought it was worth it as it enables support for a whole language, even if it is just one package but I get completely get why you wouldn't want it for single packages generally.

Anyway, I'll keep it in my personal config then. Thanks for all your hard work on Doom - I just switched from Spacemacs and I'm loving it so far :)

PR #3223, opened 2020-05-26
Update link to FAQ answer for removing evil

duchainer:
The past link was directing to wiki while FAQ is now extracted out.


hlissner:
Hi! Thanks for the PR, but I must refuse it. You have targeted the `master` branch, which isn't our default branch and is nearly (if not more than) 2 years out of date. If you are using `master`, I'd highly recommend switching to `develop`.

`master` will not be maintained/updated until our v3.0 release.

PR #3239, opened 2020-05-27
Add gdscript module

### LINKED ISSUE #3229: Add GDScript Module
**Describe the feature**
With the new GDScript language server support, and the now somewhat stable `emacs-gdscript-mode` package, I think it makes sense to combine these into a new lang module.

NathanLovato:
This commit adds the gdscript language module. It's a simple one as we packed all functionality in the gdscript-mode module, and lsp-support is built into lsp-mode.

Would it make sense to e.g. add an automatic lsp mode hook when the lsp module is available? or should there be a `+lsp` for that?

Closes #3229

### DIFF
--- a/modules/lang/gdscript/config.el
+++ b/modules/lang/gdscript/config.el
@@ -1,0 +1,29 @@
+;;; lang/gdscript/config.el -*- lexical-binding: t; -*-
+
+(after! projectile
+  (add-to-list 'projectile-project-root-files "project.godot"))
+
+(use-package! gdscript-mode
+  :defer t
+  :config
+  (set-lookup-handlers! 'gdscript-mode
+    :documentation #'gdscript-docs-browse-symbol-at-point)
+  (map! :localleader
+        :map gdscript-mode-map
+
+        (:prefix ("r" . "run")
+         "e" #'gdscript-godot-open-project-in-editor
+         "p" #'gdscript-godot-run-project
+         "d" #'gdscript-godot-run-project-debug
+         "s" #'gdscript-godot-run-current-scene)
+
+        (:prefix ("h" . "help")
+         "b" #'gdscript-docs-browse-api
+         "f" #'gdscript-docs-browse-symbol-at-point)
+
+        (:prefix ("f" . "format")
+         "b" #'gdscript-format-buffer
+         "r" #'gdscript-format-region))
+
+  (when (featurep! +lsp)
+    (add-hook 'gdscript-mode-local-vars-hook #'lsp!)))
--- a/init.example.el
+++ b/init.example.el
@@ -122,6 +122,7 @@
        ;;faust             ; dsp, but you get to keep your soul
        ;;fsharp           ; ML stands for Microsoft's Language
        ;;fstar             ; (dependent) types and (monadic) effects and Z3
+       ;;gdscript          ; the language you waited for
        ;;(go +lsp)         ; the hipster dialect
        ;;(haskell +dante)  ; a language that's lazier than I am
        ;;hy                ; readability of scheme w/ speed of python
--- a/modules/lang/gdscript/README.org
+++ b/modules/lang/gdscript/README.org
@@ -1,0 +1,51 @@
+#+TITLE:   lang/gdscript
+#+DATE:    May 27, 2020
+#+SINCE:   v2.0.9
+
+* Table of Contents                                                     :TOC:
+- [[#description][Description]]
+  - [[#module-flags][Module Flags]]
+  - [[#plugins][Plugins]]
+- [[#prerequisites][Prerequisites]]
+- [[#features][Features]]
+- [[#configuration][Configuration]]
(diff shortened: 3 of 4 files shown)


modules/lang/gdscript/config.el
```diff
@@ -0,0 +1,18 @@
+;;; lang/gdscript/config.el -*- lexical-binding: t; -*-
+
+(map! :localleader
+      :map gdscript-mode-map
```
yoavm448, line 4: This needs to run only after gdscript-mode is loaded, otherwise `gdscript-mode-map` isn't defined. The way to do this is either adding a `:after gdscript-mode` inside the `map!` expression, or putting it all inside a `use-package!`:
```lisp
(use-package! gdscript-mode
 :config
 (map! ...))
```
With `use-package!` you can also easily set an a file extension to open with `gdscript-mode` and such.

modules/lang/gdscript/config.el
```diff
@@ -0,0 +1,18 @@
...
+      "d" #'gdscript-godot-run-project-debug
+      "s" #'gdscript-godot-run-current-scene)
+
+      (:prefix ("h" . "help")
+      "b" #'gdscript-docs-browse-api
+      "f" #'gdscript-docs-browse-symbol-at-point)
+
+      (:prefix ("f" . "format")
+      "b" #'gdscript-format-buffer
+      "r" #'gdscript-format-region))
```
yoavm448, line 18: This could be better achieved using `set-formatter!`, look up `set-formatter!`'s definition

yoavm448:
> Would it make sense to e.g. add an automatic lsp mode hook when the lsp module is available? or should there be a +lsp for that?

Yes, there should be a flag to enable lsp. Its impolite otherwise, certainly if a user doesn't have lsp-mode installed through `:tools lsp`.

modules/lang/gdscript/config.el
```diff
@@ -0,0 +1,18 @@
+;;; lang/gdscript/config.el -*- lexical-binding: t; -*-
+
+(map! :localleader
```
hlissner, line 3: Please run `M-x indent-region` (or `evil-indent`) on this whole `map!` call. Its keybinds' indentation is off by one (one space too few).

modules/lang/gdscript/config.el
hlissner: I suggest using a `use-package!`. `gdscript-mode` will need more configuration in the future to integrate it into some of other modules (to set up lookup handlers with `set-lookup-handlers!`, a formatter with `set-formatter!`, and perhaps `set-eval-handler!` too). You can leave implementing those to me though.

modules/lang/gdscript/config.el
hlissner: You can leave this to me. The format module is about to undergo some big changes, along with `set-formatter!`. I'll revisit this module once that's done.

NathanLovato: The formatter is community-made and can currently have errors. For example, it doesn't support `match` blocks, a keyword built-into gdscript. That's why I left it manual.

NathanLovato:
> I've noticed lsp-mode comes with a lsp-gdscript. Do we need special support for it? Does it provide anything gdscript-mode doesn't?

gdscript-mode doesn't support the language server in itself. The gdscript-lsp client just adds lsp support to the language and sets up the connection. It's included in lsp-mode so as long as you run `lsp` in a gdscript-mode buffer it'll try to connect to Godot.

NathanLovato:
Just updated the PR. Regarding the formatter, while I could add it, this community-driven formatter has some known limitations. It's only out since 3 months ago, made by the community in Python... there are still some cases where it'll fail. At least with the `match` keyword, for sure. Actually, gdscript-mode doesn't support indentation well for that keyword either.

There are plans to have a formatter built into the engine, but it won't be available until a while.

modules/lang/gdscript/config.el
```diff
@@ -0,0 +1,24 @@
+;;; lang/gdscript/config.el -*- lexical-binding: t; -*-
+
+(use-package! gdscript-mode
+  :defer t
+  :config
+  (map! :localleader
```
hlissner, line 6: Prepend these lines (from #3265) to this `:config` block:

```elisp
(set-lookup-handlers! 'gdscript-mode
  :documentation #'gdscript-docs-browse-symbol-at-point)
```

modules/lang/gdscript/config.el
```diff
@@ -0,0 +1,24 @@
+;;; lang/gdscript/config.el -*- lexical-binding: t; -*-
+
```
hlissner, line 2: At top level, add these lines (from #3265):

```elisp
(after! projectile
  (add-to-list 'projectile-project-root-files "project.godot"))
```

NathanLovato:
I just addressed the requested changes, thanks! And thanks @otavioschwanck for the extra feature.

hlissner:

PR #3243, opened 2020-05-28
Feat: add counsel-jq

Emiller88:
Co-authored-by: rushsteve1 <rushsteve1@rushsteve1.us>

Wasn't sure if this was flag worthy or not. I also wasn't sure if my loading and mapping was correct I just tried to follow the `python` config.

### DIFF
--- a/modules/lang/json/config.el
+++ b/modules/lang/json/config.el
@@ -11,4 +11,13 @@
   (map! :after json-mode
         :map json-mode-map
         :localleader
-        "s" #'jsons-print-path))
+        "p" #'jsons-print-path))
+
+
+(use-package! counsel-jq
+  :defer t
+  :init
+  (map! :after json-mode
+        :map json-mode-map
+        :localleader
+        "s" #'counsel-jq))
--- a/modules/lang/json/config.el
+++ b/modules/lang/json/config.el
@@ -14,10 +14,12 @@
         "p" #'jsons-print-path))
 
 
-(use-package! counsel-jq
-  :defer t
-  :init
-  (map! :after json-mode
-        :map json-mode-map
-        :localleader
-        "s" #'counsel-jq))
+
+(when (featurep! :completion ivy)
+  (use-package! counsel-jq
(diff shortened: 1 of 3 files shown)


modules/lang/json/config.el
```diff
@@ -11,4 +11,13 @@
   (map! :after json-mode
         :map json-mode-map
         :localleader
-        "s" #'jsons-print-path))
+        "p" #'jsons-print-path))
+
+
+(use-package! counsel-jq
+  :defer t
```
hlissner, line 18: This block needs a `:when (featurep! :completion ivy)` as well.

hlissner:
Perhaps https://github.com/Lompik/x-path-walker would make a good helm alternative?

Emiller88:
> This package first started with the jq backend for JSON. But the python version is more efficient. It could be made optional though.
That's fine with me I'll give it a shot in a bit.

Emiller88:
@hlissner Set it up, but ended up with running it https://github.com/Lompik/x-path-walker/issues/3

It doesn't work for `json` so not in the scope of this module.

modules/lang/json/config.el
```diff
@@ -11,4 +11,15 @@
   (map! :after json-mode
         :map json-mode-map
         :localleader
-        "s" #'jsons-print-path))
+        "p" #'jsons-print-path))
+
+
+
+(when (featurep! :completion ivy)
+  (use-package! counsel-jq
```
hlissner, line 19: Rather than wrapping the `use-package!` in a `(when ...)`, please use `use-package`'s `:when` property.

```elisp
(use-package! counsel-jq
  ...)
```

This is necessary because the byte-compiler will evaluate `:when`, but not the enclosing `(when ...)` at compile time.

Emiller88: Cool, I wasn't aware of that!

Emiller88:
Should be good to go!

hlissner:
Great work. Thanks again!

PR #3249, opened 2020-05-29
tools/lsp: Improve eglot support

sei40kr:
# Changes

* Fix evil key bindings for LSP shown even when lsp module is not enabled
* Make `SPC c f` use eglot LSP formatter when `+format-with-lsp` is `t` (like lsp-mode)
* Bind general implementation lookup handler

# TODOs

* [x] ~Fix some modules overrides eglot's lookup handlers (e.g. Go)~
-> I do another PR:
   https://github.com/hlissner/doom-emacs/issues/3251

### DIFF
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -55,7 +55,6 @@
          :desc "Jump to symbol in any workspace"     "J"   #'helm-lsp-global-workspace-symbol))
        (:when (featurep! :tools lsp +eglot)
         :desc "LSP Execute code action"              "a" #'eglot-code-actions
-        :desc "LSP Format buffer/region"             "F" #'eglot-format
         :desc "LSP Rename"                           "r" #'eglot-rename
         :desc "LSP Find declaration"                 "j" #'eglot-find-declaration
         :desc "LSP Find implementation"              "J" #'eglot-find-implementation))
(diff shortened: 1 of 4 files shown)


modules/config/default/+emacs-bindings.el
```diff
@@ -34,6 +34,7 @@
        :desc "Evaluate buffer/region"                "e"   #'+eval/buffer-or-region
        :desc "Evaluate & replace region"             "E"   #'+eval/region-and-replace
        :desc "Format buffer/region"                  "f"   #'+format/region-or-buffer
+       :desc "Find implementations"                  "J"   #'+lookup/implementations
```
gagbo, line 37: This conflicts with `helm-lsp` and `lsp-ivy` bindings. As those are only active when `lsp-mode` is used, I chose to use the same bindings for both. (see l.56 here, and probably at the same relative location in evil-bindings)

Maybe the initial choice (I made) of putting helm-lsp/lsp-ivy on `:leader c j/J` is bad, but we should probably fix this at some point if we want to make `SPC c J` "global"

sei40kr: @hlissner @gagbo 
How about this?:
`SPC c o` -> Organize imports
`SPC c i` -> Find implementations
Find declaration -> No key binding because AFAIK most LSP servers `Find definition` finds declarations too.

sei40kr: @gagbo @hlissner 
I've updated the PR. Please review when you have time.

gagbo:
I have no issues with these, I'll adapt any way :)

gagbo:
We should merge this before we forget about it, and/or conflicts arise in the changed files.

The only "breaking" changes here are in the `doom-leader-code-map` it seems, real tooling could be created later to automatically extract those changes, but for the time being people can just be helped on Discord I suppose

hlissner:
Thanks!

PR #3253, opened 2020-05-30
Support LiveView templates in web module

jsmestad:
## Background

Elixir / Phoenix has a new file definition for LiveView templates. This ports over the same change that was added to the web-mode library in fxbois/web-mode#1070


hlissner:
Thanks!

PR #3256, opened 2020-05-30
tools/lsp: Make lsp checker chained to other checkers

sei40kr:
## Description

Make LSP checker in lsp-mode work with other checkers.


hlissner:
Makes sense to me. Thanks!

PR #3262, opened 2020-05-31
bump rustic

brotzeit:
I forgot to fix a mistake I did some time ago...

Since I don't use rust at the moment, it would be really great to get some help with the package, even if it's only for reviewing and testing the stuff I'm pushing. So if anybody here wants to help me I would very much appreciate it and gladly give write access to others. It seems the number of users has been increasing(I think doom is a major reason) and it makes it even more annoying when I introduce bugs that I don't notice because I don't use the package.

### DIFF
--- a/modules/lang/rust/packages.el
+++ b/modules/lang/rust/packages.el
@@ -1,6 +1,6 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; lang/rust/packages.el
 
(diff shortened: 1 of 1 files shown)


hlissner:
Sorry for the knitpick, but could you rewrote the heading of the commit message to `Bump :lang rust`?

brotzeit:
Done.

hlissner:
Thanks again!

PR #3263, opened 2020-05-31
Add proper syntax highlighting for Python-related configuration files

vathv:
This pull request will add syntax highlighting for `.flake8` and `Pipfile`. Just a small QOL improvement.

### DIFF
--- a/modules/lang/python/config.el
+++ b/modules/lang/python/config.el
@@ -17,6 +17,9 @@ called.")
 
 (use-package! python
   :defer t
+  :mode
+  ("\\.flake8" . conf-mode)
+  ("Pipfile$" . conf-mode)
   :init
   (setq python-environment-directory doom-cache-dir
         python-indent-guess-indent-offset-verbose nil)


modules/lang/python/config.el
```diff
@@ -17,6 +17,9 @@ called.")
 
 (use-package! python
   :defer t
+  :mode
+  ("\\.flake8" . conf-mode)
+  ("Pipfile$" . conf-mode)
```
hlissner, line 22: Could you remove the `:defer t` (`:mode` implies it)

hlissner: Also, change these two lines into:

```elisp
:mode ("[./]flake8\\'" . conf-mode)
:mode ("/Pipfile\\'" . conf-mode)
```

Since flake8 looks for `~/.flake8` and `~/.config/flake8`, I've adjusted the regexp to match both.

And pipfile's regexp was changed to match `Pipfile` atomically.

hlissner:
Great. thanks for your help!

PR #3276, opened 2020-06-02
Remove space from mc prefix map

DidacGit:
Change the space of the multiple-cursors keymap to a "-".

### DIFF
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -369,7 +369,7 @@
 
       ;;; <leader> m --- multiple cursors
       (:when (featurep! :editor multiple-cursors)
-       (:prefix-map ("m" . "multiple cursors")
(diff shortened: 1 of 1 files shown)


hlissner:
Hi! Thanks for the PR, but I cannot accept it. This is not an error. The label is general and not specific to the "multiple-cursors" package, and could one day be replaced, where a better implementation to appear.

DidacGit:
I'm sorry but I don't understand. This is the reason why Doom's multiple cursors map is called "doom-leader-multiple cursors-map" (with a space), isn't it? Why can't it be changed?

hlissner:
Ah, my bad! I misunderstood the PR. I've merged it now. Sorry for the trouble and thanks for the help!

DidacGit:
No problem! Thank you as always for your amazing work!

PR #3291, opened 2020-06-04
Bump :tools lsp

### LINKED ISSUE #3269: Eglot fails with (void-function project-root) (when unpinned)
Short story : using the pinned version of eglot *should* prevent the issue

The backtrace finishes with `(void-function project-root)`, and this issue may come from 2 different problems, which were uncovered by a recent eglot change that used a new version of `project.el` with a breaking API change.

Trying to find out the correct solution (and more details are provided also) in those upstream issues : 
joaotavora/eglot#492
raxod502/straight.el#531

gagbo:
Forcefully install the gnu-elpa-mirror recipe of project makes it move
emacsmirror/project@0d31141 (0.1.3) -> emacs-straight/project@da0333a (0.3.0)

Fixes #3269

### DIFF
--- a/modules/tools/lsp/packages.el
+++ b/modules/tools/lsp/packages.el
@@ -2,7 +2,10 @@
 ;;; tools/lsp/packages.el
 
 (if (featurep! +eglot)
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks!

PR #3293, opened 2020-06-04
Nix and Guix is a great idea!

grahamc:
Just being a bit cheeky :)


jcguu95:
Thought development around guix has been progressed... :( 
Well, it is a bit of improvement at least.

Will there be supports around Guix? Or it seems more likely to be around Nix?

PR #3302, opened 2020-06-05
(org +roam): use org-roam-buffer-window-parameters

jethrokuan:
Replaces the now obsolete org-roam-buffer-no-delete-other-windows


hlissner:
Thanks!

jethrokuan:
oh wow that was fast!

PR #3305, opened 2020-06-05
Bump web-mode for Elixir font lock improvements

jsmestad:


hlissner:
Done in 1d99b63. I did it myself because bump commits need to follow a particular format (so that they are machine readable for future CI). Thanks for bringing it to my attention!

PR #3307, opened 2020-06-05
Add lsp and repl to +light modeline

ericdallo:
This PR improves `+light` modeline:
* Checks for LSP status 
* Checks CIDER if enabled for clojure buffers.

I used the same :rocket: icon from `doom-modeline`, but if you think that makes sense change to another, let me know it.
image

Just following the [idea from doom modeline](https://github.com/seagle0128/doom-modeline/issues/347), when CIDER is available/connected/disconnected we show an icon:
image

### DIFF
--- a/modules/lang/clojure/config.el
+++ b/modules/lang/clojure/config.el
@@ -110,7 +110,7 @@
                                      (+modeline-format-icon 'faicon "terminal" "" face label -0.0575)
                                      " "))
           (add-to-list 'global-mode-string
-                       'cider-modeline-icon
+                       '(t (:eval cider-modeline-icon))
                        'append)))))
 
   ;; The CIDER welcome message obscures error messages that the above code is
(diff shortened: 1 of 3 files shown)


hlissner:
Could you have each particular module add a segment to `global-mode-string`, rather than add segments to the modeline?

ericdallo:
@hlissner do you mean `setq` `global-mode-string` on `:lang clojure` and `:tools lsp` modules?

ericdallo:
@hlissner just followed your suggestion, but it seems that the colors are not working via the `global-mode-string`:
image
Do you think what could it be?

ericdallo:
Ping. @hlissner

ericdallo:
@hlissner Fixed all issues, now it is working as expected:
image

ericdallo:
@hlissner could you take a look at this, please?

thiagokokada:
@hlissner

hlissner (approved):

hlissner:
Sorry for the late response. This looks good to me, thanks!

PR #3314, opened 2020-06-06
Add missing gdscript listing in modules.org

zackteo:


zackteo:
Sorry, wasn't too sure if I should make the pull request directly to master. But now that I think about it, I probably should've :x

hlissner:
Thanks!

PR #3317, opened 2020-06-06
Add lispy to evil-mc-incompalible-minor-modes

yoavm448:
Lispy commands in insert-state register only on one cursor, so this disables lispy temporarily when using evil-mc.


hlissner:
Thanks!

PR #3319, opened 2020-06-07
tools/taskrunner: Set popup rule for taskrunner buffer

sei40kr:
## Description

Add popup rule for taskrunner buffer so that we can close it by pressing Esc.


hlissner:
Thanks!

PR #3323, opened 2020-06-07
Fixing the rg-based query construction for doom--help-searchprompt

samwhitlock:
`counsel-rg-base-command` is a list, so concat needs it to string-join first.

As the comment says above, deadgrep is probably the way to go in the future. This is a quick fix for now.

### DIFF
--- a/core/autoload/help.el
+++ b/core/autoload/help.el
@@ -633,7 +633,7 @@ config blocks in your private config."
     (user-error "Can't find ripgrep on your system"))
   (if (fboundp 'counsel-rg)
       (let ((counsel-rg-base-command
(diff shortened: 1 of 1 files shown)


samwhitlock:
I'm not sure how "best" to do this string join op (new to Emacs). I've seen that `s.el` package used as a more intuitive way to do string operations in many packages, but I don't know if/how to declare that as a dependency if it isn't already.

hlissner:
This demanded a different solution, which I've pushed in 5092bd8. Thanks for bringing it to my attention!

Side-note: Emacs has a built-in `string-join` function, so using `s.el` is unnecessary here (the same is true for most of s.el's API, for that matter).

PR #3333, opened 2020-06-08
Fix typo in function name binded to jump to an IRC channel

Bounga:
This bug exists only in Emacs flavor keybindings


Bounga:
I don't know if I have to create an issue for this bug or if the PR is enough.

hlissner:
Thanks!

Issue #3332, opened 2020-06-08
+word-wrap-mode bound with emacs keybindings even if module isn't loaded


Bounga:
**What did you expect to happen?**

When using Emacs keybindings, `+word-wrap-mode` shouldn't be bound if the according module is not enabled in `init.el`. 
(:when (featurep! :editor word-wrap)
        :desc "Soft line wrapping"         "w" #'+word-wrap-mode)

**What actually happened?**

`+word-wrap-mode` is always bound to `C-c t w` so if we try to use it without having the module enabled it throws an error (`command-execute: Wrong type argument: commandp, +word-wrap-mode`) 

We should add a guard to avoid binding this function if the module isn't loaded, something like:

```elisp
(:when (featurep! :editor word-wrap)
        :desc "Word-wrap mode"               "w" #'+word-wrap-mode)
```
**Steps to reproduce:**
1. be sure to use emacs bindings and to disable `word-wrap` module,
2. start emacs,
3. use `C-c t w`
4. an error will be displayed.

**System information:**
<details>

```
emacs   version    28.0.50
        features   JPEG TIFF GIF PNG RSVG IMAGEMAGICK DBUS GLIB NOTIFY KQUEUE ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES THREADS XWIDGETS PDUMPER LCMS2 GMP
        build      Apr 24, 2020
        buildopts  (--enable-locallisppath=/usr/local/share/emacs/site-lisp --infodir=/usr/local/Cellar/emacs-head/HEAD-bef617d_1/share/info/emacs --prefix=/usr/local/Cellar/emacs-head/HEAD-bef617d_1 --without-x --with-dbus --with-imagemagick --with-modules --without-pop --with-gnutls --with-rsvg --with-xml2 --with-xwidgets --with-harfbuzz --with-ns --disable-ns-self-contained)
        windowsys  ns
        daemonp    server-running
doom    version    2.0.9
        build      HEAD -> develop 87364138d 2020-06-03 03:46:58 -0400
        dir        ~/.doom.d/
system  type       darwin
        config     x86_64-apple-darwin19.3.0
        shell      /usr/local/bin/fish
        uname      Darwin 19.4.0 Darwin Kernel Version 19.4.0: Wed Mar  4 22:28:40 PST 2020; root:xnu-6153.101.6~15/RELEASE_X86_64 x86_64
[...]

PR #3335, opened 2020-06-08
Fix word wrap binding in emacs keybindings

Bounga:
As I explained in https://github.com/hlissner/doom-emacs/issues/3332 word-wrapping mode toggling is bound in emacs keybindings even if the word-wrap module is disable which result in an error when using it.

Here's my attempt at fixing this bug.


hlissner:
Duplicate of #2267. See https://github.com/hlissner/doom-emacs/pull/2267#issuecomment-569536183

Issue #3328, opened 2020-06-08
unicode module doesn't work with with daemon mode 


guraltsev:
**What did you expect to happen?**
unicode fonts should be setup appropriately both in normal startup and when launching emacs as a daemon e.g. with `emacsclient -a="" test.org`

**What actually happened?**
`unicode-fonts-setup` does not get run

**Steps to reproduce:**
1. Use emacs 26.3
2. enable `unicode` in private init.el
3. launch emacs using `emacsclient -a="" test.org`
4. type or paste text from non-latin unicode
5. Note the font using `describe-char`
6. Now launch emacs normally, not as a daemon
7. type or paste text from non-latin unicode
8. Note the font using `describe-char`. It is a good font setup by unicode-fonts-setup

**System information:**
<details>

```
emacs   version    26.3
        features   XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS MODULES THREADS LCMS2
        build      Dec 25, 2019
        buildopts  (--without-compress-install --without-dbus --with-modules 'CFLAGS=-O2 -g3')
        windowsys  w32
        daemonp    daemon
doom    version    2.0.9
        build      HEAD -> develop, origin/develop, origin/HEAD
        dir        ~/.config/doom/
system  type       windows-nt
        config     x86_64-w64-mingw32
        shell      C:/bin/emax64-26.3/libexec/emacs/26.3/x86_64-w64-mingw32/cmdproxy.exe
        uname      n/a
[...]

PR #3336, opened 2020-06-08
Make unicode-fonts-setup run when emacs is started in daemon mode

guraltsev:
Fixes https://github.com/hlissner/doom-emacs/issues/3328.

The problem was simple: testing the variable `initial-window-system` is deprecated. One should call `(display-graphic-p)`


hlissner:
Thanks!

PR #3344, opened 2020-06-09
Update org-gcal

maikol-solis:
Makes `org-gcal-post-at-point` work again.


hlissner:
Bump commits must follow a particular format (to be machine-readable for future CI). I've bumped the `:app calendar` module in 67c53a2, which should resolve your issue.

PR #3347, opened 2020-06-10
Don't set prefix-arg when delegating to ivy/helm from +default/search-project

leoalekseyev:
The prefix arg for +default/search-project is intended to enable the
user to pick search location, however when it propagates to the ivy
search function it turns on the "search ignored files" functionality,
which us typically not intended and, furthermore, can be (if desired)
turned on simply by typing the search command in, for exampe,
counsel-rg's case as `-uu -- mysearchterm`

----

#


hlissner:
Thanks!

PR #3349, opened 2020-06-10
Bump :tools lsp

sei40kr:
## Description

Bump `:tools lsp`.
Add lsp-metals to `:lang scala` because lsp-mode has removed metals support.


hlissner:
Bump commits must follow a strict format, so I did it myself in 6c80419. Thanks for bringing it to my attention!

PR #3353, opened 2020-06-11
Fix CLI error messages for unrecognized commands

danielkoning:
`bin/doom` is producing some odd error messages.

> $ doom nonexistent
> x Couldn’t find any "nonexistent" command
> 
> See 'doom help nil' for documentation on this command.
> 
> $ doom sync --nonexistent
> x Unrecognized switch "--nonexistent"
> 
> See 'doom help --nonexistent' for documentation on this command.

I would have expected this instead:

> $ doom nonexistent
> x Couldn’t find any "nonexistent" command
> 
> See 'doom help' for documentation.
> 
> $ doom sync --nonexistent
> x Unrecognized switch "--nonexistent"
> 
> See 'doom help sync' for documentation on this command.

The argument passed to `print!` doesn't appear to have been kept up to date with the refactoring. It looks like `(car args)` used to point to the first argument after "doom" (what a C program would call `argv[1]`), but the refactoring made `command` a separate variable, and now `(car args)` denotes `argv[2]`.

### DIFF
--- a/bin/doom
+++ b/bin/doom
@@ -106,7 +106,9 @@ with a different private module."
                                            start-time))))))
         (user-error
          (print! (error "%s\n") (error-message-string e))
-         (print! (yellow "See 'doom help %s' for documentation on this command.") (car args))
+         (if (doom-cli-get command)
+             (print! (yellow "See 'doom help %s' for documentation on this command.") command)
+           (print! (yellow "See 'doom help' for documentation.")))
          (error "")) ; Ensure non-zero exit code
         ((debug error)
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR! There is nothing wrong with it, however I am working on a complete rewrite of Doom's CLI to resolve many of its issues (including a rewritten help system; overwriting this fix). Since it is due to be published very soon (and because this issue isn't critical), I must turn it down, as to avoid interfering with merging the rewrite. Thanks for looking into it and bringing it to my attention though!

PR #3355, opened 2020-06-11
Use upstream flycheck-guile package

flatwhatson:
I've pulled this checker out as its own package, so we don't need to ship it with Doom.


flatwhatson:
This PR indirectly fixes a bug in the `flycheck-guile.el` included with doom, where I've used an `rx` syntax that's not supported in emacs 26.  It's fixed in the upstream repo with this commit: https://github.com/flatwhatson/flycheck-guile/commit/f37b6143776d15b3e7907e7621f6f96f8b1aec48

hlissner:
Thanks!

elais:
This breaks workflows that rely on directory local instances of guile that are loaded with tools like direnv. In particular it makes it so that geiser does not load guile libraries that are declared in guix manifests or nix-shells and loaded with direnv. 

Also flycheck here seems to use the system's guile package instead of the guile package in profiles created by guix environment and I would assume nix-shell.

PR #3356, opened 2020-06-11
Fix #3348

### LINKED ISSUE #3348: winner-redo not functioning correctly
Minimal repro:
Use only one window (i.e. full-screen). Start in buffer A. Switch to buffer B. Use `SPC w C-u`, this will switch to buffer A. Now use `SPC w C-r`. This should take me to buffer B.

This is how winner undo/redo works in vanilla emacs (bound to `C-c <left>/<right>`) and in Spacemacs (bound to `SPC w u/U`).

**What actually happened?**
Instead we stay in buffer A, winner-redo has no effect.

**Additional details:**
I think there could be some additional weirdness with winner not handling which-key windows in a sensible way.

leoalekseyev:
I am not sure what the setting + removing hook code in use-package!
winner was trying to accomplish, but it was breaking winner-redo
functionality. This patch fixes the issue.

### DIFF
--- a/core/core-ui.el
+++ b/core/core-ui.el
@@ -450,10 +450,9 @@ windows, switch to `doom-fallback-buffer'. Otherwise, delegate to original
 
 (use-package! winner
   ;; undo/redo changes to Emacs' window layout
-  :hook (window-configuration-change . winner-mode)
   :preface (defvar winner-dont-bind-my-keys t) ; I'll bind keys myself
   :config
-  (remove-hook 'window-configuration-change #'winner-mode)
+  (add-hook 'doom-first-buffer-hook #'winner-mode)
   (appendq! winner-boring-buffers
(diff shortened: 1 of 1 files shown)


core/core-ui.el
```diff
@@ -450,10 +450,9 @@ windows, switch to `doom-fallback-buffer'. Otherwise, delegate to original
 
 (use-package! winner
   ;; undo/redo changes to Emacs' window layout
-  :hook (window-configuration-change . winner-mode)
   :preface (defvar winner-dont-bind-my-keys t) ; I'll bind keys myself
   :config
-  (remove-hook 'window-configuration-change #'winner-mode)
+  (add-hook 'doom-first-buffer-hook #'winner-mode)
```
hlissner, line 455: Replace this with `:hook (doom-first-buffer . winner-mode)`

leoalekseyev: PR updated

hlissner:
Thanks!

PR #3373, opened 2020-06-13
Fix visit docs for Racket & support racket-xp

sorawee:
- Fix Racket's `+lookup/documentation` in the same way as
  https://github.com/hlissner/doom-emacs/commit/f20f477a44138d3bc8b
- Support the `+xp` feature which enables racket-xp. When it's not
  enabled, default handlers to the traditional racket-repl
  (which requires users to run code first)

### DIFF
--- a/modules/lang/racket/autoload.el
+++ b/modules/lang/racket/autoload.el
@@ -12,7 +12,7 @@
                 buf)))))
 
 ;;;###autoload
-(defun +racket-documentation (thing)
+(defun +racket-lookup-documentation (thing)
   "A `+lookup/documentation' handler for Racket."
   (let ((buf (if racket-xp-mode
                  (racket-xp-describe thing)
@@ -22,7 +22,7 @@
       t)))
 
 ;;;###autoload
-(defun +racket-definition (_thing)
+(defun +racket-lookup-definition (_thing)
   "A `+lookup/definition' handler for Racket."
   (if racket-xp-mode
       (call-interactively #'racket-xp-visit-definition)
--- a/modules/lang/racket/autoload.el
+++ b/modules/lang/racket/autoload.el
@@ -10,3 +10,20 @@
               (let ((buf (get-buffer "*Racket REPL*")))
                 (bury-buffer buf)
                 buf)))))
+
+;;;###autoload
+(defun +racket-documentation (thing)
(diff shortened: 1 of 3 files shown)


modules/lang/racket/config.el
```diff
@@ -27,6 +27,9 @@
              #'rainbow-delimiters-mode
              #'highlight-quoted-mode)
 
+  (when (featurep! +xp)
+    (add-hook! 'racket-mode-hook #'racket-xp-mode))
```
hlissner, line 31: Replace `add-hook!` with `add-hook`. There is no benefit to using `add-hook!` like this.

modules/lang/racket/autoload.el
```diff
@@ -10,3 +10,20 @@
               (let ((buf (get-buffer "*Racket REPL*")))
                 (bury-buffer buf)
                 buf)))))
+
+;;;###autoload
+(defun +racket-documentation (thing)
```
hlissner, line 15: Please rename to `+racket-lookup-documentation`

modules/lang/racket/autoload.el
```diff
@@ -10,3 +10,20 @@
...
+  "A `+lookup/documentation' handler for Racket."
+  (let ((buf (if racket-xp-mode
+                 (racket-xp-describe thing)
+               (racket-repl-describe thing))))
+    (when buf
+      (pop-to-buffer buf)
+      t)))
+
+;;;###autoload
+(defun +racket-definition (_thing)
```
hlissner, line 25: Please rename to `+racket-lookup-definition`

modules/lang/racket/config.el
sorawee: Do you want me to replace nearby `add-hook!` to `add-hook` as well?

sorawee: I made the modification as you suggested, but let me know if you want to clean up other `add-hook!` (I used `add-hook!` because I wanted to be consistent with the other uses).

hlissner: For `racket-smart-open-bracket-mode`, yes, but not the one above (which adds rainbow-delimiters-mode and highlight-quoted-mode).

sorawee: Done.

hlissner:
Thanks!

PR #3375, opened 2020-06-13
Fix #3374

### LINKED ISSUE #3374: [REQUEST] provide an option to toggle vterm pop-up window even when displayed
**Describe the feature**
Currently, if vterm pop-up invoked via `SPC o t` is displayed, but another buffer is selected, a subsequent invocation of `SPC o t` selects the vterm buffer. This is a sensible default behavior, however for certain usage patterns it is inconvenient. Consider, for example, the case of tailing a log file, or issuing an ad hoc command in the terminal, going back to the code buffer, and realizing the terminal is not needed anymore and wanting to dismiss it.

The current behavior, moreover, is somewhat unexpected because a command with "toggle" in its name is expected to indeed toggle some feature consistently on subsequent invocations.

Furthermore, it is quite easy to select the vterm buffer, if so desired, via user's preferred window selection commands.  An option to always hide the vterm pop-up if already displayed using the same keybinding used to show it would provide substantially improved UX over having to think about whether or not the buffer is selected, whether or not one should be using the `o t` binding or `~` binding, etc.

betaprior:
This patch provides an option to always toggle vterm buffer, even if displayed. Note that the default behavior is not changed. This enables a more ergonomic UX where the user can rely on a single binding to open/hide a pop-up

Currently, if vterm pop-up invoked via SPC o t is displayed, but another buffer is selected, a subsequent invocation of SPC o t selects the vterm buffer. This is a sensible default behavior, however for certain usage patterns it is inconvenient. Consider, for example, the case of tailing a log file, or issuing an ad hoc command in the terminal, going back to the code buffer, and realizing the terminal is not needed anymore and wanting to dismiss it.

The current behavior, moreover, is somewhat unexpected because a command with "toggle" in its name is expected to indeed toggle some feature consistently on subsequent invocations.

Furthermore, it is quite easy to select the vterm buffer, if so desired, via user's preferred window selection commands. An option to always hide the vterm pop-up if already displayed using the same keybinding used to show it would provide substantially improved UX over having to think about whether or not the buffer is selected, whether or not one should be using the o t binding or ~ binding, etc.

### DIFF
--- a/modules/term/vterm/autoload.el
+++ b/modules/term/vterm/autoload.el
@@ -1,10 +1,11 @@
 ;;; term/vterm/autoload.el -*- lexical-binding: t; -*-
 
 ;;;###autoload
-(defun +vterm/toggle (arg)
+(defun +vterm/toggle (arg &optional force-toggle)
   "Toggles a terminal popup window at project root.
 
-If prefix ARG is non-nil, recreate vterm buffer in the current project's root."
+If prefix ARG is non-nil, recreate vterm buffer in the current project's root.
+If FORCE-TOGGLE is non-nil and the vterm buffer is visible, hide it (default: select vterm window)"
   (interactive "P")
   (unless (fboundp 'module-load)
     (user-error "Your build of Emacs lacks dynamic modules support and cannot load vterm"))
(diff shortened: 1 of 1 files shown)


wedens:
You can use C-`  to toggle any popups. But I agree that if it's named "toggle*" it should probably toggle :)

leoalekseyev:
> You can use C-` to toggle any popups. But I agree that if it's named "toggle*" it should probably toggle :)

C-` is nice but is undocumented and not discoverable (unlike the `SPC ~` binding to the same function). However as you said, a toggle should toggle -- Principle of Least Surprise is a Good Thing!

PR #3376, opened 2020-06-13
Disable projectile cache for fd

wedens:
In my experience it's more trouble than it's worth. I've also seen people on doom's discord regularly having issues caused by the cache.

`fd` should be fast enough without caching on higher than calculator-grade hardware ;)

### DIFF
--- a/core/core-projects.el
+++ b/core/core-projects.el
@@ -26,7 +26,8 @@ Emacs.")
              projectile-locate-dominating-file)
   :init
   (setq projectile-cache-file (concat doom-cache-dir "projectile.cache")
-        projectile-enable-caching doom-interactive-p
+        ;; fd is fast enough without caching
+        projectile-enable-caching (not (executable-find doom-projectile-fd-binary))
         projectile-globally-ignored-files '(".DS_Store" "Icon" "TAGS")
         projectile-globally-ignored-file-suffixes '(".elc" ".pyc" ".o")
         projectile-kill-buffers-filter 'kill-only-files


hlissner:
> In my experience it's more trouble than it's worth. I've also seen people on doom's discord regularly having issues caused by the cache.

Can you explain what trouble it gives you? Or trouble "people on doom's discord" are experience.

> fd should be fast enough without caching on higher than calculator-grade hardware ;)

Needs more testing. And a more dynamic solution, in case the user uses a different (i.e. slower) indexing method, or aren't using fd (is ripgrep fast/stable enough?). It may be sufficient to stop it from writing to file (e.g. `(advice-add #'projectile-serialize-cache :override #'ignore)`), so we only have an in-memory cache for the current session.

wedens:
> Can you explain what trouble it gives you? Or trouble "people on doom's discord" are experiencing?

One common example is external modifications. That includes changing .gitignore, for example.

 I've also experienced some problems that I can't even explain, but cleaning the cache was required to solve them.

>  It may be sufficient to stop it from writing to file (e.g. (advice-add #'projectile-serialize-cache :override #'ignore)), so we only have an in-memory cache for the current session.

Unfortunately, that's not enough. The cache can get out of sync while emacs is running.

eddiegroves:
FWIW, as a novice, I was stuck not finding files I knew were their after a `git pull`... only clearing the projectile cache allowed find files to work again

hlissner:
Another point against this: indexing and filtering file-lists over TRAMP is prohibitively slow without a cache.

jsmestad:
I think documenting this as a configuration option in the FAQ or customization docs would be good if @hlissner is against changing the default.

hoyon:
I also have caching disabled and have found that it doing so makes `+ivy/projectile-find-file` twice as expensive as it needs to be since it will run `projectile-project-files` twice. This means the file scan and file filtering (if using `hybrid) is performed twice in a row, with the first result immediately discarded after checking its length.

This isn't a problem if caching is enabled since the file scan will be run once at most.

hlissner:
@hoyon How about, instead of disabling caching altogether, you set `(setq projectile-files-cache-expire 5)`. The cache will be regenerated after a short pause, but quick, consecutive uses should still be fast enough.

hoyon:
@hlissner thanks for the suggestion! That should work nicely.

hlissner:
With an eventual switch to project.el getting closer (and because this PR targets the now-defunct develop branch), I will close it. Thanks for bringing it to my attention, in any case!

PR #3377, opened 2020-06-13
Make `fd` respect `projectile-globally-ignored-directories`

wedens:
It's already done for `rg`, and it's probably worth it to do the same for `fd` for consistency. Even though it's not the best to add ignored directories, users may expect that changing `projectile-globally-ignored-directories` will have effect of files search.


hlissner:
While this certainly fixes the problem, but I notice another issue: later changes to `projectile-global-ignored-directory` are not respected. It's earlier value is baked into `projectile-generic-command` and never updated.

Oh well. I'll figure out a better solution another time. Thanks for the PR!

hlissner:
10f5916 should fix that.

PR #3386, opened 2020-06-14
Bumping org-mode pin to avoid bad fontification breakage

leoalekseyev:
Current org pin breaks source block fontification

----

#


hlissner:
Org (among other things) were bumped in c232d0f. I've done it myself because bump commits need to follow a particular format (which I haven't documented anywhere yet), for the sake of future CI. In any case, thanks for bringing it to my attention!

PR #3387, opened 2020-06-14
Pin newer version of racket-mode to fix racket-xp-mode performance issue

dbriemann:
Newer version of racket-mode fixes this issue: https://github.com/greghendershott/racket-mode/issues/473


hlissner:
racket was bumped in 120ba7d. I've done it myself because bump commits need to follow a particular format (which I haven't documented anywhere yet), for the sake of future CI. In any case, thanks for bringing it to my attention!

PR #3393, opened 2020-06-15
Update elixir-mode to support new sigil font locks

jsmestad:
## Background

Adds font lock support for EEx and LiveView templates in `elixir-mode`


hlissner:
I've bumped the elixir module in 468d731. I've done it myself because bump commits need to follow a particular format (which I haven't documented anywhere yet), for the sake of future CI. In any case, thanks for bringing it to my attention!

PR #3404, opened 2020-06-17
Change emacs-plus icon

olav35:
Change the emacs-plus icon from spacemac's logo to a more neutral modern emacs icon.


hlissner:
Thanks!

olav35:
I came to notice that this option only is available on the version 27 and 28 and as such the current install command will not work.

I apologize for not catching this before sending the pr. I shall be more cautious in the future.

eddiegroves:
If version 27 is recommend anyway maybe change this to `emacs-plus@27`?

PR #3405, opened 2020-06-17
python: fix conda path

ztlevi:
Based on the discussion over here, the absolute path is needed to make conda work.
https://github.com/necaris/conda.el/issues/63#issuecomment-642404300


hlissner:
Thanks!

Issue #2333, opened 2020-01-08
[BUG] magit-show-commit has problem with emacs27 and emacs28


FirstLoveLife:
doom has some issue with `magit show commit ` in emacs27. 

How to reproduce: <kbd> SPC g B</kbd> <kbd> RET </kbd> 

Expected:
![image](https://user-images.githubusercontent.com/37442588/71950250-8840c380-3211-11ea-8eb2-ac97e6204f97.png)

Now: 
![image](https://user-images.githubusercontent.com/37442588/71950178-50398080-3211-11ea-85e6-8fd4d0a099b1.png)

<details>

```
emacs   version    28.0.50
        features   XPM JPEG TIFF GIF PNG SOUND DBUS GSETTINGS GLIB NOTIFY INOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE HARFBUZZ XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS LIBSYSTEMD PDUMPER GMP
        build      Jan 08, 2020
        buildopts  (--with-modules)
        windowsys  x
        daemonp    server-running
doom    version    2.0.9
        build      HEAD -> develop, origin/develop, origin/HEAD b5a279311 2020-01-07 00:14:46 -0500
        dir        ~/.doom.d/
system  type       gnu/linux
        config     x86_64-pc-linux-gnu
        shell      /usr/bin/xonsh
        uname      Linux 5.5.0-rc3master #15 SMP Thu Jan 2 20:04:16 CST 2020 x86_64
        path       (~/.cargo/bin/ ~/.emacs.d/bin/ ~/.cargo/bin/ ~/.node_modules/bin/ ~/.local/lib/ ~/.local/bin/ /usr/local/bin/ /usr/bin/ /bin/ /usr/local/games/ /usr/games/ /opt/Maintance/bin/ /opt/Maintance/bin/2f/ /opt/Maintance/swich/ /usr/bin/mh/ ~/.emacs.d/bin/ ~/projects/yslib/sysroot/usr/bin/ /usr/bin/mh/ ~/.emacs.d/bin/ ~/projects/yslib/sysroot/usr/bin/ /usr/local/libexec/emacs/28.0.50/x86_64-pc-linux-gnu/)
config  envfile    envvar-file
        elc-files  0
[...]

PR #3409, opened 2020-06-18
Gravatars now opt-in. Magit README updated.

t-e-r-m:
I expected to view my commit without delay.

**What actually happened?**
I wasn't connected to the internet and magit waited for the gravatar request to expire before it showed me the commit.

**Screenshot:**
https://i.imgur.com/twPRDAO.jpg

**Steps to reproduce:**
1. View a commit in magit

**Suggested change:**
In `modules/tools/magit/config.el`, change
```lisp
magit-revision-show-gravatars '("^Author:     " . "^Commit:     ")
```
to
```lisp
magit-revision-show-gravatars nil
```
(or remove it because the default is `nil`)

It should be opt-in because it's connecting to an external server.

https://github.com/hlissner/doom-emacs/issues/2333 might be related.

I've updated `magit/README.org` with the deleted line.

### DIFF
--- a/modules/tools/magit/README.org
+++ b/modules/tools/magit/README.org
@@ -12,6 +12,7 @@
 - [[#prerequisites][Prerequisites]]
 - [[#features][Features]]
 - [[#configuration][Configuration]]
+  - [[#enable-gravatars][Enable Gravatars]]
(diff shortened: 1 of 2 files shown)


hlissner:
Makes sense to me. Thanks!

PR #3411, opened 2020-06-18
Remove forced project.el recipe hack

gagbo:
Now that straight.el fixed the bug upstream there's no reason to force it anymore


hlissner:
Thanks!

PR #3412, opened 2020-06-18
Filter out 'eglot backend from doom-thing-at-point

gagbo:
Context : https://github.com/joaotavora/eglot/issues/503

Fixes a bug where having eglot enabled in a buffer will make `SPC *` search the whole project for literally `LSP Identifier at point.`


hlissner:
Thanks!

Issue #3413, opened 2020-06-19
ivy-avy is not loaded by default


Andre0991:
**What did you expect to happen?**
`ivy-avy` is called in the ivy minibuffer when pressing its keybinding (`C-'`).

See https://oremacs.com/swiper/#key-bindings-for-single-selection-action-then-exit-minibuffer

**What actually happened?**
```
C-' is undefined
```

**Additional details:**
My private config: https://github.com/Andre0991/dotfiles/tree/master/.doom.d

Also, `ivy-avy` is not defined.

Evaluating `(require 'ivy-avy)` fixes it, but I'm not sure what's the Doom way of fixing this - I suppose it involves modifying the ivy module.

**Steps to reproduce:**
- Press `SPC SPC`
- Press `C-'`
- `ivy-avy` is not called. 

**System information:**
<details>

```
emacs   version    26.3
        features   JPEG RSVG IMAGEMAGICK GLIB NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES THREADS LCMS2
        build      Aug 30, 2019
        buildopts  (--disable-dependency-tracking --disable-silent-rules --enable-locallisppath=/usr/local/share/emacs/site-lisp --infodir=/usr/local/Cellar/emacs-plus/26.3/share/info/emacs --prefix=/usr/local/Cellar/emacs-plus/26.3 --with-xml2 --without-dbus --with-gnutls --with-imagemagick --with-modules --with-rsvg --with-ns --disable-ns-self-contained)
        windowsys  ns
        daemonp    server-running
doom    version    2.0.9
        build      grafted, HEAD -> develop, origin/develop, origin/HEAD e7b7f67 2020-06-12 17:03:19 -0400
        dir        ~/dotfiles/.doom.d/
system  type       darwin
        config     x86_64-apple-darwin18.5.0
        shell      /usr/local/bin/bash
        uname      Darwin 19.4.0 Darwin Kernel Version 19.4.0: Wed Mar  4 22:28:40 PST 2020; root:xnu-6153.101.6~15/RELEASE_X86_64 x86_64
[...]

PR #3414, opened 2020-06-19
Add `ivy-avy`

Andre0991:
`ivy-avy` makes it possible to select a candidate with `avy` by using `C-'` in an ivy minibuffer.

This functionality used to come with `swiper`, but it was split: https://github.com/melpa/melpa/pull/6951

This could be added as a module flag as well.

Also, I'm not sure if this keybinding should be added to the ivy module documentation.


Andre0991:
Fixes https://github.com/hlissner/doom-emacs/issues/3413

hlissner (approved):

hlissner:
Sorry for the tremendously late reply. Thanks for the PR!

PR #3415, opened 2020-06-19
Fix error when opening REPL from unsupported mode

flatwhatson:
"Wrong number of arguments: (_mode fn . plist), 0"


hlissner:
Thanks!

PR #3416, opened 2020-06-19
Minor scheme module improvements

flatwhatson:
Ping @elais


hlissner:
@elais Because of [this autoload](https://gitlab.com/jaor/geiser/-/blob/master/elisp/geiser.el#L174-175).

@flatwhatson Thanks!

elais:
I see. Works for me.

PR #3420, opened 2020-06-19
Add gitflow and move worktree in magit dispatch

### LINKED ISSUE #3410: Magit keys are not consistent
Open a magit status window and hit `%`. Observe that a magit gitflow pop up is opened.

Open a mgit help window (`?`) and hit `%`. Observe a worktree pop instead.

It's possible that this is intentional, but this seems highly inconsistent to me.

yyadavalli:
- Adds `magit-gitflow-popup` as `%` keybinding in the `magit-dispatch`, mirroring the keybinding in `magit-mode-map`.
- The existing binding `magit-worktree` is moved to `*`

Fixes #3410


hlissner:
Thanks!

PR #3430, opened 2020-06-21
Org-roam: Add keybinding for org-roam-insert-immediate

rajk04:
This function allows the creation of file references in org-roam, without opening org-capture


hlissner:
Thanks!

Issue #3390, opened 2020-06-15
Language Servers hanging over TRAMP


mjlbach:
**What did you expect to happen?**
mspyls and pyright seem to work fine locally, but hang the process when connected via tramp.

See: https://github.com/emacs-lsp/lsp-mode/issues/1741

**What actually happened?**
Tramp hangs indefinitely, language server refuses to start (even if manually invoking the binary via ielm over TRAMP)

**Steps to reproduce:**
1. Connect to server via TRAMP with (python +lsp) enabled
2. Open python file (observe hang)

The following patch seems to fix it, but this messes with doom's lazy initialization

```diff
diff --git a/modules/lang/python/config.el b/modules/lang/python/config.el
index 662c92bd0..875abd26f 100644
--- a/modules/lang/python/config.el
+++ b/modules/lang/python/config.el
@@ -292,10 +292,13 @@ called.")
 
 (use-package! lsp-python-ms
   :when (and (featurep! +lsp) (not (featurep! :tools lsp +eglot)))
-  :after lsp-clients
-  :preface
-  (after! python
-    (setq lsp-python-ms-python-executable-cmd python-shell-interpreter))
+  :hook (python-mode . (lambda ()
+                           (require 'lsp-python-ms)
+                           (lsp)))  ; or lsp-deferred
+  ; :after lsp-clients
+  ;  :preface
+  ;   (after! python
+  ;     (setq lsp-python-ms-python-executable-cmd python-shell-interpreter))
   :init
   ;; HACK If you don't have python installed, then opening python buffers with
   ;;      this on causes a "wrong number of arguments: nil 0" error, because of
 ```
The hook seems to be necessary.

I've reproduced this on 27/28 on mac and linux.

When called with :after python, the highlight callbacks stops working. 

**System information:**
<details><pre>
((emacs
  (version . "28.0.50")
  (features . "XPM JPEG TIFF GIF PNG RSVG CAIRO SOUND DBUS GSETTINGS GLIB NOTIFY INOTIFY LIBSELINUX GNUTLS LIBXML2 FREETYPE HARFBUZZ M17N_FLT LIBOTF ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS LIBSYSTEMD JSON PDUMPER GMP")
  (build . "Jun 14, 2020")
[...]

PR #3437, opened 2020-06-22
Enable projectile-locate-dominating-file on tramp

sggutier:
Essentially this makes projectile work properly (again?) when using tramp.

This causes the following (all of it on tramp):
- If `projectile-locate-dominating-file` is disabled then `projectile-find-file` doesn't work, which also makes `+ivy/projectile-find-file` only able to search in the current directory.
- It allows `projectile-vc` to work again (only on machines with the same OS).
- Projects are now added properly: they now appear with `counsel-projectile-switch-project` (accessible in Open project) in the dashboard.
- LSP autocompletion now includes files from the whole project (LSP still doesn't work properly though, see #3390).

I tested it with Emacs 26.3 and 27.0.91 , both with ssh and sshx.  The infinite loop was no longer present.

### DIFF
--- a/core/core-projects.el
+++ b/core/core-projects.el
@@ -190,16 +190,7 @@ when using many of projectile's command, e.g. `projectile-compile-command',
 This suppresses the error so these commands will still run, but prompt you for
 the command instead."
     :around #'projectile-default-generic-command
-    (ignore-errors (apply orig-fn args)))
-
-  ;; Projectile root-searching functions can cause an infinite loop on TRAMP
-  ;; connections, so disable them.
-  ;; TODO Is this still necessary?
-  (defadvice! doom--projectile-locate-dominating-file-a (file _name)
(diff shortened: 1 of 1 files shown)


sggutier:
`projectile-vc` doesn't work for me using emacs locally on MacOs and a remote Linux server.  

I guess this was because `magit-git-executable` is set in `modules/tools/magit/config.el` with `(executable-find magit-git-executable)`.  For MacOs this evaluates to `/usr/local/bin/git`, but in Linux (Debian) it evaluates to `/usr/bin/git`.  Setting it manually to "git" allows to run magit with projectile through tramp properly.

hlissner:
I'm unable to test this at the moment, so I'll take your word for it until I can. I've also fixed the magit-git-executable issue in 58fa2b1. Thanks for the PR and bringing it to my attention!

PR #3447, opened 2020-06-23
change rust doc lookup to std library

### LINKED ISSUE #3446: Change rust docs lookup to standard library?
https://github.com/hlissner/doom-emacs/blob/99eea1d3e25d8bf6034ff31d231e4ecfc3dc628c/modules/tools/lookup/config.el#L31

Currently the documentation lookup URL points to the Rust Edition Guide (see above). 
I'm a newbie in Rust, so I may be wrong, but the [standard library doc](https://doc.rust-lang.org/std/index.html) may be more useful.

I may submit a pull request.

thevirtuoso1973:
Closes #3446
Just changing the default Rust Doc to be std library instead of editions guide.


hlissner:
Makes sense to me. Thanks!

PR #3449, opened 2020-06-24
tools/lookup: Update Xwidget WebKit browser support

### LINKED ISSUE #3441: [BUG] lookup +xwidget is not working
With `lookup +xwidget`, `SPC s k` opens a documentation with Xwidget browser.

**What actually happened?**
It opens a documentation in a browser window.

**Additional details:**
https://github.com/hlissner/doom-emacs/blob/develop/modules/tools/lookup/autoload/docsets.el#L89
`dash-docs-browser-func` is overridden here.

sei40kr:
Remove :tools/lookup +xwidget and provide `+lookup-xwidget-webkit-open-url-fn` instead so that we can manually configure Xwidget WebKit browser for `+lookup/online` or `+lookup/in-docsets`.

Fixes #3441.


hlissner:
Thanks!

PR #3464, opened 2020-06-25
Switch to emacs mode in mu4e view mode

hexagonal-sun:
Evil masks most of the keybindings that are set by mu4e-view-mode. Since
    this is a read-only buffer anyway use Emacs for that mode.

### DIFF
--- a/modules/email/mu4e/config.el
+++ b/modules/email/mu4e/config.el
@@ -63,6 +63,11 @@
   ;; set mail user agent
   (setq mail-user-agent 'mu4e-user-agent)
 
+  ;; evil masks most mu4e-view-mode key bindings. Switch to Emacs mode for
+  ;; this.
+  (when (featurep! :editor evil)
+    (add-hook! 'mu4e-view-mode-hook #'evil-emacs-state))
+
   ;; Use fancy icons
   (setq mu4e-use-fancy-chars t
         mu4e-headers-draft-mark '("D" . "")
--- a/modules/email/mu4e/config.el
+++ b/modules/email/mu4e/config.el
@@ -66,7 +66,7 @@
   ;; evil masks most mu4e-view-mode key bindings. Switch to Emacs mode for
   ;; this.
   (when (featurep! :editor evil)
-    (add-hook! 'mu4e-view-mode-hook #'evil-emacs-state))
+    (add-hook 'mu4e-view-mode-hook #'evil-emacs-state))
 
   ;; Use fancy icons
   (setq mu4e-use-fancy-chars t


modules/email/mu4e/config.el
```diff
@@ -63,6 +63,11 @@
   ;; set mail user agent
   (setq mail-user-agent 'mu4e-user-agent)
 
+  ;; evil masks most mu4e-view-mode key bindings. Switch to Emacs mode for
+  ;; this.
+  (when (featurep! :editor evil)
+    (add-hook! 'mu4e-view-mode-hook #'evil-emacs-state))
```
hlissner, line 69: Does replacing these two lines with `(set-evil-initial-state 'mu4e-view-mode 'emacs)` work?

hexagonal-sun: Unfortunately not; it didn't work for some reason.  I can add a comment explaining that I've tried that if it helps?

hlissner: No need. I'll look for a better solution another time. Instead, replace `add-hook!` with `add-hook` and this will be ready to merge.

hexagonal-sun: Done.

hexagonal-sun:
@hlissner Are you happy for this to get merged?  Thanks!  :+1:

hlissner:
Thanks!

agraul:
In my experience the [evil-collection](https://github.com/emacs-evil/evil-collection/blob/master/modes/mu4e/evil-collection-mu4e.el) keybindings work good enough - at least so far I did not mind them. This change surprised me, not sure about other mu4e users, but I will revert this in my private config.

hexagonal-sun:
@agraul I found that several of my most commonly used key bindings where masked even when in evil mode.  Case in point is `mu4e-view-scroll-up-or-next`.  Since these bindings didn't work I looked at other email modes to see if there is a convention on how to fix it and took heed from the wanderlust config: https://github.com/hlissner/doom-emacs/blob/576eb40b501213b4707b542b6ab5d13c1538414f/modules/email/wanderlust/config.el#L73

I would also propose that since `mu4e-view-mode` is a read-only view it makes sense that those bindings should follow what is documented in the mu4e manual.

agraul:
I see. I personally expect to be able to navigate inside the mail buffer with the same keybindings I use in other buffers, e.g. scrolling up and down with `C-f`, `C-b`,  `C-d`, `C-u`. In my opinion it would be better to take the missing keybindings (e.g. `mu4e-view-scroll-up-or-next`) upstream and add them to evil-collection than removing the evil keybindings. 

> I would also propose that since mu4e-view-mode is a read-only view it makes sense that those bindings should follow what is documented in the mu4e manual.

It is still a buffer of text, just like opening a file read-only. I don't get why it should be handled differently from other read-only files. 

In the end, this is a subjective matter. Either way is good as long as the user is happy and I don't mind having this reverted in my personal config. I just wonder if this (breaking) change is necessary in Doom upstream.

hlissner:
@agraul Then how about putting it in motion mode?

```elisp
(add-hook 'mu4e-view-mode-hook #'evil-motion-state)
```

Evil's motion mode is meant as a "normal-lite". Evil's motions should be available, without many of the keybinds for extraneous plugins.

agraul:
Motion mode overrides quite a few of the [default mu4e keybindings](https://www.djcbsoftware.nl/code/mu/mu4e/MSGV-Keybindings.html#MSGV-Keybindings), but does not provide the evil-collections keybindings. I think this is actually the worst of the three discussed options.

Someone else on Discord asked why this was decided so it seems I am not the only one that got surprised. I still think that using evil-collections' mu4e bindings should be the default for evil users. It can be overridden for evil users who prefer emacs bindings for mu4e in their personal configs and I *think* this is in more line with expectations than the other way around. Again, for me personally it is not that big of an issue and I just have it reverted, just thinking that the default should be the other way around.

tecosaur:
I just mask the few irritating evil bindings I've found, see https://tecosaur.github.io/emacs-config/config.html#viewing-mail,code--5

hlissner:
Understood. I've reverted it in 7a276db.

PR #3466, opened 2020-06-25
Clarify ace-window font size and add example

carloscheddar:
I clarified the pros and cons of the different window packages and added an example to show users how to customize the ace-window letter.

### DIFF
--- a/modules/ui/window-select/README.org
+++ b/modules/ui/window-select/README.org
@@ -48,7 +48,15 @@ The first character of the buffers changes to a highlighted, user-selectable
 character.
 
  + Pros: the content of the buffers are always visible.
- + Cons: The displayed characters are small and difficult to see.
+ + Cons: Requires custom configuration to show large letters.
+ 
+**** Custom font-face example
+This changes the ace-window display to show a white letter with a red background. The box attribute adds some padding.
+#+BEGIN_SRC elisp
+  (custom-set-faces
+   '(aw-leading-char-face
+     ((t (:foreground "white" :background "red" :weight bold :height 2.5 :box (:line-width 10 :color "red"))))))
+#+END_SRC 
 
 ** switch-window
 Replaces the entire buffer with large letters.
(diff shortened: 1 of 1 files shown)


modules/ui/window-select/README.org
```diff
@@ -48,7 +48,15 @@ The first character of the buffers changes to a highlighted, user-selectable
...
  + Pros: the content of the buffers are always visible.
- + Cons: The displayed characters are small and difficult to see.
+ + Cons: Requires custom configuration to show large letters.
+ 
+**** Custom font-face example
+This changes the ace-window display to show a white letter with a red background. The box attribute adds some padding.
+#+BEGIN_SRC elisp
+  (custom-set-faces
+   '(aw-leading-char-face
+     ((t (:foreground "white" :background "red" :weight bold :height 2.5 :box (:line-width 10 :color "red"))))))
```
hlissner, line 58: Use `custom-set-faces!` instead:

```elisp
(custom-set-faces!
  '(aw-leading-char-face 
    :foreground "white" :background "red" 
    :weight bold :height 2.5 :box (:line-width 10 :color "red")))
```

modules/ui/window-select/README.org
```diff
@@ -48,7 +48,15 @@ The first character of the buffers changes to a highlighted, user-selectable
 character.
 
  + Pros: the content of the buffers are always visible.
- + Cons: The displayed characters are small and difficult to see.
+ + Cons: Requires custom configuration to show large letters.
```
hlissner, line 51: I think this should be revised to "The display characters are small and difficult to see (see below for a way to enlarge them)", because the new message obscures the problem.

carloscheddar: Oh that makes sense. I was debating keeping that pros and cons list after that change but this message makes things clearer.

modules/ui/window-select/README.org
carloscheddar: Updated here https://github.com/hlissner/doom-emacs/pull/3466/commits/2db39eafa1e80ef981ffc76db798b047c00a5eea

modules/ui/window-select/README.org
carloscheddar: Updated here https://github.com/hlissner/doom-emacs/pull/3466/commits/2db39eafa1e80ef981ffc76db798b047c00a5eea

hlissner:
Thanks!

PR #3470, opened 2020-06-26
Emacs bindings: Conditionally bind workspace keys

tprk77:
Adds a missing `featurep!` conditional to the Emacs bindings. The check is already in place for the Evil bindings.


tprk77:
Thanks for the feedback, you're totally right.

hlissner:
Thanks!

PR #3473, opened 2020-06-26
Update :tools lsp-ivy

albttx:
Upgrate lsp-ivy version to latest commit

I need the last version for setting true to `lsp-ivy-show-symbol-filename`, could became a default value in :lsp if you wish :)


hlissner:
The rest of the module needed a bump, and bump commits must follow a particular message format to be accepted (so they can be read by future CI), so I've done so in fe37bfc. Thanks for bringing it to my attention!

PR #3476, opened 2020-06-27
Add package org-autolist

yyadavalli:


hlissner:
Thanks for the PR, but I cannot accept it. This workflow already exists with:

+ `C-RET` (appends a new plain list item)
+ `C-S-RET` (prepends a new plain list item),
+ `TAB` promotes the current list item
+ `S-TAB` demotes the current list item

yyadavalli:
Awesome, sorry didn't realize there is an existing workflow before opening this.

PR #3478, opened 2020-06-27
Bump :lang rust

brotzeit:
brotzeit/rustic@f7a94c4 -> brotzeit/rustic@7c0e113

### DIFF
--- a/modules/lang/rust/packages.el
+++ b/modules/lang/rust/packages.el
@@ -1,6 +1,6 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; lang/rust/packages.el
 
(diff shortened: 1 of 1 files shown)


brotzeit:
There were complains from users who are using the non-lsp flycheck checker with the stable toolchain.

I added the new defcustom ´rustic-flycheck-clippy-params´. Question is what should be the default value. Should it be configured for stable or nightly toolchain. I decided to include the option "-Zunstable-options" which is only available on nightly. But maybe you decide otherwise for doom-emacs.

hlissner:
> I decided to include the option "-Zunstable-options" which is only available on nightly. But maybe you decide otherwise for doom-emacs.

I think that's a sensible default. Thanks again for the help!

PR #3481, opened 2020-06-27
tools/magit: map zt/zb motions in magit-mode

vednoc:
Hi!

This PR adds `zt` and `zb` motions to magit-mode since these two keybinds aren't mapped to anything by default, and `zz` motion already exists. I've been using the following snippet in my config for a while:

``` emacs-lisp
(map! :map magit-mode-map
  :nv "zt" #'evil-scroll-line-to-top
  :nv "zb" #'evil-scroll-line-to-bottom)
```


hlissner:
Thanks!

PR #3484, opened 2020-06-29
lang/org: Format a source block

sei40kr:
## Description

Format an Org source block with `SPC m f`.


hlissner:
I've merged a slightly modified version of this PR in 72da7a3 directly into the format module. Let me know if it doesn't work. Thanks for bringing it to my attention.

PR #3485, opened 2020-06-29
core: use custom.el instead of config.el for custom-file

peterhoeg:


peterhoeg:
The `nofix` label is actually not correct. 

If your configuration is kept in an .org file, `config.el` will be continuously rewritten when the source is tangled thus overwriting all customizations that have previously been written to `config.el`. This matters for stuff like white listed directory local variables which will otherwise make emacs prompt again and again for user consent to set them.

hlissner:
I had intended to justify the `nofix` sooner, but on second thought it's less hassle to do things this way. Thanks for the PR!

PR #3500, opened 2020-07-02
Add automatic fold after save to TeX-fold

MYDavoodeh:
Since the Tex-fold-buffer bind (`C-c C-o C-b`) is really clunky and preferably we (I?) want the buffer as folded as possible, I see this as an optimal solution.

### DIFF
--- a/modules/lang/latex/config.el
+++ b/modules/lang/latex/config.el
@@ -88,6 +88,8 @@ If no viewers are found, `latex-preview-pane' is used.")
   ;; Fold after cdlatex macro insertions
   (advice-add #'cdlatex-math-symbol :after #'+latex-fold-last-macro-a)
   (advice-add #'cdlatex-math-modify :after #'+latex-fold-last-macro-a)
+  ;; Fold after saving
+  (add-hook! 'after-save-hook TeX-fold-buffer)
   ;; Fold after snippets
   (when (featurep! :editor snippets)
     (add-hook! 'TeX-fold-mode-hook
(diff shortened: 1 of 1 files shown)


modules/lang/latex/config.el
```diff
@@ -88,6 +88,8 @@ If no viewers are found, `latex-preview-pane' is used.")
   ;; Fold after cdlatex macro insertions
   (advice-add #'cdlatex-math-symbol :after #'+latex-fold-last-macro-a)
   (advice-add #'cdlatex-math-modify :after #'+latex-fold-last-macro-a)
+  ;; Fold after saving
+  (add-hook! 'after-save-hook TeX-fold-buffer)
```
hlissner, line 92: Should be `(add-hook 'after-save-hook #'TeX-fold-buffer)`, otherwise it throws a "void-variable: TeX-fold-buffer" error.

ymarco:
I'm not sure *I* always want the buffer as folded as possible. Many times I save + compile to see what my changes made to the pdf. In case the cursor was in something that would be folded (say a footnote, or in the middle of a macro), folding would kick it out of there. That's also why `+latex-fold-last-macro-a` does a wierd search instead of just folding the current paragraph - to have the smallest chance of getting in your way.


Personally, I just use the `TeX-fold-paragraph` keybind (`SPC m f`) from time to time.

MYDavoodeh:
@tecosaur
I never encountered such a large file...

How about we define a defcustom for it can be turned off/on from the user config.el?

ymarco:
> How about we define a defcustom for it can be turned off/on from the user config.el?

Seems like an overcomplication to me, instead of leaving it to the user to add the after-save-hook.

hlissner:
Since it seems to boil down to personal preference, has possible performance implications, and is trivial for users to add to `after-save-hook`, I'll turn down this PR. Thanks for bringing it to my attention, in any case!

PR #3513, opened 2020-07-04
Add pretty symbols to php-mode

MYDavoodeh:
I'm not pretty sure if Types are redundant or not since Php is not so much around types... But I added them just in case.

### DIFF
--- a/modules/lang/php/config.el
+++ b/modules/lang/php/config.el
@@ -61,9 +61,7 @@
     :or "||" :or "or"
     :for "for"
     :return "return"
-    :yield "use"
-    ;; Other
-    :dot "."))
+    :yield "use"))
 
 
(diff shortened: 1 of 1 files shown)


modules/lang/php/config.el
```diff
@@ -43,7 +43,25 @@
...
+    :int "int" :float "float"
+    :str "string"
+    :bool "list"
+    ;; Flow
+    :not "!"
+    :and "&&" :and "and"
+    :or "||" :or "or"
+    :for "for"
+    :return "return"
+    :yield "use"))
```
hlissner, line 64: Please move this up, so it's together with the other `set-*!` autodef calls.

hlissner:
Thanks!

PR #3518, opened 2020-07-05
Update Mac Install for emacs-plus

MarkPochert:
The parameter is wrong. It must be --with-modern-cg433n-icon and not --with-modern-icon-cg433n

You can check it on the github repo site https://github.com/d12frosted/homebrew-emacs-plus


hlissner:
Thanks!

PR #3519, opened 2020-07-05
Update getting_started.org

IsaacDempsey:
Seems like the words are in the wrong order - see here: https://github.com/d12frosted/homebrew-emacs-plus


marcus-crane:
For reference, the flag only changed recently as part of https://github.com/d12frosted/homebrew-emacs-plus/commit/c9a41a3abc88edd792bf282a7e3ee0637af19b5f

nickcernis:
Appears to be a duplicate of https://github.com/hlissner/doom-emacs/pull/3518.

PR #3521, opened 2020-07-06
Pin dash version

TimWSpence:
Fixes "wrong type argument listp" errors in lsp mode

See https://github.com/emacs-lsp/lsp-mode/issues/1207

### DIFF
--- a/modules/tools/lsp/config.el
+++ b/modules/tools/lsp/config.el
@@ -14,3 +14,5 @@ working on that project after closing the last buffer.")
 (if (featurep! +eglot)
     (load! "+eglot")
   (load! "+lsp"))
+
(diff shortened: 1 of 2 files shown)


TimWSpence:
Sorry if dash should be pinned somewhere else. I put it in the lsp module as that is where I encountered the issue but dash is such a generic package it should maybe be pinned somewhere else

hlissner:
Thanks for the PR, but I cannot accept it.

1. The issue has already been resolved upstream.
2. Dash is a dependency for so many packages that it doesn't make sense to pin it in the LSP module.
3. A `use-package!` block eagerly loads dash at startup and isn't necessary for pinning.

PR #3520, opened 2020-07-06
Fix brew install icon argument

acaire:
As per https://github.com/d12frosted/homebrew-emacs-plus/pull/246

PR #3522, opened 2020-07-06
Fix wrong emacs-plus installation option

littlemove:


Matsa59:
There is already a PR for that :/ 

https://github.com/hlissner/doom-emacs/pull/3520

littlemove:
I see. Thanks. I also see that there's even an older one: #3519 

I'll close this one. I hope #3519 gets merged soon...

PR #3526, opened 2020-07-07
Bump doom-modeline

ericdallo:


hlissner:
The package was bumped in 8ad7f64, because the PR's commit message doesn't conform to the format required for bump commits.

Side note: in the future, please include an explanation for why a package should be bumped.

PR #3538, opened 2020-07-11
Add pretty-symbols for Shell-mode

MYDavoodeh:


modules/lang/sh/config.el
```diff
@@ -50,7 +50,21 @@
...
+    ;; Types
+    :true "true" :false "false"
+    ;; Flow
+    :not "!"
+    :and "&&" :or "||"
+    :in "in"
+    :for "for"
+    :return "return"
+    ;; Other
+    :dot "." :dot "source"))
```
hlissner, line 67: Please move this up so it's together (below) with the other `set-*!` autodef calls.

MYDavoodeh: Thank you for the call.
Next time I will be more careful about the arrangements!

hlissner:
Thanks!

PR #3543, opened 2020-07-12
remove doctor warning message for optional vterm dependency

jtpavlock:
According to [emacs-libvterm](https://github.com/akermu/emacs-libvterm#requirements), `libvterm` is an optional dependency and not required to compile. 

I am seeing this manifest in my system on Ubuntu 20.04 with `libvterm-dev` installed. It doesn't include the `vterm-ctrl` binary, but `vterm` works just fine in my install. However, `doom doctor` persists with the warning message.


hlissner:
Makes sense to me. Thanks!

PR #3547, opened 2020-07-13
Add a flag for disabling doom-themes' org fixes.

drwells:
I want a more vanilla doom experience so I really need this module flag to prevent the default org settings from being overwritten.

As you can see, this is my first patch of doom :tada: Please let me know if there is anything else this PR needs.

### DIFF
--- a/modules/ui/doom/README.org
+++ b/modules/ui/doom/README.org
@@ -25,8 +25,17 @@ This module gives Doom its signature look: powered by the =doom-one= theme
 + "Thin bar" fringe bitmaps for ~git-gutter-fringe~
 + File-visiting buffers are slightly brighter (thanks to solaire-mode)
 
+This module also enables some display enhancements for =org-mode= included with
+the =doom-themes= package, such as:
++ Changing the color of =TODO= items that are marked done.
++ Fontifying hashtags (=#=) and at-tags (=@=)
++ Hiding leading stars
+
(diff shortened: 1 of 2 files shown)


hlissner:
Thanks for the PR, but I cannot accept it. I am resistant to adding new flags for trivial configuration. i.e. this could be accomplished with a single line of code:

```elisp
(remove-hook 'doom-load-theme-hook #'doom-themes-org-config)
```

I also want to avoid setting a precedent where we have flags for similarly trivial configuration. i.e. This could invite PRs for `+disable-neotree-config` and `+disable-treemacs-config` flags.

> to prevent the default org settings from being overwritten

How are you configuring `org`? Your private settings should have precedence unless you're applying them before doom-themes (or `org`) is loaded. i.e. Wrap them in `(after! org ...)` and ensure you aren't eagerly loading org at startup.

drwells:
> I am resistant to adding new flags for trivial configuration. i.e. this could be accomplished with a single line of code:

This makes sense and in retrospect I agree completely.

> (remove-hook 'doom-load-theme-hook #'doom-themes-org-config)

Oh, I see: I can add this in `.doom.d/config.el` to remove the hook before I load anything. Makes sense.

> How are you configuring org?

I have no org configuration: in particular I do not load the org module or execute any org-related code at all (I started my doom conversion by disabling absolutely everything I could).

Thanks for the help!

PR #3548, opened 2020-07-13
Add the BUG and XXX tags to hl-todo

MYDavoodeh:
Been crawling in [Wikipedia](https://en.wikipedia.org/wiki/Comment_(computer_programming)#Tags) and found XXX and BUG tags interesting.


hlissner:
Thanks!

PR #3551, opened 2020-07-13
Fix deadlinks on "Getting Started" documentation file

ralacerda:
There were a few links using `[[*header-name][Header Name]]`, resulting in dead links when rendered by github.

I changed then to `[[#header-name][Header Name]]`

I looked into others documentation files, and found no other instance of `[[*`


hlissner:
Do these links still work when you follow them in Emacs?

ralacerda:
It works if you click on it and with `org-open-at-point`

hlissner:

PR #3554, opened 2020-07-14
Add version (27) to macOS Homebrew guideline

lordgrenville:
`brew install emacs-plus` can take a version number (@27 or @28, otherwise defaults to 26). Since 27+ is recommended, I thought I'd add this. Tested on my machine (MBP running Catalina 10.15.5) and looks fine.

### DIFF
--- a/docs/getting_started.org
+++ b/docs/getting_started.org
@@ -243,8 +243,8 @@ to least recommended for Doom (based on compatibility).
 - [[https://github.com/d12frosted/homebrew-emacs-plus][emacs-plus]]:
   #+BEGIN_SRC bash
   brew tap d12frosted/emacs-plus
-  brew install emacs-plus --with-modern-cg433n-icon
-  ln -s /usr/local/opt/emacs-plus/Emacs.app /Applications/Emacs.app
+  brew install emacs-plus@27 --with-modern-cg433n-icon
+  ln -s /usr/local/opt/emacs-plus@27/Emacs.app /Applications/Emacs.app
   #+END_SRC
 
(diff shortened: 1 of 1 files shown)


docs/getting_started.org
```diff
@@ -243,8 +243,8 @@ to least recommended for Doom (based on compatibility).
 - [[https://github.com/d12frosted/homebrew-emacs-plus][emacs-plus]]:
   #+BEGIN_SRC bash
   brew tap d12frosted/emacs-plus
-  brew install emacs-plus --with-modern-cg433n-icon
-  ln -s /usr/local/opt/emacs-plus/Emacs.app /Applications/Emacs.app
+  brew install emacs-plus@27 --with-modern-cg433n-icon
+  ln -s /usr/local/opt/emacs-plus@27/Emacs.app /Applications/Emacs.app
```
d1egoaz, line 247: wondering if this icon should also be removed `--with-modern-cg433n-icon`

default icon is:

which is very nice, if user wants a different icon they could use any `--with-modern-cg433n-icon` option

lordgrenville: Agree! Not sure what the default is. Peronsally I think `--with-modern-sexy-v1-icon` is way better anyway ;)

image

hlissner: Remove `--with-modern-cg433n-icon`. The default seems reasonable and it's one less thing to maintain.

hlissner:
Thanks for the PR, but I can't accept it as it is. You'll notice instructions for other OSes/distros focus on installing Emacs 26.x, and 27 instructions are included in post-script -- if an easy method exists for that system. I'll incorporate more 27 instructions later, but until we drop 26 support, I don't want to sacrifice the 26-centricity of our guide.

PR #3557, opened 2020-07-14
Set company-capf as backend for ledger-mode.

johanthoren:
This provides pretty, and very useful completion of accounts.
Inspired by the Finance layer for Spacemacs.

Signed-off-by: Johan Thoren <johan@thoren.dev>

### DIFF
--- a/modules/lang/ledger/config.el
+++ b/modules/lang/ledger/config.el
@@ -12,6 +12,8 @@
             "hledger"
           "ledger"))
 
+  (set-company-backend! 'ledger-mode 'company-capf)
+
   (defadvice! +ledger--check-version-a (orig-fn)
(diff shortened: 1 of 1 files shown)


modules/lang/ledger/config.el
```diff
@@ -12,6 +12,9 @@
             "hledger"
           "ledger"))
 
+  (when (featurep! :completion company)
```
hlissner, line 15: Please remove the `(when (featurep! :completion company) ...)`. `set-company-backend!` is an autodef. It will no-op if `:completion company` isn't enabled, so this check is redundant.

johanthoren: Good point, thanks for pointing that out.

hlissner:
Thanks!

PR #3562, opened 2020-07-15
Add support for lua-langserver in lsp-mode

gagbo:
Register a `lsp-mode` client for `lua-langserver`. 

This could go upstream eventually, but the quickest patch to get it working was to use the same doom utils as the `eglot` config


hlissner:
Thanks!

PR #3565, opened 2020-07-15
Only fallback to hledger if ledger isn't installed

JRigotti:
Most of the report commands fail with `hledger`, so if the user has both programs installed it should use ledger as default (or simply not set another value).

### DIFF
--- a/modules/lang/ledger/config.el
+++ b/modules/lang/ledger/config.el
@@ -7,10 +7,9 @@
         ledger-mode-should-check-version nil)
 
   :config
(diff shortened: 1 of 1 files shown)


hlissner:
I ended up doing away with the hledger->ledger fallback altogether. Another plugin (hledger-mode) exists. Perhaps one day I'll extend this module to support it (if it works), but for now, if hledger support is incomplete for ledger-mode, I'd rather not depend on it at all. Thanks for bringing it to my attention!

PR #3570, opened 2020-07-16
Fix emacs binding for toggling org-tree-slide-mode

Bounga:
org-tree-slide-mode toggling binding was calling a non-existing (obsolete?) function.

It now call `org-tree-slide-mode` as evil bindings does.


hlissner:
Thanks!

PR #3571, opened 2020-07-16
Make gpg files incompatible with undo-fu session

tylerware:
Since gpg files are explicitly meant to be encrypted it doesn't make a lot of sense to store undo data unencrypted. This commit adds gpg files to the `undo-fu-session-incompatible-files` list.


hlissner:
Good idea! Thanks!

PR #3574, opened 2020-07-17
Support fcitx5

bennyyip:
fcitx5 uses `fcitx5-remote` to control its status.
see:
https://github.com/cute-jumper/fcitx.el/issues/47
https://github.com/cute-jumper/fcitx.el/commit/c89c90cca6dc148647d8bc7b4766c84ddd988e34


hlissner:
A slightly different solution was needed, which was implemented in e567cc5. Thanks for bringing it to my attention however!

PR #3580, opened 2020-07-17
remove doom/open-manual, fix doom/describe-module

LoganBarnett:
The `doom/open-manual function` seems to no longer exist. The `doom/describe-module` key binding is fixed since `SPC h d` is the doom help prefix. Under this prefix is `doom/describe-module`.

I am not aware if `doom/open-manual` was replaced by something else, but I'm happy to include that in the PR if someone can point me to it :)

Thanks for all the work on Doom Emacs!


hlissner:
A few more updates were needed for that section, which were done in 9518852 and fc2634d. Also, `doom/open-manual` is now `doom/help`. In any case, thanks for bringing it to my attention!

PR #3583, opened 2020-07-18
Override * in magit-status-mode-map

agraul:
`*` is bound to magit-worktree in magit-dispatch, but that requires opening magit-dispatch before interacting with Git worktrees (e.g. by pressing `?`). With this change, `*` is available anywhere in the magit-status buffer. 

The downside is that it can't be used to `evil-ex-search-word-forward` in magit-status anymore.


hlissner:
Thanks for the PR, but since `? *` exists I'd rather not sacrifice evil-ex-search-word-forward. I'll leave it to users to make this choise.

PR #3584, opened 2020-07-18
update indent-visualisation on indent-style change

Mic92:
Requires `doom-highlight-non-default-indentation-h` to be idempotent though. Can you help me with this one? Right now it will just highlight both tabs and spaces after a switch.

### DIFF
--- a/core/autoload/text.el
+++ b/core/autoload/text.el
@@ -286,6 +286,7 @@ Respects `require-final-newline'."
   "Switch between tabs and spaces indentation style in the current buffer."
   (interactive)
   (setq indent-tabs-mode (not indent-tabs-mode))
+  (doom-highlight-non-default-indentation-h)
   (message "Indent style changed to %s" (if indent-tabs-mode "tabs" "spaces")))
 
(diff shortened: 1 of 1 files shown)


hlissner:
Just cycle whitespace-mode, if it's active, instead of calling `doom-highlight-non-default-indentation-h`.

```elisp
(when whitespace-mode
  (whitespace-mode -1)
  (whitespace-mode +1))
```

Mic92:
> Just cycle whitespace-mode, if it's active, instead of calling `doom-highlight-non-default-indentation-h`.
> 
> ```emacs-lisp
> (when (bound-and-true-p whitespace-mode)
>   (whitespace-mode -1)
>   (whitespace-mode +1))
> ```

Thanks, I just tried that but it just stays to the current default instead of switching.

Mic92:
don't know how to solve this.

luisgerhorst:
What about turning doom-highlight-non-default-indentation into a minor mode? This would also allow users to easily disable it (I would love to use it but creating "fix indentation" commits on large legacy files is unfortunately not an option as it completely break git blame).

PR #3594, opened 2020-07-20
Disable flycheck when racket-xp-mode is enabled

sorawee:
Enabling both flycheck and racket-xp-mode would create redundant
error tooltips. Since racket-xp-mode provides a higher quality
error message, prefer it over flycheck.


sorawee:
Let me know if there's anything I should change. If this is something that should be in personal config, not Doom Emacs, I'm OK to close the PR, too (though the redundant error tooltip would then be a problem that needs to get fixed).

hlissner:
Thanks for the PR! A little more was needed for this PR, so I did it in e697916 (with credit given. Thanks for bringing it to my attention!

PR #3597, opened 2020-07-20
Ensure org-roam-directory is a directory

### LINKED ISSUE #3596: org-roam-directory missing trailing slash(/) in Windows

`org-roam` require `org-roam-directory` set with a  "/" . the following forms will return wrong path if `org-roam-directory`  without trailing "/".  That can cause `org-roam-jump-to-index` wrong.
```emacs lisp
(org-roam--get-index-path)
``` 
e.g.  when `org-directory` is "E:/emacs/org", `org-roam-directory`  will be "E:/emacs/org/roam"(need trailing /),`(org-roam--get-index-path)` will return "E:/emacs/org/roamindex.org"

gagbo:
Fixes #3596


hlissner:
Thanks!

Issue #3266, opened 2020-05-31
Literate config + pdf module causes issue when config.org is tangled


RBckmnn:
**What did you expect to happen?**
On a fresh install, when saving a literate config.org, the file should be tangled correctly.

**What actually happened?**
On a fresh install, when saving a literate config.org, interacting with the prompt asking to build the epdfinfo binary causes an error (see below for backtrace and precise instructions to reproduce).
Subsequently, most of the keybindings in emacs are lost. In particular it is no longer possible to enter any text in the minibuffer or when in insert state.

**Steps to reproduce:**
1. Start with a fresh install of doom emacs, enable literate config and the pdf module
2. Open config.org, make some changes, then save the file
3. A prompt is displayed claiming that epdfinfo needs to be (re)build, asking if you want to build now.
4. Answer either yes or no or abort using C-g
5. An error with the backtrace attached below occurs

**Backtrace**
![backtrace](https://user-images.githubusercontent.com/53320014/83363124-47e80a80-a397-11ea-92ec-8765dab6e44c.png)

I apologize for attaching the backtrace as an image but after the error occurred, I'm no longer able to issue any command in emacs to copy text.

I might also be worth mentioning, that invoking `pdf-tools-install` directly also leads to the above prompt. The error however does not occur in this situation and the binary can be build successfully.

Please let me know if you need more information.

Cheers!

**System information:**
<details>

```
emacs   version    26.3
        features   XPM JPEG TIFF GIF PNG RSVG SOUND GPM DBUS GLIB NOTIFY ACL GNUTLS LIBXML2 FREETYPE XFT ZLIB GTK3 X11 XDBE XIM MODULES THREADS XWIDGETS LCMS2
        build      Jan 03, 2020
[...]

PR #3599, opened 2020-07-20
fix for #3266: wrong-type-arg stringp from format

hperrey:
Literate configurations trigger issue #3266 when tangling on interactive save and epdfinfo is missing. For some reason, pdf-tools-install's prompt for installation ends up calling `message(nil)` which our `letf!`version does not handle well. My patch ignores these cases, which results in the prompt working and both epdfinfo installation and tangling succeeding when saving a literate config.

### DIFF
--- a/modules/config/literate/autoload.el
+++ b/modules/config/literate/autoload.el
@@ -27,7 +27,7 @@ byte-compiled from.")
              (and (require 'ob-tangle)
                   (letf! (defun message (msg &rest args)
                            (when msg
-                           (print! (info "%s") (apply #'format msg args))))
+                             (print! (info "%s") (apply #'format msg args))))
                     (org-babel-tangle-file org dest))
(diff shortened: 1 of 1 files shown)


modules/config/literate/autoload.el
```diff
@@ -26,7 +26,8 @@ byte-compiled from.")
              ;; would load all of :lang org (very expensive!).
              (and (require 'ob-tangle)
                   (letf! (defun message (msg &rest args)
-                           (print! (info "%s") (apply #'format msg args)))
+                           (when msg
+                           (print! (info "%s") (apply #'format msg args))))
```
hlissner, line 30: Please indent `print!` with two more spaces.

hperrey:
Good catch! Sorry for missing that -- fixed!

hlissner:
Thanks!

PR #3600, opened 2020-07-20
change pinned agda version

citrusmunch:
Previous pin copies from agda/agda: `src/data/emacs-mode/agda2-mode.el` which checks if version 2.6.2 is installed, giving errors on the stable release version.

Pin the agda version at the v2.6.1 release.


hlissner:
Thanks for the PR. I've bumped the module myself, in ddc3a95, because bump commits must follow a particular message format so that they're machine-readable by future CI. In any case, thanks for bringing it to my attention!

PR #3610, opened 2020-07-22
lang/common-lisp: fix SPC m ; regression

paulapatience:
Commit a9bf0b89851af0a6a1f763d8e4053502aa04f74d replaced call-interactively with funcall-interactively in λ!!, which breaks SPC m ; for two reasons. First, function sly presumes it was called interactively when its INTERACTIVE argument is non-nil, but funcall-interactively bypasses the interactive form, changing only the result of called-interactively-p (which sly eschews). Second, sly does not accept the prefix argument as an argument; instead, it directly checks current-prefix-arg.

### DIFF
--- a/modules/lang/common-lisp/config.el
+++ b/modules/lang/common-lisp/config.el
@@ -89,7 +89,10 @@
         (:localleader
          :map lisp-mode-map
          :desc "Sly"          "'" #'sly
-         :desc "Sly (ask)"    ";" (λ!! #'sly '-)
+         :desc "Sly (ask)"    ";" (lambda ()
+                                    (interactive)
+                                    (let ((current-prefix-arg '-))
+                                      (call-interactively #'sly)))
          :desc "Expand macro" "m" #'macrostep-expand
          (:prefix ("c" . "compile")
           :desc "Compile file"          "c" #'sly-compile-file


paulapatience:
Although I do prefer the solution in commit 9073a5d4028d2b0939f166057621f0f24739439c to mine above, it does not entirely address the issue. As I mentioned above, funcall-interactively bypasses the interactive form (and thus the default interactive arguments) of interactive functions. Here, sly expects `nil nil t`, so the correct mapping would be:

```elisp
         :desc "Sly (ask)"    ";" (cmd!! #'sly '- nil nil t)
```

Function call-interactively truly acts as an interactive call, but unlike funcall-interactively (and I presume for this reason you favor the latter), it has no way to pass on any (non-prefix) arguments to the called function.

The documentation for called-interactively-p explicitly recommends accepting an argument in the interactive function that will signal its interactiveness rather than relying on called-interactively-p. This presumably means funcall-interactively is discouraged in favor of call-interactively.

Therefore, I propose that we have a version of cmd!! which converts noninteractive functions (or interactive functions requiring customized arguments that an interactive call has no way to provide) into interactive functions with a mere funcall that accepts any arguments required, and a version of cmd!! which calls call-interactively and handles only the prefix argument (as cmd!! used to do before commit a9bf0b89851af0a6a1f763d8e4053502aa04f74d). In this way, we need not manually verify the arguments that interactive functions in the style of sly require, and also we can make interactive any function that was not so initially (and in which case funcall-interactively would have been misleading and subjectively incorrect).

paulapatience:
Commit 2441d28ad15a9a2410f8d65641961ea5b1d39782 has resolved the problem (and once again with a better solution than mine).

hlissner:
Ah, sorry for the late response: aye, I realized the points you mentioned in a discussion with another user on Discord (which I had wrongfully assumed was you, for some reason). The initial change to `cmd!!` was an oversight on my part. 2441d28 reverts it to suit its original purpose (to bake the prefix arg into a partial application of an interactive function).

PR #3613, opened 2020-07-22
Support number highlighting for long Elixir numbers

jsmestad:
## Background
The built-in Elixir formatter uses an underscore to delimit long numbers. However, the generic regex for `highlight-numbers-mode` does not recognize them. This adds a mode-specific regex for `elixir-mode`.

#### Before
Before

#### After
After

## Note

I have also added direct support for font face on numbers to elixir-editors/emacs-elixir#459 if you have any feedback on that as well as I don't want things to conflict with `highlight-numbers-mode`

### DIFF
--- a/modules/lang/elixir/config.el
+++ b/modules/lang/elixir/config.el
@@ -3,6 +3,14 @@
 (after! projectile
   (add-to-list 'projectile-project-root-files "mix.exs"))
 
+(after! highlight-numbers
+  (puthash 'elixir-mode
+    (rx (and symbol-start
+            (? "-")
+            (+ digit)
+            (0+ (and "_" (= 3 digit)))
+            symbol-end))
+    highlight-numbers-modelist))
 
 ;;
 ;;; Packages
--- a/modules/lang/elixir/config.el
+++ b/modules/lang/elixir/config.el
@@ -5,12 +5,8 @@
 
 (after! highlight-numbers
   (puthash 'elixir-mode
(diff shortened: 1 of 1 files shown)


modules/lang/elixir/config.el
```diff
@@ -3,6 +3,14 @@
 (after! projectile
   (add-to-list 'projectile-project-root-files "mix.exs"))
 
+(after! highlight-numbers
+  (puthash 'elixir-mode
+    (rx (and symbol-start
+            (? "-")
+            (+ digit)
+            (0+ (and "_" (= 3 digit)))
+            symbol-end))
```
hlissner, line 12: Please change this to

```elisp
(after! highlight-numbers
  (puthash 'elixir-mode
           "\\_<-?[[:digit:]]+\\(?:_[[:digit:]]\\{3\\}\\)*\\_>"
           highlight-numbers-modelist))
```

1. Doom doesn't use/load `rx`, so it's not right to use it as if it's guaranteed.
2. `puthash` was strangely indented, so I reindented it

jsmestad:
@hlissner is using `rx` considered bad form or just something doom does not guarantee is loaded? Wondering so I can learn 😄

hlissner:
> is using rx considered bad form or just something doom does not guarantee is loaded? Wondering so I can learn smile

Just something I can't guarantee will be loaded (or installed).

Anyhow thanks for the PR!

PR #3621, opened 2020-07-24
Small documentation touchup for literate module

CarlQLange:


hlissner:
There was much more to add to the literate module's readme so I've pushed them in ff21dfe with credit. Thanks for bringing it to my attention!

CarlQLange:
Sorry, I think I could have done a much better job of that, forgive me!

PR #3623, opened 2020-07-25
update emacs-lisp outline regexp to fold defs

jethrokuan:
Allows outline-mode to also fold defs, which can be useful when developing large emacs-lisp files. I've had this in my config for a while and find it useful, was wondering if upstream would like to have it as well:

image


hlissner:
Hi! Thanks for the PR but I cannot accept it. It is redundant with `hideshow` which the `:editor fold` module configures, and will handle folding lisp blocks, while outline-minor-mode is relegated to handling comment headings, and vimish-fold handling arbitrary folds.

jethrokuan:
I see, thanks! I don't use evil, so there weren't any accessible bindings, which was probably why I didn't notice.

PR #3637, opened 2020-07-27
tools/magit: map Forge jumps in evil-mode

vednoc:
Hi!

This PR will add `gi` and `gm` mnemonics in magit-mode to quickly jump to issues and pull requests respectively. I chose `m` for pull requests because `p` was already taken and it makes sense for GitLab users, not sure if GitHub users will agree.

P.S. I'm not sure if this is the proper way to go about it as I'm new to Elisp; please let me know if/how it could be improved.

### DIFF
--- a/modules/tools/magit/config.el
+++ b/modules/tools/magit/config.el
@@ -101,6 +101,11 @@
   :preface
   (setq forge-database-file (concat doom-etc-dir "forge/forge-database.sqlite"))
   :config
+  ;; Map Forge jumps in `evil-mode'
+  (when (featurep! :editor evil +everywhere)
+    (map! :map magit-mode-map
+      :n "gi" #'forge-jump-to-issues
+      :n "gm" #'forge-jump-to-pullreqs))
   ;; All forge list modes are derived from `forge-topic-list-mode'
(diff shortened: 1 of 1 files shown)


modules/tools/magit/config.el
```diff
@@ -101,6 +101,11 @@
   :preface
   (setq forge-database-file (concat doom-etc-dir "forge/forge-database.sqlite"))
   :config
+  ;; Map Forge jumps in `evil-mode'
+  (when (featurep! :editor evil +everywhere)
+    (map! :map magit-mode-map
+      :n "gi" #'forge-jump-to-issues
+      :n "gm" #'forge-jump-to-pullreqs))
```
hlissner, line 108: Could you move this into the `:config` body of `evil-magit`, at the bottom of this file?

Also, remove the `(when (featurep! :editor evil +everywhere) ...)` check, since `evil-magit` isn't loaded without it.

vednoc: I think this should do it! Later today I'll have a look about making a PR to the upstream. 😊

hlissner:
Thanks!

PR #3641, opened 2020-07-28
[UI] Set prompt to readonly in comint buffers

Compro-Prasad:
Comint buffers like *Python*, *shell*, etc. will benifit from this
change as the prompt wont be deletable.


hlissner:
Makes sense to me. Thanks!

PR #3643, opened 2020-07-29
org-roam: check for sqlite3 executable in doctor

wedens:
org-roam fails with [very unhelpful](https://discord.com/channels/406534637242810369/406554085794381833/737720230507315200) errors when `sqlite3` executable is not found.

Perhaps checking it in `doom doctor` will make it a bit easier to find the problem.


wedens:
fixed a typo

PR #3646, opened 2020-07-29
Change :app calendar README and add cfw:org-create-file-source to list of commands

MaxSchlueter:
The example in the README to include org files from other sources in the calendar view doesn't really work, because the binding to `org-agenda-files` is lost e.g. when calendar view is refreshed.
Fortunately, there is the function `cfw:org-create-file-source` that allows us to list other org files in the content sources.


hlissner:
Thanks!

Issue #3331, opened 2020-06-08
Calendar module doesn't work out-of-the-box with example config provided


Bounga:
**What did you expect to happen?**
I assumed that `=calendar` command would start the calendar showing org entries, diary entries and gcal ones by copy / pasting the [provided config example](https://github.com/hlissner/doom-emacs/tree/develop/modules/app/calendar#changing-calendar-sources).

**What actually happened?**

It crashes because `cfw:cal-create-source` and `cfw:ical-create-source` aren't defined.

Too make it work as expected I had to:

- add the packages to `packages.el`

```elisp
(package! calfw-cal :pin "8de82be1e67b6114e9661b55a92f996553a1da9d")
(package! calfw-ical :pin "263d5f37fff53f16165f347ee86dd35e54a93693")
```
then `doom sync`. I was then able to configure calendars but there's one little gotcha, I had to require `calfw-cal` and `calfw-ical`:

```elisp
(after! calfw
  (require 'calfw-cal)
  (require 'calfw-ical)
  (defun +calendar/open-calendar ()
    "My calendar definition"
    (interactive)
    (cfw:open-calendar-buffer
     :contents-sources
     (list
      (cfw:org-create-source "Green")  ; orgmode source
      (cfw:cal-create-source "Orange") ; diary source
      (cfw:ical-create-source "gcal" "https://calendar.google.com/calendar/ical/…/basic.ics" "IndianRed")))))
```

**Steps to reproduce:**
1. enable `calendar` module,
2. copy / paste the provided config and change urls / filenames,
3. run `=calendar`.

**System information:**
<details>

```
emacs   version    28.0.50
        features   JPEG TIFF GIF PNG RSVG IMAGEMAGICK DBUS GLIB NOTIFY KQUEUE ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES THREADS XWIDGETS PDUMPER LCMS2 GMP
        build      Apr 24, 2020
        buildopts  (--enable-locallisppath=/usr/local/share/emacs/site-lisp --infodir=/usr/local/Cellar/emacs-head/HEAD-bef617d_1/share/info/emacs --prefix=/usr/local/Cellar/emacs-head/HEAD-bef617d_1 --without-x --with-dbus --with-imagemagick --with-modules --without-pop --with-gnutls --with-rsvg --with-xml2 --with-xwidgets --with-harfbuzz --with-ns --disable-ns-self-contained)
[...]

PR #3648, opened 2020-07-29
Add calfw-cal dependency to allow local diary source

Bounga:
As mentioned in https://github.com/hlissner/doom-emacs/issues/3331 there's a dependency missing in calendar app module to be able to follow the README along and configure local diary source.

This PR fixes it.


hlissner:
Ah, my bad; your response to #3331 flew under my radar. Thanks for the PR!

PR #3656, opened 2020-07-31
More convenient search buffer

nhamlh:


hlissner:
Makes sense to me. Thanks!

facelesspanda:
awww, now I have to *select* what I want to search? How is this convenient? Guess I have to add it back in my own config then...

hlissner:
@facelesspanda My bad. I've added back `SPC s S` in 76eae7b

facelesspanda:
> @facelesspanda My bad. I've added back `SPC s S` in 76eae7b

That's great news! Thanks 😄

PR #3665, opened 2020-08-01
fix: bump ranger (fixes paste error)

JakDar:
Fixes pasting file to its original directory. Previously it failed with
error (undefined variable), now it works  (adds suffix ~1 to filename).


hlissner:
I've bumped the module in 8a0fd0b because bump commit messages must adhere to a strict, machine-readable format, so that is readable by future CI. It's not your fault -- I haven't documented it anywhere yet. In any case, thanks for the PR and bringing this to my attention!

PR #3666, opened 2020-08-01
Update README.org

evb-gh:
Clarify that you have to manually install the lang server in doom.


hlissner:
Thanks for the PR, but I cannot accept it. This isn't the only way to install an LSP server (much less the ones for javascript modes) and this advice applies to many modules, other than the javascript module, and so doesn't belong here.

It would be better suited to [this question in the FAQ](https://github.com/hlissner/doom-emacs/blob/develop/docs/faq.org#how-do-i-enable-lsp-support-for-insert-language-here) and/or the documentation for the `:tools lsp` module.

PR #3668, opened 2020-08-02
ui(tabs): use a smaller dot to show unsaved file status

willbchang:
The default modified marker looks unfit to the tab's height and font size, the smaller one is more appropriate.

## Before
before

## After
after


hlissner:
Makes sense to me. Thanks!

PR #3669, opened 2020-08-02
Docs update add to path

mikeo85:
Updated the instructions for adding Doom to PATH during Linux installation, based on recommendations from Ubuntu Documentation to use .profile instead of .bashrc or .zshrc.


hlissner:
Hi! Thanks for the PR, but I have to turn it down. While Ubuntu's advice makes sense, I think it's extraneous detail for Doom's installation process. `bin/doom` is intended for interactive use, from your shell. Whether or not the `PATH` entry is available to programs other than your shell isn't terribly important to the average user. Less so because this PATH entry is only a convenience.

I feel this footnote about `.profile` adds a source of confusion for beginners and noise to experienced users. In any case, thanks for bringing it to my attention.

PR #3673, opened 2020-08-02
Add some documentation

jgoerzen:
This PR adds some introductory material and docs in migration from Emacs.  It also adds descriptions for localleader in various module doc files.  This was a particular puzzle for me as I was getting started ("what the heck does localleader mean") and I hope to smooth the way for the next person.

### DIFF
--- a/docs/getting_started.org
+++ b/docs/getting_started.org
@@ -45,6 +45,7 @@ us know!
   - [[#menus--commands][Menus & Commands]]
     - [[#context-specific-commands][Context-specific commands]]
   - [[#getting-help][Getting Help]]
+  - [[#tour-of-features][Tour of features]]
 - [[#migrate][Migrate]]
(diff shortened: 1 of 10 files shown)


chrisglass:
While this might seem useless for emacs veterans, finding this PR helped me getting started.

I for one think it should be merged! Thanks a lot!

hlissner:
Hi! Thanks for the PR. I can't accept it as it is because I don't think this belongs in getting_started.org -- the concern is that we'll overloading new users by mixing up installation with configuration with usage documentation. That said, there is good material in here, so I've archived it and will refer to it when I get around to restructuring our docs, possibly move it to a tutorial.org or workflows.org. I'll follow up once there's more movement here.

hlissner:
New docs will be up soon, so I'm closing docs PRs.

PR #3680, opened 2020-08-03
Bump :lang clojure

zackteo:
CIDER just made its 0.26 (Nesebar) release


hlissner:
I bumped cider in 9ac9deb because bump commit messages must follow a particular format, so that they remain machine readable to future CI. It's not your fault -- I haven't publicly documented this anywhere yet, but thanks for bringing this to my attention!

PR #2430, opened 2020-01-26
Set up word-wrap bindings only when the feature is enabled

r-darwish:

PR #3682, opened 2020-08-04
bind deft only when deft is enabled

sorawee:


hlissner:
Hi! Thanks for the PR but I cannot accept it. This is a duplicate of #3335, #2430, #2390, #2320, and #2267. See https://github.com/hlissner/doom-emacs/pull/2267#issuecomment-569536183 for an explanation.

PR #3684, opened 2020-08-04
Fix Elisp file template header in doom-private-dir.

Janfel:
Currently, when I create a `.el` file anywhere in `doom-private-dir`, the automatically generated file header contains an absolute path. This PR fixes this issue by extending `+file-templates-get-short-path` to return a path relative to `doom-private-dir`, where applicable, as it already does with `doom-emacs-dir`.


hlissner:
Thanks!

PR #3690, opened 2020-08-05
racket-mode: avoid using flycheck-disable-checker

sorawee:
The problem is that at this point, flycheck-mode might not be active
yet. Calling flycheck-disable-checker will always call (flycheck-buffer)
which would result in an error: (user-error "Flycheck mode disabled")

### DIFF
--- a/modules/lang/racket/config.el
+++ b/modules/lang/racket/config.el
@@ -34,7 +34,8 @@
     (when (featurep! :checkers syntax)
       (add-hook! 'racket-xp-mode-hook
         (defun +racket-disable-flycheck-h ()
-          (flycheck-disable-checker 'racket)))))
+          (unless (memq 'racket flycheck-disabled-checkers)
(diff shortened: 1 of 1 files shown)


modules/lang/racket/config.el
```diff
@@ -34,7 +34,8 @@
     (when (featurep! :checkers syntax)
       (add-hook! 'racket-xp-mode-hook
         (defun +racket-disable-flycheck-h ()
-          (flycheck-disable-checker 'racket)))))
+          (unless (memq 'racket flycheck-disabled-checkers)
+            (push 'racket flycheck-disabled-checkers))))))
```
hlissner, line 38: Could you replace this with `(cl-pushnew 'racket flycheck-disabled-checkers)`?

sorawee: Done!

hlissner:
Thanks!

PR #3701, opened 2020-08-07
New language module for developing windows scripts PowerShell and Dos/cmd

samwdp:
Not sure how to do the :pin in the packages.el file so didn't add them in


gagbo:
Suggestion : we could replace importing the whole dos.el file with using a [fork that would allow a github recipe in packages.el](https://github.com/timepp/emacs_bat)

With that being said, there’s a chance that the PR is refused, to try to limit the amount of "trivial" modules (where the config.el is small and doesn’t do too much)

hlissner:
Hi! Thanks for the PR, but I cannot accept it for several reasons:

1. I cannot accept including the full source code of a package that could be installed through `package!`.
2. Emacs has a decent built-in major mode for .bat/.cmd files (bat-mode).
3. All three packages (dos, powershell, and bat-mode) autoload `auto-mode-alist` entries for the ps1, psm1, bat and cmd file extensions (so all the `:mode`s in this PR are redundant).
4. The module is installing and configuring `lsp-mode` beyond adding `lsp` to a hook. This is not its responsibility. Doom has a `:tools lsp` module that does this, whose presence should signal to other modules whether to enable lsp or not.
5. I already have powershell + batch script support in the works for the `:lang sh` module under a `+windows` flag.

In any case, thanks for bringing it to my attention.

PR #3703, opened 2020-08-07
Fix deduplication in add-load-path!

Janfel:
Currently, `add-load-path!` doesn’t check for duplicates in `load-path`, because `cl-pushnew` tests for equality with `eql`. This changes the predicate to `string=`, fixing the bug.


hlissner:
Ah, good catch! Thanks!

PR #3706, opened 2020-08-08
Add evil-tex package to :lang latex

ymarco:
[evil-tex](https://github.com/iyefrat/evil-tex) is a plugin for better latex editing in evil-mode. @iyefrat and I worked very hard on it, and I think its a fine addition to the latex module.


ymarco:
Done

hlissner:
Makes sense to me. Thanks!

PR #3708, opened 2020-08-09
Fix typo in new os/tty module readme

reedspool:
Simple typo: +osx to +osc


hlissner:
Whoops! Good catch. Thanks!

PR #3711, opened 2020-08-09
Enable native ligature support on emacs-28/mac

dlesl:
This works for me using emacs HEAD (`emacsGit` from the [Nix emacs overlay](https://github.com/nix-community/emacs-overlay)) with Fira Code and Iosevka.

### DIFF
--- a/modules/ui/pretty-code/config.el
+++ b/modules/ui/pretty-code/config.el
@@ -172,10 +172,11 @@ Otherwise it sets the buffer-local composition table to a composition table enha
  ((and IS-MAC (fboundp 'mac-auto-operator-composition-mode))
   (mac-auto-operator-composition-mode))
 
- ;; Harfbuzz builds do not need font-specific ligature support
+ ;; Harfbuzz and Mac builds do not need font-specific ligature support
  ;; if they are above emacs-27
  ((and EMACS28+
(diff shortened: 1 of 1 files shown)


gagbo:
You don't have "HARFBUZZ" in your `system-configuration-features` variable ? Can you paste the output of `M-x doom/info` please ?

gagbo:
Looks like you should have harfbuzz in the features from my very light understanding of Nix eco system https://github.com/nix-community/emacs-overlay/blob/b839a0ed8bdcceec92bc4a6a93211503501f7dd4/default.nix#L49

dlesl:
`system-configuration-features`: `JPEG TIFF PNG RSVG GLIB NOTIFY KQUEUE ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES THREADS JSON PDUMPER`.

AFAIK on mac emacs uses native text rendering (but I am not sure). However I guess it's possible that someone might be running emacs under X11 on mac, in which case harfbuzz would be needed -- maybe this is worth checking for?

dlesl:
I've updated the PR to check for `(featurep 'ns)` or harfbuzz, I think this should cover the X11-on-mac case.

hlissner:
Makes sense to me. Thanks!

PR #3716, opened 2020-08-10
Fix python-pytest-popup

benbovy:
It is now obsolete, it has been renamed to `python-pytest-dispatch` in
https://github.com/wbolster/emacs-python-pytest/commit/ee61741d4e35dc639b5e269efaa488d743da9c35

Not sure if we need to update the keybindings, though.

### DIFF
--- a/modules/lang/python/config.el
+++ b/modules/lang/python/config.el
@@ -183,7 +183,7 @@ called.")
 
 
 (use-package! python-pytest
-  :defer t
+  :commands python-pytest-dispatch
   :init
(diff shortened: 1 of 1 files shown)


benbovy:
@hlissner is there anything else I can do here? Or do you have other plans?

Unfortunately, we need to explicitly do `M-x python-pytest-dispatch` once if we want `SPC m t p` to work.

hlissner:
Sorry for the tremendously late response! Thanks for the help!

PR #3725, opened 2020-08-11
Updates getting_started.org following the release of Emacs 27.1

adamroyjones:
Emacs 27.1 has been released and emacs-plus (https://github.com/d12frosted/homebrew-emacs-plus) now installs 27.x by default.

PR #3733, opened 2020-08-12
Updated Mac installation instructions

eh8:
d12frosted's homebrew-emacs-plus was updated to Emacs 27, so the addendum is no longer necessary


hlissner:
This is a duplicate of #3725

PR #3735, opened 2020-08-12
ui/modeline: add +keycast feature

vednoc:
Hi,

I figured this would be a nice addition to Doom. Many thanks to Taco and Prot since I borrowed code from their configurations. 

Let me know if you'd like to change something and I'll happily make the necessary adjustments.

P.S. There is a minor issue with `doom-modeline` not displaying `keycast-mode` on the bottom right buffer, meanwhile it works fine on `+light` mode-line (which is what I use). It _does_ work when the focus is on the right window. Here's a couple of screenshots:

image
image


vednoc:
@hlissner I rebased this branch to avoid merge conflicts. Just pinging you in case you haven't seen this PR. 😃

PR #3740, opened 2020-08-12
Add doctor check for native JSON feature

gagbo:
I thought it might be time with emacs 27 out and first packages without json being distributed


hlissner:
That timing! I'd _just_ implemented a native json check an hour ago in https://github.com/hlissner/doom-emacs/commit/11be796c9b15b1943a5aa81d794ef7e157091725. Thanks in any case!

Issue #3728, opened 2020-08-11
Support bépo keyboard layout


xtab:
I’ve just had a look at Doom Emacs and it looks very promising! However, just like #767 and #783, I’m using a non-{AZ,QW}ERT{Y,Z} layout: instead, I’m using the [Bépo](http://bepo.fr/wiki/Accueil) keyboard layout and I’m used to swapping HJKL and CTSR when using either Vim or Spacemacs, along with some other keybindings (like “é” as a synonym for “w”).

Spacemacs [already natively supports](https://github.com/syl20bnr/spacemacs/tree/master/layers/%2Bintl/keyboard-layout) such a feature.

For now, I seem to be able to have roughly what I want by creating a `bepo.el` file in my `~/.doom.d` directory like so:
 
```elisp
;;; bepo.el -*- lexical-binding: t; -*-

(map! :nvm "h" 'evil-replace
      :nvm "H" 'evil-replace-state
      :nvm "j" 'evil-find-char-to
      :nvm "J" 'evil-find-char-to-backward
      :nvm "k" 'evil-substitute
      :nvm "K" 'evil-change-whole-line
      :nvm "l" 'evil-change
      :nvm "L" 'evil-change-line

      :nvm "c" 'evil-backward-char
      :nvm "C" 'evil-window-top
      :nvm "t" 'evil-next-line
      :nvm "T" 'evil-join
      :nvm "s" 'evil-previous-line
      :nvm "r" 'evil-forward-char
      :nvm "R" 'evil-window-bottom

      :nvm "é" 'evil-forward-word-begin
      :nvm "É" 'evil-forward-WORD-begin

      :map evil-snipe-local-mode-map
      (:nm "k" 'evil-snipe-s
       :nm "K" 'evil-snipe-S
       :nm "s" nil
       :nm "S" nil)
      :map evil-snipe-override-local-mode-map
      (:m "j" 'evil-snipe-t
       :m "J" 'evil-snipe-T
       :m "t" nil
       :m "T" nil))
```

But I’m not sure if that’s sustainable, since it means that I’d need to swap hjkl and ctsr in any major mode map.

In the end, I’d just like to have `(doom! :input bepo)` in `~/.doom.d/init.el` and have my Bépo bindings ready.

gagbo:
[...]

PR #3754, opened 2020-08-14
Add support for bépo layout in a dedicated module

gagbo:
Urgent question : The module *needs* to be loaded after `:config (default +bindings)` so I'm not sure using the same `:input` category as the one at the very top of the `doom!` macro is relevant.

## Add support for bépo layout

Hopefully this will show how it is possible to implement mappings for different keybindings

### Features
Everything should be in the README of the module so I will not repeat myself

### TODO Status
- [x] Make sure the name of the module is correct
- [x] Test if the `doom-bepo-cr-rotation-style` variable is respected if set from a user's config.el
- [x] Highlight non-breaking spaces properly when the bépo variant is chosen and expose a switch as a setting and an interactive function


modules/input/layout/+bepo.el
```diff
@@ -0,0 +1,97 @@
+;;; input/layout/+bepo.el -*- lexical-binding: t; -*-
+
+;; Eagerly evaluated for now
+;; TODO: Check if the doom-bepo-cr-rotation-style variable can have an effect if set in $DOOMDIR/config.el
+;; NOTE: the evaluation loads the whole autoload/bepo.el file but it doesn't really matter as other
+;; functions are eagerly called in this block
+(fset 'doom-bepo--evil-collection-hook
+      (doom-bepo-rotate-collection-keymaps-h-builder doom-bepo-cr-rotation-style))
+
+(setq avy-keys '(?a ?u ?i ?e ?c ?, ?t ?s ?r ?n))
```
ymarco, line 10: Might as well set `aw-keys`

gagbo:
Will fix #3728

PR #3764, opened 2020-08-17
adding another trouble shooting step to make sure mu4e is loaded properly 

jeetelongname:
As Ubuntu does not include mu4e with the mu package I added a step to
tell the use to install mu4e and add it to their load-path. I have also
added a find command to find the directories likely to have mu4e in them


hlissner:
Makes sense to me. Thanks!

PR #3776, opened 2020-08-19
Improve highlighting of Julia operators

ronisbr:
Hi!

This commit fix some problems related to highlighting math operators in Julia that created some problems like https://github.com/JuliaEditorSupport/julia-emacs/issues/141

The problem is that, different from MATLAB, Julia can have function and variable names with `!`. Thus, I tried to address some cases, like `!` before an empty space or beginning of the line. However, there things like `a==!var` or `(!var)` is also valid but cannot be treated inside this function. For a full support, we would need a full parser.

I also add some tweaks for `:` (which defined a symbol) and also added the operator `===`.

Before:

Captura de Tela 2020-08-18 às 20 59 07

After:

Captura de Tela 2020-08-18 às 21 00 10

### DIFF
--- a/modules/lang/julia/config.el
+++ b/modules/lang/julia/config.el
@@ -12,20 +12,30 @@
      mode
      `((,(let ((OR "\\|"))
            (concat "\\("  ; stolen `matlab.el' operators first
-                   "[<>!]=?" OR
+                   "[<>]=?" OR
                    "\\.[/*^']" OR
+                   "===" OR
                    "==" OR
                    "=>" OR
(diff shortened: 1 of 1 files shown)


ronisbr:
Hi @hlissner 

Sorry! My eyes missed that `:` in `<:` was not highlighted. This should be fixed now:

Captura de Tela 2020-08-18 às 21 29 03

hlissner:
Looks good to me. Thanks you for the help and for the thorough explanation with screenshots; it is much appreciated!

PR #3778, opened 2020-08-19
Add projectile dired action in workspaces module

t-e-r-m:
Adds the "open project in dired" action to projectile (`C-o`)


hlissner:
Makes sense to me. Thanks!

PR #3779, opened 2020-08-19
Add org-toggle-checkbox binding (SPC m x)

t-e-r-m:


hlissner:
A more convenient keybind already exists for evil users (RET in normal mode), but I suppose this could help non-evil users. Thanks!

PR #3780, opened 2020-08-19
Configure dired-git-info variables

t-e-r-m:
`dgi-commit-message-format` set to

`%h`  - Abbreviated commit hash
`%cs` - committer date, short format (YYYY-MM-DD)
`%s`  - subject

First two are fixed-width, which makes it look neater.

`dgi-auto-hide-details-p` set to `nil`, assuming a wide window.

I'm not sure if these are supposed to be inside `use-package!` after reading the previous commits to the file.

Screenshot:


hlissner:
Looks good to me. Thanks!

Issue #3781, opened 2020-08-19
source not found error from line 10 of bin/doom executable when running doom commands.


ludwj61:
After the most recent updates to doom, I noticed this error whenever running doom:
![image](https://user-images.githubusercontent.com/42984254/90654663-9f677b00-e20e-11ea-830f-ca4877759bbc.png)
It occurs with all doom commands, regardless if the command succeeds or not.

The issue comes from this commit: https://github.com/hlissner/doom-emacs/commit/55b87b3a94d65104841ca0b36b03ed9ed4d49597

I found the issue to be that the executable many distros use for ```sh``` do not support the ```source``` command, and rather use ```.```
Replacing the ```#!/usr/bin/env sh``` with ```#!/usr/bin/bash``` makes the issue go away, but understandably doom does not like .emacs.d being messed with and shows a warning about changed files.

**Additional details:**
- This occurs with a completely vanilla config/fresh install

**Steps to reproduce:**
1. Install doom
2. Run a doom command.

**System information:**
<details><pre>
Pop OS 20.04
Emacs 27.1
</pre></details>

gagbo:
The script should be posix compliant and use sh, for systems that only have sh and/or dash. No idea about why source doesn't work I don't remember the posix builtins

digizeph:
Same issue on PopOS 20.04 (based on Ubuntu 20.04):
```
❯ sh 
$ source
sh: 1: source: not found
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Pop
Description:	Pop!_OS 20.04 LTS
Release:	20.04
Codename:	focal
```

digizeph:
Found this answer in StackOverflow:
https://stackoverflow.com/a/13702462/768793

> In Bourne shell(sh), use the . command to source a file

hlissner:
As of 9b991fc this issue should be resolved. Let me know if that isn't the case and I'll reopen it. Thanks for bringing it to my attention!

TotNotTac:
Hi, after 9b991fc I get this error
```
$ bash doom install                                                                 
> Compiling your literate config...                                                                                       
- Tangled 41 code blocks from config.org.LL7Uah                                                                         
- Restarting...                                                                                                       
sh: 2: doom install: not found 
```

digizeph:
@hlissner @TotNotTac 
Yeah, same issue after 9b991fc:
```
❯ doom update
> Compiling your literate config...
[...]

PR #3782, opened 2020-08-19
Use `.` instead of `source` for `sh` shell

digizeph:
Related to issue #3781 

> In Bourne shell(sh), use the . command to source a file

- https://stackoverflow.com/a/13702462/768793
- https://stackoverflow.com/a/4732283/768793


hlissner:
Ah, my bad. I'd already reverted it before I saw the PR. In any case, this particular error was fixed in 3694f51. Thanks for bringing it to my attention!

PR #3785, opened 2020-08-20
Use separate geiser REPLs per project

flatwhatson:
By default geiser will re-use an existing REPL with the same interpreter.  This doesn't work well when with multiple projects using dir-locals or direnv to provide project-specific settings that affect the REPL (ie. binary and library paths).

Support for per-project REPLs was added upstream; this patch enables it by default for doom.

Also disabling `geiser-smart-tab-p` because TAB needs to stay in its lane.

@elais ping


elais:
Looks like a great change to me. I don't care about the smart tabs bit, could be either/or for me.

hlissner:
Makes sense to me, thanks!

PR #3786, opened 2020-08-20
ArchLinux packages now have 27.1 as default

axelson:
https://www.archlinux.org/packages/extra/x86_64/emacs/

We could point to emacs-git AUR package for emacs 28:
https://aur.archlinux.org/packages/emacs-git/

But I think the reference to emacs 27 was there simply because emacs 27
is the recommended version so I didn't include a replacement AUR link in this PR.


hlissner:
Thanks!

axelson:
Thank you!

PR #3793, opened 2020-08-20
Fix macos keybindings after refactor

pharcosyle:


hlissner:
Good catches, thanks!

PR #3795, opened 2020-08-20
Add nim definition and documentation keybindings

mcbloch:


ymarco:
You're treating the `after!` block like a `use-package!`, they're not equal. `after!` is simple, it doesn't accept stuff like `:init` and `:config`.
tldr: Change `after!` to `use-package!`.

modules/lang/nim/config.el
```diff
@@ -1,6 +1,7 @@
 ;;; lang/nim/config.el -*- lexical-binding: t; -*-
 
 (after! nim-mode
```
hlissner, line 3: Change `after!` to `use-package!`

hlissner:
Thanks!

PR #3796, opened 2020-08-21
Add nimble keybinding

mcbloch:
I'm not sure about the exact keybinding to use. I see that 'b' is sometimes to group build commands, but nim provides building from a file also and `b` is bound to that command. So I thought that nimble could be put under 'p' for project (as it is a project manager for nim).

### DIFF
--- a/modules/lang/nim/config.el
+++ b/modules/lang/nim/config.el
@@ -29,6 +29,8 @@ windows."
   (map! :localleader
         :map nim-mode-map
         "b" #'nim-compile
+        (:prefix ("p" . "project")
+         :desc "nimble run" "r" #'nimble-run)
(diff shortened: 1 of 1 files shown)


mcbloch:
A bit of a premature pr maybe. The command can be found here [nim-mode#230](https://github.com/nim-lang/nim-mode/pull/230#issuecomment-678799395), but there is discussion about the implementation. I will make this pr a draft while the other one is open.

hlissner:
I am closing this PR to prepare to move our modules into a separate repo. Unfortunately, upstream hasn't merged your nimble commands yet, so I cannot merge this now, but I will add them should this change. Thanks for bringing it to my attention, in any case!

PR #3806, opened 2020-08-22
Fix #3783: advice org-completing-read in bepo

### LINKED ISSUE #3783: In Bépo, cannot invoke org-insert-link
With `:input (layout +bepo)` in org-mode, I cannot use `org-insert-link`, either from normal mode keybindings or from `execute-extended-command`.

Instead of being prompted for the link I see this error with the associated backtrace : 
```
Debugger entered--Lisp error: (error "Key sequence C-c ! starts with non-prefix key C-c")
  org-completing-read("Link: " ("doom-repo:" "wolfram:" "wikipedia:" "duckduckgo:" "gmap:" "gimages:" "google:" "youtube:" "github:" "bibtex:" "download:" "pdf:" "orgit-rev:" "orgit-log:" "orgit:" "id:" "yt:" "img:" "doom-modules:" "doom-docs:" "doom:" "org:" "attachment:" "file+sys:" "file+emacs:" "shell:" "news:" "mailto:" "https:" "http:" "ftp:" "help:" "file:" "elisp:" "doi:") nil nil nil org-link--history nil)
  org-insert-link(nil)
  funcall-interactively(org-insert-link nil)
  command-execute(org-insert-link)
```

Since C-c is remapped as part of the bindings in the evil-collection hook, I suspect this is why C-c is a non-prefix key. I don’t see why something is being mapped on `C-c !` at this particular function call though.

The weirdest thing is that (using `:completion ivy`) I have no issues with any other completing-read function that I use otherwise. It needs a little more investigation

gagbo:
The original function tries to map `C-c !` in
minibuffer-local-completion-map although `C-c` is a non-prefix key in
this map for bepo users

### DIFF
--- a/modules/input/layout/+bepo.el
+++ b/modules/input/layout/+bepo.el
@@ -83,6 +83,23 @@
     (general-translate-key '(normal motion) 'lispyville-mode-map
       "«" "<"
       "»" ">"))
+  (after! org
+    (defadvice! doom-bepo--org-completing-read (&rest args)
+      "Completing-read with SPACE being a normal character, and C-c mapping left alone."
+      :override #'org-completing-read
+      (let ((enable-recursive-minibuffers t)
+            (minibuffer-local-completion-map
+             (copy-keymap minibuffer-local-completion-map)))
+        (define-key minibuffer-local-completion-map " " 'self-insert-command)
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks!

PR #3811, opened 2020-08-23
ui/nav-flash add vterm to +nav-flash-blink-cursor-maybe

Sbozzolo:
`vterm-mode` is morally equivalent to `term-mode` (but `vterm-mode` is not derived from `term-mode`), so I would expect functions that exclude `term-mode` to exclude `vterm-mode` as well.


hlissner:
Makes sense to me. Thanks!

PR #3814, opened 2020-08-23
Add Rusty Object Notation to the Rust module

enderger:
# Add RON support to the Rust module
#### About RON
Rusty Object Notation (RON) is a new JSON-like format used by several projects written in the Rust programming language (most predominantly, it is the main configuration format in the Amethyst game engine). This project adds support for it into Doom EMACS.

---
#### Why not add it into it's own module?

1. It is an exceptionally small project: The package does not exceed 70 lines at the time of writing.

2. It is only used within the Rust ecosystem

### DIFF
--- a/modules/lang/rust/config.el
+++ b/modules/lang/rust/config.el
@@ -80,3 +80,5 @@
   (set-lookup-handlers! 'rustic-mode
     :definition '(racer-find-definition :async t)
     :documentation '+rust-racer-lookup-documentation))
+
+(use-package! ron-mode)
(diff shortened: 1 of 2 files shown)


enderger:
Sorry for the bump, but this PR kind of fell through the cracks.

hlissner:
Hi! Thanks for the PR (and sorry for the late response, I've been on break), but I cannot accept it. 1) It is too small to warrant its inclusion (a user could trivially add it themselves) and 2) I'm not convinced that RON is ubiquitous or popular enough to warrant it's inclusion, by default.

And, for future reference, `(use-package X)` without any additional properties or configuration should be avoided. For one, it will eagerly load `X`.

PR #3819, opened 2020-08-24
[Fix] be lazier for minibuffer maps in :layout bepo

gagbo:
Contains 2 fixes

## helm-rg-map in +default-minibuffer-maps

It looks like we are not using helm-ag anyway

## Lazier loading of keymap rotation for minibuffer maps in :layout bepo

The old method was too eager and, when using `:completion helm`, no completing-read can be performed because of void variables on the maps (helm-rg and helm-read-file)


hlissner:
Thanks!

PR #3822, opened 2020-08-24
[Fix] string comparison in TERM check in org-capture

zzamboni:
The `bin/org-capture` script was using `-eq`, which is a numeric comparison, and was producing an "integer expression expected" error. Changed to `=`, the string comparison operator.


hlissner:
Whoops! Thanks!

PR #3823, opened 2020-08-24
Add latex faces to +spell-excluded-faces-alist

ymarco:
This list is not exhaustive, but its a lot better than nothing at all.


iyefrat:
shouldn't this be `LaTeX-mode`?

ymarco:
No, that symbol is checked against the `major-mode` variable which is `latex-mode` in LaTeX buffers.

hlissner:
Thanks!

PR #3825, opened 2020-08-24
If +flyspell enabled, disable spell-fu keybindings and use flyspell ones

JakDar:
Fixing keybindings using recently introduced +flyspell flag ;)


hlissner:
Whoops. Forgot about those. Thanks!

PR #3839, opened 2020-08-26
fix ivy not working well with lsp-ivy-workspace-symbol

cxb811201:
More details here: emacs-lsp/lsp-ivy#18 (comment)


hlissner:
Thanks!

PR #3843, opened 2020-08-26
Change the warning prefix to a non-unicode character

t-e-r-m:
It's faster and stylistically compliant.


hlissner:
Ah, been meaning to do that. Thanks!

PR #3846, opened 2020-08-26
Automatically download reveal.js and use local copy, default to v4

zzamboni:
The +present option now automatically checks out the reveal.js
repository and configures org-re-reveal to use it. It also now uses
reveal.js 4 instead of 3.9.2.

### DIFF
--- a/modules/lang/org/contrib/present.el
+++ b/modules/lang/org/contrib/present.el
@@ -14,8 +14,8 @@
 (use-package! org-re-reveal
   :after ox
   :init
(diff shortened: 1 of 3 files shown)


jeetelongname:
I really hope this solves my reveal.js woes

modules/lang/org/contrib/present.el
```diff
@@ -14,8 +14,8 @@
 (use-package! org-re-reveal
   :after ox
   :init
```
hlissner, line 16: Could you change `:init` to `:config`?

zzamboni: Done!

hlissner:
Nice work! Thanks!

PR #3847, opened 2020-08-26
Initial implementation of dired-sidebar module

precompute:
I've been using [dired-sidebar](https://github.com/jojojames/dired-sidebar) in lieu of `neotree` and `treemacs` for a while and since I had a working config, I decided to try my hand at a module.

The configuration might not warrant an entire module, but what it delivers for it is indispensable (for me). It could definitely be a flag under `dired`.

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -43,6 +43,7 @@
        (popup +defaults)   ; tame sudden yet inevitable temporary windows
        ;;tabs              ; a tab bar for Emacs
        ;;treemacs          ; a project drawer, like neotree but cooler
+       ;;dired-sidebar     ; forget neotree and treemacs, pledge your unwavering loyalty to dired
        ;;unicode           ; extended unicode support for various languages
        vc-gutter         ; vcs diff in the fringe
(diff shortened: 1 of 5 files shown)


jeetelongname:
If this is merged I would think we should get a `sidebar` module and then we enable the one we want through a flag

t-e-r-m:
Why use two lines when you can use just one?
Regardless, `+`es are features and not choices.

hlissner:
Sorry for the late response; I was on break. 

I don't think a `:ui dired-sidebar` is warranted, and would be better suited to `:emacs (dired +sidebar)`, perhaps a `:ui (sidebar +dired)` to combine the three. I'll think about it while I resolve the backlog I've returned to.

> Regardless, `+`es are features and not choices.

They can be either. Frankly I prefer that modules represent features rather than implementations.

hlissner:
Sorry for the _tremendously_ late response. I'm closing this PR in preparation for moving our modules into their own repo, but putting together a `:ui sidebar` or (`:emacs (dired +sidebar)`) is still on my TODO list, so I'll return to this when I can. Thanks for bringing it to my attention in any case!

PR #3848, opened 2020-08-26
Set navkeys for evil-markdown in bepo

gagbo:
No need to fiddle with g-maps here


hlissner:
Looks good to me. Thanks!

PR #3849, opened 2020-08-27
modules/lang: Add GraphQL module

thiagokokada:
Based on e24b8cd1d1e4ceb7e4454a267985b639d5e7eb06, the only difference is support for `.graphql` extension.

Also, cleanup `core/autoload/help.el`, it was pointing for support for some files in `data` module that is not support since the commit above, and was making me confused about why GraphQL was not working.


hlissner:
Hi. Thanks for the PR, but I cannot accept it. The config for graphql is too small to warrant its own module. The updates to autoload/help.el is useful, however. I'll accept a PR for those separately.

Issue #3852, opened 2020-08-27
 Rust module: cargo audit fails to execute correctly


NoxPardalis:
**What did you expect to happen?**
Executing `cargo audit` should successfully run the `cargo-audit` tool if installed.

**What actually happened?**
The execution was unsuccessful resulting in the following error being reported by `cargo-audit`.
```
error: missing argument to option `-f`
```

**Additional details:**
My version of `cargo-audit` is 0.10.0.

The issues lies in the implementation of `+rust/cargo-audit` in `modules/lang/rust/autoload.el`.

```elisp
;; Is this
(rustic-run-cargo-command "cargo audit -f")
```
``` elisp
;; Should probably be this
(rustic-run-cargo-command "cargo audit")
```

I wanted to know if there is a specific reason `-f` is used? Is it simply a holdover from when `cargo-audit` had a different interface?

**Steps to reproduce:**
1. Ensure [`cargo-audit`](https://github.com/RustSec/cargo-audit) is installed.
2. Open a Rust file in a Cargo project.
3. Execute `+rust/cargo-audit`

**System information:**
<details>

```
emacs   version    27.1
        features   XPM JPEG TIFF GIF PNG RSVG CAIRO SOUND DBUS GSETTINGS GLIB NOTIFY INOTIFY LIBSELINUX GNUTLS LIBXML2 FREETYPE HARFBUZZ M17N_FLT LIBOTF ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS LIBSYSTEMD JSON PDUMPER GMP
        build      Aug 27, 2020
        buildopts  (--prefix=/nix/store/grs0hrgnp2hxhlkvqm8z28p80f7kas48-emacs-unstable-emacs-27.1 --disable-build-details --with-modules --with-x-toolkit=gtk3 --with-xft --with-cairo CFLAGS=-DMAC_OS_X_VERSION_MAX_ALLOWED=101200)
        windowsys  x
        daemonp    server-running
doom    version    2.0.9
        build      HEAD -> develop f454d5fac 2020-08-20 21:40:06 -0400
        dir        ~/.doom.d/
system  type       gnu/linux
        config     x86_64-pc-linux-gnu
        shell      /run/current-system/sw/bin/bash
        uname      Linux 5.4.60 #1-NixOS SMP Fri Aug 21 11:05:39 UTC 2020 x86_64
[...]

PR #3853, opened 2020-08-27
Rust module: Modify command for executing `cargo audit`

NoxPardalis:
I'm not sure if the execution of `cargo audit` with the `-f` flag has/had a specific reason. I opened an issue for this on #3852 to discuss this.


hlissner (approved):

hlissner:
Sorry for the _super_ late response, but thanks for the help!

PR #3869, opened 2020-08-29
configure dired-aux variables

t-e-r-m:
`dired-create-destination-dirs` : set to "ask", non-existent directories will be created when copying, if the user consents.
`dired-vc-rename-file` : rename file using the vc system, if present

### DIFF
--- a/modules/emacs/dired/config.el
+++ b/modules/emacs/dired/config.el
@@ -168,6 +168,10 @@ we have to clean it up ourselves."
   (global-set-key [remap find-dired] #'fd-dired)
   (set-popup-rule! "^\\*F\\(?:d\\|ind\\)\\*$" :ignore t))
 
+(use-package! dired-aux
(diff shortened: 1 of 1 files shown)


modules/emacs/dired/config.el
```diff
@@ -168,6 +168,10 @@ we have to clean it up ourselves."
   (global-set-key [remap find-dired] #'fd-dired)
   (set-popup-rule! "^\\*F\\(?:d\\|ind\\)\\*$" :ignore t))
 
+(use-package! dired-aux
```
ymarco, line 171: Add `:defer t` here, otherwise `dired-aux` would be loaded at startup

t-e-r-m: Done! Thanks for the heads up!

hlissner:
Thanks!

PR #3873, opened 2020-08-31
nox doesn't sp xref-backend-identifier-at-point

xuning97:
https://github.com/manateelazycat/nox is a lightweight lsp client.


hlissner:
I've incorporated this fix into a similar fix for elpy in 60ba42c. Thanks for bringing it to my attention.

PR #3876, opened 2020-09-01
Allow :no-native-compile in package recipes

flatwhatson:
This was added to straight with the original native-compilation support, but apparently never worked for Doom package definitions!


hlissner:
Thanks!

PR #3878, opened 2020-09-01
Fix link in gettting_started.org

smsegal:
Original link to company module in section "Writing your own modules" was pointing to a malformed path. This changes it to point to the correct path of the company module readme.


hlissner:
Thanks!

PR #3879, opened 2020-09-01
Add ccls-specific configuration

gagbo:
Most of these changes come from MaskRay's private configuration found
here :
https://github.com/MaskRay/Config/blob/c078dfad340041673f2e056729a44c3df05a6248/home/.config/doom/modules/private/my-cc/config.el

- Limit ccls threads to be maximum half the available core count
- Add wrappers to call ccls LSP extensions
- Enable ccls-code-lens-mode
- Set initial state to emacs when navigating ccls-tree
- Use 'font-lock for semantic highlighting, making the speed choice
  according to variable's docstring

### DIFF
--- a/modules/lang/cc/config.el
+++ b/modules/lang/cc/config.el
@@ -230,7 +230,26 @@ If rtags or rdm aren't available, fail silently instead of throwing a breaking e
   (add-hook! '(c-mode-local-vars-hook
                c++-mode-local-vars-hook
                objc-mode-local-vars-hook)
-    #'lsp!)
+             #'lsp!)
+  (map!
+   (:after ccls
+    :map (c-mode-map c++-mode-map)
+    :n "C-h" (λ! (ccls-navigate "U"))
+    :n "C-j" (λ! (ccls-navigate "R"))
+    :n "C-k" (λ! (ccls-navigate "L"))
+    :n "C-l" (λ! (ccls-navigate "D"))
+    (:localleader
+     :desc "Preprocess file" :n "lp" #'ccls-preprocess-file
+     :desc "Reset cache and reload CCLS" :n "lf" #'ccls-reload)
+    (:after lsp-ui-peek
+     (:localleader
+      :desc "Callers list" :n "c" #'+ccls/caller
+      :desc "Callees list" :n "C" #'+ccls/callee
+      :desc "References (address)" :n "a" #'+ccls/references-address
+      :desc "References (not call)" :n "f" #'+ccls/references-not-call
+      :desc "References (Macro)" :n "m" #'+ccls/references-macro
+      :desc "References (Read)" :n "r" #'+ccls/references-read
+      :desc "References (Write)" :n "w" #'+ccls/references-write))))
 
   (when (featurep! :tools lsp +eglot)
     ;; Map eglot specific helper
@@ -249,7 +268,6 @@ If rtags or rdm aren't available, fail silently instead of throwing a breaking e
                                                               "-isystem/usr/local/include"]
                                                   :resourceDir (cdr (doom-call-process "clang" "-print-resource-dir"))))))))))))
 
-
 (use-package! ccls
   :when (featurep! +lsp)
   :unless (featurep! :tools lsp +eglot)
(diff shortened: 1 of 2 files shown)


gagbo:
Documentation on the initialisation options :
 https://github.com/MaskRay/ccls/wiki/Customization#indexthreads
https://github.com/MaskRay/ccls/wiki/Customization#indextrackdependency

gagbo:
looks like parametersInDeclaration is actually still a valid value, but unsure about what it does https://github.com/MaskRay/ccls/blob/3e04997061bd1313676dafb9897f4a9590f90eeb/src/config.hh#L285

@MaskRay (sorry for the ping) do you put `parametersInDeclaration` to false to get faster indexing ? Does it prevent getting the parameters in the declaration when we ask the server process later ? I could not find doc on the wiki, and I can't understand what it does from looking at the code

modules/lang/cc/config.el
```diff
@@ -260,9 +288,17 @@ If rtags or rdm aren't available, fail silently instead of throwing a breaking e
     (add-to-list 'projectile-project-root-files-bottom-up ".ccls-root")
     (add-to-list 'projectile-project-root-files-top-down-recurring "compile_commands.json"))
   :config
+  (add-hook 'lsp-after-open-hook #'ccls-code-lens-mode)
+  (when (featurep! :editor evil)
+    (evil-set-initial-state 'ccls-tree-mode 'emacs))
```
hlissner, line 293: Use `set-evil-initial-state!`, then you don't need a `(when (featurep! ...) ...)` check.

modules/lang/cc/config.el
```diff
@@ -260,9 +288,17 @@ If rtags or rdm aren't available, fail silently instead of throwing a breaking e
...
+  (setq ccls-sem-highlight-method 'font-lock
+        ccls-initialization-options
+        `(:index (:trackDependency 1
+                  :threads ,(max 1 (/ (string-to-number (shell-command-to-string "nproc")) 2)))))
```
hlissner, line 297: This doesn't seem like a platform independent option. Can this be generalized? Or perhaps wrapped in a `(and IS-MAC IS-LINUX)` check?

hlissner: Also, please use `(cdr (doom-call-process "nproc"))` instead, as it doesn't pull the shell into the mix.

modules/lang/cc/config.el
```diff
@@ -230,7 +230,35 @@ If rtags or rdm aren't available, fail silently instead of throwing a breaking e
...
+      :desc "References (address)"
+      :n "a" #'+ccls/references-address
+      :desc "References (not call)"
+      :n "f" #'+ccls/references-not-call
+      :desc "References (Macro)"
+      :n "m" #'+ccls/references-macro
+      :desc "References (Read)"
+      :n "r" #'+ccls/references-read
+      :desc "References (Write)"
+      :n "w" #'+ccls/references-write))))
```
hlissner, line 261: To be consistent everywhere where :desc is used, please inline the keys and their descriptions.

hlissner (changes requested):

modules/lang/cc/config.el
gagbo: I thought it was universal. It is very much not.

hlissner (changes requested):
Both this and the later usage of `doom-call-process` is incorrect. It takes a list of arguments, not a single command. e.g.

```lisp
(doom-call-process "sysctl" "-n" "hw.ncpu")
;; and
(doom-call-process "clang" "-print-resource-dir")
```

Please correct these. `apply` may help:

```elisp
(apply #'doom-call-process '("sysctl" "-n" "hw.ncpu"))
```

modules/lang/cc/config.el
```diff
@@ -260,9 +278,20 @@ If rtags or rdm aren't available, fail silently instead of throwing a breaking e
...
     (setq ccls-initialization-options
-          `(:clang ,(list :extraArgs ["-isystem/Library/Developer/CommandLineTools/usr/include/c++/v1"
-                                      "-isystem/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include"
-                                      "-isystem/usr/local/include"]
-                          :resourceDir (cdr (doom-call-process "clang" "-print-resource-dir")))))))
+          (append ccls-initialization-options
+                  `(:clang ,(list :extraArgs ["-isystem/Library/Developer/CommandLineTools/usr/include/c++/v1"
+                                              "-isystem/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include"
+                                              "-isystem/usr/local/include"]
+                                  :resourceDir (cdr (doom-call-process "clang -print-resource-dir"))))))))
```
hlissner, line 297: This `doom-call-process` still needs to be fixed to: `(doom-call-process "clang" "-print-resource-dir")`

hlissner (changes requested):

hlissner (approved):

hlissner:

laduke:
hi, just updated doom and saw this in my *Messages*: 

`File local-variables error: (doom-hook-error lsp! (file-missing Searching for program No such file or directory sysctl -n hw.ncpu))`

gagbo:
Yup sorry. fixed in #4075 . You can apply the change manually in your repo and it will work after a restart

PR #3888, opened 2020-09-04
Return org-journal-is-journal to magic-mode-alist.

mprihoda:
After first lazy load, put the `org-journal-is-journal` back to recognize
correctly journal files.

### DIFF
--- a/modules/lang/org/contrib/journal.el
+++ b/modules/lang/org/contrib/journal.el
@@ -18,6 +18,7 @@
                       buffer-file-name (expand-file-name org-journal-dir org-directory))
                      (require 'org-journal nil t)))
         (delq! '+org-journal-p magic-mode-alist 'assq)
+        (add-to-list 'magic-mode-alist '(org-journal-is-journal . org-journal-mode))
         (org-journal-is-journal))))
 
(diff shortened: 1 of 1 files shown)


hlissner:
This shouldn't be necessary. Loading `org-journal` should re-add the `magic-mode-alist` entry when [this line](https://github.com/bastibe/org-journal/blob/master/org-journal.el#L454) is evaluated. If this isn't the case for you, it's likely you are eagerly loading org-journal too early in your private config.

mprihoda:
Thank you, I'll investigate and resubmit only if it's really needed.

mgarabed:
This commit https://github.com/hlissner/doom-emacs/commit/1456108d5bef89195b2b63f962e0ae9d24f4c652 has saved me so many headaches since doing a recent doom upgrade, thank you!

For other people having the issue, the symptom for me was: 
`C-p` and `C-n` would stop working when transitioning buffers and the major mode went from `Journal` to `Org`.  At that point the mappings changed to `evil-paste-pop` and `evil-paste-after`.

Now the major mode stays `Journal` as expected through all the buffer changes.

PR #3899, opened 2020-09-05
Do not remove the journal file detector in :config

mprihoda:
The journal file detector is removed in config section, but it seems
that with current configuration, the evaluation is in this order:

- contrib/journal.el :init section
- org-journal, when triggered
- contrib/journal.el :config section, after org-journal

The org-journal-is-journal is removed in :config section, which means
that journal detection stops working.


mprihoda:
The problematic behaviour seems to be fixed in another way (changing the order of things using require) in 1456108. Not necessary any more.

PR #3907, opened 2020-09-07
Added warning for clang-format in doom doctor.

mlavaert:
Code formatting in Java doesn't work without clang-format, but no
warning is shown in `doom doctor`.


hlissner (changes requested):
Hi! Thanks for the PR and sorry for the late response; I've been on break. 

Doesn't this require `:editor format` enabled? If so, please wrap the check in 

```lisp
(when (featurep! :editor format)
  ...)
```

hlissner (approved):

hlissner:
Thanks!

PR #3920, opened 2020-09-09
modules/checkers/spell: remove tex option

asymmetric:
It causes aspell 0.60.8 to fail with message `the key "check-tex-comments" is unknown`.


hlissner:
I can't reproduce this, but I'll take your word for it. Thanks!

PR #3925, opened 2020-09-10
Fix magic-mode-alist for org-journal after loading

yangsheng6810:
In order for org-journal files to be opened in org-mode, some function needs to be hooked with `magic-mode-alist`. Org-journal handles it with function `org-journal-is-journal`, while Doom overrides the behavior due to lazy-loading concern, and add `+org-journal-p` instead. When `+org-journal-p` decides the loading of org-journal is inevitable, it loads it, and remove itself from `magic-mode-alist`.

However, it didn't add `org-journal-is-journal` back, so there is nothing in `magic-mode-alist` that handles the mode for an org-journal file. This PR fixes the behavior.


hlissner:
This is a duplicate of #3888. Fixed in 1456108

PR #3927, opened 2020-09-11
nix: remove company-nixos-options

Mic92:
This package blocks emacs for several seconds or even more if no network
is available. I have not seen any way to disable it.
It is unlikely that upstream will fix this anytime soon,
hence removing this package:

https://github.com/travisbhartwell/nix-emacs/issues/28

Upstream is not working on this package anymore:

https://github.com/travisbhartwell/nix-emacs/pull/46

### DIFF
--- a/modules/lang/nix/config.el
+++ b/modules/lang/nix/config.el
@@ -5,7 +5,6 @@
   :mode "\\.nix\\'"
   :config
   (set-repl-handler! 'nix-mode #'+nix/open-repl)
-  (set-company-backend! 'nix-mode 'company-nixos-options)
   (set-lookup-handlers! 'nix-mode
     :documentation '(+nix/lookup-option :async t))
 
--- a/modules/lang/nix/packages.el
+++ b/modules/lang/nix/packages.el
@@ -4,8 +4,5 @@
 (package! nix-mode :pin "0cf1ea1e0ed330b59f47056d927797e625ba8f53")
 (package! nix-update :pin "fc6c39c2da3fcfa62f4796816c084a6389c8b6e7")
 
-(when (featurep! :completion company)
-  (package! company-nixos-options :pin "977b9a505ffc8b33b70ec7742f90e469b3168297"))
-
 (when (featurep! :completion helm)
   (package! helm-nixos-options :pin "977b9a505ffc8b33b70ec7742f90e469b3168297"))


Mic92:
@magnetophon is also supporting this: https://github.com/travisbhartwell/nix-emacs/issues/28

hlissner:
Perhaps a better option would be to `(setq company-idle-delay nil)` in `nix-mode`? I have this set by default, so completion doesn't get in my way much, until I really need (then I'm willing to wait the seconds it takes to generate the initial list).

Mic92:
> Perhaps a better option would be to `(setq company-idle-delay nil)` in `nix-mode`? I have this set by default, so completion doesn't get in my way much, until I really need (then I'm willing to wait the seconds it takes to generate the initial list).

I would be ok with waiting if it was not blocking the whole editor.

Mic92:
The other question is, can I somehow disable this backend in my own configuration:
I think tried to disable it with:

```
(set-company-backend! 'nix-mode nil)	
```

but it would be still enabled.

magnetophon:
> @magnetophon is also supporting this: [travisbhartwell/nix-emacs#28](https://github.com/travisbhartwell/nix-emacs/issues/28)

I did imply that, but I'm not so sure anymore.
I think the advantages are bigger than the disadvantages.

Mic92:
I am a lot happier with doom-emacs since I applied this patch before my emacs would hang quite often.

magnetophon:
@Mic92 So how do you search for help on specific options? Like the one under the cursor?
How do you browse options?
Just trying to find out what the alternatives are.

The NixOS website?

hlissner:
> I would be ok with waiting if it was not blocking the whole editor. 

@Mic92 I think you misunderstood: I suggested setting `company-idle-delay` to `nil` so that code-completion doesn't happen automatically, as you type (which would invoke `company-nixos-options` and cause Emacs to block).

With `(setq-hook! 'nix-mode-hook company-idle-delay nil)`, code completion doesn't happen (in nix-mode buffers) until you manually invoke it with `C-SPC`. Then you can control when you suffer the initial delay. This way, we don't have to throw out `company-nixos-options`, which I find still find useful when I do need it.

> So how do you search for help on specific options? Like the one under the cursor?
> How do you browse options?
> Just trying to find out what the alternatives are.

Doom is exploiting `nixos-options` to provide a lookup-documentation command for nix (and other languages). Press `K` in normal mode (or `C-c c k` for non-evil users) and ivy/helm consults `nixos-options` and will throw up its results in a popup. Not perfect, but it's something.

Mic92:
Ok, adding `(setq-hook! 'nix-mode-hook company-idle-delay nil)` in my configuration seems to help

Issue #3188, opened 2020-05-21
+notmuch/update () does not close window after finish


sonnius:
**What did you expect to happen?**
...
When calling this function I expect the opened window to show me information, once the functions exits, the window should close when successful

**What actually happened?**
...
The split window stays open, I have to manually close it

**Additional details:**

 (after! notmuch
   (setq +notmuch-sync-backend 'custom
         notmuch-command "~/bin/remote-notmuch.sh"
         +notmuch-sync-command "~/bin/remote-mail-sync.sh"))

**Steps to reproduce:**

run Mx - +notmuch/update

**System information:**
<details><pre>

```
emacs   version    26.3
        features   JPEG RSVG IMAGEMAGICK GLIB NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES THREADS LCMS2
        build      Nov 18, 2019
        buildopts  (--disable-dependency-tracking --disable-silent-rules --enable-locallisppath=/usr/local/share/emacs/site-lisp --infodir=/usr/local/Cellar/emacs-plus/26.3/share/info/emacs --prefix=/usr/local/Cellar/emacs-plus/26.3 --with-xml2 --without-dbus --with-gnutls --with-imagemagick --with-modules --with-rsvg --with-ns --disable-ns-self-contained)
        windowsys  ns
        daemonp    server-running
doom    version    2.0.9
        build      HEAD -> develop 885197bd0 2020-05-20 16:08:52 -0400
        dir        ~/.doom.d/
system  type       darwin
        config     x86_64-apple-darwin19.0.0
        shell      /usr/local/bin/zsh
        uname      Darwin 19.4.0 Darwin Kernel Version 19.4.0: Wed Mar  4 22:28:40 PST 2020; root:xnu-6153.101.6~15/RELEASE_X86_64 x86_64
[...]

PR #3933, opened 2020-09-13
Notmuch -- make +afew feature and  fix +notmuch/update buffer kill

nackjicholson:
Addressing #3188 

Using `compile` approach similar to `doom sync`. It runs in a buffer and
deletes the buffer when complete.

`afew` is and should be optional. Putting it behind a feature flag so
that if you don't have afew installed, it won't be run. It's easily
available for those that want it, and easy to ignore for everyone else.

added documentation.


hlissner:
Some more work was needed to complete notmuch's readme, which I've collect into 494d87d. Thanks for your help and bringing it to my attention!

PR #3934, opened 2020-09-13
Add `org-journal-new-scheduled-entry` Keybindings

theutz:
Added a keybinding `SPC n j J`` to create a scheduled entry in Org Journal.


hlissner (approved):

hlissner:
Thanks!

Issue #3935, opened 2020-09-13
Wrong type argument on +spell/add-word


zeronone:
**What actually happened?**
When using `+spell/add-word` the following error happens.
```
Debugger entered--Lisp error: (wrong-type-argument characterp "Karatsuba")
  concat("*" ("Karatsuba" 9009 9018) "\n")
  (ispell-send-string (concat "*" word "\n"))
  (cond ((null scope) (ispell-send-string (concat "*" word "\n")) (ispell-send-string "#\n") (flyspell-unhighlight-at (point)) (setq ispell-pdict-modified-p '(t))) ((memq scope '(buffer session)) (ispell-send-string (concat "@" word "\n")) (add-to-list 'ispell-buffer-session-localwords word) (or ispell-buffer-local-name (setq ispell-buffer-local-name (buffer-name))) (flyspell-unhighlight-at (point)) (if (null ispell-pdict-modified-p) (setq ispell-pdict-modified-p (list ispell-pdict-modified-p))) (if (eq replace 'buffer) (ispell-add-per-file-word-list word))))
  +spell/add-word(("Karatsuba" 9009 9018) nil)
  #<subr funcall-interactively>(+spell/add-word ("Karatsuba" 9009 9018) nil)
  apply(#<subr funcall-interactively> (+spell/add-word ("Karatsuba" 9009 9018) nil))
  funcall-interactively(+spell/add-word ("Karatsuba" 9009 9018) nil)
  #<subr call-interactively>(+spell/add-word nil nil)
  apply(#<subr call-interactively> (+spell/add-word nil nil))
  explain-pause--wrap-call-interactively(#<subr call-interactively> +spell/add-word nil nil)
  apply(explain-pause--wrap-call-interactively #<subr call-interactively> (+spell/add-word nil nil))
  call-interactively(+spell/add-word nil nil)
  command-execute(+spell/add-word)
```

**Steps to reproduce:**
1. Place pointer at a word
2. `z q` for `+spell/add-word`

**System information:**
<details>

```
SYSTEM  type       darwin
        config     x86_64-apple-darwin19.5.0
        shell      /bin/zsh
        uname      Darwin 19.6.0 Darwin Kernel Version 19.6.0: Thu Jun 18 20:49:00 PDT 2020; root:xnu-6153.141.1~1/RELEASE_X86_64 x86_64
[...]

Issue #3890, opened 2020-09-04
`+spell/add-word` yields an error when `:checkers (spell +flyspell)` is enabled


karvus:
**What did you expect to happen?**

The word under the point being added to my personal dictionary.

**What actually happened?**

An error is signalled.

**Additional details:**

`:checkers (spell +flyspell)` is present in my `init.el`

The key combination is `z q` which is bound to `+spell/add-word`

The following backtrace is produced.

<details><pre>
Debugger entered--Lisp error: (wrong-type-argument characterp "frop")
  concat("*" ("frop" 47 51) "\n")
  (ispell-send-string (concat "*" word "\n"))
  (cond ((null scope) (ispell-send-string (concat "*" word "\n")) (ispell-send-string "#\n") (flyspell-unhighlight-at (point)) (setq ispell-pdict-modified-p '(t))) ((memq scope '(buffer session)) (ispell-send-string (concat "@" word "\n")) (add-to-list 'ispell-buffer-session-localwords word) (or ispell-buffer-local-name (setq ispell-buffer-local-name (buffer-name))) (flyspell-unhighlight-at (point)) (if (null ispell-pdict-modified-p) (setq ispell-pdict-modified-p (list ispell-pdict-modified-p))) (if (eq replace 'buffer) (ispell-add-per-file-word-list word))))
  +spell/add-word(("frop" 47 51) nil)
  funcall-interactively(+spell/add-word ("frop" 47 51) nil)
  call-interactively(+spell/add-word nil nil)
  command-execute(+spell/add-word)
</pre></details>

It seems that `+spell/add-word` expects `flyspell-get-word` to return as a string the current word.  Instead, it returns `(<word> <start-pos> <end-pos>)`.

I tried the following modification, and it seems to produce the intended behavior.

<details>
<pre>
diff --git a/modules/checkers/spell/autoload/+flyspell.el b/modules/checkers/spell/autoload/+flyspell.el
index 97787d1a6..e555ebd92 100644
--- a/modules/checkers/spell/autoload/+flyspell.el
+++ b/modules/checkers/spell/autoload/+flyspell.el
@@ -35,7 +35,7 @@ SCOPE can be `buffer' or `session' to exclude words only from the current buffer
 or session. Otherwise, the addition is permanent."
   (interactive
    (list (progn (require 'flyspell)
-                (flyspell-get-word))
+                (car (flyspell-get-word)))
          (cond ((equal current-prefix-arg '(16))
                 'session)
[...]

PR #3938, opened 2020-09-14
Fix `+spell/add-word` when using flyspell

karvus:
This addressed #3890 and #3935

``+spell/add-word`` expects `flyspell-get-word` to return as a string
the current word. Instead, it returns a list of the form
`(<word> <start-pos> <end-pos>)`.


kartikynwa:
Can confirm this works

hlissner (approved):

hlissner:
Good catch, thanks!

PR #3947, opened 2020-09-15
org-trello: ignore credentials

peterhoeg:
org-trello forces credentials to be stored under user-emacs-directory using
defconst, so ignore the directory to avoid accidentally committing secrets.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .local/
 var/
 modules/private
+.trello/
 .yas-compiled-snippets.el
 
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Hi! Thanks for the PR. I was a bit hesitant to accept this, since Doom doesn't configure org-trello, but I've come up with a more general solution: to ignore all dotfiles in root, which should resolve this issue. Thanks for bringing it to my attention!

PR #3966, opened 2020-09-21
:input layout bepo enhancements

### LINKED ISSUE #3858: :input layout improvements
Evil collection hook seem to eagerly run the hook on too many maps, even when not defined. [Code to check](https://github.com/emacs-evil/evil-collection/blob/567ee24d443906fbcd8b192f8505a3b4b0ee3129/evil-collection.el#L551-L586)

Also the rotation functions used in the hooks are from general.el API. It would probably be better to use evil-collection-translate-key

gagbo:
Fixes #3858

## Evil-collection instead of general for key translation

This is the biggest, most meaningful change. Using `evil-collection-translate-key` instead of `general-translate-key` prevents all the void-variable errors that happened because the general utility doesn't defer the key remapping on keymaps that aren't defined yet. See emacs-evil/evil-collection#368

## Split evil-specific changes
There are changes that are not necessarily oriented only to evil users, so splitting these off makes sense to know what you get if you enable this module without evil enabled.

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -68,9 +68,10 @@
       (:after geiser-doc :map geiser-doc-mode-map
        :n "o"    #'link-hint-open-link)
 
-      (:after (evil-org evil-easymotion)
-       :map evil-org-mode-map
-       :m "gsh" #'+org/goto-visible)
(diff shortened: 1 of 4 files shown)


hlissner (approved):

hlissner:
Thanks!

PR #3980, opened 2020-09-26
Load yas when yas-activate-extra-mode is called

yangsheng6810:
Auto load when `yas-activate-extra-mode` or `yas-deactivate-extra-mode` is called. These functions are called in `.emacs.d/modules/editor/snippets/autoload/snippets.el` by function `+snippets-enable-project-modes-h`, which may be called before yasnippet is loaded.


hlissner (approved):

hlissner:
Thanks!

PR #3981, opened 2020-09-26
Warn when .emacs.elc file is found upon install

Remmy:


hlissner:
Hi! Thanks for the PR. I'd like to know more about the circumstances that led up to an `~/.emacs.elc` existing. When and where is this file created? Is there really an installer for emacs out there for some distro which both deploys a `~/.emacs` _and_ byte-compiles it?

Remmy:
>Hi! Thanks for the PR. I'd like to know more about the circumstances that led up to an `~/.emacs.elc` existing. When and where is this file created? Is there really an installer for emacs out there for some distro which both deploys a `~/.emacs` _and_ byte-compiles it?

I must admit I byte compiled it by myself. Not sure why, I think it may have been beginner error on my part or something which I took from some random webpage about emacs. It gave issues starting doom emacs up, but if I am the only person doing that sort of silly thing, I'd be fine with leaving things as is.

Thanks!

hlissner:
That does sound like likely (since this is the first time I've heard of it). No worries! For now I'll chalk it up as a local issue. Thanks for letting me know.

PR #3815, opened 2020-08-23
doc/adding editorconfig and hl-column-fill documentation

jeetelongname:
Added docs for editorconfig and hl-column-fill

----

#

PR #3982, opened 2020-09-26
Improve fill-column module for Emacs 27+

rushsteve1:
Emacs 27 introduced `display-fill-column-indicator-mode` which replaces `hl-fill-column`. This commit adds some conditionals to use `display-fill-column-indicator-mode` on Emacs 27+, or falls back to `hl-fill-column-mode` on older versions.

This also fixes the deprecation message so I removed it. Once Emacs 27 is more common I would suggest marking it as deprecated again and removing the module entirely. I'll add the message back if you still want it there.

I suggest merging #3815 first since that adds documentation which I can update to match this change.

Any and all suggestions welcome! I'm not great at Elisp yet and in particular the `add-hook!` macro gave me trouble which led to a bit of repetition in the code.

### DIFF
--- a/modules/ui/fill-column/autoload.el
+++ b/modules/ui/fill-column/autoload.el
@@ -1,9 +1,12 @@
 ;;; ui/fill-column/autoload.el -*- lexical-binding: t; -*-
 
-;; DEPRECATED Replaced by `display-fill-column-indicator-mode' in Emacs 27+
-
 ;;;###autoload (autoload 'hl-fill-column-mode "hl-fill-column" nil t)
 
 ;;;###autoload
-(add-hook! '(text-mode-hook prog-mode-hook conf-mode-hook)
-           #'hl-fill-column-mode)
+;; Emacs 27 introduced `display-fill-column-indicator-mode' which should be
+;; used instead of `hl-fill-column-mode'
+(if (>= emacs-major-version 27)
+    (add-hook! '(text-mode-hook prog-mode-hook conf-mode-hook)
+               #'display-fill-column-indicator-mode)
+    (add-hook! '(text-mode-hook prog-mode-hook conf-mode-hook)
+               #'hl-fill-column-mode))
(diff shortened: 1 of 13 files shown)


rushsteve1:
Two additional commits with changes suggested on Discord.

modules/ui/fill-column/autoload.el
```diff
@@ -1,9 +1,12 @@
...
 ;;;###autoload
-(add-hook! '(text-mode-hook prog-mode-hook conf-mode-hook)
-           #'hl-fill-column-mode)
+;; Emacs 27 introduced `display-fill-column-indicator-mode' which should be
+;; used instead of `hl-fill-column-mode'
+(if EMACS27+
+    (add-hook! '(text-mode-hook prog-mode-hook conf-mode-hook)
+               #'display-fill-column-indicator-mode)
+    (add-hook! '(text-mode-hook prog-mode-hook conf-mode-hook)
+               #'hl-fill-column-mode))
```
hlissner, line 12: Could you replace this duplication with a delegate? e.g.

```elisp
;;;###autoload
(defun +fill-column-enable-h (&optional arg)
  (interactive "p")
  (if (fboundp 'display-fill-column-indicator-mode)
      (display-fill-column-indicator-mode arg)
    (hl-fill-column-mode arg)))
```

Then we can have only one `add-hook!`, and one `:when` check in config/default/+{evil,emacs}-bindings.el.

hlissner (changes requested):

modules/ui/fill-column/autoload.el
rushsteve1: Ah, I spent forever trying to reduce that duplication. Good to know that trick, should help me a lot.

rushsteve1:
Thanks for the suggestions @hlissner, I made them and pushed the changes.

Also ignore the force-push, I had some weird merge commits I needed to remove.

hlissner:
Looks good to me. Thanks!

PR #3989, opened 2020-09-27
tools/lsp: minor improvements

ericdallo:
After discussing some of these settings with @yyoncho (thanks for the help), we as `lsp-mode` maintainers think that these changes could be a better default/performance for doom users.

### DIFF
--- a/modules/tools/lsp/+lsp.el
+++ b/modules/tools/lsp/+lsp.el
@@ -42,7 +42,6 @@ should be a deliberate act (as is flipping this variable).")
   ;; Disable features that modify our code without our permission.
   (setq lsp-enable-indentation nil
         lsp-enable-on-type-formatting nil)
-
   :config
   (pushnew! doom-debug-variables 'lsp-log-io 'lsp-print-performance)
 
@@ -62,10 +61,6 @@ should be a deliberate act (as is flipping this variable).")
     :type-definition #'lsp-find-type-definition
     :references #'lsp-find-references)
 
-  (when lsp-auto-configure
-    (mapc (lambda (package) (require package nil t))
-          (cl-remove-if #'featurep lsp-client-packages)))
-
   (defadvice! +lsp--dont-auto-install-servers-a (orig-fn &rest args)
     "Replace auto-install behavior with warning and support indirect buffers."
     :around #'lsp
@@ -83,21 +78,11 @@ should be a deliberate act (as is flipping this variable).")
               (defun lsp--completing-read (msg clients &rest _)
                 (lsp--warn (concat msg "%s\n"
                                    "  Use `M-x lsp-install-server' to install a supported server, or read "
-                                   "https://emacs-lsp.github.io/lsp-mode/page/languages for more.")
+                                   "https://emacs-lsp.github.io/lsp-mode/page/languages for more details.")
                            (mapcar #'lsp--client-server-id clients))
                 (throw 'not-installed nil)))
         (catch 'not-installed
           (apply orig-fn args)))))
-
-  (defadvice! +lsp--respect-user-defined-checkers-a (orig-fn &rest args)
-    "Set up flycheck-mode or flymake-mode, depending on `lsp-diagnostic-package'."
-    :around #'lsp-diagnostics--flycheck-enable
-    (if flycheck-checker
-        ;; Respect file/dir/explicit user-defined `flycheck-checker'.
-        (let ((old-checker flycheck-checker))
-          (apply orig-fn args)
-          (setq-local flycheck-checker old-checker))
-      (apply orig-fn args)))
 
   (add-hook! 'lsp-mode-hook
     (defun +lsp-display-guessed-project-root-h ()
--- a/modules/tools/lsp/README.org
+++ b/modules/tools/lsp/README.org
@@ -76,6 +76,12 @@ As of this writing, this is the state of LSP support in Doom Emacs:
   that it is better etiquette to let the user decide when their development
   environment is modified. Use ~M-x lsp-install-server~ to install LSP servers
   manually.
++ ~lsp-mode~ has also been modified to not automatically format your code. If you
+  want to re-enable that behaviour, you can add:
+#+BEGIN_SRC elisp
+(setq lsp-enable-indentation t
+      lsp-enable-on-type-formatting t)
+#+END_SRC
 
 * Prerequisites
 This module has no direct prerequisites, but major-modes require you to install
--- a/modules/tools/lsp/config.el
+++ b/modules/tools/lsp/config.el
@@ -18,7 +18,7 @@ killing and opening many LSP/eglot-powered buffers.")
             (bound-and-true-p lsp-mode))
     ;; `read-process-output-max' is only available on recent development
     ;; builds of Emacs 27 and above.
-    (setq-local read-process-output-max (* 1024 1024))
+    (setq read-process-output-max (* 1024 1024))
     ;; REVIEW LSP causes a lot of allocations, with or without Emacs 27+'s
     ;;        native JSON library, so we up the GC threshold to stave off
     ;;        GC-induced slowdowns/freezes. Doom uses `gcmh' to enforce its GC


modules/tools/lsp/+lsp.el
```diff
@@ -4,12 +4,11 @@
   "The backends to prepend to `company-backends' in `lsp-mode' buffers.
 Can be a list of backends; accepts any value `company-backends' accepts.")
 
-(defvar +lsp-auto-install-servers nil
+(defvar +lsp-auto-install-servers t
```
ericdallo, line 7: For a better IDE like, we think that is better ask to user that lsp-mode has a server available and that can install automatically, the user can input `no` if he doesn't want to install it, or disable this variable.
This will help beginner users that don't know about this LSP world and just want things to just work.

modules/tools/lsp/+lsp.el
```diff
@@ -39,10 +38,6 @@ should be a deliberate act (as is flipping this variable).")
         lsp-enable-folding nil
         lsp-enable-text-document-color nil)
 
-  ;; Disable features that modify our code without our permission.
-  (setq lsp-enable-indentation nil
-        lsp-enable-on-type-formatting nil)
```
ericdallo, line 44: If user is using `+lsp` flag, he probably wants an IDE like setup, so format is something that could help.

modules/tools/lsp/+lsp.el
```diff
@@ -62,10 +57,6 @@ should be a deliberate act (as is flipping this variable).")
     :type-definition #'lsp-find-type-definition
     :references #'lsp-find-references)
 
-  (when lsp-auto-configure
-    (mapc (lambda (package) (require package nil t))
```
ericdallo, line 66: lsp-mode already does that automatically.

modules/tools/lsp/+lsp.el
```diff
@@ -83,22 +74,12 @@ should be a deliberate act (as is flipping this variable).")
...
 
-  (defadvice! +lsp--respect-user-defined-checkers-a (orig-fn &rest args)
-    "Set up flycheck-mode or flymake-mode, depending on `lsp-diagnostic-package'."
-    :around #'lsp-diagnostics--flycheck-enable
-    (if flycheck-checker
-        ;; Respect file/dir/explicit user-defined `flycheck-checker'.
-        (let ((old-checker flycheck-checker))
-          (apply orig-fn args)
-          (setq-local flycheck-checker old-checker))
-      (apply orig-fn args)))
```
ericdallo, line 100: This is not working [since months](https://github.com/emacs-lsp/lsp-mode/pull/1990), also lsp-mode has `lsp-checker-provider` which user could change if he wants another checker.

modules/tools/lsp/+lsp.el
gagbo: Pretty sure it goes against https://github.com/hlissner/doom-emacs/blob/develop/README.md#introduction

> Your system, your rules. You know better. At least, Doom hopes so! It won't automatically install system dependencies (and will force plugins not to either). Rely on doom doctor to tell you what's missing.

And the actual command to install servers is in the readme https://github.com/hlissner/doom-emacs/tree/develop/modules/tools/lsp#hacks

If beginners can't/don't want to check the modules' documentation I think it should be fixed first

modules/tools/lsp/+lsp.el
gagbo: Pretty sure the exact opposite can be said as well. I'm a user and I don't want those in my setup.

The main issue with changing these defaults is the storm of issues that come just after it.

I'd feel better if setting those would just be added in the readme of the module, or just linking to lsp-mode documentation so anyone can go on the great documentation site to know what they can customize so the y can tweak it.

modules/tools/lsp/+lsp.el
gagbo: [For reference, on current lsp-mode pin](https://github.com/emacs-lsp/lsp-mode/blob/4145a70ce1d4bfb2463606ba34c5965080b080d9/lsp-mode.el#L7449-L7453)

modules/tools/lsp/+lsp.el
ericdallo: I see, if this is something that doom wants, I'll revert this change.

modules/tools/lsp/+lsp.el
ericdallo: Ok, I'll revert this change and add a link on the module's README

ericdallo:
Done @gagbo, thanks for the review

modules/tools/lsp/+lsp.el
gagbo: Just to be clear, I don't choose if the change gets included or not. I'm just trying to explain a rationale against that change.

In any case, I'm pretty sure there are as much people for `(t t)` as there are for `(nil nil)` or any other combination. So my argument was really more on "changing defaults" and less on the actual values. If the support cost of changing anything default in Doom wasn't so steep I wouldn't mind at all.

But hopefully documentation for Doom can be seen enough to help users tweak it to their liking. But currently, besides locking Emacs until new users can prove they know how to search documentation in Emacs I don't know what could be done.

(And maybe the CLI rework will provide a platform we can leverage to communicate those changes on update, similar to `home-manager news` or `guix pull --news`)

ericdallo: > Just to be clear, I don't choose if the change gets included or not. I'm just trying to explain a rationale against that change.

Yeah, I changed after re-checking the doom's "mantra" and keep this following those rules.

> But hopefully documentation for Doom can be seen enough to help users tweak it to their liking. But currently, besides locking Emacs until new users can prove they know how to search documentation in Emacs I don't know what could be done.

Yes, I agree, thanks :)

yyoncho: > The main issue with changing these defaults is the storm of issues that come just after it.

I think that there is a lack of understanding of what these settings and the rest of the changes will cause. E. g. not setting lsp-enable-indentation will just make `indent-region` command to work differently. And the effect of those "optimizations" is in lsp-mode gitter which is full of users complaining that something does not work and that they have spent hours figuring out why it does not work and at the same time the lsp-mode defaults would handle that case well.

gagbo: I see. It's weird that it behaves differently. Do you have an example from the gitter where it does not behave ?

When lsp-mode is active, the formatter [eventually runs lsp-format-region](https://github.com/hlissner/doom-emacs/blob/1456108d5bef89195b2b63f962e0ae9d24f4c652/modules/editor/format/autoload/format.el#L253-L256) as well. Might be an issue around these LOCs actually 

the indirection is added to provide a single entrypoint to various formatters I guess ? If `format-all` [1] actually sets indent-region-function on top of providing the format-all-* entrypoint, it's probably worth to refactor this whole code to be closer to basic emacs. But I don't know if it's the case right now

[1] which has been a little buggy a while ago, don't know status yet

gagbo: (For the record, the flag in the code linked before does [defaults to t](https://github.com/hlissner/doom-emacs/blob/c4e8aa0baa4d044092e9021e9309e20ca042722b/modules/editor/format/config.el#L26-L30) )

modules/tools/lsp/README.org
```diff
@@ -76,6 +76,12 @@ As of this writing, this is the state of LSP support in Doom Emacs:
   that it is better etiquette to let the user decide when their development
   environment is modified. Use ~M-x lsp-install-server~ to install LSP servers
   manually.
++ Also ~lsp-mode~ has ben modified to not automatically format your code. If you
```
aclonegeek, line 79:
```suggestion
+ Also ~lsp-mode~ has been modified to not automatically format your code. If you
```
I prefer this wording personally:

```suggestion
+ ~lsp-mode~ has also been modified to not automatically format your code. If you
```

modules/tools/lsp/+lsp.el
yyoncho: > When lsp-mode is active, the formatter [eventually runs lsp-format-region](https://github.com/hlissner/doom-emacs/blob/1456108d5bef89195b2b63f962e0ae9d24f4c652/modules/editor/format/autoload/format.el#L253-L256) as well. Might be an issue around these LOCs actually

The calling `indent-region` will be broken not the adhoc doom implementation. And again, this is the only thing that you are preventing. Similarly, instead of letting lsp-mode warn about file count and asking the user whether he wants to watch the project, you turn off file watches which results in a broken java experience. E. g. after each change of resource users should perform a clean build which may take minutes, the projects are not reimported automatically. Similarly, lsp-mode does not install anything automatically but the user is asked to install the server if the wants, and this is not replaced from lsp-install-server because it does not provide the list of servers supported by the current file. 

Quote from lsp-mode gitter:

```
that's a price I'm willing to pay if it helps me not have these random (not really random) issues
I spent a good 1h tracking down the issue with the test just now, which might've not happened if lsp was watching the files
so 4-5s every now and then is a small price to pay considering that I might be spending 1+h debugging non-issues :)
```

ericdallo: I can confirm this happens to `lsp-clojure` and `lsp-dart` too

numkem:
When running this with unpinned `lsp-mode` and `lsp-ui` with `gopls` `master` I end up with a broken LSP that says:
```
LSP :: There are no language servers supporting current mode `go-mode' registered with `lsp-mode'.
This issue might be caused by:
1. You haven't registered/loaded external language server package which doesn't ship with `lsp-mode'(e. g. `lsp-java', `lsp-metals').
2. The language server that you expect to run is not configured to run for major mode `go-mode'. You may check that by checking the `:major-modes' that are passed to `lsp-register-client'.
3. `lsp-mode' doesn't have any integration for the language behind `go-mode'. Refer to https://emacs-lsp.github.io/lsp-mode/page/languages and https://langserver.org/ .
```

gagbo:
> When running this with unpinned `lsp-mode` and `lsp-ui` with `gopls` `master` I end up with a broken LSP that says:


Hello,

Thanks for the report but it's not related to this PR (unless this happens on the lsp-mode pin in the PR). It's better to track those separately

numkem:
It doesn't happen when I run the same environment without this PR. It seems to be related to the removal of lsp.el L65, re-adding those lines fixes it for me.

ericdallo:
This is odd as lsp-mode already have that code, are you sure you have re-byte-compiled your `lsp-mode`?

hlissner:
Sorry for the late reply; I've been burning through a massive backlog after returning from a long break. There's a lot to unpack here, but a little more needed to be done to accommodate some of these changes (which I've done in 22b6eae).

In a nutshell:

1. I've decided to do away with suppressing installation. At least `lsp-mode` is prompting for confirmation. 
2. I've rewritten how the optimizations are applied (they were buffer-local in an attempt to isolate the change -- as I can't anticipate what side effects it could have elsewhere.
3. I've enabled file-watchers -- if this turns out to be as big a performance issue as I fear, I'll turn it off on a per-client basis later.
4. ```lisp
     (when lsp-auto-configure
       (mapc (lambda (package) (require package nil t))
   ```

   These lines are no longer necessary after no.1 above.

That should address everything that's been brought up here. Let me know if I've missed something.

> This is not working since months, also lsp-mode has lsp-checker-provider which user could change if he wants another checker.

I haven't looked into this yet. A quick search in lsp-mode's source produces no results for a `lsp-checker-provider`. Is there somewhere I can read up on this so I can update our documentation?

ericdallo:
Thank you @hlissner!

> I haven't looked into this yet. A quick search in lsp-mode's source produces no results for a lsp-checker-provider. Is there somewhere I can read up on this so I can update our documentation?

Sorry, the correct variable is `lsp-diagnostics-provider` from `lsp-diagnostics.el`

hlissner:
> Sorry, the correct variable is lsp-diagnostics-provider from lsp-diagnostics.el.
We have this tutorial for enabling/disabling lsp features and you can find more info about lsp-diagnostics-provider here

Oh, but that's not what this advice was dealing with (the docstring was misleading though, my bad):

```lisp
  (defadvice! +lsp--respect-user-defined-checkers-a (orig-fn &rest args)
    "Set up flycheck-mode or flymake-mode, depending on `lsp-diagnostic-package'."
    :around #'lsp-diagnostics--flycheck-enable
    (if flycheck-checker
        ;; Respect file/dir/explicit user-defined `flycheck-checker'.
        (let ((old-checker flycheck-checker))
          (apply orig-fn args)
          (setq-local flycheck-checker old-checker))
      (apply orig-fn args)))
```

The point of this is, if the user has an explicit `flycheck-checker` set for lsp-mode buffers, it wouldn't be overwritten with `lsp`.

ericdallo:
I see, sorry, I think we can keep it so

PR #3991, opened 2020-09-28
Add lookup handler to Agda module

bottomtype:
agda2-mode has its own function for jumping to a definition, which this adds as a lookup handler.


hlissner (approved):

hlissner:
Thanks!

PR #3992, opened 2020-09-28
Add keybinding for tear-off-window

rushsteve1:
Added they Evil keybinding `SPC w T` for `tear-off-window` since that is a fairly useful windowing function.

This is just a simple 2-line change, but I thought it would be a nice addition.


hlissner (approved):

hlissner:
Makes sense to me. Thanks!

PR #3995, opened 2020-09-29
Adjusting org journal functions to new commands

samwhitlock:
Silences some obsolescence warnings when navigating among entries


hlissner (approved):

hlissner:
Thanks!

Issue #2542, opened 2020-02-16
Spacemacs Keybinding Migration Guide


howdoicomputer:
So I just migrated from Spacemacs over to Doom and... I'm *really* enjoying the experience so far. Everything seems so much more tightly integrated and fluid and stable and fast that I am just *enamored* so far.

However, I cam across this blurb in the README:

```
Have you migrated from Spacemacs? Help me flesh out this section by letting me know what kind of hurdles you faced in doing so. You’ll find me on our Discord server.
```

I didn't know what channel to use in the Discord for this type of feedback so I thought I would make it an issue.

The biggest problem that I'm having right now with the transition from Spacemacs to Doom is adjusting to the key bindings. As most people probably know, the key bindings to Spacemacs are mnemonic - which really helps with the whole "memorizing a billion keybindings" thing.

Now that I'm transitioning, I'm having a difficult time learning the key combinations and not instinctively using the old key combos. I find myself trying `SPC p t` to bring up treemacs and `SPC g s` to bring up a Magit status buffer. Oh, and `SPC c l` to comment out lines - I've been trying to use the comment line keybind (`C-x C-;`) and keep hitting (`C-x C-l`) instead lol.

My suggestion is to have migration documentation that helps with learning new keybindings. Perhaps a cheat sheet that details common actions taken and translates them to Doom keybindings? I've also been thinking about writing a private module that just replaces all the Doom keybindings with the ones from Spacemacs. Your thoughts?

howdoicomputer:
(also I can help with this if it becomes a markdown page)

hlissner:
> I didn't know what channel to use in the Discord for this type of feedback so I thought I would make it an issue.

The #development channel is best suited for these kinds of discussions.

> My suggestion is to have migration documentation that helps with learning new keybindings. Perhaps a cheat sheet that details common actions taken and translates them to Doom keybindings?

I'm all for it. Trouble is: I'm clueless on the matter -- I've never used Spacemacs at length. It'd be a great help if you could flesh out [the Spacemacs migration stub in our Getting Started page](https://github.com/hlissner/doom-emacs/blob/develop/docs/getting_started.org#from-spacemacs). Or at least include a link there to a spacemacs cheatsheet from there.

> I've also been thinking about writing a private module that just replaces all the Doom keybindings with the ones from Spacemacs. Your thoughts?

[...]

PR #4003, opened 2020-10-01
Add SPC / in evil-bindings

ctjhoa:
Add SPC / alias for "Search project" (SPC s p)

- Searching is one recurring action by developers

- Spacemacs has the same keybinding,
  so it close one gap between Spacemacs and Doom keybindings (#2542)

- It's consistent with SPC * "Search for symbol in projet"
  as it's Vim "*" and "/" equivalent but project-wide


hlissner (approved):

hlissner:
Makes sense to me. Thanks!

PR #3988, opened 2020-09-27
exwm module

chayward1:
https://github.com/hlissner/doom-emacs/issues/3117

Here is my implementation of an exwm module for doom emacs, some of the features:

- Multi-monitor support
- Auto Pluggable HDMI support

(Optional)
- Firefox browser support for evil/non-evil users via exwm-firefox-*
- Statusbar via doom/modeline and display-time-mode and display-battery-mode both of which are included in emacs

PR #4004, opened 2020-10-01
New module: os/exwm

chayward1:
https://github.com/hlissner/doom-emacs/pull/3988

Hello again, I have fixed the issues I was having with module flags. I was unable to reopen the previous PR so I linked to the old one here.

Here is my implementation of an exwm module for doom emacs, some of the features:
- Multi-monitor support
- Auto Pluggable HDMI support

(Optional)
- Firefox browser support for evil/non-evil users via exwm-firefox-*
- Statusbar via doom/modeline and display-time-mode and display-battery-mode both of which are included in emacs

### DIFF
--- a/modules/os/exwm/config.el
+++ b/modules/os/exwm/config.el
@@ -1,69 +1,78 @@
 ;;; os/exwm/config.el -*- lexical-binding: t; -*-
 
+;; Confgure `exwm' the X window manager for emacs.
+(use-package! exwm
+  :config
+  ;; Configure global key bindings.
+  (setq exwm-input-global-keys
+        `(([?\s-r] . exwm-reset)
+          ([?\s-w] . exwm-worksplace-switch)
+          ([?\s-&] . (lambda (command)
+                       (interactive (list (read-shell-command "$ ")))
+                       (start-process-shell-command command nil command)))
+          ,@(mapcar (lambda (i)
+                      `(,(kbd (format "s-%d" i)) .
+                        (lambda ()
+                          (interactive)
+                          (exwm-workspace-switch-create ,i))))
+                    (number-sequence 0 9))))
+
+  ;; Configure the default buffer behaviour. All buffers created in `exwm-mode'
+  ;; are named "*EXWM*". Change it in `exwm-update-class-hook' and `exwm-update-title-hook'
+  ;; which are run when a new X window class name or title is available.
+  (add-hook 'exwm-update-class-hook
+            (lambda ()
+              (unless (or (string-prefix-p "sun-awt-X11-" exwm-instance-name)
+                          (string= "gimp" exwm-instance-name))
+                (exwm-workspace-rename-buffer exwm-class-name))))
+  (add-hook 'exwm-update-title-hook
+            (lambda ()
+              (when (or (not exwm-instance-name)
+                        (string-prefix-p "sun-awt-X11-" exwm-instance-name)
+                        (string= "gimp" exwm-instance-name))
+                (exwm-workspace-rename-buffer exwm-title))))
+
+  ;; Enable the window manager.
+  (exwm-enable))
+
 ;; Use the `ido' configuration for a few configuration fixes that alter
 ;; 'C-x b' workplace switching behaviour. This also effects the functionality
 ;; of 'SPC .' file searching in doom regardless of the users `ido' configuration.
-(require 'exwm)
-(require 'exwm-config)
-(exwm-config--fix/ido-buffer-window-other-frame)
+(use-package! exwm-config
+  :after exwm
+  :config
+  (exwm-config--fix/ido-buffer-window-other-frame))
 
 ;; Configure `exwm-randr' to support multi-monitor setups as well as
-;; hot-plugging HDMI outputs. Read more at
+;; hot-plugging HDMI outputs. Read more at:
 ;; https://github.com/ch11ng/exwm/wiki#randr-multi-screen
-(require 'exwm-randr)
-(add-hook 'exwm-randr-screen-change-hook
-          (lambda ()
-            (let ((xrandr-output-regexp "\n\\([^ ]+\\) connected ")
-                  default-output)
-              (with-temp-buffer
-                (call-process "xrandr" nil t nil)
-                (goto-char (point-min))
-                (re-search-forward xrandr-output-regexp nil 'noerror)
-                (setq default-output (match-string 1))
-                (forward-line)
-                (if (not (re-search-forward xrandr-output-regexp nil 'noerror))
+(use-package! exwm-randr
+  :after exwm
+  :config
+  (add-hook 'exwm-randr-screen-change-hook
+            (lambda ()
+              (let ((xrandr-output-regexp "\n\\([^ ]+\\) connected ")
+                    default-output)
+                (with-temp-buffer
+                  (call-process "xrandr" nil t nil)
+                  (goto-char (point-min))
+                  (re-search-forward xrandr-output-regexp nil 'noerror)
+                  (setq default-output (match-string 1))
+                  (forward-line)
+                  (if (not (re-search-forward xrandr-output-regexp nil 'noerror))
+                      (call-process
+                       "xrandr" nil nil nil
+                       "--output" default-output
+                       "--auto")
                     (call-process
                      "xrandr" nil nil nil
-                     "--output" default-output
-                     "--auto")
-                  (call-process
-                   "xrandr" nil nil nil
-                   "--output" (match-string 1) "--primary" "--auto"
-                   "--output" default-output "--off")
-                  (setq exwm-randr-workspace-monitor-plist
-                        (list 0 (match-string 1))))))))
-(exwm-randr-enable)
+                     "--output" (match-string 1) "--primary" "--auto"
+                     "--output" default-output "--off")
+                    (setq exwm-randr-workspace-monitor-plist
+                          (list 0 (match-string 1))))))))
+  (exwm-randr-enable))
 
-;; Configure global key bindings.
-(setq exwm-input-global-keys
-      `(([?\s-r] . exwm-reset)
-        ([?\s-w] . exwm-worksplace-switch)
-        ([?\s-&] . (lambda (command)
-                     (interactive (list (read-shell-command "$ ")))
-                     (start-process-shell-command command nil command)))
-        ,@(mapcar (lambda (i)
-                    `(,(kbd (format "s-%d" i)) .
-                      (lambda ()
-                        (interactive)
-                        (exwm-workspace-switch-create ,i))))
-                  (number-sequence 0 9))))
-
-;; Configure the default buffer behaviour. All buffers created in `exwm-mode'
-;; are named "*EXWM*". Change it in `exwm-update-class-hook' and `exwm-update-title-hook'
-;; which are run when a new X window class name or title is available.
-(add-hook 'exwm-update-class-hook
-          (lambda ()
-            (unless (or (string-prefix-p "sun-awt-X11-" exwm-instance-name)
-                        (string= "gimp" exwm-instance-name))
-              (exwm-workspace-rename-buffer exwm-class-name))))
-(add-hook 'exwm-update-title-hook
-          (lambda ()
-            (when (or (not exwm-instance-name)
-                      (string-prefix-p "sun-awt-X11-" exwm-instance-name)
-                      (string= "gimp" exwm-instance-name))
-              (exwm-workspace-rename-buffer exwm-title))))
-
-;; Here we configure the rudamentary status bar.
+;; Configure the rudamentary status bar.
 (when (featurep! +status)
   (setq display-time-default-load-average nil)
   (display-time-mode +1)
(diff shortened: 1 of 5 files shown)


modules/os/exwm/doctor.el
```diff
@@ -0,0 +1,5 @@
+;;; -*- lexical-binding: t; no-byte-compile: t; -*-
+;;; os/exwm/doctor.el
+
+(unless (executable-find "firefox")
+  (warn! "Couldn't find firefox. exwm-firefox will not work"))
```
loafofpiecrust, line 5: Maybe only run this warning if `+firefox` is enabled?

modules/os/exwm/config.el
```diff
@@ -0,0 +1,90 @@
+;;; os/exwm/config.el -*- lexical-binding: t; -*-
+
+;; Use the `ido' configuration for a few configuration fixes that alter
+;; 'C-x b' workplace switching behaviour
+(require 'exwm)
+(require 'exwm-config)
+(exwm-config-ido)
+(ido-mode +1)
```
loafofpiecrust, line 8: For `ivy`/`helm` users, is `ido-mode` needed?

modules/os/exwm/config.el
```diff
@@ -0,0 +1,90 @@
...
+                        (exwm-workspace-switch-create ,i))))
+                  (number-sequence 0 9))))
+
+;; Configure the default buffer behaviour. All buffers created in `exwm-mode'
+;; are named "*EXWM*". Change it in `exwm-update-class-hook' and `exwm-update-title-hook'
+;; which are run when a new X window class name or title is available.
+(add-hook 'exwm-update-class-hook
+          (lambda ()
+            (unless (or (string-prefix-p "sun-awt-X11-" exwm-instance-name)
+                        (string= "gimp" exwm-instance-name))
```
loafofpiecrust, line 57: These rules should probably live in a `defcustom` or `defvar` because I can foresee wanting to change them.

modules/os/exwm/config.el
```diff
@@ -0,0 +1,90 @@
+;;; os/exwm/config.el -*- lexical-binding: t; -*-
+
+;; Use the `ido' configuration for a few configuration fixes that alter
+;; 'C-x b' workplace switching behaviour
+(require 'exwm)
```
loafofpiecrust, line 5: I suspect using `use-package` would fit this repo's standards better. You should also put it all behind a trigger (like environment variable or new function) so that I can run emacs within another window manager without `exwm` getting enabled, but then make a session file that runs `EMACS_EXWM=1 emacs` or similar.

modules/os/exwm/config.el
chayward1: Great suggestion, I overlooked this completely. I think a decent solution would be to define two variables to store a list of prefixes and names that all get handled.

```el
(defvar "exwm/ignore-class-prefixes" '(("sun-awt-X11-")) "")
(defvar "exwm/ignore-class-names" '(("gimp")) "")
```
Something like that, and users could push to the list to configure it.

modules/os/exwm/doctor.el
chayward1: Fixed
```el
(when (featurep! +firefox)
  (unless (executable-find "firefox")
    (warn! "Couldn't find firefox. exwm-firefox will not work")))
```

modules/os/exwm/config.el
chayward1: The short version is yes.

I don't personally have ido enabled in my configuration but when I use SPC . to nagivate the local directory it indeed is broken.

I did however manage to have the same effect by using
```el
(exwm-config--fix/ido-buffer-window-other-frame)
```
instead of the ido configuration, which now leaves the users to set their own configurations for exwm in their own configs

modules/os/exwm/config.el
loafofpiecrust: Definitely! Though a list of regex's seems more flexible and a frequent way packages do it.

loafofpiecrust:
Input methods are important for typing other scripts and I use this snippet to enable Emacs input methods in all X windows:
```el
(use-package! exwm-xim
  :after exwm
  :config
  ;; These variables are required for X programs to pick up Emacs IM.
  (setenv "XMODIFIERS" "@im=exwm-xim")
  (setenv "GTK_IM_MODULE" "xim")
  (setenv "QT_IM_MODULE" "xim")
  (setenv "CLUTTER_IM_MODULE" "xim")
  (setenv "QT_QPA_PLATFORM" "xcb")
  (setenv "SDL_VIDEODRIVER" "x11")
  (exwm-xim-enable))
```
This should probably be behind a feature flag like `+xim` because I'm some people prefer `ibus` or `fcitx`.

modules/os/exwm/config.el
chayward1: Done

chayward1:
>   (add-hook 'exwm-mode-hook #'doom-mark-buffer-as-real-h)

Added to the normal config

> If you want to add it, this should probably be behind a feature flag like +xim because I'm sure some people prefer ibus or fcitx.

Added under the +xim flag

> Finally the following code makes sure exwm windows can be restored when switching workspaces

Added to the normal config

modules/os/exwm/config.el
chayward1: > You should also put it all behind a trigger (like environment variable or new function) so that I can run emacs within another window manager without exwm getting enabled, but then make a session file that runs EMACS_EXWM=1 emacs or similar.

According to the EXWM documentation it's always safe to have exwm-enable called even if running in a different window manager, but having tried myself under xmonad and on a TTY it appears emacs loads for a split second as vanilla emacs before it reverts back to the expected doom configuration.

I am wondering though if it would be better to use a value such as NO_EXWM=1 emacs instead, since users who are likely using this module will be using it as their WM. Just a thought!

chayward1:
> Finally the following code makes sure exwm windows can be restored when switching workspaces:
> 
> ```emacs-lisp
>   (defun exwm--update-utf8-title-advice (oldfun id &optional force)
>     "Only update the window title when the buffer is visible."
>     (when (get-buffer-window (exwm--id->buffer id))
>       (funcall oldfun id force)))
>   ;; Allow persp-mode to restore window configurations involving exwm buffers by
>   ;; only changing names of visible buffers.
>   (advice-add #'exwm--update-utf8-title :around #'exwm--update-utf8-title-advice)
> ```

Hey so this actually breaks pdf-view for me so I'm not going to include it in the module for compatability

DhashS:
Random stuff, but I’ve been trying this as well.

1. Using `SPC` as a leader key doesn’t seem to work no matter what I try, EXWM also hijacks `C-c` so I’m still not sure about how to input normal doom things in an EXWM buffer.
2. There’s the two config lines of `(setq exwm-layout-show-all-buffers t)` and `(setq exwm-workspace-show-all-buffers t)` that I find extremely useful
3. Still not sure about how to kill an EXWM buffer besides killing the async terminal, since I can’t use my leader key.
4. Exwm breaks `+childframe` on ivy, the workaround is to use `(setq ivy-posframe-parameters ‘((parent-frame nil)))` 
5. I got the standard evil “i” for insert working with pushing “i” onto `exwm-input-prefix-keys` and adding a `map!` binding to exwm-mode-map in normal mode to switch EXWM to char-mode upon hitting i. I have not yet found a way to make esc go back to line mode, I’ve been using `s-r`/exwm-reset to do this.

I’m new to emacs but I’m down to help test this!

loafofpiecrust:
> 1. Using `SPC` as a leader key doesn’t seem to work no matter what I try, EXWM also hijacks `C-c` so I’m still not sure about how to input normal doom things in an EXWM buffer.

You could bind `M-SPC` as a global exwm binding so it works in char mode.
In line mode I've had at least `C-c` work.

> 5. I got the standard evil “i” for insert working with pushing “i” onto `exwm-input-prefix-keys` and adding a `map!` binding to exwm-mode-map in normal mode to switch EXWM to char-mode upon hitting i. I have not yet found a way to make esc go back to line mode, I’ve been using `s-r`/exwm-reset to do this.

So I've found but haven't tried yet this package: https://github.com/domenzain/evil-exwm-state
I would also be really curious to look at the code for `exwm-xim` because it has to know when you're actually typing text in the X app to apply input methods correctly. Leveraging this to treat evil as a kind of input method that's aware of when you've selected a text box sounds like a boon.

> 7. Ace-window can jump to X windows, but the character highlight does not work. I turned on the hint in the modeling

Yeah this is an issue for me too because I disable the mode line in exwm. A decent solution might be to make a PR to ace-window or make a new package that renders the hints with posframe. Then we could apply the `parent-frame . nil` hack to show them on top of X windows.

djeis97:
So, I've got my own doom module for exwm I've been tinkering with for a while (not public anywhere, it's still a bit of a mess...) but I figure I might be able to contribute some ideas.

In my module I have
```
(defvar +exwm-doom-leader nil)
```
and
```
  (exwm-input-set-key (kbd (or +exwm-doom-leader doom-leader-alt-key)) doom-leader-map)
```

in the config section of the `use-package!`. Then in my config I set `+exwm-doom-leader` to `"f20"` b/c I have an f20 key in my custom keyboard layout. That lets me use `f20` as my leader key everywhere.

I also use the systray and attached minibuffer mode, so I have
```
  (when (featurep! +systray)
    (require 'exwm-systemtray)
    (exwm-systemtray-enable)
    (setq exwm-systemtray-height 11))

  (when (featurep! +attached-minibuffer)
    (add-hook! exwm-init #'exwm-workspace-attach-minibuffer))
```
in the config section as well.

Not sure if this still makes sense, but I use
```
  (setq mouse-autoselect-window t
        focus-follows-mouse t
        use-dialog-box nil)
```
I set those because I prefer follow focus when I do use the mouse (which is admittedly not a lot). It's possible that doom turns off dialog boxes itself, come to think of it- parts of my config have been copied wholesale from a vanilla config which itself was put together from the remains of a spacemacs config.

Regarding xim, I had to
```
(push ?\C-\\ exwm-input-prefix-keys)
```
to get `C-\` to be useable in exwm buffers.

Another thing I do in my actual config is use `exwm-init-hook` to launch processes like polybar and dunst when exwm launches.

Lastly, I have a totally different strategy for handling randr: I configured as many workspaces as there are outputs, and I keep that mapping in sync using the randr hook. I remember taking parts of this code from another config but I cannot remember where.

```
(defun exwm//flatenum (i ls)
  (if ls (cons i (cons (car ls) (exwm//flatenum  (1+ i) (cdr ls)))) (list)))

(defun djeis97/exwm-update-randr ()
  (let ((exwm--randr-displays (split-string
                               (shell-command-to-string
                                "xrandr | grep ' connected' | cut -d' ' -f1 "))))
    (setq exwm-workspace-number (length exwm--randr-displays))
    (setq exwm-randr-workspace-monitor-plist (exwm//flatenum 0 exwm--randr-displays))))
```

Ideally that code would also add/remove workspaces to match the number of displays but I've never gotten around to actually setting that up.

modules/os/exwm/config.el
```diff
@@ -0,0 +1,134 @@
...
+  (when (get-buffer-window (exwm--id->buffer id))
+    (funcall oldfun id force)))
+
+;; Confgure `exwm' the X window manager for emacs.
+(use-package! exwm
+  :config
+  ;; Configure global key bindings.
+  (setq exwm-input-global-keys
+        `(([?\s-r] . exwm-reset)
+          ([?\s-w] . exwm-worksplace-switch)
```
saep, line 22:
```suggestion
          ([?\s-w] . exwm-workspace-switch)
```

worldofgeese:
Thank you for this module! It's been the kick in the butt I've needed to switch to EXWM full-time.

I have two things, one a request and another an issue.

I'd love to have out of the box access to the doom leader in exwm buffers. I've written a bunch of super key binds to things I normally quick access but sometimes you have a need for something not often accessed and it always starts from the leader.

I own a Samsung C49 ultrawide and everything just works when I plug in to the monitor: buffers resize appropriately, my laptop’s screen turns off. It's glorious. The issue is when I unplug, buffers don't fully conform to the smaller display. I'm speaking specifically of the rightmost window which is "cut” off. All buffers moved to that window have their contents partially cut. I have to exit my session and login again for windows to full conform.

martinbaillie:
For the first question, I just bind M-SPC as doom-leader-map as an EXWM input key

loafofpiecrust:
My so-far favorite trick for having an all-powerful leader key accessible anywhere is to transform the super key from a modifier into a normal key:
```
xmodmap -e "keycode 133 = F19"
xmodmap -e "clear Mod4"
```
I put that in my `autostart.sh` script that runs whenever exwm starts.
Then, I bind `<f19>` to `doom-leader-map` just like @martinbaillie suggested.
This beats sticky keys for me because the other modifiers are left alone and I don't have to make any other bindings.

djeis97:
I do swapcaps and run xcape to make tapping control (that is, tapping my caps key) simulate an f20 and bind that to doom-leader-map with exwm-input-set-key.

EquusRaphanus:
I'm having issues opening files from the command line after installing this module. Is anybody else experiencing similar, or do I have an issue somewhere else in my config?

I'm trying to work around it by modifying this config.el file to conditionally `(exwm-enable)` when emacs is launched only when there is no existing window manager. So far, I haven't gotten that logic to work yet, but conditionally disabling `(exwm-enable)` has made it possible to `emacs -nw /some/file` again.

djeis97:
I sort that by not including the `(exwm-enable)` in my config at all. Instead I put `exec emacs --eval "(exwm-enable)"` in my xinitrc.

chayward1:
I am wondering if this type of module deserves more attention from a more seasoned elisp developer. It seems to me that everyone has wildly different ways of implementing exwm in their config (nothing wrong with that) and I think this module as a one-size-fits-all is a poor idea. I am open to suggestions, but I will be unable to support this module in the long term.

PR #4007, opened 2020-10-02
Adds transparency module

brorbw:
I have wanted this for quite a while of course I can just keep it in my own config but others might like it.


hlissner:
Hi! Thanks for the PR, but I cannot accept it because a) `M-x doom/set-frame-opacity` already exists and b) this configuration is too little to warrant its own module. Thanks for the PR, in any case!

brorbw:
No worries, couldn't find it in the docs

Issue #2552, opened 2020-02-18
Add a shell.nix to `:lang` modules


hlissner:
As Doom continues to grow, so have incoming bug reports for modules I do not use and toolchains I know little about, like `:lang clojure` or `:tools ein`, and so I struggle to support them.

If you know a little nix and use Doom, consider contributing a `shell.nix` file to the module(s) you use the most. Particularly `:lang` modules, so that I can get a test environment up and running to test against.

These shell.nix files should set up a minimum viable environment in which to use and test that module (not including Doom/Emacs itself). For example, this is what `lang/clojure/shell.nix` might look like:

```nix
with import <nixpkgs> {};
stdenv.mkDerivation rec {
  name = "lang-clojure";
  env = buildEnv { name = name; paths = buildInputs; };
  buildInputs = [
    clojure
    joker
    leiningen
  ];
}
```

hlissner:
This issue has been mentioned on **Doom Emacs Discourse**. There might be relevant details there:

https://doomemacs.discourse.group/t/add-a-shell-nix-to-modules-for-easier-testing/157/1

a-gilbert:
Just to promote further isolation, are you open to adding a compatible docker layer on top of the nix specs? That way contributors can be local environment agnostic while replicating your desired test setup.

PR #4011, opened 2020-10-02
Add nix-shell for Clojure

hackeryarn:
Part of  #2552.

I decided to re-use the setup that's already there in `test/shell.nix`. This just adds the required Clojure dependencies. I think this approach can greatly reduce the amount of setup for the individual nix environments.

### Unsupported Feature

`clj-kondo`, a requirement for linting, is currently broken on `nix` so it is missing here. I will look into fixing it on the nix end. Once that's done, we can just add it here.

### DIFF
--- a/modules/lang/clojure/test/clojure.nix
+++ b/modules/lang/clojure/test/clojure.nix
@@ -1,0 +1,6 @@
+let pkgs = import <nixpkgs> { };
+
+in pkgs.stdenv.mkDerivation rec {
+  name = "clojure";
+  buildInputs = with pkgs; [ jdk11 leiningen clojure-lsp ];
+}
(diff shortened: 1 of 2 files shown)


modules/lang/clojure/test/clojure.nix
```diff
@@ -0,0 +1,6 @@
+let pkgs = import <nixpkgs> { };
+
+in pkgs.stdenv.mkDerivation rec {
+  name = "clojure";
+  buildInputs = with pkgs; [ jdk11 leiningen clojure-lsp ];
```
loganlinn, line 5: while I'm not exactly clear on what we are going after, a couple other executables that are used/supported in clojure module are: `clj-kondo`, `niel`, `jet`

hlissner:
This was auto-closed because I deleted the `develop` branch, but I'm still interested in the functionality. I will revisit it after v3.

PR #4012, opened 2020-10-02
lang/julia: improve lsp support

hanmertens:
The package `lsp-julia` was not included when `julia +lsp` was selected, because it also included its own bundled `LanguageServer.jl` and dependencies of that. However, with a nonrecursive clone we can include `lsp-julia` without the bundled language server. Users can overwrite the `lsp-julia` package in `~/.doom.d/packages.el` to be able to use the bundled language server anyway.

---

I've now also added eglot support with `eglot-jl`, so now it's more of a general attempt to improve lsp support for Julia.


modules/lang/julia/README.org
```diff
@@ -26,42 +29,75 @@ Adds Julia support to Doom Emacs
...
+(setq eglot-jl-language-server-project "~/.julia/environments/v1.0")
+#+END_SRC
+
+Alternatively, you can let ~eglot-jl~ handle the installation of
+~LanguageServer.jl~ and its dependencies by putting the following in your
+personal ~config.el~.
+
+#+BEGIN_SRC elisp
+;; ~/.doom.d/config.el
+(setq eglot-jl-language-server-project (eval (car (get 'eglot-jl-language-server-project 'standard-value))))
```
benneti, line 85: Do you not set this as the default to not have problems with the default timeout?
Additionally it might be worth it to note in this case it might be necessary to increase the `eglot-connect-timeout`.
(Probably it could anyhow be noted that issues with julia and the lsp modes can be solved by increasing `eglot-connect-timeout` as it takes quiet long to start `LanguageServer.jl`.

benneti: Additionally for me 
`(eq (setq eglot-jl-language-server-project (eval (car (get 'eglot-jl-language-server-project 'standard-value)))) eglot-jl-language-server-project) => t`
so would it not be sufficient and simpler to just directly use the latter?

hanmertens: [The timeout is handled elsewhere,](https://github.com/hlissner/doom-emacs/blob/a4c5c4cee79120486d8614abeef33301bd648e09/modules/lang/julia/config.el#L84-L87) because that's indeed necessary to have a pleasant experience with `eglot-jl`. This particular snippet is there because [I chose to overwrite `eglot-jl-language-server-project` by default](https://github.com/hlissner/doom-emacs/blob/a4c5c4cee79120486d8614abeef33301bd648e09/modules/lang/julia/config.el#L82). I did this to prevent `eglot-jl` from installing `LanguageServer.jl` on its own as I believe that goes against the "your system, your rules" mantra of Doom. However, that means getting `eglot-jl` to do install `LanguageServer.jl` is a bit cumbersome.

benneti: ah thanks, overlooked this when lookibg through your changes . in my opinion this PR looks great, better than my take at it.

modules/lang/julia/config.el
```diff
@@ -66,9 +66,24 @@
...
+  :when (and (featurep! +lsp) (featurep! :tools lsp +eglot))
+  :after eglot
+  :preface
+  ;; Prevent auto-install of LanguageServer.jl
+  (setq eglot-jl-language-server-project "~/.julia/environments/v1.0")
+  ;; Prevent timeout while installing LanguageServer.jl
+  (add-hook! 'julia-mode-hook
+    (defun +julia-eglot-timeout-h ()
+      (when (< eglot-connect-timeout 60)
+        (setq eglot-connect-timeout 60))))
```
benneti, line 87: I would suggest to use 
``` elisp
(defun +julia-eglot-timeout-h ()
        "Ensure that the `eglot-connect-timeout' is longer than 60s in `julia-mode' buffers."
        (when (< eglot-connect-timeout 60)
          (setq-local eglot-connect-timeout 60)))
```
instead.
We want to ensure the timeout is large enough in julia buffers such that a buffer local seems like the best solution.

hanmertens: Thanks for the suggestion, I didn't know about `setq-local`. It's indeed nice if the larger timeout is set only for Julia buffers.

benneti: glad to help

hlissner:
Sorry for the tremendously late response. I've correct some issues and merged this. Thanks for your help!

hanmertens:
Thanks for merging! I see you made using the bundled `LanguageServer.jl` the default, but [`packages.el` still contains `:recipe (:nonrecursive t)` for `lsp-julia`](https://github.com/hlissner/doom-emacs/blob/ba0f791b05f284ad308e2ba898a54eb1fc8f0149/modules/lang/julia/packages.el#L10), which was there to exclude the git submodules containing the bundled language server and dependencies, so that should be deleted if the bundled server is to be used. It's a rather small change, so I'd thought to report it like this instead of through a new PR, but if you prefer that let me know.

samhedin:
> Thanks for merging! I see you made using the bundled `LanguageServer.jl` the default, but [`packages.el` still contains `:recipe (:nonrecursive t)` for `lsp-julia`](https://github.com/hlissner/doom-emacs/blob/ba0f791b05f284ad308e2ba898a54eb1fc8f0149/modules/lang/julia/packages.el#L10), which was there to exclude the git submodules containing the bundled language server and dependencies, so that should be deleted if the bundled server is to be used. It's a rather small change, so I'd thought to report it like this instead of through a new PR, but if you prefer that let me know.

I asked about this in the discord and it was suggested to create a new PR.

hlissner:
> Thanks for merging! I see you made using the bundled `LanguageServer.jl` the default, but [`packages.el` still contains `:recipe (:nonrecursive t)` for `lsp-julia`](https://github.com/hlissner/doom-emacs/blob/ba0f791b05f284ad308e2ba898a54eb1fc8f0149/modules/lang/julia/packages.el#L10), which was there to exclude the git submodules containing the bundled language server and dependencies, so that should be deleted if the bundled server is to be used. It's a rather small change, so I'd thought to report it like this instead of through a new PR, but if you prefer that let me know.

Done.

PR #4015, opened 2020-10-03
Update getting_started.org

haringsrob:
(not a hacktober commit)

I was following along with the documentation and found that emacs27 can be installed using the ppa repo.


hlissner:
Hi! Thanks for the PR, but I cannot accept it. I have no objection with the contents, just that I'm already working on a rewrite of the install instructions for all OSes, so that the user will install emacs 27 by default, which makes this PR redundant. In any case, thanks for bringing it to my attention!

PR #4019, opened 2020-10-04
Add keybind to close calendar details in evil mode

thevirtuoso1973:
Currently, after pressing SPC to open the details for a selected day in a calfw calendar, an evil user cannot quit the details buffer conveniently with 'q'.
This PR adds this.

### DIFF
--- a/modules/app/calendar/config.el
+++ b/modules/app/calendar/config.el
@@ -22,7 +22,11 @@
         cfw:fchar-top-left-corner ?┏
         cfw:fchar-top-right-corner ?┓)
 
+  ;; Keybindings
   (define-key cfw:calendar-mode-map "q" #'+calendar/quit)
+  (when (featurep! :editor evil +everywhere)
+    (evil-define-key 'normal cfw:details-mode-map
+      "q" #'cfw:details-kill-buffer-command))
 
   (add-hook 'cfw:calendar-mode-hook #'doom-mark-buffer-as-real-h)
   (add-hook 'cfw:calendar-mode-hook 'hide-mode-line-mode)


modules/app/calendar/config.el
```diff
@@ -22,7 +22,11 @@
         cfw:fchar-top-left-corner ?┏
         cfw:fchar-top-right-corner ?┓)
 
+  ;; Keybindings
   (define-key cfw:calendar-mode-map "q" #'+calendar/quit)
+  (when (featurep! :editor evil +everywhere)
+    (evil-define-key 'normal cfw:details-mode-map
+      "q" #'cfw:details-kill-buffer-command))
```
hlissner, line 29: Any reason this should be evil-specific. The `q` convention is shared across both vanilla and evil key schemes. Is `q` already bound to something, normally, in `cfw:details-mode-map`?

thevirtuoso1973: Yes, switching to Emacs state and pressing `q` closes the details, so it seems it's already bound to `cfw:details-kill-buffer-command` normally.

I tried replacing the evil specific binding with `define-key` and `map!` but surprisingly neither of them worked to replace q in evil mode.
For example, replacing lines 27-29 with this does *not* work for me:
```
(map! :mode cfw:details-mode-map
       "q" #'cfw:details-kill-buffer-command)
```

hlissner: Could you try replacing `:mode` with `:map` in your snippet?

thevirtuoso1973: That does not work either.

Adding `:n` does make the map work, but of course that's evil specific.

hlissner: It's fine to bind evil keys when evil isn't present. `map!` no-ops in that case.

stites:
Calendar mode also takes over the global leader key if it is `SPC` -- changing it to "change to detail view." Can this also be fixed in this PR? changes to both the enter- and exit- of details?

thevirtuoso1973:
@stites I think that's working as intended (I always use `SPC` to enter calendar details even though my leader key is `SPC`). Perhaps it could be mapped to something else like the enter key, but that would be for another PR.

iyefrat:
@thevirtuoso1973 I can't reproduce this problem. After I do `M-x cfw:open-calendar-buffer` and press `SPC` to get to a detailed view, `q` quits. If it still exists, can you give a more detailed explanation?

thevirtuoso1973:
@iyefrat Can you confirm you use EVIL? Following those steps and then pressing `q` does not quit for me, instead `q` is bound to the usual `evil-record-macro` function in that detailed view buffer.

Also, I believe `M-x =calendar` is the preferred way of opening calendar.

iyefrat:
I am on evil mode. `q` brings me back from the org-agenda view to the calendar, but `q` again doesn't quit the calendar itself. However this might be the difference, it seems that you don't get `org-agenda` after pressing space.

thevirtuoso1973:
> I am on evil mode. `q` brings me back from the org-agenda view to the calendar, but `q` again doesn't quit the calendar itself. However this might be the difference, it seems that you don't get `org-agenda` after pressing space.

It appears so, `SPC` is mapped to `cfw:show-details-command` for me.

hlissner:
Sorry for the tremendously late response, but I've figured out the cause of this issue and deployed a more comprehensive fix in b5f2589: I added a full evil-ified keybinding scheme for the module. The keys are largely the same as the vanilla keybinds, with a few evil tweaks:

+ RET replaces SPC for both calendar and details views (`SPC` will still work for folks who've changed their leader keys)
+ Adds `C-{h,j,k,l}` for traversing weeks and months in calendar view, and days/items in details view.
+ Adds `H` for goto-first-date and `L` for goto-last-date
+ `gr` refreshes instead of `r` (as per evil-collection convention)

And, of course, this fixes `q` in the details buffer (as well as fixes `q` for folks who opened the calendar using `M-x cfw:open-calendar-buffer` instead of `=calendar`).

Let me know if you encounter any other issues!

PR #4030, opened 2020-10-05
[bepo] Map ace window to <leader>éé (from ww)

gagbo:
w has the same location as é and é is not used so might as well go all the way

### DIFF
--- a/modules/input/layout/+bepo.el
+++ b/modules/input/layout/+bepo.el
@@ -27,6 +27,8 @@
   (map!
    :leader
    :desc "Window" "é" 'evil-window-map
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Thanks!

gagbo:
I forgot it was linked to window-select (just saw the tag), it'll need to be featurep! gated probably

hlissner:
Actually, it should be bound to `other-window` (no check necessary). The window-select module remaps other-window, but that doesn't have to matter to the input layout module.

PR #4032, opened 2020-10-06
Feature/bepo ace window

gagbo:
I was utterly confused by a Discord message.

`ace-window` is just a clean remap of `other-window`

`other-window` was mapped to `evil-window-map C-w` and never `evil-window-map w`. So this is why I couldn't find it under `é` in `evil-window-map`

Therefore
- The changes in #4030 are reverted
- `evil-window-map` is additionally bound to `C-é` so that `C-é C-é` gives `other-window`

I also transformed an open question I had in FIXME, although to be honest it is not urgent at all.


hlissner (approved):

hlissner:
Looks good to me. Thanks!

PR #4034, opened 2020-10-06
(org +roam): use company-capf instead of company-org-roam

jethrokuan:
company-org-roam is now deprecated, completions are provided via capf instead.

I updated the pin to the latest Org-roam (release v1.2.2), which has some fixes for completions.

### DIFF
--- a/modules/lang/org/contrib/roam.el
+++ b/modules/lang/org/contrib/roam.el
@@ -51,6 +51,9 @@
               ((featurep! :completion ido) 'ido)
               ('default)))
 
+  (when (featurep! :completion company)
+    (set-company-backend! 'org-mode '(company-capf company-yasnippet company-dabbrev)))
+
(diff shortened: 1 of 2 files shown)


jethrokuan:
In addition to this, Doom can consider:

1. setting `org-roam-db-location`: this is the path to the org-roam db. I believe Doom also moves cache files from projectile to a `var` folder

2. setting of `org-roam-completions-everywhere`: this provides as-you-type completions (outside of link brackets)

hlissner:
Hi! Thanks for the PR. I have no objection to it, but it is redundant with some work that I'm _amost_ ready to push (the lang org module was overdue a fair bit of updating, let alone org-roam). The update already addresses everything in this PR and then some. I'll be sure to add `org-roam-db-location` and `org-roam-completions-everywhere` to it too.

In any case, thank you for bringing it to my attention! I'll add a `co-authored-by` heading to the commit to credit you for the help.

PR #4039, opened 2020-10-06
update idris module to new format

iyefrat:
also add witty comment


hlissner (approved):

hlissner:
Thanks! Linking #1166 for posterity.

PR #4042, opened 2020-10-06
Add ligratures for emacs-lisp-mode

Davoodeh:
It's a bit daring to propose changes for languages you don't know... But I did it anyways. I just want my few lines of config to be prettier...

Please double check every symbol and keyword... I simply guessed most of them since I'm not familiar with Lisps at all :D

Could also add `()` as a False symbol tho I doubted maybe that's a pushing... So I didn't


hlissner:
Hi! Thanks for the PR, but I am not accepting default ligature configurations for lisps, because lisps tend to be much better structured than other languages, I feel they bring a lot less value to the table. I leave it to the user to implement if they want them. Thanks for the PR in any case!

Issue #4061, opened 2020-10-09
Incorrect use of `plist-put` in documentation


MithicSpirit:
### What did you expect to happen?
Using `plist-put` as shown in [modules/ui/ligatures/README.org](https://github.com/hlissner/doom-emacs/blob/develop/modules/ui/ligatures/README.org#changing-ligatures) to change the default ligatures (as a workaround for #4054) should work fine:
```elisp
(plist-put +ligatures-extra-symbols
  '(:map           "→"
    :return        "↦"
    :yield         "↤"
    :tuple         "⊗"))
```

### What actually happened?
There is an error that I'm passing the wrong number of arguments to `plist-put`. Looking at the documentation for `plist-put` it appears that each new property must be added individually; also, it says that it should be stored back into the `plist` because it "is modified by side effects." This is the supposed correct way to use `plist-put`: 
```elisp
(setq PLIST (plist-put PLIST PROP1 VAL1))
(setq PLIST (plist-put PLIST PROP2 VAL2))
```

### Additional details:
- [.doom.d in my dotfiles repo](https://github.com/MithicSpirit/dotfiles/tree/aceb4cb36757205e8054cbafa7ea972b3440bf0a/.doom.d)
- Error message: `error in process sentinel: Error in private config: "config.el", (wrong-number-of-arguments plist-put 2)`

#### Backtrace
<details><pre>
Debugger entered--Lisp error: (wrong-number-of-arguments plist-put 2)
  (plist-put +ligatures-extra-symbols '(:map "→" :return "↦" :yield "↤" :union "∪" :tuple "⊗"))
  #<subr eval-buffer>(#<buffer  *load*> nil "/home/mithic/.doom.d/config.el" nil t)
  apply(#<subr eval-buffer> (#<buffer  *load*> nil "/home/mithic/.doom.d/config.el" nil t))
  eval-buffer(#<buffer  *load*> nil "/home/mithic/.doom.d/config.el" nil t)  ; Reading at buffer position 4545
  load-with-code-conversion("/home/mithic/.doom.d/config.el" "/home/mithic/.doom.d/config.el" t t)
  #<subr load>("/home/mithic/.doom.d/config" t nomessage nil nil)
  apply(#<subr load> ("/home/mithic/.doom.d/config" t nomessage))
  load("/home/mithic/.doom.d/config" t nomessage)
  doom-initialize-modules(force)
  doom-initialize(force)
[...]

PR #4063, opened 2020-10-09
Fix ligature documentation typo

gilbertw1:
Fixes documentation type found in #4061

### DIFF
--- a/modules/ui/ligatures/README.org
+++ b/modules/ui/ligatures/README.org
@@ -172,45 +172,44 @@ if you don't like the symbols chosen you can change them by using...
 
 #+BEGIN_SRC emacs-lisp
 ;; you don't need to include all of them you can pick and mix
-(plist-put +ligatures-extra-symbols
-  '(;; org
-    :name          "»"
-    :src_block     "»"
-    :src_block_end "«"
-    :quote         "“"
-    :quote_end     "”"
-    ;; Functional
-    :lambda        "λ"
-    :def           "ƒ"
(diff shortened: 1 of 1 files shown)


hlissner (changes requested):
Thanks for catching this. This is half the fix however. `plist-put!` is variadic so this should instead be:

```lisp
(plist-put! +ligatures-extra-symbols
  ;; org
  :name          "»"
  :src_block     "»"
  :src_block_end "«"
  :quote         "“"
  :quote_end     "”"
  ;; Functional
  :lambda        "λ"
  :def           "ƒ"
  :composition   "∘"
  :map           "↦"
  ;; Types
  :null          "∅"
  :true          "𝕋"
  :false         "𝔽"
  :int           "ℤ"
  :float         "ℝ"
  :str           "𝕊"
  :bool          "𝔹"
  :list          "𝕃"
  ;; Flow
  :not           "￢"
  :in            "∈"
  :not-in        "∉"
  :and           "∧"
  :or            "∨"
  :for           "∀"
  :some          "∃"
  :return        "⟼"
  :yield         "⟻"
  ;; Other
  :union         "⋃"
  :intersect     "∩"
  :diff          "∖"
  :tuple         "⨂"
  :pipe          "" ;; FIXME: find a non-private char
  :dot           "•")
```

gilbertw1:
Ah changed, thanks!

hlissner (approved):

hlissner:
Thanks!

PR #4068, opened 2020-10-10
Bumping ocamlformat package to latest

nate:
Bumping to the latest version because the current pinned version of ocamlformat in doom is broken on macos, formatting does some weird things. This PR fixes the issue https://github.com/ocaml-ppx/ocamlformat/pull/1481 and we may as well go to the latest anyways.


hlissner:
Hi! Thanks for the PR, but I already had a full bump of the module in the pipeline. Also, bump commits must follow a strict message format for machine readability by future CI. In any case, thanks for bringing this to my attention!

PR #4072, opened 2020-10-10
Bump :lang latex

acristoffers:
jwiegley/auctex@59e251c -> jwiegley/auctex@9e7406c
iyefrat/evil-tex@06234c9 -> iyefrat/evil-tex@a3b6875

iyefrat/evil-tex#3 fixes selection/navigation in LaTeX files which are not the root document and makes section regex-matching more strict, to avoid partial matching.

### DIFF
--- a/modules/lang/latex/packages.el
+++ b/modules/lang/latex/packages.el
@@ -1,11 +1,11 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; lang/latex/packages.el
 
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks!

hlissner:
Just realized that this was failing to update because straight.el installs auctex from emacs-straight/auctex, not jwiegley/auctex, and 9e7406c doesn't exist in the mirror. I've reverted auctex's bump in 7138cab.

PR #4075, opened 2020-10-11
[:lang cc] Fix list in call for cpu count command

gagbo:
I should find some way to test on Mac OS I guess, or test in CI or something at least. That would make those small errors easier to catch than trying to review it 1000 times


hlissner (approved):

hlissner:
Thanks!

PR #4076, opened 2020-10-11
Add table describing keybindings for clojure module

lunik1:
Adds a table describing the key bindings under `localleader` in clojure-mode.

I had to use zero width spaces to get the table to render correctly for bindings ending in single or double quoutes. [The org-mode mailing list explains the problem and workaround.](https://lists.gnu.org/archive/html/emacs-orgmode/2010-04/msg00330.html).


hlissner (approved):

hlissner:
Thanks!

PR #4090, opened 2020-10-13
[:lang python] Update README

brianmcgillion:
Notify the users that updating the doom env is necessary
after choosing to use a virtual environment such as pyenv

### DIFF
--- a/modules/lang/python/README.org
+++ b/modules/lang/python/README.org
@@ -120,11 +120,16 @@ To enable support for auto-formatting with black enable ~:editor format~ in
 | ~<localleader> g u~ | ~anaconda-mode-find-references~  |
 
 * Configuration
-This module has the following variables to set extra arguments to [[https://ipython.org/][ipython]] and
++ This module has the following variables to set extra arguments to [[https://ipython.org/][ipython]] and
 [[https://jupyter.org/][jupyter]] shells:
 
 #+BEGIN_SRC elisp
 ;; ~/.doom.d/config.el
 (setq +python-ipython-repl-args '("-i" "--simple-prompt" "--no-color-info"))
 (setq +python-jupyter-repl-args '("--simple-prompt"))
+#+END_SRC
+
++ After installing the virtual environments be sure to regenerate your doom env
+#+BEGIN_SRC sh
+~/.emacs.d/bin/doom env
 #+END_SRC


modules/lang/python/README.org
```diff
@@ -120,11 +120,16 @@ To enable support for auto-formatting with black enable ~:editor format~ in
...
 #+BEGIN_SRC elisp
 ;; ~/.doom.d/config.el
 (setq +python-ipython-repl-args '("-i" "--simple-prompt" "--no-color-info"))
 (setq +python-jupyter-repl-args '("--simple-prompt"))
 #+END_SRC
+
++ After installing the virtual environments be sure to regenerate your doom env
+#+BEGIN_SRC sh
+~/.emacs.d/bin/doom env
+#+END_SRC
```
gagbo, line 135: This is misleading.

Running `doom env` in a terminal will indeed load the current PATH for further doom invocations, but that will only work for the currently loaded environment.

If a project uses a different virtualenv, then the PATH will actually be different and a lot of things will work weirdly and trigger bugs.

I don't use venvs this way so I can't tell how to handle this (the beauty of Python I suppose). Since I use conda for python envs, I have 2 solutions to get the correct path in Doom :

- I set up `(eval . (conda-env-activate "env_name"))` in my `.dir-locals.el`, or
- I use direnv and `:tools direnv`

hlissner:
Hi! Thanks for the PR, but I cannot accept it. `doom env` is not an acceptable solution for "injecting" virtualenv environments into Doom, because virtualenvs are per-project, and `doom env` seeds Emacs' global default environment at startup. 

As gagbo explained, it is better to use `conda-env-activate` in `.dir-locals.el`, perhaps on python-mode-hook (slow), or (my recommendation) by using direnv.

In any case, thanks for the PR.

brianmcgillion:
thanks for explaining! I only tested with pyenv as I installed to the "`suggested path`" from their install guide, which was not in PATH.  it was picked up and *seems* to be working fine. Also, I noticed when running `doom upgrade` or `doom sync` that it is calling `doom env` would this not mean that the user will end up with whichever virtenv is in their env - everytime they run upgrade (or was the `doom env` triggered for a different reason? i.e. it is not triggered on all upgrade/sync/build cycles?

Thanks again for the detailed explanation.

PR #4091, opened 2020-10-13
added remove ~/.emacs in README

MrCow3:
When I first downloaded doom I was not aware that you need to delete .emacs so I added it to the read me in case anyone else is confused.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,7 @@ doctor` to check for any that you may have missed.
 ``` sh
 git clone --depth 1 https://github.com/hlissner/doom-emacs ~/.emacs.d
 ~/.emacs.d/bin/doom install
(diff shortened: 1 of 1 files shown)


MrCow3:
Just so you know this is my first contribution on git

hlissner:
Hi! Thanks for the PR, but sorry to say I cannot accept it because:

1. There is work going on in `doom install` to warn users about `~/.emacs`, 
2. `doom doctor` already does.
3. It's mentioned [in the FAQ](https://github.com/hlissner/doom-emacs/blob/develop/docs/faq.org#i-get-the-vanilla-emacs-splash-screen-at-startup).
4. Not all installations of emacs creates the `~/.emacs` file. Some create a `~/.emacs.d`. Some create neither. This isn't general enough to be included in our "quick start" instructions.

PR #4094, opened 2020-10-14
Add scala module documentation

gilbertw1:
This pull request adds basic documentation for the `scala-mode` and specifically using it with LSP. This should contain enough information to get users up and running with LSP properly along with pointers to the metals emacs documentation.

I modeled this readme after the one for `java-mode`. Let me know if I should make any changes to it.

### DIFF
--- a/modules/lang/scala/README.org
+++ b/modules/lang/scala/README.org
@@ -1,0 +1,94 @@
+#+TITLE:   lang/scala
+#+DATE:    October 14, 2020
+#+SINCE:   v1.3
+#+STARTUP: inlineimages
+
+* Table of Contents :TOC_3:noexport:
+- [[#description][Description]]
+  - [[#module-flags][Module Flags]]
+- [[#lsp-integration][=+lsp= Integration]]
+  - [[#prerequisites][Prerequisites]]
+    - [[#coursier][Coursier]]
+    - [[#metals][Metals]]
+  - [[#features][Features]]
+  - [[#notes][Notes]]
+
+* Description
+This module adds [[https://www.scala-lang.org][scala]] and [[https://www.scala-sbt.org/][sbt]] support to Doom Emacs.
+
+** Module Flags
++ =+lsp= Enables integration for the metals LSP server.
+
+
+* =+lsp= Integration
+
+** Prerequisites
+
+*** Coursier
+Note: Coursier is only required to install Metals. If system is running arch
+linux this step can be safely skipped.
+
+**** Ubuntu / Debian / Fedora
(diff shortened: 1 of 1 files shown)


modules/lang/scala/README.org
```diff
@@ -0,0 +1,123 @@
...
+
+*** Arch
+#+BEGIN_SRC sh
+yay -S coursier
+#+END_SRC
+
+** Metals (lsp only)
+This module only requires [[https://scalameta.org/metals][metals]] if utilizing LSP.
+
+Note: Update following command to latest version of metals [[https://scalameta.org/metals/docs/editors/emacs.html][found here]].
```
ashfaqfarooqui, line 87: Maybe it's worth mentioning here that Metals can be installed from Arch aur if one does not want to manually handle the update process?  `yay -S metals`

modules/lang/scala/README.org
```diff
@@ -0,0 +1,123 @@
...
+
+* Description
+This module adds [[https://www.scala-lang.org][scala]] support to Doom Emacs.
+
+** Module Flags
++ =+lsp= Enables integration for the metals LSP server.
+
+* Prerequisites
+
+This module requires the [[https://www.scala-sbt.org/][sbt]] build tool.
```
ashfaqfarooqui, line 32: Is this always true? Scala works without SBT too.  You can always use other build tools. 

But does Doom require SBT as a prerequisite? I often use Mill as my build tool and have faced no issues. 

If you agree with me, I suggest putting this under a "build tool" header; the prerequisite would be to have scala installed. That can be linked to the official install scala page or the [one click installer](https://www.scala-lang.org/2020/06/29/one-click-install.html) the seem to recommend these days.

gilbertw1:
Hey, thanks for the feedback!

You're absolutely right about SBT. While `sbt-mode` is included and sbt support is available by default it is by no means a requirement to use the mode. I'll update the documentation to reflect that SBT support is available, but that SBT itself is not required for the module to function properly.

I'll also update the documentation to mention that metals can be installed via AUR.

gilbertw1:
@ashfaqfarooqui What do you think of the lastest version?

I removed the explicit section about installing `sbt` and just noted that the module added support for `sbt` in the description. Without lsp enabled the module requires no prerequisites to function, so I simply placed the rest of the documentation under an `+lsp Integration` header because it's the only part that takes some work to use successfully.

hlissner (approved):

hlissner:
Thanks! Tagging #1166 for posterity.

PR #4095, opened 2020-10-14
popup: add rule for list-processes

snizovtsev:
Its content is typically thrown away after a quick glance, so make it popup.


hlissner (approved):

hlissner:
Thanks!

PR #4097, opened 2020-10-15
Bump coq

brettgilio:
(package! proof-general :pin "cd6955d5a7fa35ab1aa6a7754d194a87f7c9aedf")

### DIFF
--- a/modules/lang/coq/packages.el
+++ b/modules/lang/coq/packages.el
@@ -1,5 +1,5 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; lang/coq/packages.el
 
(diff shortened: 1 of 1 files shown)


hlissner (changes requested):
Hi, thanks for the PR, but I bump packages each weekend. When PRing a bump request please include an explanation why this package should be bumped ahead of schedule.

Second, bump commits must follow a strict format. Please reword the commit to

```lisp
Bump :lang coq

ProofGeneral/PG@4f6b602 -> ProofGeneral/PG@cd6955d

<EXPLANATION FOR WHY THIS IS HAPPENING EARLY>
```

Issue #3891, opened 2020-09-04
lsp, Flycheck checker and Autocomplete don't work if file is passed as an argument to Emacs


shans10:
**What did you expect to happen?**
I have enabled lsp for python and C/C++ and it works flawlessly if I manually open associated files in emacs but whenever I open it using **emacs "filename"** or **emacs -nw "filename"** from terminal lsp and flycheck should work. 

**What actually happened?**
When filename is given as argument in terminal Flycheck, lsp and autocomplete do not work and I have to enable them manually!

**Steps to reproduce:**
1. Open a terminal and type emacs "filename" or emacs -nw "filename".
2. Flycheck, lsp and autocomplete are disabled, need to enble them manually.

**System information:**
<details><pre>
((system
  (type . gnu/linux)
  (config . "x86_64-pc-linux-gnu")
  (shell . "/usr/bin/fish")
  (uname . "Linux 5.8.5-arch1-1 #1 SMP PREEMPT Thu, 27 Aug 2020 18:53:02 +0000 x86_64")
  (path "~/.config/xmobar/scripts/" ".bin/" ".cargo/bin/" ".emacs.d/bin/" "/usr/local/sbin" "/usr/local/bin" "/usr/bin" "/usr/bin/site_perl" "/usr/bin/vendor_perl" "/usr/bin/core_perl" "/usr/lib/emacs/27.1/x86_64-pc-linux-gnu"))
 (emacs
  (dir . "~/.emacs.d/")
  (version . "27.1")
  (build . "Aug 28, 2020")
  (buildopts . "--prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib --localstatedir=/var --with-x-toolkit=gtk3 --with-xft --with-wide-int --with-modules --with-cairo --with-harfbuzz 'CFLAGS=-march=x86-64 -mtune=generic -O2 -pipe -fno-plt' CPPFLAGS=-D_FORTIFY_SOURCE=2 LDFLAGS=-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now")
[...]

PR #4104, opened 2020-10-17
Fix lsp not triggering when opening a file from cli

innerout:
This PR closes issue #3891 
Just a note I am a beginner, so there might be better ways to solve it.


hlissner:
Hi, thanks for the PR but I cannot accept it. The `*-local-vars-hook` family of hooks were created specifically for launching servers (and LSP). Moving them off of them is not an acceptable solution and will introduce a multitude of other issues. Another solution is needed.

Thanks for the attempt, in any case.

PR #4105, opened 2020-10-17
Print diff url in terminal in case emacs is in a remote server

### LINKED ISSUE #2936: Print Github compare URL during doom upgrade
It would be great if, during `doom upgrade`, besides printing the old and new revisions and opening the browser, the URL is [printed](https://github.com/hlissner/doom-emacs/blob/5d51d4e4a662410679a5678fcc3f5251a8f33a66/core/cli/upgrade.el#L98) as well. My reason is that I primarily use Doom through SSH. Going through the history of upgrade.el, I see that the URL was printed in the past, and I'd like to consider bringing it back for the SSH use case. Perhaps we can detect SSH with `(getenv "SSH_CONNECTION")` and only print it then?

innerout:
Closes #2936


hlissner (approved):

hlissner:
Thanks!

PR #4113, opened 2020-10-19
A comment for users and tide package pin

Sleepful:
Pinning the new tide commit due to: https://github.com/ananthakumaran/tide/issues/401#event-3888907902
And adding the `doom sync -u` command as a comment on the packages.el settings file because it went over my head when I was trying to change the recipe for the tide package. The comment about using `doom sync -u` is in the docs tho: https://github.com/hlissner/doom-emacs/blob/develop/docs/getting_started.org#changing-a-recipe-for-a-included-package

### DIFF
--- a/core/templates/packages.example.el
+++ b/core/templates/packages.example.el
@@ -32,6 +32,11 @@
 ;(package! builtin-package :recipe (:nonrecursive t))
 ;(package! builtin-package-2 :recipe (:repo "myfork/package"))
 
+;; Remember to run 'doom sync -u' after changing recipes for existing packages.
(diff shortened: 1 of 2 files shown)


hlissner:
Hi! Thanks for the PR, unfortunately I must refuse it because a) I already had a bump commit in the pipeline which makes it redundant and b) I am working on a revamp of our documentation that would delete your contribution to our packages.el template anyway.

Two things for the future:

1. These two commits serve separate purposes, please PR distinct concerns separately.the future.
2. Bump commits must follow a strict format so it is machine readable for future CI. See b726adac8 as an example.

PR #4115, opened 2020-10-19
Bump lsp-metals

gregghz:


hlissner (changes requested):
Hi, when PRing a bump request please include an explanation why this package should be bumped ahead of schedule. Unless the reason is compelling, I'd rather not bump them early.

Second, bump commits must follow a strict format. Please reword the commit to

```lisp
Bump :lang scala

emacs-lsp/lsp-metals@8f8471c -> emacs-lsp/lsp-metals@e42c0b2

<EXPLANATION FOR WHY THIS IS HAPPENING EARLY>
```

`Bump :lang scala` can be `Bump lsp-metals` _only_ if other packages in the module have pending updates, but you've selectively chosen only to update that package. That is not the case here.

gregghz:
Thanks for the feedback.

> include an explanation why this package should be bumped ahead of schedule

What's the normal schedule? I was unaware that packages were automatically bumped on a schedule.

If that's the case, then this can wait for that. It adds a minor new feature and I can just bump the version in my local config instead of here.

hlissner:
> What's the normal schedule?

Packages are bumped every weekend, so long as they pass some rudimentary sanity checks.

In this case, lsp-metals didn't make last weekend because there were upstream breakages in lsp-mode that bubbled down into its clients which I'm still working on. I'll bump lsp-metals (and other other lsp client packages) once it is done.

> If that's the case, then this can wait for that.

Understood. Thanks for bringing it to my attention, in any case!

PR #4120, opened 2020-10-20
Format with LSP if supported on save

### LINKED ISSUE #3626: how to format buffer on save using lsp formatter
I want to make emacs format the buffer using lsp formatter on save.
I tried to enable (format +onsave) in init.el and set `+format-with-lsp` to t, but only `+format/buffer` can invoke lsp format and format-on-save still use some third party formatters(eg. prettier for js).
I also tried `(add-hook 'js2-mode-hook #'+format|enable-on-save)` but doesn't work.
How can I correctly set lsp format on save?

systemctl603:
Fixes #3626

To fix this, 
```lisp
(format-all-mode +1)
```
Was replaced with:
```elisp
(+format-on-save-h)
```
This ensures that `+format/buffer` is used instead of `format-all-buffer`
In the function `+format-on-save-maybe-h` located in `modules/format/config.el`

### DIFF
--- a/modules/editor/format/config.el
+++ b/modules/editor/format/config.el
@@ -44,7 +44,7 @@ This is controlled by `+format-on-save-enabled-modes'."
                      (memq major-mode (cdr +format-on-save-enabled-modes)))
                     ((not (memq major-mode +format-on-save-enabled-modes))))
               (not (require 'format-all nil t)))
-    (format-all-mode +1)))
+    (add-hook 'before-save-hook #'+format/buffer nil t)))
 
 (when (featurep! +onsave)
   (add-hook 'after-change-major-mode-hook #'+format-enable-on-save-maybe-h))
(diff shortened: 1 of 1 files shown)


systemctl603:
Sorry, accidentally pressed draft button.

hlissner (approved):

hlissner:
Thanks!

ehamberg:
Is it expected that this overrides a formatter that was set explicitly with `set-formatter!`? If so, how can a specific formatter be used for a given mode, even when the corresponding LSP supports formatting (with a different formatter)?

systemctl603:
Yes, `set-formatter!` defines a CLI tool, which is overridden by the LSP formatter.
To use a specific formatter, see [this](https://github.com/hlissner/doom-emacs/tree/develop/modules/editor/format#disabling-the-lsp-formatter)

ehamberg:
> To use a specific formatter, see [this](https://github.com/hlissner/doom-emacs/tree/develop/modules/editor/format#disabling-the-lsp-formatter)

Thanks! That worked. 👍

PR #4123, opened 2020-10-21
Bump iedit

gilbertw1:
Bumping `iedit` version to fix `Symbol's function definition is void: iedit-lib-cleanup` error preventing exit of `iedit` mode. `iedit` was updated to have it's cleanup function renamed and `evil-iedit-state` was updated to account for this change. However, Doom's version of `iedit` did not contain the newer function, thus introducing this error.

Details of the issue can be found here: https://github.com/syl20bnr/evil-iedit-state/issues/27

### DIFF
--- a/modules/editor/multiple-cursors/packages.el
+++ b/modules/editor/multiple-cursors/packages.el
@@ -3,9 +3,7 @@
 
 (cond
  ((featurep! :editor evil)
-  ;; REVIEW Broken in 8abf2c1f4f0ade64cbb06c8f47055f04ab83e8d6 (latest commit at
-  ;;        time of writing). Revisit later.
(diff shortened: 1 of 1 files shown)


hlissner:
Hi! Thanks for the PR but I cannot accept it yet. What's holding me back from bumping iedit is it introduces [a breaking change for evil-multiedit](https://github.com/hlissner/evil-multiedit/issues/39) which I haven't pushed my fix for yet. I will bump this package once it has.

In any case, the current commit of evil-multiedit doesn't call iedit-lib-cleanup or iedit-cleanup, so this isn't an issue for Doom at the moment.

Thanks for bringing it to my attention.

PR #4136, opened 2020-10-22
Enable mouse scrolling in tty module. 

smsegal:
When the `:os tty` module is enabled, the mouse should also scroll the window Emacs. This PR introduces that behaviour. 

Only thing I'm not sure of is if `<mouse-4>` and `<mouse-5>` are scroll-up and scroll-down for every terminal emulator / mouse. 
I only have the one mouse, but I've tested it in kitty terminal and gnome terminal and it works the same in both for me.


hlissner:
Mouse wheel scrolling already works. Have you disabled evil?

smsegal:
Oh, hm. No, I haven't disabled evil. I suppose I should transition this to a bug report instead :)

smsegal:
Bug report opened at #4137

PR #4139, opened 2020-10-23
Add exunit mode for Elixir

jsmestad:
I modeled this off the definition for `rspec-mode` in the `ruby` module. Let me know if there is anything I missed or need to adjust.

🍻

### DIFF
--- a/modules/lang/elixir/config.el
+++ b/modules/lang/elixir/config.el
@@ -15,17 +15,17 @@
   (provide 'smartparens-elixir)
   :config
   (set-ligatures! 'elixir-mode
-                  ;; Functional
-                  :def "def"
-                  :lambda "fn"
-                  ;; :src_block "do"
-                  ;; :src_block_end "end"
-                  ;; Flow
-                  :not "!"
-                  :in "in" :not-in "not in"
-                  :and "and" :or "or"
-                  :for "for"
-                  :return "return" :yield "use")
+    ;; Functional
+    :def "def"
+    :lambda "fn"
+    ;; :src_block "do"
+    ;; :src_block_end "end"
+    ;; Flow
+    :not "!"
+    :in "in" :not-in "not in"
+    :and "and" :or "or"
+    :for "for"
+    :return "return" :yield "use")
 
   ;; ...and only complete the basics
   (sp-with-modes 'elixir-mode
@@ -74,15 +74,14 @@
 
 (use-package! exunit
   :hook (elixir-mode . exunit-mode)
-  :defer t
-  :config
+  :init
   (map! :after elixir-mode
-        :localleader
-        :map elixir-mode-map
-        :prefix "t"
-        "a" #'exunit-verify-all
-        "r" #'exunit-rerun
-        "v" #'exunit-verify
-        "T" #'exunit-toggle-file-and-test
-        "t" #'exunit-toggle-file-and-test-other-window
-        "s" #'exunit-verify-single))
+    :localleader
+    :map elixir-mode-map
+    :prefix "t"
+    "a" #'exunit-verify-all
+    "r" #'exunit-rerun
+    "v" #'exunit-verify
+    "T" #'exunit-toggle-file-and-test
+    "t" #'exunit-toggle-file-and-test-other-window
+    "s" #'exunit-verify-single))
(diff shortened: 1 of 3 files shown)


modules/lang/elixir/config.el
```diff
@@ -15,17 +15,17 @@
...
+                  :def "def"
+                  :lambda "fn"
+                  ;; :src_block "do"
+                  ;; :src_block_end "end"
+                  ;; Flow
+                  :not "!"
+                  :in "in" :not-in "not in"
+                  :and "and" :or "or"
+                  :for "for"
+                  :return "return" :yield "use")
```
hlissner, line 28: Please restore the original (and correct) indentation

modules/lang/elixir/config.el
```diff
@@ -71,3 +71,18 @@
   (let ((fn (byte-compile (lambda () (add-to-list (make-local-variable 'company-backends) 'alchemist-company)))))
     (remove-hook 'alchemist-mode-hook fn)
     (remove-hook 'alchemist-iex-mode-hook fn)))
+
+(use-package! exunit
+  :hook (elixir-mode . exunit-mode)
+  :defer t
```
hlissner, line 77: `:defer t` is redundant with a `:hook` property. Please remove it.

modules/lang/elixir/config.el
```diff
@@ -71,3 +71,18 @@
...
+  :config
```
hlissner, line 78: Replace `:config` with `:init`. Otherwise the `map!`'s deferral happens too late (making it impossible for users to override those keybinds, because `(map! :after elixir ...)` won't run late enough.

hlissner (changes requested):

modules/lang/elixir/config.el
jsmestad: Whooops not sure why that happend. 🙃

jsmestad: @hlissner I had to manually correct the indentation. What would doom be doing that would cause what you are calling the incorrect indentation? I had to switch to fundamental mode (out of elisp mode) to get the format on save to stop messing with this.

hlissner: It's possible that you are running a formatter that reformats the whole file? Emacs won't know how to indent macros that haven't been loaded yet (like `set-ligatures!`), so it'll guess.

I would highly recommend not using a formatter with elisp. [It is normally inhibited in emacs-lisp-mode for this reason](https://github.com/hlissner/doom-emacs/blob/develop/modules/editor/format/config.el#L4).

modules/lang/elixir/config.el
```diff
@@ -71,3 +71,17 @@
...
+  (map! :after elixir-mode
+    :localleader
+    :map elixir-mode-map
+    :prefix "t"
+    "a" #'exunit-verify-all
+    "r" #'exunit-rerun
+    "v" #'exunit-verify
+    "T" #'exunit-toggle-file-and-test
+    "t" #'exunit-toggle-file-and-test-other-window
+    "s" #'exunit-verify-single))
```
hlissner, line 87: Please correct the indentation here. Should be

```lisp
(use-package! exunit
  :hook (elixir-mode . exunit-mode)
  :init
  (map! :after elixir-mode
        :localleader
        :map elixir-mode-map
        :prefix "t"
        "a" #'exunit-verify-all
        "r" #'exunit-rerun
        "v" #'exunit-verify
        "T" #'exunit-toggle-file-and-test
        "t" #'exunit-toggle-file-and-test-other-window
        "s" #'exunit-verify-single))
```

hlissner (changes requested):

modules/lang/elixir/config.el
jsmestad, line 87:
```suggestion
(use-package! exunit
  :hook (elixir-mode . exunit-mode)
  :init
  (map! :after elixir-mode
             :localleader
             :map elixir-mode-map
             :prefix "t"
             "a" #'exunit-verify-all
             "r" #'exunit-rerun
             "v" #'exunit-verify
             "T" #'exunit-toggle-file-and-test
             "t" #'exunit-toggle-file-and-test-other-window
             "s" #'exunit-verify-single))
```

jsmestad:
@hlissner alright I should have the indent sorted

hlissner (approved):

hlissner:
Thanks!

PR #4141, opened 2020-10-23
julia with lsp +eglot config

benneti:
Update the README and config to consider the lsp +eglot option.
Additionally now the packages eglot-jl or lsp-julia are packaged
depending on the featurerep of lsp.
They are not pinned as they can be seen as more of a extended
configuration of the julia and emacs site 
(if you think even for these packages pinning would be preferred let me know).

Finally as the LanguageServer.jl needs comparably long to start even after the first run it might be nice to add a note to this in the README as well, where would you put this or do you think it would be worth it to change the defaults in julia buffers directly from the module config.el?

### DIFF
--- a/modules/lang/julia/README.org
+++ b/modules/lang/julia/README.org
@@ -8,9 +8,8 @@
   - [[#module-flags][Module Flags]]
   - [[#plugins][Plugins]]
 - [[#prerequisites][Prerequisites]]
-  - [[#language-server][Language Server]]
 - [[#features][Features]]
-  - [[#language-server-1][Language Server]]
+  - [[#language-server][Language Server]]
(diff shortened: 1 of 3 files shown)


benneti:
Sorry, this is a duplicate effort of #https://github.com/hlissner/doom-emacs/pull/4012, as I overlooked it before creating the PR.

I however did it such that by default eglot uses its onw LanguageServer.jl installation.

gagbo:
Hello,

You should always pin all the things. It's when exceptions start to be added that everything falls apart :(

benneti:
Closing in favor of #https://github.com/hlissner/doom-emacs/pull/4012

PR #4143, opened 2020-10-23
Add emacs bindings and documentation in fold module 

innerout:
This PR adds emacs bindings for vimish-fold when evil is disabled. It also adds documentation for the bindings. Finally, it reports a weird bug and explains a way to bypass it.

### DIFF
--- a/modules/editor/fold/config.el
+++ b/modules/editor/fold/config.el
@@ -85,7 +85,3 @@
     "zE" #'vimish-fold-delete-all)
   :config
   (vimish-fold-global-mode +1))
-
-(use-package! vimish-fold
-  :unless (featurep! :editor evil)
-  :defer t)
--- a/modules/editor/fold/config.el
+++ b/modules/editor/fold/config.el
@@ -85,3 +85,14 @@
     "zE" #'vimish-fold-delete-all)
   :config
   (vimish-fold-global-mode +1))
+
+(when (not (featurep! :editor evil))
+  (use-package! vimish-fold
+    :init
+    (define-key! 'global
+      "C-c f v"   #'vimish-fold
+      "C-c f a v" #'vimish-fold-refold-all
+      "C-c u v"   #'vimish-fold-unfold
+      "C-c u a v" #'vimish-fold-unfold-all
+      "C-c d v"   #'vimish-fold-delete
+      "C-c d a v" #'vimish-fold-delete-all)))
--- a/modules/editor/fold/config.el
+++ b/modules/editor/fold/config.el
@@ -86,13 +86,6 @@
   :config
   (vimish-fold-global-mode +1))
 
-(when (not (featurep! :editor evil))
-  (use-package! vimish-fold
-    :init
-    (define-key! 'global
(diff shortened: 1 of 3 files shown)


modules/editor/fold/config.el
```diff
@@ -85,3 +85,14 @@
     "zE" #'vimish-fold-delete-all)
   :config
   (vimish-fold-global-mode +1))
+
+(when (not (featurep! :editor evil))
+  (use-package! vimish-fold
+    :init
```
hlissner, line 91: Please use `(unless (featurep! ...)` instead of `(when (not (featurep! ...))`

hlissner: Please lazy load the package by adding `:defer t` before `:init`.

hlissner: Please use `use-package`'s `:unless` property instead of wrapping a `use-package!` block in a `when`/`unless`. e.g.

```diff
(use-package! vimish-fold
  :unless (featurep! :editor evil)
  :defer t
  :init
```

modules/editor/fold/config.el
```diff
@@ -85,3 +85,14 @@
...
+    (define-key! 'global
+      "C-c f v"   #'vimish-fold
+      "C-c f a v" #'vimish-fold-refold-all
+      "C-c u v"   #'vimish-fold-unfold
+      "C-c u a v" #'vimish-fold-unfold-all
+      "C-c d v"   #'vimish-fold-delete
+      "C-c d a v" #'vimish-fold-delete-all)))
```
hlissner, line 98: 1. Please move these keybinds to `modules/config/default/+emacs-bindings.el`.
2. Use `map!` instead of `define-key!`.
3. Use the `+fold/*` commands where appropriate.
4. Consolidate these keys under fewer prefixes. Perhaps `C-c C-f`, since `C-c f` is used, and `C-c {f,u,d}` is a wide spread. e.g.

```elisp
(map! :leader
      (:when (featurep! :editor fold)
        (:prefix ("C-f" . "fold")
         "C-d"     #'vimish-fold-delete
         "C-a C-d" #'vimish-fold-delete-all
         "C-f"     #'+fold/toggle
         "C-a C-f" #'+fold/close-all
         "C-u"     #'+fold/open
         "C-a C-u" #'+fold/open-all)))
```

hlissner (changes requested):

innerout:
By mistake, I deleted the branch.

modules/editor/fold/config.el
```diff
@@ -85,3 +85,7 @@
     "zE" #'vimish-fold-delete-all)
   :config
   (vimish-fold-global-mode +1))
+
+(use-package! vimish-fold
+  :unless (featurep! :editor evil)
+  :defer t)
```
hlissner, line 91: Please remove this entirely. It does nothing.

hlissner (changes requested):

hlissner (approved):

hlissner:
Thanks!

PR #4145, opened 2020-10-24
Clarify doom-store-put docstring

gagbo:
Explicitely states that a `nil` TTL means "no expiration"

Make it clearer that TTL is an offset to current-time, and not a timestamp in seconds


hlissner (approved):

hlissner:
Thanks!

Issue #3826, opened 2020-08-25
Activate Poetry virtualenv automatically


sei40kr:
Poetry creates a virtualenv for each project, and it can be activated by `poetry-venv-workon` (like `pyvenv-venv-workon`).
I want to activate it automatically like the other virtual environment managers do.

lqmanh:
Just toggle `poetry-tracking-mode`

PR #4146, opened 2020-10-24
Enable poetry-tracking-mode

sei40kr:
# Description

Implements #3826.

Enable `poetry-tracking-mode` in the first `python-mode` buffer (if `python +poetry` enabled).

### DIFF
--- a/modules/lang/python/config.el
+++ b/modules/lang/python/config.el
@@ -289,7 +289,7 @@
   :when (featurep! +poetry)
   :after python
   :init
-  (add-hook! 'python-mode-hook (poetry-tracking-mode +1)))
+  (add-hook 'python-mode-hook #'poetry-tracking-mode))
 
 
 (use-package! cython-mode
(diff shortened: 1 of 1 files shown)


modules/lang/python/config.el
```diff
@@ -287,7 +287,9 @@
 
 (use-package! poetry
   :when (featurep! +poetry)
-  :after python)
+  :after python
+  :init
+  (add-hook! 'python-mode-hook (poetry-tracking-mode +1)))
```
hlissner, line 292: This unnecessarily abuses inline lambdas. Use `(add-hook 'python-mode-hook #'poetry-tracking-mode)` instead.

hlissner (changes requested):

modules/lang/python/config.el
sei40kr: @hlissner 
Thank you. I've fixed the PR.

hlissner (approved):

hlissner:
Thanks!

PR #4147, opened 2020-10-24
Escape skk-mode on doom/escape

sei40kr:
# Description

Escape skk-mode on `doom/escape`.

### DIFF
--- a/modules/input/japanese/config.el
+++ b/modules/input/japanese/config.el
@@ -36,10 +36,8 @@
 
 (use-package! ddskk
   :general ("C-x j" #'skk-mode)
-  :config
-  (add-hook! 'doom-escape-hook
-    (defun +japanese-escape-skk-mode-h ()
-      (skk-mode -1))))
+  :init
+  (add-hook 'doom-escape-hook #'skk-mode-exit))
 
 
 ;;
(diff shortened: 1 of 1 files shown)


modules/input/japanese/config.el
```diff
@@ -35,7 +35,11 @@
 
 
 (use-package! ddskk
-  :general ("C-x j" #'skk-mode))
+  :general ("C-x j" #'skk-mode)
+  :config
+  (add-hook! 'doom-escape-hook
+    (defun +japanese-escape-skk-mode-h ()
+      (skk-mode -1))))
```
hlissner, line 42: Is there any reason you can't bind `skk-mode-exit` to it instead? e.g.

`(add-hook 'doom-escape-hook #'skk-mode-exit)`

hlissner (changes requested):

modules/input/japanese/config.el
sei40kr: @hlissner 
`skk-mode-exit` looks better! I've updated the PR. Thank you.

hlissner (approved):

hlissner:
Thanks!

PR #4166, opened 2020-10-27
Disable lsp indentation in scala mode

gilbertw1:
Metals (the scala LSP server) does not support formatting blocks of code and does appear to be interested in adding it in the future. This pull requests disables lsp indentation when using `+lsp` in `scala-mode` so that emacs will fall back to using indentation provided by `scala-mode` versus providing no indentation at all.

Relevant issue on formatting within Metals: https://github.com/scalameta/metals/issues/1649


hlissner (approved):

hlissner:
Thanks!

PR #4170, opened 2020-10-28
Add single quotes to ‘any quote’ text objects

ehamberg:
Add typographic single quotation marks to ‘any quote’ text objects.

(It could be argued that «guillemets» should be added too, but what is considered the left and right varies between languages, so I decided against adding them.)


hlissner (approved):

hlissner:
Thanks!

PR #4184, opened 2020-10-30
Fix typo in faq

innerout:


hlissner:
Hi! Thanks for the PR, but I cannot accept it at this time. I am moving a chunk of the documentation to our Discourse (will be public within the next week). Our faq.org will be the first to go, so there's no point in making any changes to this document. Thanks for the help, in any case!

PR #4193, opened 2020-11-01
insert-heading-respect-content for C-RET in org

### LINKED ISSUE #3306: C-RET in org-mode should insert new heading
When I'm in a list in an org file, M-RET adds another list item. Sometimes I would like to end the list and add a new heading instead. The default org keybinding for this is C-RET, which calls org-insert-heading-respect-content. In doom, C-RET seems to duplicate M-RET when I'm in a list item, and inserts an empty line when I'm in a new line (or several lines) below the list.

yogsototh:
> Close #3306

According to https://orgmode.org/manual/Structure-Editing.html
C-RET should call `insert-heading-respect-content`.

I think this should work, but as this is my first PR for doom-emacs it might be completely wrong.
In particular I considered that I shouldn't use/create a function with the `+org` namespace.

Note I tried to add:

```
(map! :map org-mode-map
        "C-RET" #'org-insert-heading-respect-content
        [C-return]   #'org-insert-heading-respect-content
        (:when IS-MAC
         [s-return]   #'org-insert-heading-respect-content))
```

To my `config.el` and but this still does not work as I expected.
So I might have done something wrong here.
Mainly if I call `insert-heading-respect-content` it works as expected.
But when I type `control-return` it still add an item and not a new heading.

### DIFF
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -648,14 +648,14 @@ between the two."
         "C-c C-i"    #'org-toggle-inline-images
         ;; textmate-esque newline insertion
         "S-RET"      #'+org/shift-return
-        "C-RET"      #'+org/insert-item-below
+        "C-RET"      #'org-insert-heading-respect-content
         "C-S-RET"    #'+org/insert-item-above
         "C-M-RET"    #'org-insert-subheading
-        [C-return]   #'+org/insert-item-below
+        [C-return]   #'org-insert-heading-respect-content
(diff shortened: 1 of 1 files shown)


hlissner:
Hi! Thanks for the PR, but I'm afraid I cannot accept it. This is not the correct solution for this problem. That `C-RET` isn't splitting lines is a design oversight which I will address when I have time, but otherwise it functioning correctly. To throw it away for this one edge case is overkill. In any case, thanks for the help!

PR #4197, opened 2020-11-01
Suppress format-all-mode error when no formatter is available

offbyone:
Having this on renders it difficult to work with debug-on-error, since a
_lot_ of filetypes don't have formatters


hlissner (approved):

offbyone:
Something is not quite right with this, as ... after a restart it's not suppressing the error. I'm not entirely sure why either.

(I swear I'm decent at this coding thing, but damn, does working on emacs configs confound me :) )

offbyone:
... except it _is_ working now? For a moment it actually stopped suppressing the original errors. So, it's not my patch that's wrong, it's something in emacs' signal handling. So, please ignore my comment and feel free to merge; this works, it's _computers_ that suck :)

hlissner:
Thanks!

PR #4210, opened 2020-11-04
Link from the FAQ to a list of built-in themes

axelson:
Makes it easier for newbies to know what themes are available after reading the FAQ. I debated about linking to https://github.com/hlissner/doom-emacs/blob/develop/modules/ui/doom/README.org#changing-theme instead but the doom-themes list seemed more comprehensive (and easier to link to)

### DIFF
--- a/docs/faq.org
+++ b/docs/faq.org
@@ -573,6 +573,8 @@ can easily change how Doom uses ~doom-theme~, but I can't (easily) control how
 you use the ~load-theme~ function.
 #+end_quote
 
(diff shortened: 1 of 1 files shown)


hlissner:
Hi! Thanks for the PR, but as I explained in https://github.com/hlissner/doom-emacs/pull/4184#issuecomment-719898404

> Hi! Thanks for the PR, but I cannot accept it at this time. I am moving a chunk of the documentation to our Discourse (will be public within the next week). Our faq.org will be the first to go, so there's no point in making any changes to this document. Thanks for the help, in any case!

axelson:
Okay, make sense. Thanks!

PR #4219, opened 2020-11-05
Remove eshell obsolete variable

condy0919:
`eshell-buffer-shorthand` was removed in Emacs 25, see https://github.com/emacs-mirror/emacs/blob/master/etc/NEWS.25#L1078-L1082


hlissner (approved):

hlissner:
Thanks!

PR #4225, opened 2020-11-06
Update getting_started.org

kshenoy:
Added clarification about updating an existing doom config to use Chemacs


hlissner:
Hi! Thanks for the PR, but I cannot accept it because an overhaul of our documentation is in the works (and already includes this tidbit), which makes this PR redundant. In any case, thanks for the help!

PR #4243, opened 2020-11-10
Switch to straight.el built-in builder for org

SebastienWae:
Since Doom updated the pinned commit for `straight.el` a few days ago we can now use their built-in builder for `org`.
See [https://github.com/raxod502/straight.el/pull/535](https://github.com/raxod502/straight.el/pull/535) for more info.

Two questions remain:
- Should we keep using the github mirror of `org` or use the default option of `straight` (https://code.orgmode.org/bzg/org-mode.git)
- Should we keep the [`:shadow` keyword](https://github.com/hlissner/doom-emacs/blob/92c9127b86f0f843cb47d20c77cb5ad82dc73f1a/core/core-packages.el#L199) of `package!` since it's no longer used anywhere in the code.

### DIFF
--- a/core/core-packages.el
+++ b/core/core-packages.el
@@ -240,7 +240,8 @@ processed."
 
 (defun doom-package-recipe-repo (package)
   "Resolve and return PACKAGE's (symbol) local-repo property."
-  (if-let* ((recipe (cdr (straight-recipes-retrieve package)))
+  (if-let* ((retrieved (straight-recipes-retrieve package))
+            (recipe (cdr (if (straight--quoted-form-p retrieved) (eval retrieved) retrieved)))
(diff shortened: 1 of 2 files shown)


hlissner:
Hi! Thanks for the PR, but I cannot accept it.

1. I want to continue using the mirror because (as the comments over our `package!` say), the original's server is unreliable compared to github's.
2. The built-in :build command does not fix the problem for shallow clones, which Doom defaults to. It only addresses full clones.
3. `:shadow` [is still in use](https://github.com/hlissner/doom-emacs/blob/develop/core/core-packages.el#L195-L208) and necessary.
4. Your change to `doom-package-recipe-repo` is unnecessary as well, because `straight-recipes-retrieve` evaluates quoted recipes internally.

PR #4254, opened 2020-11-11
:lang org: (+roam) Replace deprecated functions

brianmcgillion:
Update to the latest functions - after the following commit:
https://github.com/org-roam/org-roam/commit/47e83f7d

Org-Roam-Dailies deprecated the older forms of the functions
and replaced with the new constructs.


hlissner (approved):

hlissner:
Thanks!

PR #4255, opened 2020-11-11
Do not overwrite user's value of org-roam-db-location

RBckmnn:
This PR makes it so that `org-roam-db-location` is treated in the same way as `org-roam-directory`.


hlissner (approved):

hlissner:
Thanks!

PR #4261, opened 2020-11-13
Revert Rails command change

otavioschwanck:
Rails <= 5.2 dont have this options, the users cannot use this feature without searching in the doc.  I think is better to leave enabled. 

If possible, make something to limit the `projectile-rails-server-mode` @hlissner .  Like, if has more then 1.5k lines, when create new logs, erase the old ones.

### DIFF
--- a/modules/lang/ruby/README.org
+++ b/modules/lang/ruby/README.org
@@ -7,7 +7,6 @@
 - [[#description][Description]]
   - [[#module-flags][Module Flags]]
   - [[#packages][Packages]]
-  - [[#hacks][Hacks]]
 - [[#prerequisites][Prerequisites]]
   - [[#ubuntu][Ubuntu]]
   - [[#macos][MacOS]]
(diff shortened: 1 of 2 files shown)


hlissner:
> If possible, make something to limit the projectile-rails-server-mode

Please request this [upstream](https://github.com/asok/projectile-rails). I will remove this option once we have a solution in place. I'm currently swamped with other tasks so it'll be a while until I can look at it.

otavioschwanck:
> > If possible, make something to limit the projectile-rails-server-mode
> 
> Please request this [upstream](https://github.com/asok/projectile-rails). I will remove this option once we have a solution in place. I'm currently swamped with other tasks so it'll be a while until I can look at it.

https://github.com/asok/projectile-rails/issues/148

hlissner (approved):

hlissner:
With the suggestion posted upstream we now have a viable solution to this issue and this PR can be merged. Thanks for bringing this to my attention!

PR #4274, opened 2020-11-15
Remove unicode input mode default from Racket REPL

EoinDoherty:
`racket-mode` has an optional unicode input mode that replaces Greek letter names with unicode Greek characters (ex: 'pi' -> 'π'). Doom enables this input mode by default for `racket-repl-mode` buffers, but not for `racket-mode` buffers. This PR removes the `racket-repl-mode` hook and adds some documentation about `racket-mode`'s unicode input mode to Doom's Racket plugin readme. An alternative would be to enable unicode input for both modes, but some Racket DSLs don't interpret `λ` as `lambda`.


hlissner (approved):

hlissner:
Makes sense to me. Thanks!

Issue #3983, opened 2020-09-26
Cannot install doom using current (2020-09-24 or newer) gccEmacs


vlaci:
Doom cannot be installed in a clean environment if gccEmacs is compiled from https://github.com/emacs-mirror/emacs/commit/db354ffd578a46d898cac161ea1de1b42f96d2a0 or newer revisions.

**Steps to reproduce:**
The easiest way to reproduce is to use [this](https://hub.docker.com/r/andreacorallo/emacs-nativecomp) docker image:
```terminal
$ docker run -ti --network host andreacorallo/emacs-nativecomp
# apt update
# apt install git
# git clone --depth 1 https://github.com/hlissner/doom-emacs ~/.emacs.d
#  ~/.emacs.d/bin/doom -y install
```

After a while, doom install will stall, and the compiler worker processes are exited.
A workaround is to ctrl+c and run `doom install` again.

Aplying the following patch to Emacs also gets rid of the problem:

```patch
diff --git a/lisp/emacs-lisp/comp.el b/lisp/emacs-lisp/comp.el
index cac63a5978..6b7616ccba 100644
--- a/lisp/emacs-lisp/comp.el
+++ b/lisp/emacs-lisp/comp.el
@@ -2597,7 +2597,7 @@ Return the its filename if found or nil otherwise."
 ;;;###autoload
 (defun comp-subr-safe-advice (subr-name)
   "Make SUBR-NAME effectively advice-able when called from native code."
-  (unless (or (memq subr-name comp-never-optimize-functions)
+  (unless (or t (memq subr-name comp-never-optimize-functions)
               (gethash subr-name comp-installed-trampolines-h))
     (let ((trampoline-sym (comp-trampoline-sym subr-name)))
       (cl-assert (subr-primitive-p (symbol-function subr-name)))
```

**Additional details:**
Tracing the build process with strace reveals that the builders exit because of a `native-compiler-error-empty-byte` signal during compiling autoloads. 

I am aware that this issue may be with gccEmacs and not Doom itself but I am yet to reproduce it outside of Doom.

AndreaCorallo:
Do you know what's `subr-name` in the last call to `comp-subr-safe-advice` before crashing?
May be sufficient to just add like `(message "XXX %s" subr-name)` at the beginning of  `comp-subr-safe-advice` and see what's the sequence of calls.

vlaci:
subr-name is `make-process`.

FilipeAzero:
[...]

Issue #3655, opened 2020-07-31
doom build or doom upgrade stuck on "Waiting for 1 async jobs..." when using emacs native-comp branch.


h0cheung:
**What did you expect to happen?**
It should finish building normally.

**What actually happened?**
It stuck on "Waiting for 1 async jobs...", and the CPU usage of Emacs is almost zero.

**Additional details:**
- Include a link to your private config
<https://gitlab.com/h-cheung/dotfiles/-/tree/master/doom.d>

**System information:**
<details><pre>
Package subdirs is deprecated
((emacs
  (version . "28.0.50")
  (features . "XPM JPEG TIFF GIF PNG RSVG CAIRO SOUND GPM DBUS GLIB NOTIFY INOTIFY ACL GNUTLS LIBXML2 FREETYPE HARFBUZZ LIBOTF ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES NATIVE_COMP THREADS XWIDGETS LIBSYSTEMD JSON PDUMPER LCMS2")
  (build . "7月 28, 2020")
  (buildopts "--prefix=/usr --sysconfdir=/etc --libexecdir=/usr/lib --localstatedir=/var --mandir=/usr/share/man --with-gameuser=:games --with-sound=alsa --with-modules --without-gconf --without-gsettings --with-nativecomp --with-x-toolkit=gtk3 --without-xaw3d --without-m17n-flt --with-cairo --with-xwidgets --without-compress-install 'CFLAGS=-march=native -O2 -pipe -fuse-ld=gold' CPPFLAGS=-D_FORTIFY_SOURCE=2 LDFLAGS=-Wl,-O2,--sort-common,--as-needed,-z,relro,-z,now")
  (windowsys . batch)
  (daemonp . server-running))
 (doom
  (version . "2.0.9")
  (build . "HEAD -> develop 092480152 2020-07-30 18:57:53 -0400")
  (dir . "~/.doom.d/"))
 (system
  (type . gnu/linux)
  (config . "x86_64-pc-linux-gnu")
  (shell . "/usr/bin/zsh")
  (uname . "Linux 5.7.9-zen1-1-zen-anbox #1 ZEN SMP PREEMPT Sun, 19 Jul 2020 13:08:36 +0000 x86_64")
[...]

PR #4275, opened 2020-11-16
Bootstrap trampolines to prevent doom build hang

flatwhatson:
This is a workaround for `doom build` hanging on native-comp Emacs when run with an empty cache.  We're effectively automating the process of killing `doom build` and trying again.

Re: #3655
Re: #3983


hlissner (approved):

hlissner:
Very cool. Thanks!

PR #4276, opened 2020-11-16
Add with-editor.el to the compilation black-list

flatwhatson:
Sometimes compiling this file hangs; adding it to the compilation black-list prevents that.

Re: #3655


hlissner:
I'm not confident this fixes the issue (rather than masking the symptom), but I'll merge this for now and investigate it another time. Thanks for the help!

PR #4277, opened 2020-11-16
Adding a keyboard shortcut for CMake mode help

samwhitlock:
Adding a command in normal mode `Spc m h` (via localleader) to access the built-in cmake help.

This is a really good way to access the documentation for CMake commands without leaving to go to the browser.

I just tested this by making sure the command was available in the cmake mode after opening a cmake file, and that it did the correct thing (open the help command).

### DIFF
--- a/modules/lang/cc/config.el
+++ b/modules/lang/cc/config.el
@@ -161,6 +161,10 @@ This is ignored by ccls.")
   :after cmake-mode
   :config (set-company-backend! 'cmake-mode 'company-cmake))
 
+(map! :after cmake-mode
+      :map cmake-mode-map
+      :localleader
+      (:desc "CMake Help" :n "h" #'cmake-help))
 
 (use-package! demangle-mode
   :hook llvm-mode)
--- a/modules/lang/cc/config.el
+++ b/modules/lang/cc/config.el
@@ -154,17 +154,13 @@ This is ignored by ccls.")
 ;; Major modes
 
 (after! cmake-mode
-  (set-docsets! 'cmake-mode "CMake"))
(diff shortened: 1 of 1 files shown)


hlissner:
Would `cmake-help` work as a lookup handler? i.e.

```lisp
(after! cmake-mode
  (set-lookup-handler! 'cmake-mode
    :documentation #'cmake-help))
```

I'd much prefer this over a localleader keybind if it's possible. Then our universal "Look up documentation" key (on <kbd>K</kbd> for evil users and <kbd>C-c</kbd><kbd>c</kbd><kbd>k</kbd> for non-evil users) would work as expected in cmake-mode.

samwhitlock:
I changed to the lookup handler method per your suggestion. It's definitely nicer, but invoking `cmake-help` via `M-x` (or the `map!` keybind) causes [a window to be opened with the buffer contents](https://github.com/emacsmirror/cmake-mode/blob/master/cmake-mode.el#L258).

It looks like `+lookup--jump-to` expects to manage the buffer / window jumping (the buffer that cmake-mode normally jumps to is created, but not displayed / jumped to; it's in the background), but the `result` that the function gets when it calls the cmake handler is the same marker for the point in the Cmake buffer (_not_ the documentation buffer the command creates).

Is there some way to impose on `cmake-command-run` with some advice function (or some other workaround)?
My emacs-fu is not at that level to see how to fix this.

hlissner:
The lookup handler just needed a wrapper. I've implemented one in 315ae16 (and added a popup rule for it so, like other doc popups, they're displayed at the bottom of the frame). Thanks for bringing it to my attention!

PR #4280, opened 2020-11-17
Prevent native compilation for emacs-jupyter

hanmertens:
Something in `jupyter-channel.el` of emacs-jupyter does not get compiled correctly and raises void-variable jupyter-channel errors, see https://github.com/nnicandro/emacs-jupyter/issues/297.

I initially fixed it for myself by putting this in my `packages.el`:
```elisp
(package! jupyter :recipe (:no-native-compile t))
```
But some more testing suggested that preventing native compilation for just `jupyter-channel.el` also works, so I put that in this PR.


hlissner (approved):

hlissner:
Thanks!

PR #4281, opened 2020-11-18
New module: lang/d

fabianhjr:
Add d-mode for https://dlang.org/

### DIFF
--- a/modules/lang/d/packages.el
+++ b/modules/lang/d/packages.el
@@ -1,0 +1,4 @@
+;; -*- no-byte-compile: t; -*-
+;;; lang/d/packages.el
+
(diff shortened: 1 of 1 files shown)


hlissner:
Hi! Thanks for the PR, but I'm afraid I cannot accept it -- I do not accept modules that are so trivial that a user could implement it themselves in one or two lines (as is the case here). If there is more that can be added to it, like lookup and repl handlers, company plugins, etc. I'd be open to it, but it doesn't make the cut otherwise.

fabianhjr:
Hi, thanks for taking a look and maintaining doom. Will keep that in mind for future PRs and if d gets to the point of having more integrations I will resubmit.

PR #4283, opened 2020-11-18
Enable god-mode support for which-key if required

darth10:
In this change:

* Bump version pin of `god-mode`
* Call [`which-key-enable-god-mode-support`](https://github.com/justbur/emacs-which-key#god-mode) after loading `which-key` to enable support for `god-mode`. Prior to this change, keys entered in `god-local-mode` will not invoke `which-key` :no_entry:

As always, any feedback is more than welcome!

### DIFF
--- a/modules/editor/god/config.el
+++ b/modules/editor/god/config.el
@@ -4,4 +4,7 @@
   :hook (doom-after-init-modules . god-mode-all)
   :config
   (add-hook 'post-command-hook #'+god--configure-cursor-and-modeline-h)
-  (add-hook 'overwrite-mode-hook #'+god--toggle-on-overwrite-h))
+  (add-hook 'overwrite-mode-hook #'+god--toggle-on-overwrite-h)
+
+  (after! which-key
+    (which-key-enable-god-mode-support)))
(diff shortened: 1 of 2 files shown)


core/core-keybinds.el
```diff
@@ -203,7 +203,10 @@ localleader prefix."
   (setq-hook! 'which-key-init-buffer-hook line-spacing 3)
 
   (which-key-add-key-based-replacements doom-leader-key "<leader>")
-  (which-key-add-key-based-replacements doom-localleader-key "<localleader>"))
+  (which-key-add-key-based-replacements doom-localleader-key "<localleader>")
+
+  (when (featurep! :editor god)
+    (which-key-enable-god-mode-support)))
```
hlissner, line 209: Please move this to editor/god/config.el, in an `after! which-key` block inside `god-mode`'s `:config` block.

```lisp
(after! which-key
  (which-key-enable-god-mode-support))
```

hlissner (changes requested):

core/core-keybinds.el
darth10: Done!

hlissner (approved):

hlissner:
Thanks!

PR #4284, opened 2020-11-18
input/japanese: Do not eval skk-mode-exit if ddskk not loaded

sei40kr:
## Description

If we enable `:input/japanese` and press Esc before autoloading `skk`, an error occurs: 

```
Symbol's function definition is void: skk-mode-exit
```

This PR forces to load `skk` when calling `skk-mode-exit`.

### DIFF
--- a/modules/input/japanese/config.el
+++ b/modules/input/japanese/config.el
@@ -34,9 +34,9 @@
         pangu-spacing-real-insert-separtor t))
 
 
-(use-package! ddskk
+(use-package! skk
   :general ("C-x j" #'skk-mode)
-  :init
+  :config
   (add-hook 'doom-escape-hook #'skk-mode-exit))
 
 


hlissner:
If `skk-mode-exit` belongs to the `skk` package, why is it being used in the `ddskk` package? Should we not do this instead?

```diff
-(use-package! ddskk
+(use-package! skk
   :general ("C-x j" #'skk-mode)
   :init
   (add-hook 'doom-escape-hook #'skk-mode-exit))
```

sei40kr:
@hlissner 
I read the source code and couldn't find any part where declares `(provide 'ddskk)`.
(Still the package name on MELPA is ddskk)

So I changed it to skk and made not to eval `exit-skk-mode` when `(featurep 'skk)` is nil.

hlissner (approved):

hlissner:
Thanks!

PR #4285, opened 2020-11-19
core-lib: Add docstring for add-hook-trigger!

akirabaruah:
I was browsing the source trying to understand how/when these doom hooks get triggered and found this TODO. Hope this description is accurate and helpful!


hlissner (approved):

hlissner:
Thanks!

PR #4292, opened 2020-11-20
[WIP] config/default: Replace magit-clone with +magit/clone

sei40kr:
# Description

Replace `magit-clone` with `+magit/clone`.

# TODOs

- [ ] Handle the case `magit-clone-default-directory` is a function
  ```emacs-lisp
  (if (functionp magit-clone-default-directory)
      (funcall magit-clone-default-directory url-or-repo)
    magit-clone-default-directory)
  ```
- [ ] Declare `+magit-default-clone-url`


hlissner:
Could you explain why you are proposing this?

hlissner:
On second look, it appears `magit-clone` now supports what `+magit/clone` offers, so it'd be better to remove `+magit/clone` altogether.

PR #3971, opened 2020-09-22
Mu4e Module Revamp

tecosaur:
# Mu4e4u

A whole raft of changes.

## Features

- [x] Better handling of Gmail
- [x] Minor restyling of mu4e's headers view
- [x] Configure `org-msg`
  - [x] nicer styling
  - [x] nicer LaTeX fragments
  - [x] per-message toggling
- [x] Configure `mu4e-alert`
- [x] Add attachments with dired
- [x] Detail faster syncing solution (goimapnotify)
- [x] maybe add cooperative locking from https://tecosaur.github.io/emacs-config/config.html#process-control
- [x] Fashionable (and compact) account stripe

PR #4295, opened 2020-11-20
Update org-msg org-msg variable.

### LINKED ISSUE #4294: org-msg does not send plain text alternative due to change in variables
The `org-msg-text-plain-alternative` variable has been deprecated in org-msg.

The new way to have org-msg send html + text is to use: `org-msg-default-alternatives `

Ref: https://github.com/jeremy-compostella/org-msg/pull/71

ptoal:
Fixes #4294.

### DIFF
--- a/modules/email/mu4e/config.el
+++ b/modules/email/mu4e/config.el
@@ -110,7 +110,7 @@
   :config
   (setq org-msg-startup "inlineimages"
         org-msg-greeting-name-limit 3
-        org-msg-text-plain-alternative t))
+        org-msg-default-alternatives '(html text)))
 
 
 


modules/email/mu4e/config.el
```diff
@@ -110,7 +110,7 @@
   :config
   (setq org-msg-startup "inlineimages"
         org-msg-greeting-name-limit 3
-        org-msg-text-plain-alternative t))
+        org-msg-default-alternatives `(html text)))
```
hlissner, line 113: Please replace the backquote with a single quote.

```diff
-       org-msg-default-alternatives `(html text)))
+       org-msg-default-alternatives '(html text)))
```

hlissner (changes requested):

hlissner (approved):

ptoal:
Sorry about that.  Having a bad git day.

hlissner:
No worries! Thanks for the help!

tecosaur:
Why do people keep messing with my PR :cry: 
I don't like dealing with conflicts

ptoal:
I'm sorry.  I don't understand the issue?  I found problem that was caused by upstream change in org-msg, and fixed it.  No idea what you mean about your PR?

tecosaur:
Oh, don't mind me that much - I just have a big mu4e PR that has a bunch of stuff like this already taken care of, see: https://github.com/hlissner/doom-emacs/pull/3971

arvindsv:
I'm not sure if this is the right place to mention this or if I should create a new issue, but there is a difference between:

```
org-msg-default-alternatives '(html text)
```

and

```
org-msg-default-alternatives '(text html)
```

In section 7.2.3 of , [RFC-1341](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html) it says:

```
In general, user agents that compose multipart/alternative entities should place the body parts in increasing order of preference, that is, with the preferred format last. For fancy text, the sending user agent should put the plainest format first and the richest format last. Receiving user agents should pick and display the last format they are capable of displaying. In the case where one of the alternatives is itself of type "multipart" and contains unrecognized sub-parts, the user agent may choose either to show that alternative, an earlier alternative, or both. 
```

Since the default value was set to `'(html text)`, this caused my emails sent to favour `text over `html` in the display. Especially in GMail, it caused all formatting to be lost. I had to change it to `'(text html)` to get it to show properly.

I'm not sure what the earlier behaviour was, when `org-msg-text-plain-alternative` was set to `t`.

PR #4296, opened 2020-11-20
Set default directory in MacOS

### LINKED ISSUE #4224: LSP error when file path begins with `//`
When I open Doom it starts from the path `//` (filesystem root, with double slashes for some reason)

Screenshot 2020-11-06 at 16 02 25

if then I open some file, it causes an error in LSP mode

Screenshot 2020-11-06 at 16 04 16

If I restart the editor using `SPC q r`, it will restart in the user home folder

Screenshot 2020-11-06 at 16 09 56

and in this case, there is no error.

That seems to be a bug.

innerout:
Closes #4224

### DIFF
--- a/modules/os/macos/config.el
+++ b/modules/os/macos/config.el
@@ -7,6 +7,10 @@
 ;; variants thereof), since it requires no additional setup.
 (setq locate-command "mdfind")
 
+;; When Emacs is launched from the dock the environment variables are not populated.
+;; Emacs does not try to figure out the HOME directory anymore,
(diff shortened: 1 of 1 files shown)


hlissner:
Hi, thanks for the PR. I'm afraid this is not an acceptable solution to the problem. This causes issues for folks who use Emacs in the terminal -- where Emacs takes the `default-directory` from the current working directory the user started it from. This also causes issues for GUI users with sane Emacs installations on macOS. It only works as a workaround. I'll follow up with an explanation in #4224.

Issue #4303, opened 2020-11-23
Error running `doom build` on gccemacs 


bnjmnt4n:
**What did you expect to happen?**

I upgraded my gccemacs (pgtk version) to a newer version, and ran doom build, expecting a seamless upgrade.

**What actually happened?**

After running doom build, I get the following error:
```
Debugger entered--Lisp error: (error "Don’t call me!")
  error("Don't call me!")
  define-package("smartparens" "1.11.0" "Automatic insertion, wrapping and paredit-like nav..." ((dash "2.13.0") (cl-lib "0.3")))
  native-elisp-load("/home/bnjmnt4n/.emacs.d/.local/cache/eln/28.0.50-x..." nil)
  #f(compiled-function (process event) #<bytecode -0xdcd8999f5f7a232>)(#<process Compiling: /home/bnjmnt4n/.emacs.d/.local/straight/build-28.0.50/smartparens/smartparens-pkg.el> "finished\n")
  sleep-for(0.1)
  (let ((inhibit-message t)) (sleep-for 0.1))
  (if (/= previous pending) (progn (doom--print (doom--format (format (doom--output-class 'info "\33[KWaiting for %d async jobs...\33[1A" pending)))) (setq previous pending)) (let ((inhibit-message t)) (sleep-for 0.1)))
  (while (progn (setq pending (doom--native-compile-jobs)) (not (= 0 pending))) (if (/= previous pending) (progn (doom--print (doom--format (format (doom--output-class 'info "\33[KWaiting for %d async jobs...\33[1A" pending)))) (setq previous pending)) (let ((inhibit-message t)) (sleep-for 0.1))) (setq --cl-var-- nil))
  (let* ((pending nil) (previous 0) (--cl-var-- t)) (while (progn (setq pending (doom--native-compile-jobs)) (not (= 0 pending))) (if (/= previous pending) (progn (doom--print (doom--format (format (doom--output-class ... "\33[KWaiting for %d async jobs...\33[1A" pending)))) (setq previous pending)) (let ((inhibit-message t)) (sleep-for 0.1))) (setq --cl-var-- nil)) nil)
  doom--wait-for-compile-jobs()
  (progn (doom--compile-site-packages) (doom--wait-for-compile-jobs) (doom--write-missing-eln-errors) (doom--print (doom--format (format (doom--output-class 'success "\33[KRebuilt %d package(s)") (length built)))))
[...]

PR #4309, opened 2020-11-24
Update for the latest native-compile-async API

flatwhatson:
Renamed `comp-deferred-compilation-black-list` to `comp-deferred-compilation-deny-list`.

Removed the `late` load flag which is no longer required.

Added a check against the deny list when compiling all Elisp on the load-path, so we don't inadvertently compile something we shouldn't.

This change depends on https://github.com/raxod502/straight.el/pull/637.  I've added a hacky commit here to depend on the patched version of straight; that needs to become a proper bump commit once it's merged upstream.

### DIFF
--- a/core/cli/packages.el
+++ b/core/cli/packages.el
@@ -225,9 +225,12 @@ list remains lean."
                                    collect path)
              for file in (doom-files-in paths :match "\\.el\\(?:\\.gz\\)?$")
              if (and (file-exists-p (byte-compile-dest-file file))
-                     (not (doom--find-eln-file (doom--eln-file-name file)))) do
+                     (not (doom--find-eln-file (doom--eln-file-name file)))
+                     (not (cl-some (lambda (re)
+                                     (string-match-p re file))
+                                   comp-deferred-compilation-deny-list))) do
              (doom-log "Compiling %s" file)
-             (native-compile-async file nil 'late))))
+             (native-compile-async file))))
 
 (defun doom--bootstrap-trampolines ()
   "Build the trampolines we need to prevent hanging."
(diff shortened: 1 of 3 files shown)


h0cheung:
I tried to use this, but it report `Symbol’s value as variable is void: comp-deferred-compilation-deny-list` when `doom sync`.

bnjmnt4n:
@h0cheung you probably need to get the latest native-comp and rebuild Emacs - this variable renaming change was just made today: https://github.com/emacs-mirror/emacs/commit/6104ab0f35e10c4d61c8e8774aa246e6630c8ac0.

hlissner:
Since your PR was merged and gccemacs was updated, are we ready to merge this?

If so, could you change your DO NOT MERGE HACK commit to a bump commit for straight? If not, why not?

flatwhatson:
The only problem is the hard break for old versions of native-comp.  If you're concerned about breakage we could add some aliases and whatnot to smooth the transition.  If not, I think this is ready to go!

hlissner:
> If you're concerned about breakage we could add some aliases and whatnot to smooth the transition

Yeah, we may as well. Is it only for `comp-deferred-compilation-black-list`?

flatwhatson:
There's also the new selector argument of `native-compile-async` which straight makes use of.

flatwhatson:
@hlissner I've added the necessary stubs and tested against the following native-comp builds:
 - native-comp@3ae309bd (the last working version)
 - native-comp@033e9605 (version that introduced #4303)
 - native-comp@7a8370ed (latest version with API changes)

hlissner (approved):

hlissner:
Looks good to me. Thanks for looking into this!

Issue #4129, opened 2020-10-22
error "Don’t know how to format org-mode code"


brianmcgillion:
**What did you expect to happen?**

Saving an org file should work

**What actually happened?**
error "Don’t know how to format org-mode code"

**Additional details:**

Started after updating to d1effe2ddd3293547216ee74e68db60ea06bcf05

<details><pre>
If an error message is involved include a backtrace of it.

Debugger entered--Lisp error: (error "Don’t know how to format org-mode code")
  signal(error ("Don’t know how to format org-mode code"))
  error("Don't know how to format %S code" org-mode)
  format-all-buffer()
  funcall-interactively(format-all-buffer)
  call-interactively(format-all-buffer)
  +format-buffer-h()
  run-hooks(before-save-hook)
  basic-save-buffer(nil)
  save-buffer()
  #f(compiled-function (&optional stay-with-capture) "Finalize the capture process.\nWith prefix argument STAY-WITH-CAPTURE, jump to the location of the\ncaptured item after finalizing." (interactive "P") #<bytecode 0x55611a1>)(nil)
  apply(#f(compiled-function (&optional stay-with-capture) "Finalize the capture process.\nWith prefix argument STAY-WITH-CAPTURE, jump to the location of the\ncaptured item after finalizing." (interactive "P") #<bytecode 0x55611a1>) nil)
  org-capture-finalize(nil)
  funcall-interactively(org-capture-finalize nil)
  call-interactively(org-capture-finalize nil nil)
  command-execute(org-capture-finalize)

**Steps to reproduce:**
Enable roam
configure a roam capture
perform a capture
Confirm the capture
Backtrace

**System information:**
<details><pre>
```
SYSTEM  type       gnu/linux
	config     x86_64-pc-linux-gnu
	shell      /bin/bash
	uname      Linux 5.4.0-52-generic #57-Ubuntu SMP Thu Oct 15 10:57:00 UTC 2020 x86_64
	path       (~/.pyenv/shims ~/.pyenv/bin ~/.local/bin ~/bin /usr/local/sbin /usr/local/bin /usr/sbin /usr/bin /sbin /bin /usr/games /usr/local/games /snap/bin ~/.emacs.d/bin /snap/bin /var/lib/snapd/snap/bin /usr/lib/emacs/27.1/x86_64-linux-gnu)
EMACS   dir        ~/.emacs.d/
	version    27.1
	build      Sep 19, 2020
[...]

PR #4319, opened 2020-11-25
Prevent formatting error message (org-mode) #4129

credmp:
Prevent formatting error message (unsupported) when formatting org-mode #4129 . 

Running `format-all-buffer` yields the error message `format-all-buffer: Don’t know how to format org-mode code`. This same functionality is called when running the `before-save-hook` that comes from `(format +onsave)`. Although the document saves, the error message is unnecessary. Added `org-mode` to the exclusion list.

### DIFF
--- a/modules/editor/format/config.el
+++ b/modules/editor/format/config.el
@@ -4,6 +4,7 @@
   '(not emacs-lisp-mode  ; elisp's mechanisms are good enough
         sql-mode         ; sqlformat is currently broken
         tex-mode         ; latexindent is broken
+        org-mode         ; prevents error messages when saving org-mode
         latex-mode)
   "A list of major modes in which to reformat the buffer upon saving.
 


hlissner:
Can you still format src blocks for supported languages (e.g. python) with org-mode in the black list?

credmp:
That is a very good point. Until you just noted it I had never seen that happen.

 Currently

1. when saving inside a code block the code will be formatted with +onsave 
2. When saving outside the code block, the code block is not formatted.

With org-mode in the exclusions

1. the code block is not formatted when saved while in the code block.
2. the code block is not formatted when saved outside the code block

Although it removes the error message it does remove functionality it seems. Any suggestion on a better "fix" for this issue? Am happy to work on something that maintains the functionality.

credmp:
Judgin from the code this is the intended way it should work. From a user perspective it might be better to "catch" the error from format-all and deal with it silently instead of letting the error propagate, as it is not an "error" case imo.

credmp:
I have been giving this some thought. What is annoying about how it now works is that you will get error messages when the formatter does not know have access to the proper tools. So for me I have the tools for a few languages, but do not care about others.

So I would suggest to do the following:
- Add an config option to show error messages
- Catch the error message that is thrown and based on the option to show this information, display it (or not)

What do you think?

hlissner:
A better solution was found in #4425 that won't disable formatting in org-mode altogether, but should fix this issue, so I'll consider this resolved. Thanks for bringing it to my attention!

PR #4325, opened 2020-11-26
[Quickfix] Follow evil-magit migration to evil-collection

gagbo:
I don't know why yet, but while unpinned, having `:input (layout +bepo)` and `:editor (evil +everywhere)` fails to start because of a bad `evil-magit-change-popup-key` call. I removed evil-magit to only use evil-collection since evil-magit migrated last year, and now it seems to work


gagbo:
I think the issue was that the hooks to rotate keymaps were run twice since I am unpinned, and that now evil-magit code is also living in evil-collection-magit (triggering the evil-collection hook).

I'll try to clean this up to make a proper PR

modules/input/layout/+bepo.el
```diff
@@ -131,7 +132,8 @@
     (after! (evil org evil-org-agenda)
       (doom-bepo-rotate-bare-keymap '(org-agenda-keymap) doom-bepo-cr-rotation-style)
       (doom-bepo--evil-collection-hook nil '(evil-org-agenda-mode-map)))
-    (after! (evil magit evil-magit)
+
+    (after! (evil magit evil-collection)
```
gagbo, line 136: This is dangerous : what happens with the global evil-collection keymap hook then ?

hlissner (changes requested):
Migrating to evil-collection-magit needed a bit more finesse, which I've done in c1bdd63dc. Could you rebase the evil-collection/magit changes out of this commit, and reduce it to the fixes for `:input layout`?

gagbo:
Easier to close it and see if the issues are solved. I think this was the only issue (maybe I’ll need to replace `evil-magit` with `evil-collection-magit` in the `after!` call, but I actually don’t know the interactions between this manual hook and the big `evil-collection-keymap-hook` I also installed

PR #4334, opened 2020-11-29
Improve byte compilation message

rgrinberg:
When compiling standalone files such as `init.el`, instead of saying
that it belongs to module `nil nil`, say that these files are
standalone.


hlissner (approved):

hlissner:
Thanks!

PR #4335, opened 2020-11-29
Elisp `eldoc': fix "Invalid format operation %-"

nbfalcon:
In `+emacs-lisp-append-value-to-eldoc-a`, if the `frame-width` of the minibuffer
is smaller than the length of the documentation + " [...]" + 1, a negative
maximum %s bound is passed, causing the error in the title.

Fix this by clamping the computed LIMIT to 0.

If needed, I can make a full issue, but I hope that from my description, the cause is clear: `limit` = `(frame-width)` (minibuffer, in eldoc) - `(length str + length " [...]" + 1)` -> can be negative -> error.

### DIFF
--- a/modules/lang/emacs-lisp/config.el
+++ b/modules/lang/emacs-lisp/config.el
@@ -105,7 +105,7 @@ employed so that flycheck still does *some* helpful linting.")
                      (str (symbol-value sym))
                      (str (prin1-to-string str))
                      (limit (- (frame-width) (length ret) (length truncated) 1)))
-                (format (format "%%0.%ds%%s" limit)
+                (format (format "%%0.%ds%%s" (min limit 0))
                         (propertize str 'face 'warning)
                         (if (< (length str) limit) "" truncated))))))
 
--- a/modules/lang/emacs-lisp/config.el
+++ b/modules/lang/emacs-lisp/config.el
@@ -105,7 +105,7 @@ employed so that flycheck still does *some* helpful linting.")
                      (str (symbol-value sym))
                      (str (prin1-to-string str))
                      (limit (- (frame-width) (length ret) (length truncated) 1)))
-                (format (format "%%0.%ds%%s" (min limit 0))
+                (format (format "%%0.%ds%%s" (max limit 0))
                         (propertize str 'face 'warning)
                         (if (< (length str) limit) "" truncated))))))
 


modules/lang/emacs-lisp/config.el
```diff
@@ -105,7 +105,7 @@ employed so that flycheck still does *some* helpful linting.")
                      (str (symbol-value sym))
                      (str (prin1-to-string str))
                      (limit (- (frame-width) (length ret) (length truncated) 1)))
-                (format (format "%%0.%ds%%s" limit)
+                (format (format "%%0.%ds%%s" (min limit 0))
```
hlissner, line 108: Function should be `max`, not `min`.

hlissner (changes requested):

modules/lang/emacs-lisp/config.el
nbfalcon: This is an awkward blunder... Fixed.

hlissner (approved):

hlissner:
Thanks!

nbfalcon:
As a side note, what `git-commit-major-mode` should I use in my commit messages when contributing? Currently, I assume and use `git-commit-elisp-text-mode`.

nbfalcon:
You may want to add that to `dir-locals`.

hlissner:
> As a side note, what git-commit-major-mode should I use in my commit messages when contributing? Currently, I assume and use git-commit-elisp-text-mode.

I've simply used the default (`text-mode`), but `git-commit-elisp-text-mode` looks interesting. I'll give it a spin before adding it to our dir-locals. Thanks!

sxjune:
This fix may cause the following problems ?
On Windows 10 Emacs 27.1, Upgrade from `Doom v2.0.9 (HEAD -> develop, origin/develop, origin/HEAD 88e18fc7c 2021-02-27 23:56:15 -0500)` to `Doom v2.0.9 (HEAD -> develop 46dedb3e3 2021-03-13 16:35:21 -0500)`.

nbfalcon:
@sxjune I don't think your issue is caused by my PR - it is very much unrelated. You'll need to upgrade `magit` for your issue to go away, since recently we switched to the upstream safe-variable predicates. You can do this using `./bin/doom sync -u`

PR #4336, opened 2020-11-29
Changing elisp macro expand key to avoid collision with parinfer mode toggle

samwhitlock:
`Spc m m` toggles parinfer mode and expands macros in elisp mode.

This is just a proposed solution to change the macro expand to `Spc m b`.

If this would be too destructive to folks' workflows (I use macro expanding a lot myself; only discovered the parinfer conflict when helping someone out on discord just now), please close this unmerged.

### DIFF
--- a/modules/lang/emacs-lisp/config.el
+++ b/modules/lang/emacs-lisp/config.el
@@ -111,7 +111,7 @@ employed so that flycheck still does *some* helpful linting.")
 
   (map! :localleader
         :map emacs-lisp-mode-map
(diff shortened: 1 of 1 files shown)


hlissner:
`SPC m m` is the canonical keybind for macro expansion in most (if not all of) our lisp modules, not just elisp. So I think parinfer `SPC m m` needs a new keybind instead. I do not use the parinfer module so this slipped under my radar. I'll look into a better keybind soon. Thanks for bringing it to my attention!

PR #4341, opened 2020-11-29
Fix doom upgrade

rgrinberg:
Previously, `doom upgrade` would emit:

```
New revision: fatal: amb (fatal: ambiguous argument '_upgrade/HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
```

It appears it's not possible to remote revisions with
`<remote>/<revision>`. Instead, we fetch the upgrade remote into a named
revision and then use it.


hlissner (approved):

hlissner:
Thanks!

PR #4343, opened 2020-11-30
Fix typo

robbinch:
I think this is a typo? At least I got my `magit-status` to work over TRAMP after this change.

### DIFF
--- a/core/core-lib.el
+++ b/core/core-lib.el
@@ -689,7 +689,7 @@ REMOTE is non-nil, search on the remote host indicated by
                     command
                     (mapcar
                      (lambda (x) (concat (file-remote-p default-directory) x))
-                     (exec-path))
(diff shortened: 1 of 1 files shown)


hlissner:
This is not a typo. `exec-path` is a function and a variable. The function fetches the `PATH` from the remote session (over TRAMP). `exec-path` returns your local `PATH`, which is less correct. The real issue lies elsewhere. In any case, thanks for the help!

robbinch:
Thanks for the quick response! Learned something today. The error I got was
something to the effect of symbol's function definition is void for
exec-path (not at comp now so no trace). I'm using Emacs 26.3, but in any
case you gave me an idea on how to work around this (define the function in
my config). Thanks again for doom-emacs!

PR #4347, opened 2020-12-01
Moved emojis to .local/

catap:
As soon as I enabled `ui/emoji` with images it downloaded images into `emojis` folder inside doom-emacs source tree and `git status` starts to complain.

Let move images to right place that is `.local/`.

### DIFF
--- a/modules/ui/emoji/config.el
+++ b/modules/ui/emoji/config.el
@@ -8,4 +8,5 @@
          nil (list (if (featurep! +ascii) 'ascii)
                    (if (featurep! +github) 'github)
                    (if (featurep! +unicode) 'unicode))))
+  (setq emojify-emojis-dir (concat doom-local-dir "emojis"))
   (emojify-set-emoji-styles emojify-styles))


gagbo:
Hello,

I don't use emojis, but usually the files downloaded by Emacs should not be at the root of the repository, but probably in a subdirectory of `doom-etc-dir` (`~/.emacs.d/.local/etc`) or `doom-local-dir` more generally (`~/.emacs.d/.local`) which are properly ignored.

In my opinion, a better fix would then be to check the configuration of `:ui emoji` so that it uses these folders and avoid polluting the root. But it is something that needs fixing for sure :+1:

catap:
@gagbo very good point. Thanks.

hlissner:
c517be8 should indirectly fix this issue (for this and many other packages). Let me know if that isn't the case. Thanks for bringing this to my attention!

catap:
@hlissner I'd love to confirm that this is fixed.

PR #4349, opened 2020-12-01
grammar: macports compatibility

catap:
Macports is quite popular a way to install some application to macOS and it is shame that it doesn't support such way by default.


hlissner (approved):

hlissner:
Thanks!

Issue #4346, opened 2020-12-01
doom up removes git's remote for upgrade


catap:
```
√ doom-emacs % ~/src/doom-emacs/bin/doom up                                                                                                 (git)-[develop]
> Preparing to upgrade Doom Emacs and its packages...
x There was an unexpected error
  Message: error
  Data: (error . "Failed to fetch from upstream")
  Backtrace:
    (error "Failed to fetch from upstream")
    (or (= 0 (car (setq result (doom-call-process "git" "fetch" "--tags" doom-
    (let (result) (or (= 0 (car (doom-call-process "git" "remote" "add" doom-r
    (unwind-protect (let (result) (or (= 0 (car (doom-call-process "git" "remo
    (let* ((branch (replace-regexp-in-string "^\\(?:[^/]+/[^/]+/\\)?\\(.+\\)\\
    (let ((default-directory doom-emacs-dir) process-file-side-effects) (doom-
    (doom-cli-upgrade nil nil)
    (cond (packages-only-p (doom-cli-execute "sync" "-u") (doom--print (doom--
    (let ((doom-auto-discard force-p)) (cond (packages-only-p (doom-cli-execut
    (let ((force-p (cdr (assq 'force-p --alist--))) (packages-only-p (cdr (ass
! Extended backtrace logged to .local/doom.error.log
? doom-emacs % git remote                                                                                                                   (git)-[develop]
origin
√ doom-emacs % git remote add _upgrade https://github.com/hlissner/doom-emacs                                                               (git)-[develop]
√ doom-emacs % git remote update                                                                                                            (git)-[develop]
Fetching origin
Fetching _upgrade
From https://github.com/hlissner/doom-emacs
 * [new branch]          develop     -> _upgrade/develop
 * [new branch]          master      -> _upgrade/master
 * [new branch]          screenshots -> _upgrade/screenshots
√ doom-emacs % ~/src/doom-emacs/bin/doom up                                                                                                 (git)-[develop]
> Preparing to upgrade Doom Emacs and its packages...
x There was an unexpected error
  Message: error
  Data: (error . "Failed to fetch from upstream")
  Backtrace:
    (error "Failed to fetch from upstream")
    (or (= 0 (car (setq result (doom-call-process "git" "fetch" "--tags" doom-
    (let (result) (or (= 0 (car (doom-call-process "git" "remote" "add" doom-r
    (unwind-protect (let (result) (or (= 0 (car (doom-call-process "git" "remo
[...]

PR #4350, opened 2020-12-01
upgrade.el: force fetch

catap:
`doom-cli-upgrade` calls `git fetch` to fetch remote branch and tags and check output code of this command to complain when it failed.

It fails when remote server isn't available that is good.

But it also fails when it can't to fast-forward a branch that will be used to upgrade doom-emacs.

The last case created a loop and the only way to broke this loop is remote a branch that should be used for upgrade.

Unfortunately user has no idea about branch name and can't do anything, just enjoy crash with `Failed to fetch from upstream` message.

This commit added `--force` flag to `git fetch` that forces it to overwrite branch that is used to upgrade that prevent such loop.

It also fixed https://github.com/hlissner/doom-emacs/issues/4346


hlissner (approved):

hlissner:
Thanks for the help!

PR #4358, opened 2020-12-03
Bump :lang scheme

flatwhatson:
jaor/geiser@2accab7 -> jaor/geiser@cd00be6
flatwhatson/flycheck-guile@2940f16 -> flatwhatson/flycheck-guile@e46d6e5


hlissner:
Thanks!

PR #4373, opened 2020-12-04
Add (some) evil vterm bindings

mjlbach:
See https://github.com/akermu/emacs-libvterm/pull/455 and https://github.com/akermu/emacs-libvterm/issues/313#issuecomment-738842507

We should probably add few more... Suggestions?

Edit: This isn't actually editing buffer contents and is broken atm

### DIFF
--- a/modules/term/vterm/config.el
+++ b/modules/term/vterm/config.el
@@ -22,8 +22,59 @@
     hscroll-margin 0)
 
   ;; Restore the point's location when leaving and re-entering insert mode.
+  ; (when (featurep! :editor evil)
+  ;   (add-hook! 'vterm-mode-hook
+  ;     (defun +vterm-init-remember-point-h ()
+  ;       (add-hook 'evil-insert-state-exit-hook #'+vterm-remember-insert-point-h nil t)
+  ;       (add-hook 'evil-insert-state-entry-hook #'+vterm-goto-insert-point-h nil t)))))
+
   (when (featurep! :editor evil)
+      (defun vterm-evil-insert ()
+        (interactive)
+        (vterm-goto-char (point))
+        (call-interactively #'evil-insert))
+
+      (defun vterm-evil-append ()
+        (interactive)
+        (vterm-goto-char (1+ (point)))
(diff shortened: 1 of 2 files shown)


mjlbach:
*draft, some things aren't working

modules/term/vterm/config.el
```diff
@@ -26,4 +26,34 @@
...
+
+      ;; Add evil specific bindings that work with vterm mode
+      (defun vterm-evil-insert ()
+        (interactive)
+        (vterm-goto-char (point))
+        (call-interactively #'evil-insert))
+      (defun vterm-evil-append ()
+        (interactive)
+        (vterm-goto-char (point))
+        (call-interactively #'evil-append))
```
jixiuf, line 39: this should be （1+ （point））

mjlbach:
Thanks @jixiuf! (I copied from you, so not sure how that happened). I'm not quite sure why, but the delete bindings are still not working (deletes in rendered buffer but not on the cli)

I added this to my zshrc
```
vterm_printf(){
    if [ -n "$TMUX" ]; then
        # Tell tmux to pass the escape sequences through
        # (Source: http://permalink.gmane.org/gmane.comp.terminal-emulators.tmux.user/1324)
        printf "\ePtmux;\e\e]%s\007\e\\" "$1"
    elif [ "${TERM%%-*}" = "screen" ]; then
        # GNU screen (screen, screen-256color, screen-256color-bce)
        printf "\eP\e]%s\007\e\\" "$1"
    else
        printf "\e]%s\e\\" "$1"
    fi
}

vterm_prompt_end() {
    vterm_printf "51;A$(whoami)@$(hostname):$(pwd)";
}
setopt PROMPT_SUBST
PROMPT=$PROMPT'%{$(vterm_prompt_end)%}'
```

jixiuf:
That‘s intended behavior. Only the area after last prompt are editable, you can only delete/change there.

mjlbach:
Yes, it wasn't working on the last prompt, but it seems to be fine now.

PR #4381, opened 2020-12-06
Add word existence checking for +spell/correct

iwahbe:
`(bounds-of-thing-at-point 'word)` can return `nil`. This checks that it
does not, giving an `user-error` if no word is found. I think this is better
then the `(wrong-number-of-arguments (start . end) nil)` error message.

### DIFF
--- a/modules/checkers/spell/autoload/+spell-fu.el
+++ b/modules/checkers/spell/autoload/+spell-fu.el
@@ -64,9 +64,9 @@
   (if (not (or (featurep! :completion ivy)
                (featurep! :completion helm)))
       (call-interactively #'ispell-word)
-    (let ((current-point (bounds-of-thing-at-point 'word)))
-      (if current-point
-    (cl-destructuring-bind (start . end) current-point
+    (cl-destructuring-bind (start . end)
+        (or (bounds-of-thing-at-point 'word)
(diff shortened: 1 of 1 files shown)


modules/checkers/spell/autoload/+spell-fu.el
```diff
@@ -64,8 +64,9 @@
   (if (not (or (featurep! :completion ivy)
                (featurep! :completion helm)))
       (call-interactively #'ispell-word)
-    (cl-destructuring-bind (start . end)
-        (bounds-of-thing-at-point 'word)
```
hlissner, line 68: This is a roundabout way of doing this. Could you use this approach instead?

```diff
     (cl-destructuring-bind (start . end)
-        (bounds-of-thing-at-point 'word)
+        (or (bounds-of-thing-at-point 'word)
+            (user-error "No word at point"))
```

hlissner (changes requested):

modules/checkers/spell/autoload/+spell-fu.el
iwahbe: I didn't know that `user-error` short circuited the function. That would work.

hlissner (approved):

hlissner:
Thanks!

PR #4389, opened 2020-12-08
Added new package to csharp lang module

samwdp:
Added sharper as an interface to interact with dotnet cli through doom. Created new flag to enable this option in the init file


hlissner (approved):

hlissner:
Sorry for the late response. I'm not sure sure about the flag -- I want to avoid setting a precedent for adding one flag per package, but I'll revisit this another time. In any case, thanks for the help!

PR #4398, opened 2020-12-09
Bump js2-mode in :lang javascript

MageJohn:
This commit from js2-mode adds support for recently added logical assignment operators, such as `&&=`.


hlissner:
Hi! Thanks for your PR, but bump commits must follow a strict format (so they can be machine readable by future CI), so I've done so in c9dff2f. Thanks for bringing it to my attention!

PR #4409, opened 2020-12-12
Add "python3" to possible names for the python binary

icmor:
Followed lang/python/config.el which already checks for python3
when setting the python-shell-interpreter.

Python version naming is a hot mess:
https://www.python.org/dev/peps/pep-0394/


hlissner (approved):

hlissner:
Looks good to me, thanks!

PR #4414, opened 2020-12-12
Enable `smartparens' for `evil'

nbfalcon:
febf960 introduced a regression, in that it caused `smartparens-mode` to
no longer be enabled for `evil-ex`. This is because the latter doesn't
call `read--expression`, but calls the minibuffer directly instead.

Return `doom-init-smartparens-in-minibuffer-maybe-h`, enabling
`smartparens` for `evil-ex`. The hook responsibly for enabling it in
`eval-expression-minibuffer-setup-hook` was renamed to
`doom-init-smartparens-in-eval-expression-h`, since otherwise the naming
would have been awkward.

Also explicity enable `smartparens-mode` instead of toggling it.

### DIFF
--- a/core/core-editor.el
+++ b/core/core-editor.el
@@ -505,10 +505,19 @@ files, so we replace calls to `pp' with the much faster `prin1'."
     (setf (alist-get key sp-message-alist) nil))
 
   (add-hook! 'eval-expression-minibuffer-setup-hook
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Thanks!

PR #4415, opened 2020-12-12
Workaround: fix goto definition on modules

nbfalcon:
`+emacs-lisp-lookup-definition` does not work when browsing the directory of a
module, because, due to a possible bug in `counsel`, the visited buffer is not
immediately visible after calling `counsel-find-file` (unlike with `find-file`).
As such, the backend should return `deferred` for that case.

See abo-abo/swiper#2752. This should be removed once that PR is merged.

### DIFF
--- a/modules/lang/emacs-lisp/autoload.el
+++ b/modules/lang/emacs-lisp/autoload.el
@@ -61,7 +61,9 @@ to a pop up buffer."
 (defun +emacs-lisp-lookup-definition (_thing)
   "Lookup definition of THING."
   (if-let (module (+emacs-lisp--module-at-point))
-      (doom/help-modules (car module) (cadr module) 'visit-dir)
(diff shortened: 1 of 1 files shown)


nbfalcon:
Would you prefer to fix this in upstream and to not have this workaround at all, or would you rather have this workaround in temporarily? Note that it is still unclear whether that should be considered a bug in `ivy`, resolved by my PR.

hlissner:
Seems a resolution upstream will take a while, so may as keep the workaround for the time being. Thanks for the help!

nbfalcon:
@hlissner My upstream PR was merged. However, I believe this change is still necessary: `+lookup/definition` followed by `C-SPC` (`ivy-call)` to open a file in the module followed by `ESC` would cause the newly-opened file to be buried. Selecting a file using `RET` would work as it does now, though.

What do you think? The comment stating that this is a workaround should be removed, though.

PR #4416, opened 2020-12-12
`doom/help-modules': C-u -> browse directory

nbfalcon:
If called with a C-u prefix, `doom/help-modules` now browses the module's
directory instead of opening its documentation. This exposes the VISIT-DIR
argument to `interactive` use.

### DIFF
--- a/core/autoload/help.el
+++ b/core/autoload/help.el
@@ -351,13 +351,15 @@ module derived from a `featurep!' or `require!' call, c) the module that the
 current file is in, or d) the module associated with the current major mode (see
 `doom--help-major-mode-module-alist')."
   (interactive
-   (mapcar #'intern
-           (split-string
-            (completing-read "Describe module: "
-                             (doom--help-modules-list)
-                             nil t nil nil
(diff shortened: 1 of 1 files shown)


nbfalcon:
As a side note, in latest `develop`, magit seems to no longer work like it did previously: I am now in `evil-emacs-state` in the magit status buffer and `V` is no longer bound to `evil-visual-state`; `evil-visual-state` also no longer has magit commands available, which is annoying since I use it for `magit-spinout` quite a lot.

nbfalcon:
I found the issue: 6665d2b6c235d80f5126bb328123cf4188527d43. `evil-collection` needs to be bumped, because `evil-collection-magit` is too old.

hlissner:
> I found the issue: 6665d2b. evil-collection needs to be bumped, because evil-collection-magit is too old.

Sounds like were either `git pull`ing instead of `doom upgrade`ing. Update your local packages with `doom sync -u`. evil-collection has been bumped several times since your linked commit, and none of the commits since touch evil-collection-magit.

PR #4424, opened 2020-12-13
popup.el: consider auto-saving indirect buffers too

mskorzhinskiy:
Function (buffer-file-name) always returns nil with indirect buffer as
an argument. Extracting base buffer and checking if base buffer visiting
file fixes data loss in indirect buffers showed in popups.

E-mail thread in the org-mode mailing list:

  https://lists.gnu.org/archive/html/emacs-orgmode/2020-12/msg00085.html


hlissner (approved):

hlissner:
Makes sense to me. Thanks!

PR #4425, opened 2020-12-13
A workaround for formatting org mode

samwhitlock:
Only attempt to format the org source block if the point is in it, else do nothing.

`format-all-buffer` failing to look up the formatter for Org mode is causing issues like #4319 . Instead of just excluding this, this PR uses a compromise: only format the code block region the point is currently in, if any.

I don't like this workaround, but it is the best compromise I can think of and it prevents this common error.
I'm not sure what the idiomatic way to format normal org text on save would be, but `format-all-buffer` seems to be for LSP formatters, not for internal Emacs formatting (as far as I can tell).

### DIFF
--- a/modules/editor/format/autoload/format.el
+++ b/modules/editor/format/autoload/format.el
@@ -225,9 +225,9 @@ If nil, BEG and/or END will default to the boundaries of the src block at point.
 (defun +format/buffer ()
   "Reformat the current buffer using LSP or `format-all-buffer'."
   (interactive)
-  (if (and (eq major-mode 'org-mode)
-           (org-in-src-block-p t))
-      (+format--org-region nil nil)
+  (if (eq major-mode 'org-mode)
+      (when (org-in-src-block-p t)
+        (+format--org-region nil nil))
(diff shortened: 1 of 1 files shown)


samwhitlock:
Another interesting interaction and reason why this fix is useful even in lieu of a proper fix: background saving of org buffers is affected.

I have the following bit in my config

```emacs-lisp
(after! org
  (appendq! org-after-todo-state-change-hook 'save-buffer))
```

This appears to actually save the file (based on the disk contents), but this org mode formatting failure (fixed by this commit) seems to prevent the buffer from being marked as saved (i.e. it still asks for confirmation to save before quitting).

hlissner (approved):

hlissner:
Sounds good to me. Thanks!

samwhitlock:
@hughwilliams94 it's probably best to open a new issue and mention this.

I suspect that the `(eq major-mode 'org-mode)` may need to be changed to `(derived-mode-p 'org-mode)`, but it's best to edebug this function to see what it does.

PR #4428, opened 2020-12-14
Module documentation: fix error if no description

nbfalcon:
`+emacs-lisp-lookup-documentation` does a `re-search-forward` for "*Description"
without NOERROR. This causes an `error` if the module has no "*Description", as
it the case for the `editorconfig` module.

To reproduce:
0. `M-x debug-on-error`
1. `M-x evil-lookup` on the editorconfig module errors

### DIFF
--- a/modules/lang/emacs-lisp/autoload.el
+++ b/modules/lang/emacs-lisp/autoload.el
@@ -75,7 +75,8 @@ if it's callable, `apropos' otherwise."
                (re-search-forward
                 (if (caddr module)
                     "\\* Module Flags$"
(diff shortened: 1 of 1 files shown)


hlissner:
Hi! Thanks for the PR but I cannot accept it. This error is intentional. It could be done more elegantly, but removing it altogether is less desirable than fixing the documentation for the editorconfig module, which doesn't use [our updated README template](https://github.com/hlissner/doom-emacs/issues/1166#issuecomment-462459077). In any case, thanks for the help!

PR #4433, opened 2020-12-14
Do not panic when xclip is not available

yangsheng6810:
`(xclip-mode +1)` will raise an error when `xlip` is not found, which can interrupt the whole initialization process.

### DIFF
--- a/modules/os/tty/config.el
+++ b/modules/os/tty/config.el
@@ -22,7 +22,7 @@
       ;; wl-copy, termux-clipboard-get, or getclip (cygwin); depending on what
       ;; is available.
       (and (require 'xclip nil t)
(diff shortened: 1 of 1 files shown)


hlissner (changes requested):
Please use `(ignore-errors (xclip-mode +1))` instead

hlissner:
Actually, it'd be better to do

```elisp
(and (require 'xclip nil t)
     (with-demoted-errors "%s" (xclip-mode +1)))
```

So the issue is visibly logged to \*Messages\*.

yangsheng6810:
Force push as suggested.

hlissner (approved):

hlissner:
Thanks!

Issue #4427, opened 2020-12-13
User inputs gets trapped in teal ivy-actions hydra which hides itself but does not exit after first action is executed


AloisJanicek:
**What did you expect to happen?**
When being presented with ivy actions after calling `C-o` (`ivy-dispatching-done`) from `ivy` I expect the available actions menu to de-active itself after I execute one of its actions OR to stay visible (less convenient but predictable) so its visibility reminds me I should close it by one of the listed keys.

**What actually happened?**
After the https://github.com/hlissner/doom-emacs/commit/c4a0174fe2d02741df73e1ff97fa56b2c3ea2a42 the action menu went from buggy minibuffer menu which notoriously refuses to show all action candidates (even with the `ivy-read-action-format-columns`) to `teal` hydra which hides itself after user executes one of the listed actions but still somehow remains active and traps all user keystrokes which leads to all kids of unpredictable and dangerous behavior. This invisible teal hydra can be exited only by one of the keys listed in `ivy-dispatching-done-hydra-exit-keys`, which currently are `M-o` and `C-g`.

**Steps to reproduce:**
1. `SPC f r` for `counsel-recentf`
2. `C-o` on the desired candidate
3. chose `j` "other window" action on desired candidate
4. [files opens, hydra disappears, user input remains trapped]
5. pressing `j` to navigate one line down in newly opened file leads to **complete buffer content deletion** with an error:
```
ivy--done: No catch for tag: exit, nil
```
6. pressing any other key which is not listed in the teal ivy-actions hydra leads to:
```
An amaranth Hydra can only exit through a blue head
```

**System information:**
<details><pre>
```
SYSTEM  type       gnu/linux
        config     x86_64-pc-linux-gnu
        shell      /bin/bash
        uname      Linux 5.9.13-zen1-1-zen #1 ZEN SMP PREEMPT Tue, 08 Dec 2020 12:10:52 +0000 x86_64
EMACS   dir        ~/.doom-vanilla/.emacs.d/
        version    28.0.50
        build      Dec 08, 2020
[...]

PR #4434, opened 2020-12-14
Better ivy hydras usability

AloisJanicek:
There are two hydras that can be opened from ivy interface:

1. `ivy-dispatching-done` (doom key: `"C-o"`, vanilla key: `"M-o"`)
2. `hydra-ivy/body`           (doom key: `"M-o"`, vanilla key: `"C-o"`)

Original behavior is that they both can be exited / toggled by the same key they were opened.

This PR:

1. sets hydra as default interface for `ivy-dispatching-done`
Contrary to original assumption outlined in https://github.com/hlissner/doom-emacs/issues/4427 hydra works just fine as `ivy-dispatching-done` interface. Regression introduced in https://github.com/hlissner/doom-emacs/commit/c4a0174fe2d02741df73e1ff97fa56b2c3ea2a42 was a result of assigning function to the wrong kind of variable.

 2. restores the original toggle behavior but does it without reverting the decision to swap the `"C-o"` and `"M-o"` key bindings.

### DIFF
--- a/modules/completion/ivy/config.el
+++ b/modules/completion/ivy/config.el
@@ -59,6 +59,7 @@ results buffer.")
   (setq ivy-height 17
         ivy-wrap t
         ivy-fixed-height-minibuffer t
+        ivy-read-action-function #'ivy-hydra-read-action
         ivy-read-action-format-function #'ivy-read-action-format-columns
         projectile-completion-system 'ivy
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Thanks for looking into this!

PR #4435, opened 2020-12-15
Prioritize company-capf back-end in org-journal

AloisJanicek:
This allows functionality like completion of org-roam items seamlessly work in org-journal-mode files too.

Same setting is already set for org-mode itself.


hlissner (approved):

hlissner:
Thanks!

PR #4436, opened 2020-12-15
[4415] remove FIXME comment

nbfalcon:
Due to `ivy-call` still causing files to be opened in the
background (which is not a bug in `ivy`),
`+emacs-lisp-lookup-definition` still needs to return `deferred` for
modules. The comment is as such misleading and needs to be removed (it
has been resolved).

### DIFF
--- a/modules/lang/emacs-lisp/autoload.el
+++ b/modules/lang/emacs-lisp/autoload.el
@@ -61,8 +61,6 @@ to a pop up buffer."
 (defun +emacs-lisp-lookup-definition (_thing)
   "Lookup definition of THING."
   (if-let (module (+emacs-lisp--module-at-point))
-      ;; FIXME: this is probably a bug in `counsel'. See
-      ;; https://github.com/abo-abo/swiper/pull/2752.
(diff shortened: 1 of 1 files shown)


nbfalcon:
#4415.

hlissner (approved):

nbfalcon:
Note that we could also just remove the `'deferred` return value. In this case, opening files using `C-SPC` would open buffers that would disappear after finishing the `goto-definition`, with only the selected file remaining.

hlissner:
> Note that we could also remove the `'deferred` return value. In this case, opening files using `C-SPC` would open buffers that would disappear after finishing the `goto-definition`, with only the selected file remaining. Do you think that might be better?

Sorry for the tremendously later reply, but this does sound better. Please do.

nbfalcon:
@hlissner done. C-SPC will now open the file in the foreground temporarily, restoring the old buffer after exiting the counsel-find-file dialog with "ESC". Exiting with RET keeps the selected file in the foreground. I have tested this using RET and C-SPC.

hlissner (approved):

hlissner:
Thanks for your help!

PR #4437, opened 2020-12-15
Fix eager macro expansion failed `hydra-ivy'

nbfalcon:
With the `ivy` layer enabled, `M-x doom/reload` -> causes message `eager macro expansion failed, hydra-ivy`.

### DIFF
--- a/modules/completion/ivy/config.el
+++ b/modules/completion/ivy/config.el
@@ -100,7 +100,8 @@ results buffer.")
     ;; Ensure `ivy-dispatching-done' and `hydra-ivy/body' hydras can be
     ;; exited / toggled by the same key binding they were opened
     (add-to-list 'ivy-dispatching-done-hydra-exit-keys '("C-o" nil))
-    (defhydra+ hydra-ivy () ("M-o" nil)))
+    ;; Prevent eager macro expansion warnings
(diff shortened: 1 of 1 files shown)


AloisJanicek:
Thank You for fixing my mistake!

nbfalcon:
If there is some better way to prevent eager macro expansion, that would be great. I don't really like `eval` here, but `cl-eval-when 'load` doesn't seem to work.

nbfalcon:
@hlissner why is the `if` (then) branch with `featurep` in [`after!`](https://github.com/hlissner/doom-emacs/blob/0c9256411d6bdb6cfc920bcce2216d99ff28a0c5/core/core-lib.el#L338) necessary? It causes problems like this.

syzygyzer:
Confirm, this fixed the bug for me.

hlissner:
> @hlissner why is the `if` (then) branch with `featurep` in [`after!`](https://github.com/hlissner/doom-emacs/blob/0c9256411d6bdb6cfc920bcce2216d99ff28a0c5/core/core-lib.el#L338) necessary? It causes problems like this.

Sorry for the tremendously late reply: that `if` was removed in b8b51cee7b6d. Does that fix this issue?

nbfalcon:
b8b51ce indeed fixes this.

PR #4438, opened 2020-12-15
Add workspace support to rss/elfeed

d1egoaz:
I basically looked at the calendar app and copy/pasted the code into the rss app.

I tested it on my machine and it worked.


hlissner (approved):

hlissner:

PR #4442, opened 2020-12-17
feat: run golang tests when build tags are set

MannySchneck:
---
name: General Contribution
about: A general code or documentation PR
---

Golang tests with build tags at the top of the file will only run if the `go test` command is invoked with the corresponding tag (`go test -tags=...`). This is commonly used to avoid running integration tests on every test run. 

This means that `+go/test-single` does not work when trying to run a test in such a file. 

This PR does some dirty-regex hacking to add the correct flag to the go-test-single args. I'm currently using it at work 🙂 .

### DIFF
--- a/modules/lang/go/autoload.el
+++ b/modules/lang/go/autoload.el
@@ -32,13 +32,29 @@
   (interactive)
   (+go--run-tests "./..."))
 
+
+;;;###autoload
+(defun +go--extract-test-tag ()
+  (save-excursion
+    (if-let (tag (progn
+                   (re-search-backward "^//[ ]*\\+build[ ]*\\([[:alnum:]]+\\)"
+                                       nil
+                                       t)
+                   (match-string-no-properties 1)))
+        tag
+      "")))
+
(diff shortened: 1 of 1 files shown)


modules/lang/go/autoload.el
```diff
@@ -32,13 +32,29 @@
   (interactive)
   (+go--run-tests "./..."))
 
+
+;;;###autoload
+(defun +go--extract-test-tag ()
+  (save-excursion
+    (if-let (tag (progn
```
cbio-mannyschneck, line 39: TODO: experiment with `match-string-no-properties`. If it returns empty string instead of nil (confiigurable?) I can drop the if-let.

hlissner:
Any progress on this PR?

MannySchneck:
The LSP lag when using gopls drove me to neovim. I'm not an emacs user anymore 😢, but I'll always remember Doom fondly. 

To anyone who comes across this PR, setting GOTAGS in your environment will work as well.

PR #4444, opened 2020-12-17
Add find replace all keybindings

ericdallo:
For an easy find in project and replace I use `projectile-replace` which works very nice, I noticed there's no keybinding for this command in doom, this is a suggestion.

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -586,6 +586,8 @@
        :desc "Switch project"               "p" #'projectile-switch-project
        :desc "Find recent project files"    "r" #'projectile-recentf
        :desc "Run project"                  "R" #'projectile-run-project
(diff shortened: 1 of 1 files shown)


hlissner:
Hi! Thanks for the PR, but I cannot accept it. There is no keybind for it, because the canonical way to search and replace in doom is:

1. `SPC s p`.
2. Narrow your search.
3. `C-c C-e` to generate a wgrep (writeable-grep) buffer from the results.
4. Modify the results as you would any regular buffer.
5. `C-c C-c` to commit your changes to all files.

I cannot endorse `projectile-*` commands in general, because I intend to eventually remove projectile and because there are often better commands available (usually provided by a `:completion` module; i.e. helm or ivy).

In any case, thanks for the help.

ericdallo:
It worked nice indeed, I didn't know about this, thank you!

PR #4448, opened 2020-12-17
Add `--exclude .git` to fd in `projectile-generic-command`

hoyon:
I have `projectile-index-method` set to `alien` as using `hybrid` is too slow in my project of around 3800 files when cache is disabled (I disable cache since I switch branches often and was constantly deleting my cache).

When using `alien` `.git` files are being included which would otherwise be filtered by `'hybrid` so this PR passes a flag to exclude `.git` to `fd`.

Note that this is already being done in the `rg` fallback below.

### DIFF
--- a/core/core-projects.el
+++ b/core/core-projects.el
@@ -179,7 +179,7 @@ And if it's a function, evaluate it."
                          (cl-find-if (doom-rpartial #'executable-find t)
                                      (list "fdfind" "fd"))
                        doom-projectile-fd-binary))
-              (concat (format "%s . -0 -H --color=never --type file --type symlink --follow"
+              (concat (format "%s . -0 -H --color=never --type file --type symlink --follow --exclude .git"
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
> I disable cache since I switch branches often and was constantly deleting my cache

It's odd that you have to do this. It indicates that [this line](https://github.com/hlissner/doom-emacs/blob/develop/modules/tools/magit/config.el#L38) isn't working, or are you not using magit? Perhaps a focus-in hook to check if the branch has changed is necessary.

In any case, I'll merge this PR soon. Thanks for the help!

hoyon:
I don't use magit for a lot of my git commands but might have to actually learn it properly now. 

Thanks!

hlissner:
Sorry for the long wait! Thanks again!

PR #4456, opened 2020-12-20
Fix interactive call of doom/bump-package

RBckmnn:
The function expects 'package' to be a symbol instead of a string.


hlissner (approved):

hlissner:
Good catch! Thanks!

PR #4458, opened 2020-12-20
Fix CIDER debug keybindings in evil-mode

heyarne:
See https://github.com/syl20bnr/spacemacs/issues/2356 for further information,


hlissner:
Are you sure this is necessary when [evil-collection-cider](https://github.com/emacs-evil/evil-collection/blob/master/modes/cider/evil-collection-cider.el#L92-L109) is doing much of this work already?

And if that isn't enough, perhaps this should be PRed to evil-collection instead?

PR #4461, opened 2020-12-22
Add cider pprint to local map.

slackorama:
Add  `cider-pprint-*` defuns to `<localleader> p`.

Also thought about adding these to `<localleader> e p` since they are eval and pprints.


hlissner (approved):

hlissner:
Thanks!

slackorama:
Thanks! Sorry about that dangling paren.

PR #4467, opened 2020-12-22
Disable scheme-mode for .rkt files if :lang racket

iwahbe:
With `:lang racket` enabled, `scheme-mode` still pops up on `.rkt` files.


hlissner (approved):

hlissner:
Thanks!

PR #4472, opened 2020-12-24
Fix typo for +spell/add-word binding in documentation

smallwat3r:
For evil users, it seems the default binding for `+spell/add-word` is `zq`, not `zg`.


smallwat3r:
After double-checking, it seems it's correctly bind to `zg` under `modules/editor/evil/config.el`
 https://github.com/hlissner/doom-emacs/blob/0c9256411d6bdb6cfc920bcce2216d99ff28a0c5/modules/editor/evil/config.el#L428

But for some reason, on my personal doom config, `+spell/add-word` is bind to `zq`, instead of `zg`. I've checked, but I haven't overwritten the binding anywhere.
Screenshot 2020-12-24 at 02 03 41

It might be an issue with my personal config. I will close this PR as I suppose the primary intent is to use `zg` in evil mode (as in vim).

PR #4485, opened 2020-12-26
config: move format binding behind conditional

asymmetric:
We don't want the binding if the function is not defined.

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -362,7 +362,8 @@
        :desc "Jump to references"                    "D"   #'+lookup/references
        :desc "Evaluate buffer/region"                "e"   #'+eval/buffer-or-region
        :desc "Evaluate & replace region"             "E"   #'+eval:replace-region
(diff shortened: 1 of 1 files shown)


ymarco:
I think this is a duplicate of #3682, see Henrik's response:
> Hi! Thanks for the PR but I cannot accept it. This is a duplicate of #3335, #2430, #2390, #2320, and #2267. See #2267 (comment) for an explanation.

asymmetric:
@hlissner would it make sense instead to add a `format +lsp` flag that binds to `lsp-format-buffer`?

hlissner:
> @hlissner would it make sense instead to add a `format +lsp` flag and bind here to a more generic function that then dispatches to either `lsp-format-buffer` or `+format/region-or-buffer`?

Why is that necessary? The `+format-with-lsp` variable, if non-nil, should already accomplish this. 

I'm not keen on using a module because this setting should be controllable at the per-buffer/major mode level, because LSP formatters are not universally desirable.

asymmetric:
Gotcha. Guess this can be closed then.

PR #4487, opened 2020-12-27
[doc] Fix minor grammar mistake

bhipple:


hlissner:
Hi! Thanks for the PR but I'm afraid I must turn it down. We're transitioning to Discourse later this week, and our FAQ along with it, so `faq.org` will be deleted soon, so I'm not accepting PRs to the docs for the time being. Thanks for the help in any case!

PR #4490, opened 2020-12-27
faq: document gating bindings in :config default

asymmetric:
Closes #4485, based on the comment in #2267.


hlissner:
Hi! Thanks for the PR but I'm afraid I must turn it down. We're transitioning to Discourse later this week, and our FAQ along with it (`faq.org` will be deleted soon). Therefore I'm not accepting PRs to the docs for the time being. Thanks for the help in any case!

PR #4491, opened 2020-12-29
Bump ivy-prescient in :completion/ivy

t-e-r-m:
The previous pin broke regex. Using anything other than $ would make the prompt hang. Looks like this pin fixes it.

Not sure if anyone else can reproduce this, though. I'm using native-comp on Debian Stable.


hlissner:
Hi! The rest of the module was overdue a bump and bump commits must adhere to a strict format, so I've bumped it in 4e2a473. Thanks for bringing this to my attention.

PR #4492, opened 2020-12-29
Fix `+eshell/split-window' error

nbfalcon:
`+eshell/split-window` eventually calls `+eshell--bury-buffer`, which
invoked `+eshell/here` with `t` instead of a command string, erroring.


hlissner (approved):

hlissner:
Good catch! Thanks!

PR #4493, opened 2020-12-30
Don't hide the menu bar by default.

### LINKED ISSUE #3926: Consider not disabling menu-bar-mode by default
Hi all!

As some of you might be aware, there's discussion regarding making Emacs more approachable to newcomers, over on emacs-devel. Distributions like Doom Emacs are part of the discussion, and it's a common conclusion that projects like yours are very helpful in this regard and there are ideas about collaboration and support.

As part of those discussions, trying to spot what some common patterns might be hiding some otherwise more discoverable functionality in Emacs, we came to the conclusion that the menu bar is important as a tool for discovering functionality, both a fixed set, and minor/major mode related stuff.

This is of course not a catch-all fix to the problems of the newcomers, and a lot of ideas are being produced around how to solve other, more major problems, but not disabling menu-bar-mode by default seems to be a step in the right direction.

So we thought that recommending distributions that seem to disable it by default to consider leaving that to the user instead. If you want to read more on the discussions, or contribute your views, please don't hesitate to do so on the emacs-devel mailing list.

tmtvl:
Keep menu-bar-lines at the default to not hide the menu bar.
Fixes #3926

### DIFF
--- a/core/core-ui.el
+++ b/core/core-ui.el
@@ -297,12 +297,11 @@ windows, switch to `doom-fallback-buffer'. Otherwise, delegate to original
 (setq window-resize-pixelwise t
       frame-resize-pixelwise t)
 
-(unless (assq 'menu-bar-lines default-frame-alist)
+(unless (assq 'tool-bar-lines default-frame-alist)
   ;; We do this in early-init.el too, but in case the user is on Emacs 26 we do
   ;; it here too: disable tool and scrollbars, as Doom encourages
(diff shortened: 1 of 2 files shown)


hlissner:
Hi! Thanks for the PR, but I'm afraid I must refuse it. [I explain why in #3926](https://github.com/hlissner/doom-emacs/issues/3926#issuecomment-753679968).

PR #4503, opened 2021-01-04
Emacs-lisp eldoc: fix error if symbol unbound

nbfalcon:
`+emacs-lisp-append-value-to-eldoc-a` `error`s if the symbol isn't
bound, because `symbol-value` errors in that case. Fix that bug by only
modifying the result of `elisp-get-var-docstring` if the symbol is
bound.


nbfalcon:
To reproduce (with `debug-on-error`):
```emacs-lisp
(defun q () u|nbound-variable) ; | = cursor here
```

hlissner (approved):

hlissner:
Looks good to me. Thanks!

PR #4508, opened 2021-01-04
Make `flyspell' less annoying in commit messages

nbfalcon:
A detailed description is in the `defadvice!` itself. The TLDR is that `flyspell-buffer` sometimes calls `flyspell-small-region`, which calls `sit-for` after checking every word, causing a redisplay. This causes insertions at the wrong place when typing e.g. a commit message, which is annoying (-> `undo`).

### DIFF
--- a/modules/checkers/spell/config.el
+++ b/modules/checkers/spell/config.el
@@ -192,6 +192,22 @@ e.g. proselint and langtool."
                  (featurep 'langtool))
              (setq-local flyspell-mark-duplications-flag nil))))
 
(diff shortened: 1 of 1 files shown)


hlissner:
Sorry for the late response, but is this still an issue? I am unable to reproduce it in vanilla Doom. Can you [reproduce it in the sandbox](https://github.com/hlissner/doom-emacs/blob/develop/docs/getting_started.org#testing-in-dooms-sandbox)?

PR #4512, opened 2021-01-07
Include snippets in lsp completion

jsravn:
This gives completion of snippets along with normal LSP completion,
similar to how the old company-lsp backend worked.

image


hlissner (approved):

hlissner:
Thanks!

PR #4513, opened 2021-01-07
Bump :core projects

gagbo:
bbatsov/projectile@d1cfad0 -> bbatsov/projectile@c31bd41

Fixes `(wrong-number-of-arguments (3 . 4) 2)` on startup for `trunk` users (context : https://github.com/bbatsov/projectile/issues/1635 )


hlissner:
I already had a larger bump commit in store, so I merged your PR into it. In any case, thanks for bringing it to my attention.

PR #4521, opened 2021-01-15
Add org-agenda-priority keybindings to org config

jtmoulia:
Summary: This diff adds evil-esque keybindings for the org-agenda-priority
functions. These functions allow the org subtree priority to be set from
the agenda view.

Rationale: I love the existing `org-agenda-*` keybindings, but I find myself missing `org-agenda-priority` when doing exactly what you'd guess: trying to set an item's priority while in the agenda. This adds that keybinding in, along with a prefix menu for all of the `org-agenda-priority-*` functions.

I didn't see existing keybinding documentation for org-mode so I didn't add any myself, but let me know what is necessary to get this PR up to spec.

### DIFF
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -684,7 +684,6 @@ between the two."
         "I" #'org-toggle-inline-images
         "n" #'org-store-link
         "o" #'org-set-property
-        "p" #'org-priority
         "q" #'org-set-tags-command
         "t" #'org-todo
         "T" #'org-todo-list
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Sorry for the late response. This PR made be realize that `SPC m p` was incorrectly bound. I've appended a fix for that in this PR and mirrored those keybinds to org-agenda-mode-map. Thanks for the help!

PR #4523, opened 2021-01-16
markerp is an undefined var?

blithefeng:


hlissner:
This code is incorrect and will cause an error. `markerp` is a built-in Emacs function.

PR #4529, opened 2021-01-17
Bumping lsp-haskell pinned version

wraithm:
This new version of lsp-haskell works. I tested on my company's project.


hlissner:
Sorry for the delay. A much larger effort to bump all our `:lang` module was already underway. What's more, bump commits must follow strict formatting so they remain machine readable for future CI. In any case, lsp-haskell was bumped in a79275a, which should resolve your issue. Thanks for bringing it to my attention!

PR #4533, opened 2021-01-17
Fix various broken links

danlamanna:
There were a few other links below that I was unsure of how to handle, but I think this is a strict improvement.

```
http://www.culater.net/software/SIMBL/SIMBL.php
https://github.com/hlissner/dotfiles/tree/master/shell/mu
https://github.com/maskray/ccls/blob/master/src/symbol.h
https://gitweb.gentoo.org/repo/gentoo.git/tree/net-mail/mu/files/70mu-gentoo.el
```


hlissner (approved):

hlissner:
Thanks for your help!

PR #4538, opened 2021-01-18
Fix crash on alist in emulation-mode-map-alists

jstranik:
Emulation-Mode-Map-Alists can contain alists as well as symbols.
If it is a symbol, dont crash.


hlissner:
A better solution was needed and pushed in 03929a2, which should resolve this issue. Thanks for bringing it to my attention!

PR #4540, opened 2021-01-19
Bump org-journal

mpsq:
## What

Bump org-journal by 2 commits:

[Here is the diff](https://github.com/bastibe/org-journal/compare/c0836483ae43e525bf7547b7a789d171eda84c84...08d5fce95023c015372678d353388ad0dae8952b).

## Why

For the adventurous souls using emacs devel, a [change](https://github.com/emacs-mirror/emacs/commit/32c6732d16385f242b1109517f25e9aefd6caa5c) introduced recently broke some libraries. Org-journal people already fixed the issue, hence why it needs to be bumped.

More details on [Reddit](https://www.reddit.com/r/emacs/comments/kqb9s9/cannot_recompile_packagess_error_wrong_number_of/).

### DIFF
--- a/modules/lang/org/packages.el
+++ b/modules/lang/org/packages.el
@@ -54,7 +54,7 @@
 (when (featurep! +jupyter)
   (package! jupyter :pin "360cae2c70ab28c7a7848c0c56473d984f0243e5"))
 (when (featurep! +journal)
-  (package! org-journal :pin "c0836483ae43e525bf7547b7a789d171eda84c84"))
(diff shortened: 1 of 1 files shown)


hlissner:
Sorry for the delay. A much larger effort to bump the `:lang org` module (and other `:lang` modules) was already underway. What's more, bump commits must follow strict formatting so they remain machine readable for future CI. In any case, `org-journal` was bumped in f753871, which should resolve your issue. Thanks for bringing it to my attention!

PR #4546, opened 2021-01-20
lang/terraform: Set docsets for terraform-mode

sei40kr:
## Description

Set docsets for terraform-mode.

### DIFF
--- a/modules/tools/terraform/config.el
+++ b/modules/tools/terraform/config.el
@@ -4,12 +4,15 @@
   (add-hook 'terraform-mode-local-vars-hook #'lsp!))
 
 
-(map! :after terraform-mode
-      :map terraform-mode-map
-      :localleader
-      :desc "terraform apply" "a" (cmd! (compile "terraform apply" t))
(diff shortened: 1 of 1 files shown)


modules/tools/terraform/config.el
```diff
@@ -4,12 +4,15 @@
   (add-hook 'terraform-mode-local-vars-hook #'lsp!))
 
 
-(map! :after terraform-mode
-      :map terraform-mode-map
-      :localleader
-      :desc "terraform apply" "a" (cmd! (compile "terraform apply" t))
-      :desc "terraform init"  "i" (cmd! (compile "terraform init"))
-      :desc "terraform plan"  "p" (cmd! (compile "terraform plan")))
+(use-package! terraform-mode
```
hlissner, line 7: The package is no longer lazy loaded. Please use `after!` instead:

```elisp
(after! terraform-mode
  (set-docsets! ...)
  (map! ...))
```

hlissner (changes requested):

hlissner (approved):

PR #4556, opened 2021-01-22
Bump org-mode

anonimitoraf:
emacs-straight/org-mode@3916a5 <- emacs-straight/org-mode@6b83c6

Known to have the fix for https://github.com/hlissner/doom-emacs/issues/4534


hlissner:
Sorry for the delay. A much larger effort to bump the `:lang org` module (and other `:lang` modules) was already underway, and this bump quickly became obsolete in the meantime. As of f753871 `org-mode` was bumped, which should resolve your issue. Thanks for bringing it to my attention!

PR #4566, opened 2021-01-24
smartparens: more electricity for C-style comments

nbfalcon:
- `/*` = `/* | */`
- `/*[*!]` = `/*[*!] | */`
- `/* C-6 *` = 6 stars

Typing `/**/` now yields `/* | */`, saving the user a single space keypress.
Then typing * or ! puts them at the start of the comment, i.e. /\*! =
/**! | \*/, with /* | */ as intermediate step (somewhat like
`lispy-comment`).

Expand JavaDoc and Doxygen Qt-style comments (/** \*/ and /*\*! \*/
respectively) automatically if the user invokes `newline-and-indent`
within such a comment: /** | \*/ + RET -> /**\n * |\n */. This is only
done for JavaDoc comments that are not yet expanded, such comments being
defined as spanning not more than one line.

`+default--newline-and-indent` now respects the prefix argument: /* C-6
RET now correctly yields 6 stars. This also works with JavaDoc comment
expansion.

Drop the `smartparens` pairs for /** and /\*!, since they are now handled
by /*. Drop `+default-open-doc-comments-block`, since it is unused.

----

#

### DIFF
--- a/modules/config/default/config.el
+++ b/modules/config/default/config.el
@@ -164,7 +164,7 @@
       ;; Needed so downstream `advice'
       ;; (`+default--newline-indent-and-continue-comments-a') can receive the
       ;; prefix ARG for some reaosn.
-      (interactive "*P")
+      (interactive "*p")
       ;; `javascript-mode' uses `c-do-auto-fill'. For the other cases,
       ;; `c-literal-limits' only works in cc buffers, erroring otherwise.
       (when (and (or c-buffer-is-cc-mode (derived-mode-p 'javascript-mode 'js2-mode)))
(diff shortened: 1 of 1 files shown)


nbfalcon:
As a side note, why does DOOM internally use `defvar` over `defcustom` in many places (e.g. `+default-want-RET-continue-comments`)? `counsel-projectile-rg`ing the codebase, I only found two uses of `defcustom`.

hlissner:
> As a side note, why does DOOM internally use `defvar` over `defcustom` in many places (e.g. `+default-want-RET-continue-comments`)? `counsel-projectile-rg`ing the codebase, I only found two uses of `defcustom`.

Because Doom doesn't use nor endorse the `customize` interface, so they're simply variables. The `defcustom`s in the modeline module are there because I intend for it to be an independent package one day.

In any case, sorry for my late review of this PR. My initial testing shows quite a few issues with doc comment expansion in PHP and JS. Our smartparens config is unfortunately delicate, so this will need further testing. Unfortunately I'm terribly strapped for time, so it may be some time before I get around to this.

nbfalcon:
Thank you for reviewing it anyway! This PR is not ready yet though, there are still some issues to be solved. I will look into the PHP/JS issues. Do you get errors in `js`? There is a bug with `c-literal-limits` cache not being cleared on `js2-parse-hook`, which I'll need to make a PR for.

Take your time.

PR #4567, opened 2021-01-24
Enable highlight of current line in org agenda

RBckmnn:
This did not work previously since `org-agenda-mode` does not derive from any of the mode in  `global-hl-line-modes`.


hlissner:
I moved this to core for now, since more work would be needed to load-check `hl-line`. In any case, thanks for bringing it to my attention!

PR #4571, opened 2021-01-25
company: support `prescient`

nbfalcon:
As with the `ivy` module, `+prescient` enables `prescient`-based sorting for
`company` using `company-prescient`.

----

#


hlissner:
Thanks for the suggestion but I must turn it down. The module used to use company-prescient, but I removed it because it introduced too many edge cases across too wide a range of backends. I'll leave it to users to set it up themselves.

nbfalcon:
Can you name an example of some problems you encountered? Perhaps they can be fixed.

PR #4575, opened 2021-01-27
Fix typo

yangsheng6810:
`want-byte` and `want-native` was renamed `want-byte-compile` and `want-native-compile` in commit cfb8a866dc6181889b0c056abf4fdd3f34fb144b, but one `want-native` was not updated, leading to void-variable error while `doom sync`.


hlissner (approved):

hlissner:
Thanks!

PR #4576, opened 2021-01-27
Bump inflections to get extra argument to define-obsolete-function-alias

cdegroot:
Similar to https://github.com/hlissner/doom-emacs/issues/4534. I could build doom with the current Emacs 28.0.50 native-comp branch after this without problems.


hlissner:
Sorry for the delay. A much larger effort to bump `:lang ruby` (and the rest of our `:lang` modules) was already underway. What's more, bump commits must follow strict formatting so they remain machine readable for future CI. In any case, inflections was bumped in a79275a, which should resolve your issue. Thanks for bringing it to my attention!

PR #4579, opened 2021-01-27
Fix lsp-rename doesn't work with ls-ts

aburtsev:
Fix bug described [here](https://github.com/emacs-lsp/lsp-mode/issues/2548)


hlissner:
Sorry for the late response. A larger effort to bump our LSP module was in the works at the time this was opened, and was completed in 45b68e46e and, later, a9da1fcde. Additionally, bump commits must follow strict formatting to remain machine readable for future CI (see the linked commits for examples). In any case, this and the original issue should be resolved. Thanks for bringing it to my attention!

PR #4582, opened 2021-01-28
Fix 'readme' button doom/help-packages

MageJohn:
The button was trying to open `path/to/README.org/README.org`.


hlissner (approved):

hlissner:
Good catch, thanks!

PR #4583, opened 2021-01-29
Fix for "Discover projects in folder"

smook1980:
Given the following config:
`(setq projectile-project-search-path '("~/src/"))`

I ran into a few errors when attempting to discover projects. This is my first contribution, and my Lisp skills are not strong so if there's something I'm missing please let me know.


hlissner (approved):

hlissner:
Good catch, thanks for your help!

PR #4588, opened 2021-01-29
Fix :unless in (doom!) block.

### LINKED ISSUE #4587: The :unless predicate in (doom!) might be broken.
I believe this line should contain `xor` instead of `or`.
https://github.com/hlissner/doom-emacs/blob/56e9d273e1d798df90257aa00f4e1a7b83f4c173/core/core-modules.el#L282

Janfel:
Fixes #4587.


hlissner:
Good catch, thanks.

PR #4593, opened 2021-01-30
Remap `describe-symbol` to `counsel-describe-symbol` .

datool:
"SPC h o" now uses `counsel-describe-symbol` instead of plain `helpful-symbol` (maybe with completion) to get input.

This allows us to fully take the advantage of ivy's superior completion system:

- `[I]: info` and `[d]: definition` hydra heads in `ivy-dispatching-done`

- Quickly describe symbols with "C-M-n/p" (`ivy-next(or previous)-line-and-call` or with the `calling` hydra head on

Additionally the counsel one is prettier when icons are enabled.

Users who have the `+icons` flag with ivy are expected to notice the change visually when invoking "o" in help-map as icons now appear there. But it's an arguably good change anyway since originally "f1 f" and "f1 v" have icons but "f1 o" doesn't.

### DIFF
--- a/modules/completion/ivy/config.el
+++ b/modules/completion/ivy/config.el
@@ -225,8 +225,7 @@ results buffer.")
 
   ;; Integrate with `helpful'
   (setq counsel-describe-function-function #'helpful-callable
-        counsel-describe-variable-function #'helpful-variable
-        counsel-describe-symbol-function #'helpful-symbol)
+        counsel-describe-variable-function #'helpful-variable)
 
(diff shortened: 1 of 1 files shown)


datool:
Excuse me, I didn't know that `helpful-symbol` asks an additional question when the symbol refers to both a command and a variable.
Guess I have to personally stick to the plain old `describe-symbol` instead to use the Counsel version since "C-M-n/p" seems far too useful to me.

hlissner (approved):

hlissner:
Thanks!

PR #4594, opened 2021-01-30
allow opt-out for automatic lsp/tide

dradetsky:
In the past, when editing many, many projects simultaneously, I've noticed that I've accumulated dozens of running tideserver processes, which were starting to consume nontrivial amounts of memory. Given that I don't often take advantage of tide, I prefer to start it only if I want it.

This change adds a `-lsp-or-tide` flag to disable automatically starting the langserver.

I noted that you didn't take #929. However, this version preserves the default behavior. It's opt-out, not opt-in.

Arguably `-lsp-or-tide` is a bad name. I just kinda copied the name from the hook I was changing. Possible better names include:

- `-auto-lsp-or-tide`
- `-auto-langserver`

### DIFF
--- a/modules/lang/javascript/README.org
+++ b/modules/lang/javascript/README.org
@@ -31,6 +31,8 @@ This module adds JavaScript and TypeScript support.
 + =+lsp= Enables LangServer support for this module. You must have =:tools lsp=
   enabled for this to work, as well as the langserver (e.g.
   typescript-language-server) installed on your system.
++ =-lsp-or-tide= Disables auto-starting LangServer. Useful if you are editing
+  many projects simultaneously
 
 ** Packages
 + [[https://github.com/defunkt/coffee-mode][coffee-mode]]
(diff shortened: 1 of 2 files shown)


hlissner:
Thanks for the PR, but I must refuse it, for the second reason I mentioned in #929.

> I want to avoid adding flags for individual packages, unless absolutely necessary. Flags are meant to represent features (groups of packages or non-trivial configuration), tool integration, or exceptionally popular defaults (like the +auto flag for completion/company). The idiomatic way to disable individual packages in Doom is (package! tide :disable t) in your ~/.doom.d/packages.el.

```elisp
;; add this to ~/.doom.d/packages.el
(package! tide :disable t)
```

`+javascript-init-lsp-or-tide-maybe-h` is designed to bow out if the package isn't present.

That said, I will likely be removing tide entirely from Doom within the coming months, so the js module can focus on LSP support, instead.

dradetsky:
note to self: reading is fundamental

PR #4599, opened 2021-01-31
Doom Doctor: Fix insteadOf check

MithicSpirit:
The warning would always trigger when everything was fine but never when there was an issue.


hlissner (approved):

hlissner:
Whoops. Good catch, thanks.

PR #4604, opened 2021-02-02
Fix interactive doom/retab.

Janfel:
Both `+evil:retab` and `doom/retab` are currently broken when used interactively. This PR fixes these issues and makes both commands behave like intended.


hlissner (approved):

hlissner:
Thanks!

PR #4606, opened 2021-02-02
Extract program name from inferior-lisp-program

justjoheinz:
Many users of sly also use roswell and set the inferior-lisp-program to
values like "ros run" or "ros -Q run". This is not detected correctly by
executable-find. Hence we try to extract the first part of the program
name.


hlissner (approved):

hlissner:

PR #4610, opened 2021-02-04
Fix find file in private config for emacs bindings

chuxubank:
Use default `doom/find-file-in-private-config` just like evil binding,
Or is it intended to?


hlissner (approved):

hlissner:
Good catch, thanks!

PR #4622, opened 2021-02-06
Move setq forms of recentf variables to :init

daanturo:
Move these variables to :init block so that users don't need to resort to `after!` or `with-eval-after-load` to set them.

### DIFF
--- a/core/core-editor.el
+++ b/core/core-editor.el
@@ -235,6 +235,11 @@ or file path may exist now."
   :defer-incrementally easymenu tree-widget timer
   :hook (doom-first-file . recentf-mode)
   :commands recentf-open-files
+  :init
+  (setq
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but I'm afraid I cannot accept it. The canonical way to configure packages in Doom is with `after!`, because (almost) all of them are lazy loaded. This guarantees that your settings will stick, now or in the future, if/when I change that package's defaults. The only exception to this rule are file/directory variables that are likely candidates for reconfiguration (like `org-directory` or `org-roam-directory` -- variables like `recentf-save-file` do not qualify because it's used for internal persistence storage; there's little reason for it to be changed directly).

Accepting this PR would add ambiguity to this contract and encourage folks not to use `after!`. In any case, thanks for bringing it to my attention.

PR #4627, opened 2021-02-07
[feature] Better CIDER debugging

cfeckardt:
The CIDER debugging experience needs a little bit of improvement.
This PR fixes some issues, in particular:

* Maps cider-debug-at-point to "localleader d d" - so that you can easily drop into debug!
* Overrides evil keybindings to not interfere with the cider--debug-mode
  bindings during the debug session - this overrides keys like 'j' and 'n' to allow you to 'inject expression' or 'step over'.

Thanks for a great editor Henrik!

### DIFF
--- a/modules/lang/clojure/config.el
+++ b/modules/lang/clojure/config.el
@@ -86,6 +86,14 @@
            (with-current-buffer nrepl-server-buffer
              (buffer-string)))))))
 
+  ;; When in cider-debug-mode, override evil keys to not interfere with debug keys
+  (after! evil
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
I think this should be introduced upstream to `evil-collection-cider`, rather than kept here, but I'll keep it in our module for the time being. Thanks for the help!

cfeckardt:
@hlissner okay, i'll prepare a PR upstream too.

PR #4630, opened 2021-02-07
New module, :app everywhere

tecosaur:
No need to leave Emacs any more.

Module discussed on discord. See the `README.org` for more information on this.

### DIFF
--- a/modules/app/everywhere/config.el
+++ b/modules/app/everywhere/config.el
@@ -1,0 +1,30 @@
+;;; app/everywhere/config.el -*- lexical-binding: t; -*-
+
+(use-package! emacs-everywhere
+  :defer ; entry point is autoloaded
+  :config
+  (after! doom-modeline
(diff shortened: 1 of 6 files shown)


modules/app/everywhere/config.el
```diff
@@ -0,0 +1,29 @@
+;;; app/everywhere/config.el -*- lexical-binding: t; -*-
+
+(use-package! emacs-everywhere
+  :config
```
ymarco, line 4: This would eager-load emacs-everywhere

tecosaur: Hmmm, `:defer` would probably be a good idea wouldn't it.

tecosaur: Added.

hlissner (approved):

hlissner:
Great work! Thanks again!

PR #4631, opened 2021-02-07
Enabling user config override for org-id-locations-file

samwhitlock:
The use case: if a user has their org directory stored in a synced
location (e.g. dropbox), this file is the same on all machines and is
almost always conflicted.

Given that it is just a cache on the org file locations for IDs, this enables users to move this file out of the directory.

This change enables users to override this in the :init section of their
config (or otherwise before org loading) and that the
+org-init-org-directory-h hook will detect an override.


hlissner (approved):

hlissner:
Makes sense to me. Thanks!

PR #4637, opened 2021-02-10
[doc: modules/lang/cc] Readability improvements

ghost-420:
- _sudo_ should not be assumed, as _[doas](https://github.com/Duncaen/OpenDoas)_ is an optimal and more secure alternative for workstation use cases (sudo is suited for servers, where automation is important).
- ~~It's easier to tell if a line is a command by having $ (non-escalated) or # (escalated/root) before the command, as Arch Linux's wiki does by default.~~
This causes problems if you want to execute the command(s) directly in Emacs, see: \
http://howardism.org/Technical/Emacs/literate-devops.html (visual examples) \
https://orgmode.org/manual/Working-with-Source-Code.html

### DIFF
--- a/modules/lang/cc/README.org
+++ b/modules/lang/cc/README.org
@@ -95,13 +95,15 @@ After installing its dependencies (Clang and CMake), run ~M-x irony-install-serv
 *** macOS
 Due to linking issues, macOS users must compile irony-server manually:
 
-#+BEGIN_SRC sh
-$ brew install cmake
-$ brew install llvm
+~$ brew install cmake~ \\
+~$ brew install llvm~
 
-$ git clone https://github.com/Sarcasm/irony-mode irony-mode
-$ mkdir irony-mode/server/build
-$ pushd irony-mode/server/build
+#+BEGIN_SRC sh
+#!/bin/sh
+
+git clone https://github.com/Sarcasm/irony-mode irony-mode
+mkdir irony-mode/server/build
+pushd irony-mode/server/build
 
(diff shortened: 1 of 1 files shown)


hlissner:
> sudo should not be assumed, as doas is an optimal and more secure alternative for workstation use cases (sudo is suited for servers, where automation is important).

Agreed.

> ~~It's easier to tell if a line is a command by having $ (non-escalated) or # (escalated/root) before the command, as Arch Linux's wiki does by default.~~
> See: http://howardism.org/Technical/Emacs/literate-devops.html

I welcome a convention, but not keen on `#` because it can be misinterpreted as a comment and would need to be explained. However, I welcome using `$` everywhere to denote an inline shell command (and simply avoid the subject of privilege altogether, except where absolutely necessary). Not sure what you're pointing to at that link; I see no mention of this convention.

ghost-420:
> > ~It's easier to tell if a line is a command by having $ (non-escalated) or # (escalated/root) before the command, as Arch Linux's wiki does by default.~
> > See: http://howardism.org/Technical/Emacs/literate-devops.html
>
> Not sure what you're pointing to at that link; I see no mention of this convention.


The convention breaks being able to execute commands inside Emacs the way that webpage is showcasing:
`sh: line 1 : $: command not found`
It's Org mode's code block evaluation. I learned about it last night and I'm liking it a lot.
codeblock_output
codeblock_output_raw

hlissner (approved):

hlissner:
Thanks!

PR #4642, opened 2021-02-11
Correctly gate on deft

iwahbe:
Does not provide a key binding for deft when it is not installed. Also ignores .eln files.


hlissner:
Duplicate of #4490, #4485, #3682, #3335, #2430, #2390, #2320, and #2267.

See https://github.com/hlissner/doom-emacs/pull/2267#issuecomment-569536183.

PR #4650, opened 2021-02-13
snippets.el: Prompt should say `edit` and not `alias`.

slackorama:
Defun is `+snippets/edit` so the prompt to user should say `edit`.


hlissner (approved):

hlissner:
Thanks

PR #4667, opened 2021-02-17
Remove the custom org-priority-regexp

kuba-orlik:
This regular expression broke sorting by priority in agenda view. Reverting it back to the original one fixes the issue

### DIFF
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -118,11 +118,7 @@ Is relative to `org-directory', unless it is absolute. Is used in Doom's default
           (?C . success))
         org-startup-indented t
         org-tags-column 0
-        org-use-sub-superscripts '{}
-        ;; HACK Speed up regexp for priority faces by making it a little less
-        ;;      greedy than the default.
-        ;; REVIEW May be upstreamed at some point. Keep an eye out.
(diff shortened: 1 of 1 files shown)


hlissner:
Does it work if you revert its global value to its default, but add this to your config.el?

```lisp
(setq-hook! 'org-mode-hook org-priority-regexp "^\\*+.*\\(\\[#\\([A-Z0-9]+\\)\\] ?\\)")
```

kuba-orlik:
Unfortunately, no. I've added the above snippet to `config.el`, reloaded doom emacs and opened the agenda view, and items are again not sorted by priority. When inspecting the `org-priority-regexp` variable in agenda view, it's value is `"^\\*+.*\\(\\[#\\([A-Z0-9]+\\)\\] ?\\)"`. 

My current workaround is to set the regexp as custom variable for each of my agenda views, but that's far from perfect

kuba-orlik:
Also, the commit that introduced the custom regexp broke some syntax highlighting in org agenda, causing priority characters not being highlighted

hlissner (approved):

hlissner:
Right, then reverted it is. Odd that no other org user on our discord has mentioned any of these issues, but revisiting this will have to wait until I have more time. Thanks for your help!

PR #4669, opened 2021-02-17
Update docs for nonstandard homebrew installation prefix.

dabrahams:
If one assumes the user has set up their .profile with a definition of
HOMEBREW_PREFIX per the homebrew instructions then we don't
need the initial command to get it from brew config.  
@hlissner I'd be glad of your input on how to proceed with this

Still TODO: update emacs-lisp that assumes standard homebrew installation prefix.

### DIFF
--- a/docs/getting_started.org
+++ b/docs/getting_started.org
@@ -240,18 +240,20 @@ to least recommended for Doom (based on compatibility).
 
 - [[https://github.com/d12frosted/homebrew-emacs-plus][emacs-plus]]:
   #+BEGIN_SRC bash
+  HOMEBREW_PREFIX=$(brew config | grep '^HOMEBREW_PREFIX:' | cut -f 2 -d' ')
   brew tap d12frosted/emacs-plus
   brew install emacs-plus
-  ln -s /usr/local/opt/emacs-plus/Emacs.app /Applications/Emacs.app
+  ln -s "$HOMEBREW_PREFIX/opt/Emacs.app" /Applications/Emacs.app
   #+END_SRC
 
 - [[https://bitbucket.org/mituharu/emacs-mac/overview][emacs-mac]] is another acceptable option. It offers slightly better integration
   with macOS, native emojis and better childframe support. However, at the time
   of writing, it [[https://github.com/railwaycat/homebrew-emacsmacport/issues/52][lacks multi-tty support]] (which impacts daemon usage):
   #+BEGIN_SRC bash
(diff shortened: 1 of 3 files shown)


hlissner:
I don't agree with this change; I believe this introduces unnecessary complexity for beginner readers. More advanced users know to make their own corrections. It also imposes the assumption that all users are using homebrew -- what if they use macports?

dabrahams:
Totally understandable not wanting to add complexity for beginners.  

Couple questions:
1. Suppose we leave out the variable setting line?  Then it's no more complex than using `/usr/local`, and if they've followed standard homebrew install practice as the instructions already assume they do, it will "just work."  It'll just work for more people.
2. Are you also opposed to updating the elisp to account for non-standard installation?

(FWIW, the reason some people don't install in `/usr/local` is that then brew doesn't require sudo and isn't interfered with by system installations.  If the beginner user isn't an administrator they may not even be able to install in `/usr/local`).

hlissner:
Sorry for the late response!

> Suppose we leave out the variable setting line? Then it's no more complex than using /usr/local, and if they've followed standard homebrew install practice as the instructions already assume they do, it will "just work." It'll just work for more people.

I'd argue that removing the variable setter line makes it more ambiguous (to beginners), but perhaps my common sense for what is obvious has been destroyed by the issues I've seen. I feel `/usr/local` is the least ambiguous, and least harmful to misinterpret. The variable could be interpreted as either a) something that the user is responsible for setting, b) something the reader is expected to know, or c) something homebrew should define. These would cause a beginner despair, but not phase the savvier MacOS user, but neither would they be phased by `/usr/local`. I imagine _our_ assumption would be obvious for them to recognize and correct.

Perhaps the simplest option is to reserve a section in the install instructions to discuss non-standard installations on MacOS?

That is, assuming the steps for setting up an unprivileged homebrew install are identical to a system install, and only require a mental substitution of paths as you go.

> Are you also opposed to updating the elisp to account for non-standard installation?

Not at all. I welcome that. Though, to avoid consulting `brew --prefix` at startup, I'd like to bake its value into some persisted variable at `doom sync` time. Also, how would we go about accommodating macports as well? My mac-fu is nearly 6 years out of date.

I see a few hits for /usr/local in doom which I will look into addressing. Are any of them particularly important to you?

hlissner:
Ah, side note: I'm closing this because I'm cleaning up the github for a transition to discourse, but feel free to follow up here.

PR #4671, opened 2021-02-18
Add support for parinfer-rust-mode

### LINKED ISSUE #4635: parinfer-mode has been deprecated
[parinfer-mode](https://github.com/DogLooksGood/parinfer-mode) has been deprecated and will no longer be updated by the author. Consider moving to [parinfer-rust](https://github.com/eraserhd/parinfer-rust) and [parinfer-rust-mode](https://github.com/justinbarclay/parinfer-rust-mode) which is recommended by the author of parinfer-mode.

From the author:

> Deprecated
This project is no longer maintained. I have no success for a fast enough implementation in Emacs Lisp.
Fortunately, there’s a parinfer-rust-mode!

johanthoren:
This commit adds support for the `parinfer` module flag `+rust`, which will use `parinfer-rust-mode`
instead of `parinfer-mode`. `parinfer-rust-mode` makes use of an external library named `parinfer-rust`
to do the heavy lifting. As such, `--with-modules` must be enabled.

Resolves #4635.

### DIFF
--- a/modules/editor/parinfer/config.el
+++ b/modules/editor/parinfer/config.el
@@ -33,17 +33,14 @@
           racket-mode
           hy-mode) . parinfer-rust-mode)
   :init
-  (setq! parinfer-rust-library
-         (concat user-emacs-directory
-                 ".local/etc/parinfer-rust/"
-                 (cond ((eq system-type 'darwin) "parinfer-rust-darwin.so")
-                       ((eq system-type 'gnu/linux) "parinfer-rust-linux.so")
-                       ((eq system-type 'windows-nt) "parinfer-rust-windows.dll")
-                       ((eq system-type 'berkeley-unix) "libparinfer_rust.so")))
-         parinfer-rust-auto-download
-         (if (eq system-type 'berkeley-unix)
-             nil
-           t))
+  (setq parinfer-rust-library
+        (concat user-emacs-directory
+                ".local/etc/parinfer-rust/"
+                (cond ((eq system-type 'darwin) "parinfer-rust-darwin.so")
+                      ((eq system-type 'gnu/linux) "parinfer-rust-linux.so")
+                      ((eq system-type 'windows-nt) "parinfer-rust-windows.dll")
+                      ((eq system-type 'berkeley-unix) "libparinfer_rust.so"))))
+  (setq parinfer-rust-auto-download (if (eq system-type 'berkeley-unix) nil t))
   :config
   (map! :map parinfer-rust-mode-map
         :localleader
--- a/modules/editor/parinfer/config.el
+++ b/modules/editor/parinfer/config.el
@@ -3,11 +3,11 @@
 (use-package! parinfer
   :unless (featurep! +rust)
   :hook ((emacs-lisp-mode
-            clojure-mode
-            scheme-mode
-            lisp-mode
-            racket-mode
-            hy-mode) . parinfer-mode)
+          clojure-mode
+          scheme-mode
+          lisp-mode
+          racket-mode
+          hy-mode) . parinfer-mode)
   :init
   (setq parinfer-extensions
         '(defaults
--- a/modules/editor/parinfer/config.el
+++ b/modules/editor/parinfer/config.el
@@ -1,24 +1,48 @@
 ;;; editor/parinfer/config.el -*- lexical-binding: t; -*-
 
-(use-package! parinfer
-  :hook ((emacs-lisp-mode
-          clojure-mode
-          scheme-mode
-          lisp-mode
-          racket-mode
-          hy-mode) . parinfer-mode)
-  :init
-  (setq parinfer-extensions
-        '(defaults
-          pretty-parens
-          smart-tab
-          smart-yank))
-  (when (featurep! :editor evil +everywhere)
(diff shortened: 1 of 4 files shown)


hackeryarn:
Since `parinfer` is deprecated, as per their repo readme, would it make sense to completely replace `parinfer` with `parinfer-rust-mode`?

It could probably keep the same module name, just use `parinfer-rust-mode` behind the scenes.

hackeryarn:
Sorry just noticed the discussion is taking place in another issue.

modules/editor/parinfer/config.el
```diff
@@ -1,24 +1,48 @@
...
-  (when (featurep! :editor evil +everywhere)
-    (push 'evil parinfer-extensions))
-  :config
-  (map! :map parinfer-mode-map
-        "\"" nil  ; smartparens handles this
-        :i "<tab>"     #'parinfer-smart-tab:dwim-right-or-complete
-        :i "<backtab>" #'parinfer-smart-tab:dwim-left
-        :localleader
-        :desc "Toggle parinfer-mode" "m" #'parinfer-toggle-mode))
+(if (not (featurep! +rust))
```
hlissner, line 3: Rather than wrapping them in an `if` statement, use `use-package`'s `:if` parameter instead. e.g.

```elisp
(use-package! parinfer
  :unless (featurep! +rust)
  ...)

(use-package! parinfer-rust-mode
  :when (featurep! +rust)
  :when (bound-and-true-p module-file-suffix)
  ...)
```

This has some compile-time benefits that aren't seen otherwise.

modules/editor/parinfer/config.el
```diff
@@ -1,24 +1,48 @@
...
+            :i "<tab>"     #'parinfer-smart-tab:dwim-right-or-complete
+            :i "<backtab>" #'parinfer-smart-tab:dwim-left
+            :localleader
+            :desc "Toggle parinfer-mode" "m" #'parinfer-toggle-mode))
+  (defvar parinfer-rust--doom-lib-name (cond ((eq system-type 'darwin)
+                                              "parinfer-rust-darwin.so")
+                                             ((eq system-type 'gnu/linux)
+                                              "parinfer-rust-linux.so")
+                                             ((eq system-type 'windows-nt)
+                                              "parinfer-rust-windows.dll")))
```
hlissner, line 31: Please inline this into the `setq!` statement below.

modules/editor/parinfer/config.el
```diff
@@ -1,24 +1,48 @@
...
+  (use-package! parinfer-rust-mode
+    :hook ((emacs-lisp-mode
+            clojure-mode
+            scheme-mode
+            lisp-mode
+            racket-mode
+            hy-mode) . parinfer-rust-mode)
+    :init
+    (setq! parinfer-rust-library (concat user-emacs-directory
```
hlissner, line 40: `s/setq!/setq`. There is no benefit to triggering setters before the package is loaded (they will be triggered when the package is later loaded, if any).

modules/editor/parinfer/config.el
```diff
@@ -1,24 +1,48 @@
...
+                                         ".local/etc/parinfer-rust/"
+                                         parinfer-rust--doom-lib-name))
+    :config
+    (map! :map parinfer-rust-mode-map
+          :localleader
+          :desc "Switch parinfer-rust-mode" "m" #'parinfer-rust-switch-mode
+          :desc "Temporarily disable parinfer-rust-mode" "M"
```
hlissner, line 47: Please remote these `:desc`s. Which-key doesn't support them on minor modes, and they will end up overwriting the label for these key sequences globally instead.

modules/editor/parinfer/packages.el
```diff
@@ -11,4 +11,6 @@
   ;; separate session:
   (autoload 'evil-define-key "evil-core" nil nil 'macro))
 
-(package! parinfer :pin "8659c99a9475ee34af683fdf8f272728c6bebb3a")
+(if (featurep! +rust)
+    (package! parinfer-rust-mode :pin "c825606e6aca4a2ed18c0af321df5f36a3c8c774")
+  (package! parinfer :pin "8659c99a9475ee34af683fdf8f272728c6bebb3a"))
```
hlissner, line 16: Could you replace `(package! parinfer ...)` with

```elisp
(package! parinfer
  :recipe (:host github :repo "emacsattic/parinfer")
  :pin "8659c99a9475ee34af683fdf8f272728c6bebb3a")
```

The package is no longer in MELPA or emacsmirror, so it won't be found otherwise.

hlissner (changes requested):
Please create a doctor.el file in the module with the following:

```elisp
;;; editor/parinfer/doctor.el -*- lexical-binding: t; -*-

(when (featurep! +rust)
  (unless (fboundp 'module-load)
    (warn! "Your emacs wasn't built with dynamic modules support. `parinfer-rust-mode' won't work")))
```

With that and the commented fixes we should be good to go. In the long term, the `parinfer` package is now deprecated and will remain for a year or two (or until it breaks, whichever comes first) before it is removed.

hlissner (approved):

hlissner:

PR #4675, opened 2021-02-20
+scala/open-repl don't kill non-existant sbt buffer

samspills:
## Why
On current develop trying to open a scala repl in an sbt project via `SPC o r` or calling the `+scala/open-repl` method would fail with
```
if: No buffer named *sbt*<~/project/path>
```
This change updates the check so that we don't attempt to kill an sbt buffer that doesn't exist

## What
- the original method used only `comint-check-proc` before attempting to kill the buffer
  - this method doesn't distinguish between a dead buffer process and a buffer that doesn't actually exist
- update to check that buffer exists and is dead before calling `kill-buffer`

## Questions
1. I don't ever use scala without sbt so I didn't make the same change to line 50 (I'm not sure if its even broken for non-sbt users?). Should I update the logic in both places though?

### DIFF
--- a/modules/lang/scala/autoload.el
+++ b/modules/lang/scala/autoload.el
@@ -41,7 +41,7 @@ Meant to be used for `scala-mode's `comment-line-break-function'."
   (if (and (require 'sbt-mode nil t)
            (sbt:find-root))
       (let ((buffer-name (sbt:buffer-name)))
-        (unless (comint-check-proc buffer-name)
+        (when (and (get-buffer buffer-name) (not (comint-check-proc buffer-name)))
           (kill-buffer buffer-name))
         (run-scala)
         (get-buffer buffer-name))


hlissner (approved):

hlissner (changes requested):
> I don't ever use scala without sbt so I didn't make the same change to line 50 (I'm not sure if its even broken for non-sbt users?). Should I update the logic in both places though?

Please do.

samspills:
@hlissner I just went back and I think the non-sbt scala buffer is actually okay. I saw the `comint-check-proc` call and my brain auto-completed the same logic, but it's actually doing the positive check for a live process and then if there isn't one it starts the scala repl with the given buffer name. I don't think the actual existence (or lack thereof) of the buffer makes any difference. Right? 

Here's the section of code for reference:
https://github.com/hlissner/doom-emacs/blob/394e16689418b0f9e1bf8e070c14d1ff2940aa11/modules/lang/scala/autoload.el#L48-L54

hlissner (approved):

hlissner:
> I don't think the actual existence (or lack thereof) of the buffer makes any difference

Ah, you're right! I hadn't thought it through. It's good to go as is, thanks for your help!

PR #4678, opened 2021-02-22
Set default eshell TERM

elken:
Makes eshell more ansi-friendly ootb. I'm not 100% tied to it being `xterm` if you think there's a better option.

### DIFF
--- a/modules/term/eshell/config.el
+++ b/modules/term/eshell/config.el
@@ -71,7 +71,9 @@ You should use `set-eshell-alias!' to change this.")
         eshell-prompt-function #'+eshell-default-prompt-fn
         ;; em-glob
         eshell-glob-case-insensitive t
-        eshell-error-if-no-glob t)
+        eshell-error-if-no-glob t
+        ;; Shell config
+        eshell-term-name "xterm-256color")
 
   ;; Consider eshell buffers real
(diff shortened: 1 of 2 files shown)


modules/term/eshell/config.el
```diff
@@ -90,6 +90,7 @@ You should use `set-eshell-alias!' to change this.")
 
   ;; UI enhancements
   (add-hook! 'eshell-mode-hook
+    (setenv "TERM" "xterm-256color")
```
hlissner, line 93: Use `(setq eshell-term-name "xterm-256color")` instead, and move it out of the hook as it should be set globally so that it is easily changed.

That said, this may have other side effects. Please keep a look out for them. If folks report issues after this change I may revert it later.

hlissner (changes requested):

modules/term/eshell/config.el
elken: Yeah good point, I will.

I've been using it myself and it hasn't _caused_ any issues yet but it certainly helps modernise eshell a bit. I'll add a README note too

hlissner (approved):

hlissner:
Looks good to me. Thanks!

PR #4685, opened 2021-02-23
Do not refer to coq-mode-map  without proper guard

rgrinberg:
This fixes $ doom compile with coq enabled.

### DIFF
--- a/modules/lang/coq/config.el
+++ b/modules/lang/coq/config.el
@@ -71,15 +71,17 @@
 
   (setq company-coq-disabled-features '(hello company-defaults))
 
-  (if (featurep! :completion company)
-      (define-key coq-mode-map [remap company-complete-common]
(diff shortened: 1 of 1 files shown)


hlissner:
I'm afraid I cannot accept this PR, because it convolutes the customization path of those keybinds. Users would have to `(after! (company-coq coq) ...)` in order to change them where, formerly, a `(after! company-coq ...)` (or even `(after! coq-mode ...)` was sufficient, do to how they both are loaded together in a typical, interactive session.

A more systemic solution is needed to make it cooperate with `doom compile`. Thanks for the PR in any case.

rgrinberg:
Could you give me a hint what a more systemic solution would look like? As it stands, compiling the coq module breaks all of `doom compile` since doom does not proceed with the other modules  after failing on the coq module

PR #4686, opened 2021-02-23
use-package! to configure lean

rgrinberg:


hlissner:
The change is not equivalent and prevents the package from being lazy loaded, so I cannot accept this PR.

rgrinberg:
> The change is not equivalent and prevents the package from being lazy loaded,

Could you explain why? I thought using `use-package!` with `:config` maintains lazy loading.

hlissner:
> I thought using use-package! with :config maintains lazy loading.

That is incorrect. `(use-package! X :config Y)` is equivalent to 

```elisp
(require X)
Y
```

You need a deferring keyword in order to lazy load a package. e.g. like `:hook`, `:defer`, `:after`, `:commands`, or `:mode`. The minimum equivalent to `(after! X Y)` is `(use-package! X :defer t :config Y)`.

PR #4688, opened 2021-02-24
Add xterm-color to improve colour support

elken:
Similar to the last PR, this attempts to add even better colour support. Currently trying to break it but seems to fix most of them.

https://i.imgur.com/35dhkJl.png

Do you think this should be behind a flag?

### DIFF
--- a/modules/term/eshell/config.el
+++ b/modules/term/eshell/config.el
@@ -199,3 +199,13 @@ You should use `set-eshell-alias!' to change this.")
   (defadvice! +eshell--fallback-to-bash-a (&rest _)
     :before-until #'fish-completion--list-completions-with-desc
     (unless (executable-find "fish") "")))
+
+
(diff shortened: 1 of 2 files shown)


hlissner:
Can you show a before screenshot? `rg` output seems correctly colored without.

image

Partly because [of our `rg` alias](https://github.com/hlissner/doom-emacs/blob/develop/modules/term/eshell/config.el#L32).

elken:
`rg` is a bad example then xD I will find a few tonight

elken:
image

It's mostly triggered with the `doom` cli from what I can see, a few others here and there

hlissner:
Did you figure it a fix? Is this PR safe to close?

elken:
The only one I could see was add the hook back and set TERM to `eshell-term-name`, or use this PR

PR #4690, opened 2021-02-24
Enable disabling of magit-gitflow

jcf:
Disabling magit-gitflow breaks magit-dispatch due to the unconditional application of `transient-replace-suffix`.

With this change it is possible to disable magit-gitflow by adding the following to your private Doom configuration:

    (package! magit-gitflow :disable t)

### DIFF
--- a/modules/tools/magit/config.el
+++ b/modules/tools/magit/config.el
@@ -233,7 +233,10 @@ ensure it is built when we actually use Forge."
     (evil-define-key* evil-collection-magit-state git-rebase-mode-map
       "gj" #'git-rebase-move-line-down
       "gk" #'git-rebase-move-line-up))
-  (transient-replace-suffix 'magit-dispatch 'magit-worktree
-    '("%" "Gitflow" magit-gitflow-popup))
+
+  (after! magit-gitflow
+    (transient-replace-suffix 'magit-dispatch 'magit-worktree
+      '("%" "Gitflow" magit-gitflow-popup)))
+
   (transient-append-suffix 'magit-dispatch '(0 -1 -1)
     '("*" "Worktree" magit-worktree)))


jcf:
I should add that without this change, you get an error when trying to pop open `(magit-dispatch)`.

```
Debugger entered--Lisp error: (error "Suffix magit-gitflow-popup is not defined or autol...")
  apply(debug (error (error "Suffix magit-gitflow-popup is not defined or autol...")))
  transient--exit-and-debug(error (error "Suffix magit-gitflow-popup is not defined or autol..."))
  signal(error ("Suffix magit-gitflow-popup is not defined or autol..."))
```

modules/tools/magit/config.el
```diff
@@ -233,7 +233,10 @@ ensure it is built when we actually use Forge."
     (evil-define-key* evil-collection-magit-state git-rebase-mode-map
       "gj" #'git-rebase-move-line-down
       "gk" #'git-rebase-move-line-up))
-  (transient-replace-suffix 'magit-dispatch 'magit-worktree
-    '("%" "Gitflow" magit-gitflow-popup))
+
+  (after! magit-gitflow
+    (transient-replace-suffix 'magit-dispatch 'magit-worktree
+      '("%" "Gitflow" magit-gitflow-popup)))
```
hlissner, line 239: Move the `transient-replace-suffix` call into `magit-gitflow`'s use-package! block above.

hlissner: And please remove any extraneous newlines surrounding its old and new locations.

hlissner (changes requested):

hlissner (approved):

hlissner:
Nevermind my previous review. It's fine as is. Thanks for the help!

PR #4700, opened 2021-02-27
`evil-org-mode-map`'s `CSright` conflict with `Cright`

HyunggyuJang:
I understand that what you tried to accomplish via CSright was; however, it doesn't make any difference from Cright, and so for other navigation keys -- h,j,k,l.
I fixed this discrepancy.


hlissner (approved):

hlissner:
Good catches. Thanks!

PR #4702, opened 2021-02-27
Enable lispy-mode during eval-expression.

daanturo:
`eval-expression` is basically Elisp, therefore it makes sense to enable `lispy`.


hlissner (approved):

hlissner:
Thanks!

PR #4705, opened 2021-02-27
Fixing the gpl snippet template formatting

samwhitlock:
Need to escape the backticks that should appear as literals in the GPLv3 license.


hlissner (approved):

hlissner:
Good catches. Thanks!

PR #4710, opened 2021-03-01
Update brew install emacs-mac instructions

JasonMFry:
When installing emacs-mac today with the command `brew install emacs-mac --with-modules` I got `Warning: Treating emacs-mac as a formula. For the cask, use railwaycat/emacsmacport/emacs-mac`. I'm not aware of any significant difference between formula vs cask, but most users are going to copy and paste the commands, see the warning, and be concerned that they messed up something. If the user has a reason to treat emacs-mac as a formula, they are probably savvy enough to use the other command.

### DIFF
--- a/docs/getting_started.org
+++ b/docs/getting_started.org
@@ -242,7 +242,7 @@ to least recommended for Doom (based on compatibility).
   with macOS, native emojis and better childframe support. 
   #+BEGIN_SRC bash
   brew tap railwaycat/emacsmacport
-  brew install emacs-mac --with-modules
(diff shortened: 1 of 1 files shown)


hlissner:
Hi! Thanks for the PR, unfortunately it comes at a bad time. The documentation is about to see a major overhaul -- changes like these will be swallowed up by it. Until this is done I am not accepting PRs to the documentation, but I will double check `brew install emacs-mac` on my mac VMs and update the new documentation accordingly. Sorry for the trouble, but thanks for the help!

PR #4716, opened 2021-03-02
Fix leftover function name

yyadavalli:
The function name change `doom-num-cpus` to `doom-system-cpus` was missed.


hlissner:
Already done in #4718, but thanks for the help!

PR #4718, opened 2021-03-02
Use new CPU counter function in cc module

daedric:
Broken since commit c5e3f4d6.


hlissner (approved):

hlissner:
Whoops! Thanks for that!

PR #4721, opened 2021-03-02
Fix +org/refile-to-other-window in indirect buffer

ricklupton:
`+org/refile-to-other-window` does not work when working with an indirect buffer (e.g. with `org-tree-to-indirect-buffer`). This adds a call to `buffer-base-buffer` to make it work.


hlissner (changes requested):
Would you mind extending this fix to `+org/refile-to-other-buffer` and `+org/refile-to-file`?

Please use a `let` where multiple uses are involved.

hlissner (approved):

hlissner:
Thanks!

PR #4732, opened 2021-03-04
Fix the "Yank file path" non-evil bindings

paul-wilkinson:
+default/yank-buffer-filename was replaced in
be9b3ff352f39f2f21812e13466bd6d2f26cd6e4 and +evil-bindings.el was
updated accordingly. This commit makes the corresponding updates in
+emacs-bindings.el.


hlissner (approved):

hlissner:
Thanks!

PR #4743, opened 2021-03-06
Emacs lisp: fix variable value display in eldoc

nbfalcon:
`fboundp` does not check if a variable is bound, but rather a function.
Use `boundp`.

I think this was a mistake introduced in one of my bug-fixing PRs...

----

#


hlissner (approved):

hlissner:
Good catch! Thanks!

PR #4744, opened 2021-03-07
org-capture: accept text arg non-interactively

edrex:
Change org-capture to get text from positional args if present,
otherwise stdin.

Currently it gets text from positional args if interactive, and from
stdin otherwise, which lead to me scratching my head wondering why the script was eating
the passed texted when called from qutebrowser:

    :spawn org-capture "[[{url}][{title}]]"

In its present form the script doesn't support this usage.

### DIFF
--- a/bin/org-capture
+++ b/bin/org-capture
@@ -28,7 +28,11 @@ while getopts "hk:" opt; do
 done
 shift $((OPTIND-1))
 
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Looks good to me. Thanks!

PR #4748, opened 2021-03-07
`git-commit`: fix variable predicates

nbfalcon:
`git-commit-summary-max-length` should be a number, not a symbol.

`git-commit-major-mode` should be checked more carefully, as otherwise
exploits are possible (e.g. if it is `erase-buffer` or some dangerous
function of the user's environment).

----

#


hlissner (approved):

hlissner:
Thanks!

PR #4749, opened 2021-03-07
Fix nil issue with nav-flash & ivy

elken:
Raised on discord, without this causes issues with ivy in most cases. (sorry about the dirty commits)

### DIFF
--- a/modules/ui/nav-flash/autoload.el
+++ b/modules/ui/nav-flash/autoload.el
@@ -10,8 +10,8 @@ jumping to another part of the file)."
   (unless (minibufferp)
     (nav-flash-show)
     ;; only show in the current window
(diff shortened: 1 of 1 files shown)


hlissner (changes requested):
Rather than create an overlay, it'd be better not to call `overlay-put` at all. Please wrap it in a 

```lisp
(when (overlayp compilation-highlight-overlay)
  ...)
```

Also, please rebase (squash and force-push) this PR into a single commit.

elken:
True, yeah I'll change. I did try but it wasn't having it and I didn't expect it to bomb out.

elken:
All good! Apologies, haven't had to do a rebase for a while.

hlissner:
No worries. Thanks for the help!

PR #4752, opened 2021-03-08
Fix 'too many arguments' error on org-capture

panchoh:
org-capture foo bar baz

will fail with:

org/capture: line 33: [: too many arguments

Adding quotes to the expansion of $str will ensure that test -z has
only one argument.

Hope that it helps.
Cheers!


hlissner:
Thanks!

PR #4755, opened 2021-03-08
Prioritize company-capf over company-yasnippet

saevarb:
I noticed after updating that I was seeing a lot of irrelevant results for certain completion scenarios and tracked it down to this commit. Swapping the order does the trick. This ensures that when doing things like completing on `this.` inside classes, the class fields and methods are shown before the yasnippet snippets.

Before: 

image

After:

image


hlissner:
Thanks!

PR #4756, opened 2021-03-08
fix a break change in vterm.el,  (+vterm/here) can't create new terminal any more without this pr

jixiuf:
see https://github.com/akermu/emacs-libvterm/issues/505

I am not a doom user , maybe this is not a best fix, but it works .

### DIFF
--- a/modules/term/vterm/autoload.el
+++ b/modules/term/vterm/autoload.el
@@ -56,7 +56,7 @@ If prefix ARG is non-nil, cd into `default-directory' instead of project root."
              project-root))
          display-buffer-alist)
     (setenv "PROOT" project-root)
-    (vterm)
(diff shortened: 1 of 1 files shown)


gagbo:
Note: the change is 5 commits _after_ current pin https://github.com/akermu/emacs-libvterm/commits/c2c2c8afb935c11778388721912250cd1103aaf9

hlissner:
Thanks for the PR, but I must turn it down (for now). This module will be bumped soon along with other changes to the `:term` modules. I will include this one when that happens.

ra1nb0w:
I don't see this merged; Can be merged? thank you

Dieken:
this is included in https://github.com/hlissner/doom-emacs/pull/4836, but it’s also not merged😓

PR #4760, opened 2021-03-09
`git-commit`: use upstream `magit` predicates

nbfalcon:
The same predicates are now specified in upstream magit, so there is no need to
override `safe-local-variable`.

----

#


hlissner:
Thanks for the PR, but bump commit messages must follow a strict format, so I bumped the `:tools magit` module in 14eb254 together with your fix. Thanks for bringing it to my attention!

PR #4761, opened 2021-03-09
Better defaults for dired

vherrmann:
This pr sets `dired-create-destination-dirs` per default to `'ask`, so that the behaviour is more consistent with the one of `find-file`.


hlissner:
Thanks

PR #4766, opened 2021-03-10
Fix ivy-resume with ivy-prescient

p1llule:
Hi M.Lissner

ivy-resume can be used with a argument to restore a previous ivy session.

This doesn't work with ivy-prescient installed.

Steps to reproduce :

    1. have ivy and ivy-prescient enabled.
    2. register an ivy session (eg: call M-x or wathever)
    3. C-u ivy-resume
    4. ivy-prescient-sort-function: Wrong type argument: sequencep, counsel-M-x

See https://github.com/raxod502/prescient.el/issues/102

### DIFF
--- a/modules/completion/ivy/config.el
+++ b/modules/completion/ivy/config.el
@@ -393,6 +393,10 @@ results buffer.")
     (let ((prescient-filter-method '(literal regexp)))
       (ivy-prescient-re-builder str)))
 
+  ;; Fix `ivy-resume' when invoked with an argument
+  ;; see https://github.com/raxod502/prescient.el/issues/102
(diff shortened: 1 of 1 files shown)


hlissner:
I've temporarily fixed this in 62df2c3, but since I can reproduce this in vanilla Emacs, this is an upstream issue and should be resolved more permanently there. Thanks for bringing it to my attention.

p1llule:
Indeed. 

You may also reconsider using #'+ivy-prescient-non-fuzzy in ivy-re-builders-alist as it breaks swiper-query-replace.

See https://github.com/raxod502/prescient.el/issues/93

PR #4775, opened 2021-03-11
Fix doom-{after,before}-reload-hook doc strings

ivanbrennan:
The doc strings for these hooks had "after" and "before" swapped:
- `doom-after-reload-hook`
- `doom-before-reload-hook`


hlissner (approved):

hlissner:
Good catches, thanks!

PR #4782, opened 2021-03-11
Fix `+debugger/start` not working

naistran:
Use `let*` instead of `let`, otherwise we get this error `completing-read: Symbol’s value as variable is void: result`.


hlissner (approved):

hlissner:
Thanks!

PR #4786, opened 2021-03-11
bump evil-nerd-commenter to 3.5.4

LoganBarnett:
This gives us `evilnc-yank-and-comment-operator`, which allows us to yank
a line while simultaneously commenting or uncommenting it.

This function was added just a little while ago at https://github.com/redguardtoo/evil-nerd-commenter/issues/120, and it's been tagged as 3.5.4, which is the same as the commit in this PR.

Not that Spacemacs should be a reference for authority, but I think it had used `g Y` as its prefix for similar functionality (it uses `evil-commentary`, which has had this feature for a while). Quick prodding around shows `g Y` to be free in the Doom ecosystem.

Such a mapping might be as simple as this?

``` lisp
(map!
  :n "g Y" #'evilnc-yank-and-comment-operator
  )
```
I'm happy to add it to the PR as well, if that seems reasonable/useful. But for now this is just a bump :)

I've done some quick testing with a local pin and 3.5.4 works reasonably well.

Thanks for maintaining Doom! :)

### DIFF
--- a/modules/editor/evil/packages.el
+++ b/modules/editor/evil/packages.el
@@ -11,7 +11,7 @@
 (package! evil-exchange :pin "3030e21ee16a42dfce7f7cf86147b778b3f5d8c1")
 (package! evil-indent-plus :pin "0c7501e6efed661242c3a20e0a6c79a6455c2c40")
 (package! evil-lion :pin "6b03593f5dd6e7c9ca02207f9a73615cf94c93ab")
-(package! evil-nerd-commenter :pin "2730820b9ccedf758c8a0428ee2c994c9fc415dd")
+(package! evil-nerd-commenter :pin "b8ac35fe019df5602c31912f65303a3d8ad0066c")
 (package! evil-numbers
   :recipe (:host github :repo "janpath/evil-numbers")
   :pin "006da406d175c05fedca4431cccd569e20bef92c")


hlissner:
Hi! Thanks for your PR, but i must refuse it. 

1. This PR's does not follow our strict formatting for bump commits (so they are machine readable by future CI).
2. There are other bumps in the pipeline for this module.
3. The change is not urgent enough to warrant an early bump. Packages are generally bumped every weekend, given they are ready to be bumped. I will queue `evil-nerd-commenter` for this weekend.

A side question: when you created this PR, did you see [this PR template](https://github.com/hlissner/doom-emacs/blob/develop/.github/PULL_REQUEST_TEMPLATE/bump.md)? Few bump PRs have followed it, so I fear folks are missing it or github isn't picking it up.

> but I think it had used g Y as its prefix for similar functionality

We already have a `gY` keybind: it yanks text without leading indentation. I think this is more useful than `evilnc-yank-and-comment-operator`, which already has an acceptable alternative (`y` then `gc`). I'll sit on it and see if a better solution can't be found.

LoganBarnett:
> This PR does not follow our strict formatting for bump commits (so they are machine readable by future CI).

I'm terribly sorry if I've bypassed your process! Before submitting this PR, I did some digging around and found this: https://github.com/hlissner/doom-emacs/blob/develop/docs/contributing.org#help-keep-packages-up-to-date but it didn't give me the impression of a template to follow. Perhaps it was elsewhere and I missed it? Again - my apologies if I've messed that up.

> A side question: when you created this PR, did you see this PR template? Few bump PRs have followed it, so I fear folks are missing it or github isn't picking it up.

 I've seen some pre-filled templates come up when filling out issues or pull requests in other places of Github - I try to follow those as best I can, but I don't recall seeing one as I filled this out. I tried re-creating the pull request to see if it would show the template and this is what I saw:

Screen Shot 2021-03-12 at 11 16 21

Which is just my commit message rolled out into the title/description.

> There are other bumps in the pipeline for this module.

I did do a search beforehand for `evil-nerd-commenter` pull requests and found none (which was odd to me, but I suppose that should've been cause for further investigation).

> Packages are generally bumped every weekend

Excellent! I'd have waited if I knew about that. Would it be helpful if I expanded some of the contributing documentation in the section linked above?

> We already have a `gY` keybind

Doh. Another miss :( FWIW - `y` and `gc` do work, but you have to re-perform the operation to get it. Unless there's some aspect of vim/evil that allows you to repeat the last command with the same motion but different operation. I haven't been able to find such a thing though.

Thanks again for maintaining this. Sorry that I've caused some unnecessary churn here. I agree it is good to decline the pull request.

PR #4788, opened 2021-03-12
evil: bind smerge-vc-next-conflict to ]<

elmarco:


hlissner:
Thanks for the PR, but I cannot accept this because it is incomplete. This needs more thought to accommodate the rest of smerge's commands, which I'm working on elsewhere. In any case, sorry for the trouble and thanks for the help.

elmarco:
Ok, thanks @hlissner. Is this other smerge work available somewhere? It would be great to bind the next-conflict command.

PR #4792, opened 2021-03-13
Fix non-default REPL's format of command name.

daanturo:
Existing REPL commands has a dash ("-") before the "repl" suffix, but the function format to intern doesn't, currently I always get an error `"Couldn't find a valid REPL for %s"` whenever I open a non-default REPL.

For example, choosing "ipython" using `+eval/open-repl-other-window` with an `universal-argument` in `python-mode`.

### DIFF
--- a/modules/tools/eval/autoload/repl.el
+++ b/modules/tools/eval/autoload/repl.el
@@ -71,7 +71,7 @@
                (format "+%s/open-%srepl" module
                        (if (string= repl "default")
                            ""
-                         repl))))))
+                         (concat repl "-")))))))
     (let ((region (if (use-region-p)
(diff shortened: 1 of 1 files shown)


hlissner (changes requested):
This will break the default repl. Make this change instead:

```diff
               (format "+%s/open-%srepl" module
                       (if (string= repl "default")
                           ""
-                        repl))))))
+                        (concat repl "-")))))))
```

daanturo:
Thank you, I didn't see that.

hlissner (approved):

hlissner:
Thanks for your help and quick response!

PR #4795, opened 2021-03-13
Don't set dash-docs-browser-func to eww.

daanturo:
The only reference of `dash-docs-browser-func` is a lexical binding, which is `#'browse-url` by default anyway.

Setting it to `eww` doesn't make sense because plain `counsel/helm-dash` is hardly usable unless all of local `dash-docs-docset` are installed (which is set by `set-docsets!`).

Using a modern graphical browser is far easier on the eyes.

### DIFF
--- a/modules/tools/lookup/config.el
+++ b/modules/tools/lookup/config.el
@@ -197,8 +197,7 @@ Dictionary.app behind the scenes to get definitions.")
   :config
   (setq dash-docs-enable-debugging doom-debug-p
         dash-docs-docsets-path (concat doom-etc-dir "docsets/")
-        dash-docs-min-length 2
-        dash-docs-browser-func #'eww)
+        dash-docs-min-length 2)
 
   (cond ((featurep! :completion helm)
          (require 'helm-dash nil t))


hlissner:
Hi! Thanks for the PR, but I disagree with this change. Doom uses `eww` by default so that the docset is opened _in_ Emacs (as a regular Emacs buffer which can be searched or manipulated directly). `eww` displays docsets reasonably well.

> because plain counsel/helm-dash is hardly usable unless all of local dash-docs-docset are installed (which is set by set-docsets!).

You should be using `+lookup/in-docsets` (on `SPC s k`, or `K` if the mode doesn't have a specialized documentation lookup handler). It strips out docsets that aren't installed before performing the search. `{counsel,helm}-dash` does not do this.

Perhaps this behavior should be advised into `{counsel,helm}-dash`, or changed upstream, instead of split off into a separate command?

daanturo:
> Doom uses `eww` by default

I thought that `browse-url` was the default, it's reproducible in the sandbox.
 `+lookup/in-docsets` utilizes `+lookup-open-url-fn`'s value, which is set to `browse-url` by default.

> Perhaps this behavior should be advised into `{counsel,helm}-dash`, or changed upstream, instead of split off into a separate command?

I didn't mean to use `{counse,helm}-dash`, `+lookup/in-docsets` abstracts them away anyway, which ignores `dash-docs-browser-func` therefore setting it doesn't have any effects on Doom's exposed commands.

hlissner:
Ah, I see the issue. As of 5ee8080 `+lookup/in-docsets` will respect `dash-docs-browser-func`, which should resolve this issue. Let me know if that isn't the case and I'll reopen it. Thanks for bringing it to my attention!

PR #4805, opened 2021-03-15
Fix deletion in minibuffer under cua-mode

jpcima:
With `cua-mode` globally enabled, deleting a text selection in the minibuffer using the `<delete>` key does not work as expected.
Instead, only the character under the cursor is deleted.
By editing the key override for minibuffer as indicated, this problem is resolved.


hlissner:
Sorry for the late response and thanks for the PR! I'm dropping Emacs 26.x support late next month. The problems those `doom-silence-motion-key` hacks set out to fix were addressed upstream in Emacs 27 and onwards, so it will be one of the first things to be removed once 26.x support is dropped. I'll turn down this PR only because it would be b swallowed up by that change anyway. In any case, thanks for bringing it to my attention!

PR #4806, opened 2021-03-15
Fix major mode for jupyter-R src block

sei40kr:
## Summary

`R-mode` -> `ess-r-mode`

Actually `R-mode` is a function alias for `ess-r-mode`, but with `R-mode`, `SPC c f` wouldn't work.


sei40kr:
Sorry it didn't work after restarting Emacs. I'm closing this.

PR #4808, opened 2021-03-16
Remove double underscore from c header templates

andreiltd:
Using double underscore for an identifier in C/C++ invokes undefined
behavior.

From https://eel.is/c++draft/lex.name:

Each identifier that contains a double underscore __ or begins with an
underscore followed by an uppercase letter is reserved to the
implementation for any use.


hlissner (approved):

hlissner:
Thanks!

PR #4810, opened 2021-03-16
docs(debugger): readme draft

R0flcopt3r:
Adds documentation for debugger module.

This only cover dap-mode for clang and python, but it is a start.

In the configuration heading I added the keybinds I have set.

I am not sure what to put in the `#+SINCE:` field.


jeetelongname:
relevent too #1166

hlissner (approved):

hlissner:
Sorry for the late response. I am rewriting (and reformatting) our documentation, which already includes a slightly better README.org for this module. I will incorporate some of your suggestions into it, but I must turn down the PR simply for the timing. In any case, thank you for bringing it to my attention!

PR #4814, opened 2021-03-17
reload-font is needed after setting doom-font variable

vmiheer:


hlissner:
Hi! Thanks for the PR but I'm afraid I must turn it down. We're transitioning to Discourse later this week, and our FAQ along with it (faq.org will be deleted soon). What's more, everything in docs/*.org will be rewritten, so I'm not accepting PRs to docs/ for the time being. Thanks for the help in any case!

Issue #4656, opened 2021-02-16
Module for Zig language


bnjmnt4n:
It would be great if there were a module for the Zig programming language in the future. Right now, the tooling around Zig is still a little nascent, but I've created a WIP module in my own doom configuration: https://github.com/bnjmnt4n/system/tree/ab2c5b0072d2923372bfc3966fc6c1f834173582/doom/modules/lang/zig. 

Plugins:
- https://github.com/ziglang/zig-mode
- LSP server, for use with lsp-mode: https://github.com/zigtools/zls
  Note that there still aren't pre-built binaries for zls, and Zig is changing at a pretty rapid pace too so binaries for those might be stale.

PR #4827, opened 2021-03-20
Add `:lang zig` module

bnjmnt4n:
This PR adds a language module for Zig. It turns out that there's less configuration than I expected, especially following updates to lsp-mode (https://github.com/emacs-lsp/lsp-mode/pull/2723), so let me know if you'd rather not add this as a module. Otherwise, I can write up the documentation for the module as well.

Ref. #4656.


hlissner:
Looks good to me, thanks!

PR #4829, opened 2021-03-20
Update vterm installation instructions for NixOS

vrinek:
According to `home-manager`:
```
trace: Melpa attribute 'emacs-libvterm' is a legacy alias that will be removed in 21.03, use 'vterm' instead
```

:point_up: is just a warning. `home-manager` builds without errors either with `emacs-libvterm` or `vterm`.

FYI, my home-manager config is set to `home.stateVersion = "21.03";`. I wonder if we want to include instructions for previous versions as well.


hlissner:
Resolved in #4963

PR #4835, opened 2021-03-29
Update Elixir Installation instructions

axelson:
Link to official installation guide
Mark asdf as recommended for easier switching of versions


hlissner (approved):

hlissner:
I've removed the `(recommended)` tag as the reasoning doesn't seem compelling enough for Doom (or I) to officially recommend it. It implies that Doom provides special support for `asdf`, which it doesn't. It is already the first option in the list, which should defer more importance to it than the rest, but otherwise I leave it to the user to decide if asdf suits their needs. 

Thanks for the help in any case!

axelson:
Fair enough :+1:

Issue #4273, opened 2020-11-14
Doom CLI Rewrite


hlissner:
I've been working on this CLI rewrite for a while. Over time, bug reports have trickled in about various problems that this rewrite directly or indirectly fixes. To reduce the noise I'm consolidating those and futures posts into this one, to represent the rewrite:

## Primary goals of this rewrite

+ [ ] (#4270) To make it transactional (failures should revert back to where we started).
+ [ ] (#2141) To offer a `doom rollback` command.
+ [ ] (#3490) Make `doom sync` smarter about `:recipe` changes and stale bytecode (so you don't need to do that `doom sync -u` and `doom build` dance).
+ [ ] Improve the doctor API and checks, so we get more useful checks _and_ the warnings/errors they emit are less cryptic.
  + Check for misspelled modules and flags (#2048)
  + Warn users when they use inadvertently use package.el #4157
  + Support for a private doctor.el in your DOOMDIR (#4190)
  + Perform more sanity checks on your private config (#2959)
+ [ ] To fix our unit tests and add integration tests, powered by nix and buttercup.
+ [ ] CI/CD, so bumping packages can be automated (and tested).
+ [ ] To handle straight.el's prompts better -- or prevent them entirely. Re-cloning packages is better UX than these confusing prompts.

This is not an exhaustive list. 

## Other bugs

+ [ ] To fix whatever weirdness Windows is doing in #3844 

-----

I'll keep this post updated until the rewrite is done. There is no ETA on this.

danilevy1212:
Hi! Anyway we can help you with this rewrite or features of this rewrite? Thanks for the awesome work.

the-cobalt:
> Add a new `doom package ...` family of commands for per-package management

I think it might be better to integrate this into existing commands Using your example, instead of `doom package build x`, I think `doom build x` might be cleaner.

This is just a micro-aggression, but:
> they use inadvertently use package.el
I'm pretty sure that's a typo

Also, I think it would be nice to put this in like a "cli" branch, considering how this could be a breaking change, but people might want to keep track of/try it on their own.

innerout:
If possible, I would suggest adding in the `doom upgrade` command a message informing the user that init.el has been updated with new options. This would be useful when a new module is added or when a new flag is added in a module.

hlissner:
> If possible, I would suggest adding in the doom upgrade command a message informing the user that init.el has been updated with new options. This would be useful when a new module is added or when a new flag is added in a module.

[...]

PR #4840, opened 2021-04-11
doom/sandbox: fix docstring

iyefrat:


hlissner:
Our sandbox is being rewritten as part of #4273, and the docstring (or function, for that matter) you're modifying won't survive it. Since this is minor enough, I'd rather let it slip to avoid the merge conflict. Sorry for the trouble, but thanks for the help anyway!

PR #4841, opened 2021-04-11
Improve factor configuration

rgrinberg:
* Add bindings to fuel-help-mode
* Add more bindings to factor mode
* Add more bindings to fuel-listener mode

### DIFF
--- a/modules/lang/factor/config.el
+++ b/modules/lang/factor/config.el
@@ -12,35 +12,64 @@
     :references #'fuel-show-callers
     :documentation #'fuel-help))
 
+(map! :after fuel-help
+      :map fuel-help-mode-map
+      :localleader
+      "e" #'fuel-help-edit
+      "d" #'fuel-help-delete-bookmark
+      "B" #'fuel-help-display-bookmarks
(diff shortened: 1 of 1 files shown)


modules/lang/factor/config.el
```diff
@@ -12,35 +12,64 @@
     :references #'fuel-show-callers
     :documentation #'fuel-help))
 
+(map! :after fuel-help-mode
```
hlissner, line 15: `s/fuel-help-mode/fuel-help`. [The package that defines `fuel-help-mode-map`](https://github.com/factor/factor/blob/master/misc/fuel/fuel-help.el) is called `fuel-help`, not `fuel-help-mode`, therefore these keybinds will never take effect.

hlissner (changes requested):

modules/lang/factor/config.el
rgrinberg: Fixed. Thanks for the review.

hlissner (approved):

hlissner:
Thanks!

PR #4842, opened 2021-04-13
ocamlforamt hook: buffers without a file name

rgrinberg:
The behavior before this PR would cause an error if a buffer not backed
by a file was opened. This can be trivially reproduced by opening a
merge/rebase conflict using ediff inside an .ml file.


hlissner (approved):

hlissner:
Thanks!

Issue #3939, opened 2020-09-14
Sorting files in dired mode doesn't work with windows 10


altomcat:
**What did you expect to happen?**

The dired buffer is not sort by date when pressing the 'o' key.

**What actually happened?**

In dired mode, pressing 'o' switches from"Dired by name" to "Dired by date" in the modeline with no effect.
So the dired buffer remains sorted by name.

**Additional details:**

- it doesn't work with emacs 26.3 or 27.1 on windows 10 with a fresh install of Doom 2.0.9.
- it works with 'Emacs -q' on windows 10
- it works with emacs 26.3 on Mac OS
-  `SPC h k o` returns : 

> dired-sort-toggle-or-edit is an interactive and compiled function
> defined in dired.el.
> 
> Signature
> (dired-sort-toggle-or-edit &optional ARG)
> 
> Documentation
> Toggle sorting by date, and refresh the Dired buffer.
> 
> With a prefix argument, edit the current listing switches instead.
> 
> View in manual
> 
> Key Bindings
> dired-mode-map <normal-state> o
> dired-mode-map s

- Include any warnings or errors logged to \*Messages\* (use `M-x
  view-echo-area-messages` to see it).
No popup error when enabling debug-on-error.

**Steps to reproduce:**

1. 'C-x d' to jump to dired
2. Select a directory then press Return
2. Press 'o' to toggle from 'by name' to 'by date'.

**System information:**
<details><pre>
```
SYSTEM  type       windows-nt
        config     x86_64-w64-mingw32
        shell      C:/Users/$USER/Downloads/emacs-27.1-x86_64/libexec/emacs/27.1/x86_64-w64-mingw32/cmdproxy.exe
        uname      n/a
[...]

PR #4844, opened 2021-04-14
Fix for issue #3939 for dired under Windows

matzesc:
Added code to remove the `--group-directories-first` command line option for `ls` under Windows, since Windows uses an emulated `ls` implementation that does not support this command line option. This fixes the issue #3939.


hlissner:
Thanks for the PR, but I'm afraid I must refuse it because it causes a regression: depriving WSL or cygwin users on Windows of `--group-directories-first` when users have `ls-lisp-use-insert-directory-program` disabled. I will follow up with a possible solution in #3939 in a moment. Thanks for the help in any case!

PR #4845, opened 2021-04-14
Add doom-dashboard-widget-header

haoxiangliew:
This small segment of code adds a `doom-dashboard-widget-header`, which is defined to be below `doom-dashboard-widget-header`. This would he helpful especially if Doom Emacs plans to have an official `fancy-splash-image`.


hlissner:
I don't see the benefit of this, and so must refuse it. The subtitle is part of the ascii banner, we don't need two banners, even in the presence of an official banner image.

PR #4856, opened 2021-04-14
Fix vterm popup rule

marienz:
akermu/emacs-libvterm@c2c2c8afb935c11778388721912250cd1103aaf9 changed
the default buffer name of vterm buffers, so Doom's popup-rule no longer
matches them. This causes functions like `+vterm/here` to use the entire
frame instead of creating a popup window.

Fix it by changing the popup rule to match the new naming scheme.


hlissner (approved):

hlissner:
Thanks!

PR #4861, opened 2021-04-14
doom/move-this-file now supports moving directly into a directory

NightMachinary:


hlissner:
This is not necessary. `doom/move-this-file`'s default behavior is to move the current file into a target directory, if a directory is given. Thanks for the PR in any case.

NightMachinary:
[**@hlissner**](https://github.com/hlissner) commented on [Apr 15, 2021, 4:12 AM GMT+4:30](https://github.com/hlissner/doom-emacs/pull/4861#issuecomment-819918756 "2021-04-14T23:42:15Z - Replied by Github Reply Comments"):
> This is not necessary. `doom/move-this-file`'s default behavior is to move the current file into a target directory, if a directory is given. This is due to `rename-file`'s behavior. To quote its documentation:
> 
> > If NEWNAME is a directory name, rename FILE to a like-named file under  
> > NEWNAME. For NEWNAME to be recognized as a directory name, it should  
> > end in a slash.
> 
> Thanks for the PR in any case.

‍`rename-file` behaves correctly, but the wrapper code around it fails and opens a non-existent buffer. I have had this happen all the time, as I move files into dirs a lot.

core/autoload/files.el
```diff
@@ -286,6 +286,9 @@ If FORCE-P, overwrite the destination file if it exists, without confirmation."
     (user-error "Buffer is not visiting any file"))
   (let ((old-path (buffer-file-name (buffer-base-buffer)))
         (new-path (expand-file-name new-path)))
+    (when (directory-name-p new-path)
+      (setq new-path (concat new-path (f-filename old-path)))
+      )
```
ymarco, line 291: Don't leave closing parens on their own line, people don't do that in lisp.
The closing paren for `when` should be on the same line as `setq`.

NightMachinary: I find it easier to read code this way. It's like adding a new after a block in python:
```
if True:
  pass

some_other_code
```

But of course, I can change it back if it matters.

hlissner: It does. And please replace `(f-filename old-path)` with `(file-name-nondirectory old-path)`. Doom does not guarantee f.el is present, nor is it necessary.

NightMachinary: Okay, I applied the changes.

ymarco:
> I find it easier to read code this way.

Can't comment on that, but nowhere in Doom's codebase are paren left out like that.

NightMachinary:
Can someone merge this?

hlissner (approved):

hlissner:

PR #4873, opened 2021-04-16
Bump :editor evil

jcf:
emacs-evil/evil-collection@b45ec5b → emacs-evil/evil-collection@b6025cf
emacs-evil/evil@7eace67 -> emacs-evil/evil@d998a81

The register system has been rejigged in the last week, and some
packages in the Evil ecosystem are already making use of the new vars.

emacs-evil/evil@229ebb4
mamapanda/evil-owl#10

### DIFF
--- a/modules/editor/evil/packages.el
+++ b/modules/editor/evil/packages.el
@@ -1,7 +1,7 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; editor/evil/packages.el
 
-(package! evil :pin "7eace67fc6d1f992e447f216baf9a8b7be634c1e")
+(package! evil :pin "d998a8195e404b01e2ea62a455c3dec74d0823c3")
 (package! evil-args :pin "758ad5ae54ad34202064fec192c88151c08cb387")
 (package! evil-easymotion :pin "f96c2ed38ddc07908db7c3c11bcd6285a3e8c2e9")
 (package! evil-embrace :pin "4379adea032b25e359d01a36301b4a5afdd0d1b7")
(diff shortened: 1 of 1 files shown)


jcf:
I'm not sure if I got the format of the commit message right. Apologies!

I did try to run `doom/commit-bumps` but it looks like that function needs to be written?

``` elisp
;;;###autoload
(defun doom/commit-bumps ()
  (interactive))
```

I'm guessing it won't be trivial to implement, but will take a quick stab at it.

hlissner (changes requested):
Thanks for the PR! The commit message is perfect, but would you mind also bumping `evil-collection` as well? 

There's a subtle difference between `Bump package-name` and `Bump :editor evil`. The latter implies you've bumped the whole module and any packages that haven't been bumped either a) have no updates or b) are blocked by something else. The former explicitly means "I'm only bumping this one package."

Given that evil-collection has updates and, after looking through them, there are no blockers, I'd rather expand the scope of this PR.

> re:doom/commit-bumps

Don't waste any time on `doom/commit-bumps`. The reason this isn't implemented (and may never be) is because that functionality is being implemented elsewhere (as part of #4273). 

The new CLI will introduce CI/CD for bumping packages or modules with the `bin/doom` tool. e.g. `doom bump evil` (bumps the package) or `doom bump :lang evil` (bumps all packages in the module). This will then open $EDITOR with a pre-filled commit message (plus a `-f` switch or similar to skip that step).

This will be automated much later, and tied to a web interface outdated pins can be tracked. (Not to mention, folks will be able to request out-of-band bumps on our Discourse).

jcf:
> There's a subtle difference between Bump package-name and Bump :editor evil. The latter implies you've bumped the whole module and any packages that haven't been bumped either a) have no updates or b) are blocked by something else. The former explicitly means "I'm only bumping this one package."

@hlissner that makes sense. I thought my use of `:editor evil` might be a little… excessive. I've bumped evil-collection as requested, and force pushed a new commit with an update commit message.

I'm already subscribed to #4273, so will keep an eye on things there, and help out wherever I can.

Thanks!

hlissner (approved):

hlissner:

PR #4875, opened 2021-04-17
Indicate that lsp-describe-thing-at-point is async

dsdshcym:
- otherwise, `+lookup-online-backend-fn` would be called as well
- lsp-describe-thing-at-point was marked as not async in
  https://github.com/hlissner/doom-emacs/commit/d4eb7e31ac25107a35f9de567a497d9100fcc1ff
- but `lsp-describe-thing-at-point` calls `lsp--make-request`
  internally (which is async I assume):
  https://github.com/emacs-lsp/lsp-mode/blob/eda51c21662253fd05b4f3f20ad7b029d9c2aff7/lsp-mode.el#L4677


hlissner (approved):

hlissner:
Thanks!

PR #4878, opened 2021-04-17
Fix doom-module-mplist-map obsolete module warning

ivanbrennan:
Inside `doom-module-mplist-map`, one of the code paths handling obsolete module warnings references a `key` variable, which appears to be unbound.
```elisp
(message "WARNING %s module was removed" key)
```
I changed `key` to `(list category module)`, which appears to be what was intended and is consistent with neighboring code paths.

__Note__: Untested. I don't have doom installed, but I noticed this while reading through the code.


hlissner (approved):

hlissner:
Thanks!

PR #4879, opened 2021-04-17
AWS CloudFormation module

xlarsx:
Utilities to improve the experience editing AWS CloudFormation templates.

### DIFF
--- a/modules/tools/cloudformation/doctor.el
+++ b/modules/tools/cloudformation/doctor.el
@@ -1,0 +1,4 @@
+;;; tools/cloudformation/doctor.el -*- lexical-binding: t; -*-
+
+(unless (executable-find "cfn-lint")
+  (warn! "Couldn't find cfn-lint."))
--- a/modules/tools/cloudformation/doctor.el
+++ b/modules/tools/cloudformation/doctor.el
@@ -2,3 +2,7 @@
 
 (unless (executable-find "cfn-lint")
   (warn! "Couldn't find cfn-lint."))
+
+;; The binary is with underscore but the name is with hyphen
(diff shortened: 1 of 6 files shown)


gagbo:
What's wrong with [cfn-mode](https://gitlab.com/worr/cfn-mode/tree/a4ca40978e680f9edc86c141e696e0ae57c63533) on melpa ?

xlarsx:
Nice catch @gagbo, feedback applied

thank you very much for the feedback!

modules/tools/cloudformation/doctor.el
```diff
@@ -0,0 +1,8 @@
+;;; tools/cloudformation/doctor.el -*- lexical-binding: t; -*-
+
+(unless (executable-find "cfn-lint")
+  (warn! "Couldn't find cfn-lint."))
+
+;; The binary is with underscore but the name is with hyphen
+(unless (executable-find "cfn_nag")
+  (warn! "Couldn't find cfn-nag."))
```
jcf, line 8: Couldn’t find `cfn-nag` or `cfn_nag`?

hlissner:
Hi! Thanks for the PR, but I must turn it down. The configuration is much too trivial to warrant its own module. I wouldn't mind this being a part of a larger `:tools aws` module, perhaps under a `+cloudformation` flag, but as it is now it is trivial enough for end users to implement themselves. In any case, thanks for the help; this will certainly serve as a useful reference if/when an `aws` module materializes.

xlarsx:
Hi, no problem, I understand your point, that `aws` module sounds tempting 😊, thank you very much for the feedback

Greetings!

PR #4709, opened 2021-03-01
Update URL to post on Doom

noelwelsh:
I reworked my blog and broke all links. Oops! This change updates the link to my notes on Doom.

PR #4881, opened 2021-04-17
Add link to clearify

ShalokShalom:
When you tell somebody to do something, also show them how to.


hlissner:

Issue #4452, opened 2020-12-19
Symbol’s value as variable is void: all-the-icons-install-fonts


in-in:
**What did you expect to happen?**
Successfully installed ‘all-the-icons’
...

**What actually happened?**
Symbol’s value as variable is void: all-the-icons-install-fonts
...

**Additional details:**
- Include a link to your private config: I don't have one.
- Include screenshots/casts of your issue
![203541608380215](https://user-images.githubusercontent.com/8797432/102689208-569bef00-420d-11eb-8849-1de938b783f2.png)

**Steps to reproduce:**
1. Open terminal and run `emacs --batch -f all-the-icons-install-fonts`

According to the [documentation](https://github.com/hlissner/doom-emacs/blob/develop/docs/getting_started.org#install-doom-manually) the last step of the manual installation should be `emacs --batch -f all-the-icons-install-fonts`, but this command doesn't work.
![image](https://user-images.githubusercontent.com/8797432/102689419-0a51ae80-420f-11eb-9529-8291cf110f5e.png)

**System information:**
<details><pre>
SYSTEM  type       gnu/linux
	config     x86_64-pc-linux-gnu
	shell      /usr/bin/zsh
	uname      Linux 5.4.0-58-generic #64-Ubuntu SMP Wed Dec 9 08:16:25 UTC 2020 x86_64
	path       (~/.config/emacs/bin ~/.nvm/versions/node/v14.15.1/bin ~/.local/bin ~/bin /usr/local/sbin /usr/local/bin /usr/sbin /usr/bin /sbin /bin /usr/games /usr/local/games /snap/bin /usr/lib/emacs/27.1/x86_64-linux-gnu)
EMACS   dir        ~/.config/emacs/
	version    27.1
	build      Sep 19, 2020
[...]

PR #4883, opened 2021-04-18
docs: load init for all-the-icons-install-fonts command

karlicoss:
See https://github.com/hlissner/doom-emacs/issues/4452, before init wasn't loaded, so command didn't work.


hlissner:

Issue #4771, opened 2021-03-10
Go-to-reference broken in Golang with gopls after commit #430d628


mhersson:
After upgrading and applying commit #430d628 " tools/lsp: use xref backend as lookup handler"
The go-to-reference shortcuts (gD or SPC-c-D bound to +lookup/references) does not work with gopls. 

Reverting this commit  and adding the following two lines to the lsp module remedies the problem.
```
    :definition #'lsp-find-definition
    :references #'lsp-find-references
```

**What did you expect to happen?**
Menu pops up with all references to the function, once one is selected cursor is moved to point of reference.

**What actually happened?**
Doom freezes for a couple of seconds utilizing 100% cpu, then outputs "Go guru finished".

**Additional details:**
The solution to the problem might be in the go module

From modues/lang/go/config.el
```
   6 (after! go-mode
   7   (set-docsets! 'go-mode "Go")
   8   (set-repl-handler! 'go-mode #'gorepl-run)
   9   (set-lookup-handlers! 'go-mode
  10     :definition #'go-guru-definition
  11     :references #'go-guru-referrers
  12     :documentation #'godoc-at-point)
```

I tried commenting out lines 10 and 11, and that kind of fixed the issue, but for some reason it didn't work everywhere, but Doom didn't freeze up. I couldn't make out why it only worked some places in the code.

**EDIT**: Just found out that when it workes the lookup returnes multiple references. When only 1 reference is returned cursor/focus is not moved upon selection.

**EDIT 2**: The issue isn't only with the number of references which I claimed in the previous edit. It's with where there this single reference is. If the only reference is in another file. Then It does not work. It actually does find the reference and moves the cursor there in the file, but the buffers are not switched. If the single reference is in the same file it works. If there are multiple references it always works, same file or different file, no matter.

**Steps to reproduce:**
1. Brand new install of Doom Emacs - only change for default: enable lsp and (go +lsp) in init.el
2. Create a new go project or open exsting just make sure LSP is enabled. Can be tested with something as simple as
```
   1 package main
   2 
   3 import "fmt"
   4 
   5 func Hello() {
   6     fmt.Println("Greetings!")
   7 }
   8 
   9 func main() {
  10     Hello()                                                                                                                                        
  11 }
```

3. Move cursor to Hello on line 5 and press gD or SPC-c-D

**System information**
https://pastebin.com/MxWTUjeV

markflarup:
I can verify this problem.

**System information**
```
[...]

PR #4885, opened 2021-04-18
tools/lsp: mark lookup handlers for lsp-mode as async

dsdshcym:
- otherwise, +lsp-lookup-definition-handler and +lsp-lookup-references-handler
  would fail when the definition/reference is in another file
- related:
  + https://github.com/hlissner/doom-emacs/pull/4875
  + https://github.com/hlissner/doom-emacs/issues/4771


hlissner:
Thanks for the PR but I cannot accept it. The point of removing `:async t` was so that lookup handlers would fall through to the next. This PR breaks that behavior. I'm looking for improvements to those handlers to properly return `t` if it succeeds and `nil` when it fails, so that the order is preserved.

dsdshcym:
No problem! Thanks for the response.
Glad to know you are working on this issue :).

PR #4889, opened 2021-04-19
fix and add keybinds in :lang scheme, with descriptions

hugonobrega:
Some keys were bound to no-longer-existing commands, or commands which changed
name.

Others only made sense in the REPL, or also in the REPL, so we made keybinds in
`scheme-mode-map', `geiser-repl-mode-map', or both accordingly.


modules/lang/scheme/config.el
```diff
@@ -22,46 +21,51 @@
...
-          "q" #'geiser-repl-exit
-          "l" #'geiser-load-current-buffer
-          "r" #'geiser-restart-repl
-          "R" #'geiser-reload
-          "c" #'geiser-repl-clear-buffer)))
-
+  (map!
+   :localleader
+   (:map (scheme-mode-map geiser-repl-mode-map)
+    :desc "Start a/switch to/from REPL" "'"  #'switch-to-geiser
```
ymarco, line 27: Isn't "toggle" the easier way of saying "start a/switch to/from"?

hugonobrega:
Probably, but I couldn't word that clearly while still fitting into 27
characters. Something like "Start or toggle REPL/buffer"? It feels a bit
weird to me because it isn't clear that Start refers only to REPL

ymarco:
I think just "toggle" (without "start or") is enough. All popup terminal keybinds use that.
https://github.com/hlissner/doom-emacs/blob/68d1a450345fb460761fea59b2fdd1bb5ebc0aaf/modules/config/default/+evil-bindings.el#L566-L576

hugonobrega:
Ok, if it's convention I definitely buy it of course. But it sounds a bit
weird to me as "toggle" would suggest (to me) something like on/off, so
showing/hiding the REPL instead of switching back and forth to it.

I'll change it

hugonobrega:
(closed by accident due to lack of skill with git/github)

hugonobrega:
Recreated as PR #4890

PR #4891, opened 2021-04-19
Fix treemacs + lsp doctor check

rgrinberg:
The previous check was incorrect when eglot and treemacs without lsp was
on.


hlissner:
Thanks!

PR #4905, opened 2021-04-21
Syntax highlight .sface files as html

davydog187:
[Phoenix.LiveView](https://hexdocs.pm/phoenix_live_view) has a project called [Surface](https://surface-ui.org/) that adds a `.sface` extension to files. These should be syntax highlighted as html.

First PR!

### DIFF
--- a/modules/lang/web/+html.el
+++ b/modules/lang/web/+html.el
@@ -5,6 +5,7 @@
   :mode "\\.\\(?:tpl\\|blade\\)\\(?:\\.php\\)?\\'"
   :mode "\\.erb\\'"
   :mode "\\.l?eex\\'"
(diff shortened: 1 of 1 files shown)


modules/lang/web/+html.el
```diff
@@ -5,6 +5,7 @@
   :mode "\\.\\(?:tpl\\|blade\\)\\(?:\\.php\\)?\\'"
   :mode "\\.erb\\'"
   :mode "\\.l?eex\\'"
+  :mode "\\.sface\\'"
```
davydog187, line 8: if you want i can merge this with the line above so that all the elixir extensions are collocated

hlissner:
Sorry for the late response. This looks good to me, thanks!

Issue #2182, opened 2019-12-12
[BUG] org-tree-slide mode does not show headings


dinosaur-lin:
**Describe the issue**
- What did you expect to happen?
org-tree-slide mode should show headings at the top of the slide
- What actually happened?
org-tree-slide mode does not show headings at the top of slides
- Do you have screenshots/casts of your issue?
This is the text in org file:

```org
* Introduction

  My son is reading yellow stone stories in kida-z now.

  It is a fun story
```

In the screenshot of org-tree-slide mode, there is no "Introduction"

![image](https://user-images.githubusercontent.com/1984447/70677150-4cd7c700-1c5c-11ea-924a-fc0a9be42318.png)

- What commands are bound to any keys you've referenced? (use `SPC h k` to
  inspect keys)
I just did M-x org-tree-slide-mode
- Can you include a link to your private config?
I use plain doom distribution without any changes.
- Were any warnings or errors logged to \*Messages\* (`SPC h e` or `M-x
  view-echo-area-messages` will open it).
No, I do not see any errors in \*Messages\*

hlissner:
Iirc, this is controlled by `org-tree-slide-skip-outline-level`, which doom sets to `2` by default (as the top-level headings are treated as "slides", and subheadings are treated as the actual content of the slide).

This would reverse that:

```elisp
(after! org-tree-slide
  (setq org-tree-slide-skip-outline-level 1))
```

PR #4907, opened 2021-04-22
Proper fix for "[BUG] org-tree-slide mode does not show headings"

xukai92:
https://github.com/hlissner/doom-emacs/issues/2182 still exist and the cause was not that explained in https://github.com/hlissner/doom-emacs/issues/2182#issuecomment-564899286.

The cause is from the line I'm removing in this PR, which actually skips the heading.

### DIFF
--- a/modules/lang/org/contrib/present.el
+++ b/modules/lang/org/contrib/present.el
@@ -51,7 +51,6 @@
                      (when (org-before-first-heading-p)
                        (org-next-visible-heading 1))
                      (ignore-errors (org-up-heading-all 99))
-                     (forward-line 1)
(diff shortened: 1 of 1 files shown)


xukai92:
Also, what's exactly the code inside `defadvice!` is doing? I didn't see any effect like starting the presentation from the subtree of my cursor as far as I can see.

hlissner:
As of 8bb54a9 I've turned it into an option (`+org-present-hide-first-heading`) which is disabled by default, which should resolve your issue.

xukai92:
Thanks for the proper fix!

PR #4910, opened 2021-04-22
config/default: Fix C-return binding in GUI Emacs

ellakk:
The "C-RET" counsel-company binding does only seem to work when running Emacs in a
terminal. "C-<return>" makes it work when running the desktop version too.


hlissner (approved):

hlissner:

PR #4911, opened 2021-04-22
Also kill window after +org/export-to-clipboard

siddharthshekar:
The previous implementation only closed the buffer leaving an extra
window hanging around.

### DIFF
--- a/modules/lang/org/autoload/org-export.el
+++ b/modules/lang/org/autoload/org-export.el
@@ -31,7 +31,7 @@ Prompts for what BACKEND to use. See `org-export-backends' for options."
     (unwind-protect
         (with-current-buffer buffer
           (kill-new (buffer-string)))
-      (kill-buffer (current-buffer)))))
+      (kill-buffer-and-window))))
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


hlissner (changes requested):
Could you try this approach instead?

```diff
+(defvar org-export-show-temporary-export-buffer)
 ;;;###autoload
 (defun +org/export-to-clipboard (backend)
   "Exports the current buffer/selection to the clipboard.
 
 Prompts for what BACKEND to use. See `org-export-backends' for options."
   (interactive
    (list (intern (completing-read "Export to: " org-export-backends))))
-  (let ((buffer (org-export-to-buffer backend "*Formatted Copy*" nil nil t t)))
+  (let ((buffer (org-export-to-buffer backend "*Formatted Copy*" nil nil t t))
+        org-export-show-temporary-export-buffer)
     (unwind-protect
         (with-current-buffer buffer
           (kill-new (buffer-string)))
-      (kill-buffer (current-buffer)))))
+      (kill-buffer buffer))))
```

I'd rather avoid creating the window in the first place.

PR #4913, opened 2021-04-23
lang/julia: Add support for julia-vterm when vterm is enabled.

AlexLewandowski:
There is a new Julia REPL (https://github.com/shg/julia-vterm.el) built to work specifically with the vterm backend. I have been using it for a few months and greatly prefer it over the julia-repl package (even with its own vterm backend). Here are my suggested changes to use julia-vterm when vterm is enabled.


hlissner:
Sorry for the super late response. It appears vterm support [was incorporated into julia-repl](https://github.com/tpapp/julia-repl#using-vterm-experimental) at some point, which renders julia-vterm obsolete. I will add it to the README soon. Thanks for bringing this to my attention!

PR #4917, opened 2021-04-24
rss: +org flag control installation of elfeed-org

nackjicholson:
Hey there,

In the default `~/.doom.d/init.el` this module is commented but listed with the +org flag `(rss +org)`.

I happened to try using it without `+org` and it still installs https://github.com/remyhonig/elfeed-org even though it's not needed. This PR makes it so you only need to install elfeed-org if you use +org flag


hlissner (approved):

hlissner:
Good catch, thanks!

PR #4920, opened 2021-04-24
Japanese input docs fleshed out and added mozc

brandonpollack23:
Second PR, first one isnt even done, on a roll (if it isnt terrible)

### DIFF
--- a/modules/tools/prodigy/config.el
+++ b/modules/tools/prodigy/config.el
@@ -1,8 +1,6 @@
 ;;; tools/prodigy/config.el -*- lexical-binding: t; -*-
 
 (after! prodigy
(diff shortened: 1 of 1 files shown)


iyefrat:
@brandonpollack23 I messed around with a mozc module a while back, and a significant issue was the fact that the mozc repo was like half a gig so cloning it took forever (and almost all of it is irrelevant to emacs). Did you experience this too?

brandonpollack23:
It is about half a gig, 200M of which are the git repo.  I'm not super familiar with straight, but I'm sure we can --depth 1 checkout the commit we want or put this whole feature behind a + guard?

PR #4921, opened 2021-04-24
Update org-gcal autoload fix to match upstream

brandonpollack23:
First contribution, forgive me (I rushed through the contributors.org)

Resend because now im using branches correctly


hlissner:
This issue was fixed in 89775f38d, so I'll consider this resolved. Thanks for bringing it to my attention!

PR #4923, opened 2021-04-25
Makes fd-dired compatible with BSD xargs

alexDarcy:
`Fd-dired` used the `--quoting-style=literal` option to `xargs`, which does not exist in BSD xargs. I've just removed this option.


hlissner:
Seems this was tackled upstream in https://github.com/yqrashawn/fd-dired/issues/17, so I've bumped the `:emacs dired` module in 5be988a, indirectly fixing this issue. Thanks for bringing it to my attention!

PR #4926, opened 2021-04-25
Reimplement +emacs-lisp--module-at-point to fix #4896

### LINKED ISSUE #4896: +lookup/definition in init.el breaks when using :if, :cond etc.
Calling `+lookup/definition` in `$DOOMDIR/init.el` works on all symbols in the `(doom!)` block.

**What actually happened?**
The function does not work on symbols where there is an instance of the `(:if condition module)` construct between the symbol and the corresponding module keyword. To be more exact, the helper function `+emacs-lisp--module-at-point` when called on `lang/emacs-lisp` returns `(:if emacs-lisp nil)` instead of `(:lang emacs-lisp nil)`.

**Additional details:**

**Steps to reproduce:**
1. Open `$DOOMDIR/init.el`.
2. Change the `lang/cc` line from `cc` to `(:if t cc)`.
3. Position the cursor on `emacs-lisp` on the `lang/emacs-lisp` line such that the `lang/cc` line is between the `lang/emacs-lisp` line and the `:lang` line.
4. Call `+lookup/documentation`.
5. Evaluate `(+emacs-lisp--module-at-point)`

**System information:**
https://pastebin.com/YbCbyWeg

Janfel:
This PR reimplements `+emacs-lisp--module-at-point` on top of `syntax-ppss` to make it more robust and make it recognize module flags not starting with a `+`.
Fixes https://github.com/hlissner/doom-emacs/issues/4896.

### DIFF
--- a/modules/lang/emacs-lisp/autoload.el
+++ b/modules/lang/emacs-lisp/autoload.el
@@ -29,35 +29,54 @@ to a pop up buffer."
 ;;; Handlers
 
 (defun +emacs-lisp--module-at-point ()
-  (let ((origin (point)))
-    (save-excursion
-      (goto-char (point-min))
-      (when (re-search-forward "(doom! " nil 'noerror)
-        (goto-char (match-beginning 0))
-        (cl-destructuring-bind (beg . end)
-            (bounds-of-thing-at-point 'sexp)
-          (when (and (>= origin beg)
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Fantastic work! Thanks again for your help!

PR #4930, opened 2021-04-26
Fix typo (dependecy/dependency) in error message

marienz:


hlissner:
Thanks for the PR! There's nothing wrong with it, but the CLI is currently undergoing a hefty rewrite (include the doctor and all its checks), so I am currently not accept PRs for our CLI (except for hotfixes). Thanks for bringing it to my attention, in any case!

PR #4932, opened 2021-04-27
Add Rails class template

otavioschwanck:


modules/lang/ruby/config.el
```diff
@@ -189,6 +189,11 @@
   (when (featurep! :lang web)
     (add-hook 'web-mode-hook #'projectile-rails-mode))
   :config
+  (when (featurep! :editor file-templates)
+    (defadvice! +projectile-rails-ignore-snippet-for-module-a (last-part name)
+      "Use ruby-mode's __app_class file template by returning no snippet."
```
agraul, line 194: This doc string is now a bit misleading, the `__module` file template is still used in `lib/`.

agraul:
This is what I get from your template in `app/foo/bar.rb`:
```ruby
# frozen_string: literal

class Bar < Base
  
end
```

While this is the default from projectile-rails' template:
```ruby
class Bar

end
```

The magic comment can be turned on by setting `(setq projectile-rails-expand-snippet-with-magic-comment t)` and results in:
```ruby
# frozen_string_literal: true

class Baz

end
```

The only difference with your template (apart from the slightly mis-formed magic comment) that I can see is the automatic inheritance of `Base`, which I think is not a good thing to have in Doom upstream. Not everyone who uses a Ruby project with a nested `app/` directory will have a class called `Base`.

I'd just use the file templates from `projectile-rails` for `app/` since it does a good job with its templates there. For `lib/`, I came up with another advice that might be better because `projectile-rails--snippet-for-module` is used both for `lib/` and for `app/` (if no specialized `app/` snippets applies):

```lisp
(when (featurep! :editor file-templates)
  (defadvice! +projectile-rails-ignore-snippet-for-lib-a (orig-fn)
    "Use ruby-mode's __app_class file template by returning no snippet."
    :around 'projectile-rails-corresponding-snippet
    (if (string-match "lib/\\(.+\\)\\.rb$" (buffer-file-name))
        nil
      (funcall orig-fn))))
```

It uses the same regexp that `projectile-rails-corresponding-snippet` uses to find `lib/` and will therefore only override that one. The downside is that this regexp needs to be kept up to date in case `projectile-rails` changes its regexp, but in this case I doubt there will be many changes.

otavioschwanck:
> This is what I get from your template in `app/foo/bar.rb`:
> 
> ```ruby
> # frozen_string: literal
> 
> class Bar < Base
>   
> end
> ```
> 
> While this is the default from projectile-rails' template:
> 
> ```ruby
> class Bar
> 
> end
> ```
> 
> The magic comment can be turned on by setting `(setq projectile-rails-expand-snippet-with-magic-comment t)` and results in:
> 
> ```ruby
> # frozen_string_literal: true
> 
> class Baz
> 
> end
> ```
> 
> The only difference with your template (apart from the slightly mis-formed magic comment) that I can see is the automatic inheritance of `Base`, which I think is not a good thing to have in Doom upstream. Not everyone who uses a Ruby project with a nested `app/` directory will have a class called `Base`.
> 
> I'd just use the file templates from `projectile-rails` for `app/` since it does a good job with its templates there. For `lib/`, I came up with another advice that might be better because `projectile-rails--snippet-for-module` is used both for `lib/` and for `app/` (if no specialized `app/` snippets applies):
> 
> ```lisp
> (when (featurep! :editor file-templates)
>   (defadvice! +projectile-rails-ignore-snippet-for-lib-a (orig-fn)
>     "Use ruby-mode's __app_class file template by returning no snippet."
>     :around 'projectile-rails-corresponding-snippet
>     (if (string-match "lib/\\(.+\\)\\.rb$" (buffer-file-name))
>         nil
>       (funcall orig-fn))))
> ```
> 
> It uses the same regexp that `projectile-rails-corresponding-snippet` uses to find `lib/` and will therefore only override that one. The downside is that this regexp needs to be kept up to date in case `projectile-rails` changes its regexp, but in this case I doubt there will be many changes.

about the Base, its changable, i don't use base either, when i create the file, it come selected (by the snippet), just press C-d and its gone, or tab and i can change forr my Base (in case, is ApplicationBusiness)

agraul:
You're right that nested modules don't indent well, however I think that should be fixed upstream at projectile-rails and not worked around by creating a new file template in Doom.

otavioschwanck:
> You're right that nested modules don't indent well, however I think that should be fixed upstream at projectile-rails and not worked around by creating a new file template in Doom.

this template that i made is a little better than projectile rails, but anyway, i created an issue there:
https://github.com/asok/projectile-rails/issues/155

PR #4938, opened 2021-04-27
lang/latex: Remove the -g background flag on Skim.app to foreground the viewer

jmhammond:
The Skim.app `displayline` command takes the flag `-g` which is "Do not bring Skim to the foreground"

when viewing the pdf results of the latex build, bringing skim to the foreground saves an `Alt+Tab` to open the app that I was intending to lauch by the auctex View command.


hlissner (approved):

hlissner:
Makes sense to me. Thanks!

PR #4940, opened 2021-04-27
Added instructions for Void Linux onto getting_started.org

gi-yt:
Hi,
I added instructions for installing doom in Void Linux.
Regards,
AK

### DIFF
--- a/docs/getting_started.org
+++ b/docs/getting_started.org
@@ -22,6 +22,7 @@ us know!
     - [[#on-linux][On Linux]]
       - [[#ubuntu][Ubuntu]]
       - [[#fedora][Fedora]]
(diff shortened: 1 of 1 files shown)


FollieHiyuki:
There are 2 emacs packages in Void: `emacs-x11` and `emacs-gtk3` that can be considered. `emacs-gtk3` tends to crash a lot while running in daemon mode, compared to `emacs-x11` (at least in my experience).
`emacs-common` will get pulled in as a dependency when installing either of the packages above, so we don't need to explicitly install it manually.

gi-yt:
oh ok. i use emacs-gtk3 and it is pretty smooth

hlissner:

PR #4941, opened 2021-04-28
lang/org: cleanup `+pretty` feature, mostly just adding `if` cookie

HyunggyuJang:
Add `if` cookie so that it doesn't get compiled by `doom compile`.

I know that you discourage to byte compile doom modules; nonetheless, it definitely speed up quite a lot. I felt it is annoying `pretty.el` file getting compiled with `doom compile`, so I added it to be consistent with other featured `.el` files.


hlissner (approved):

hlissner:
Makes sense to me. Thanks!

PR #4942, opened 2021-04-28
Fix `+lookup/file` so that it can look up file in selection

HyunggyuJang:
Previously, when we issue `gf` in visual mode, `+lookup/file` couldn't reflect
the region in interest.

With this commit, you can look up file based on the region activated.
I agree this can be achieved via tweaking `+lookup-file-functions`, but the end
user should have this facility by default.


hlissner:
I've had a more comprehensive improvement to `+lookup/file` in testing for some time and decided to push that instead in 259cf83, which should also address your request. Thanks for bringing it to my attention!

PR #4944, opened 2021-04-28
Fix smartparens pairing ' and ` in the minibuffer.

Janfel:
The current fix only specifies `minibuffer-inactive-mode`, which does nothing for me (Emacs 28 at https://github.com/emacs-mirror/emacs/commit/d55d5358b27dee15ebbd998131d22b221f5f4964). This PR adds `minibuffer-mode` to the mode lists, which fixes the issue for me.


hlissner:
I've been aware of this issue for some time. Seems `minibuffer-mode` was introduced very recently. Thanks for the fix!

PR #4945, opened 2021-04-28
Bind TeX-view in LaTeX-mode-map too

ymarco:
I know https://github.com/hlissner/doom-emacs/commit/0d6c32ff25292ecdef7ecdfd1542bb65807c4d9a was supposed to fix *something*, but I'm just unable to use any keybind on latex-mode-map.
Binding it on both is the safest bet I think.

### DIFF
--- a/modules/lang/latex/config.el
+++ b/modules/lang/latex/config.el
@@ -100,7 +100,11 @@ If no viewers are found, `latex-preview-pane' is used.")
     (add-hook! '(tex-mode-local-vars-hook
                  latex-mode-local-vars-hook)
                #'lsp!))
-  (map! :map latex-mode-map
+  (map! :localleader
+        :map latex-mode-map
+        :desc "View" "v" #'TeX-view)
+  (map! :after latex
+        :map LaTeX-mode-map
         :localleader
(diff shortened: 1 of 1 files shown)


ymarco:
Hold on, `map!` expands incorrectly here

hlissner (changes requested):
Does it work if you combine the two `map!` calls with `:map (latex-mode-map LaTeX-mode-map)` instead?

ymarco:
> Does it work if you combine the two `map!` calls with `:map 
> (latex-mode-map LaTeX-mode-map)` instead?

Yes, but LaTeX-mode-map might not be bound after only auctex's 
tex.el loads. It's declared in auctex's latex.el.
latex-mode-map is defined in tex-mode.el, packaged with Emacs. The 
built-in latex-mode is used by auctex's TeX-mode.
Can we maybe bind TeX-view to TeX-mode-map?
What a mess.

ymarco:
The issue is still present...

hlissner (approved):

hlissner:
Sorry for the late response. It is frustrating that this is necessary, but I'll see if I can refactor it later. Thanks for the help!

PR #4952, opened 2021-05-01
Fixing windows pathway to use latest straight API

ihdavids:
We are using the old straight--call function in the windows pathway. This switches us to using the new API


hlissner (approved):

hlissner:
Thanks!

PR #4957, opened 2021-05-01
tools/debugger: fix fringe when lsp + vc-gutter

Percee:
When (debugger +lsp) and :ui vc-gutter are both enabled, sometimes the
left fringe where breakpoints and diff are will be half-size, cutting
the breakpoint in half.

This change fixes that.

image

### DIFF
--- a/modules/tools/debugger/config.el
+++ b/modules/tools/debugger/config.el
@@ -158,3 +158,7 @@
   :when (featurep! +lsp)
   :hook (dap-mode . dap-ui-mode)
   :hook (dap-ui-mode . dap-ui-controls-mode))
+
+(when (and (featurep! +lsp) (featurep! :ui vc-gutter))
+  (add-hook 'dap-ui-mode-hook (lambda ()
+                                (setq-local left-fringe-width 10))))
--- a/modules/tools/debugger/config.el
+++ b/modules/tools/debugger/config.el
@@ -160,5 +160,7 @@
   :hook (dap-ui-mode . dap-ui-controls-mode))
 
 (when (and (featurep! +lsp) (featurep! :ui vc-gutter))
-  (add-hook 'dap-ui-mode-hook (lambda ()
-                                (setq-local left-fringe-width 10))))
+  (defadvice! +debugger--restore-fringe-width-a (&rest _)
+    "Ensure the fringe is wide enough to display breakpoints."
+    :after #'dap-ui--refresh-breakpoints
+    (set-fringe-style)))


hlissner:
Thanks for the PR, but I cannot accept it. I'd prefer the fringe be adjusted when the debugger is activated, then restored when it is deactivated, but not enlarged globally and permanently. Can you adjust this PR to do so? If not, please close it.

Percee:
Sorry about that, it should do what you want now.

hlissner:
Unfortunately it does not. `dap-mode` and `dap-ui-mode` are global modes. This is effectively the same as doing `(setq-local left-fringe-width 10)` at top level, which has very little effect because there's no guarantee what buffer will be focused when that line gets evaluated. Even if it does act on the correct buffer, ti will only do so on the first buffer.

The problem is a lot more involved than it appears. I don't have dap-mode installed to investigate, but you'll have to find a minor mode or hook that activates when a debugger is activated and killed. Or some way to catch those events and attach this logic to it (including restoring the fringe after it's done).

Percee:
The fix I propose would not be only when a debugger is activated but when there are breakpoints, even while not debugging.
I thought it would make sense to have the fringe at least the width of a breakpoint in any buffer where a breakpoint could appear. But I understand that this is not the way.

I think I found part of the problem. When opening a new .cpp file with existing breakpoints, they are already cut in half for an unknown reason. But if I set the fringe back to 10, and trigger a popup, the fringe of the cpp buffer gets axed, by the advice function +popup-adjust-fringes-h which is supposed to only modify the popup fringe.

I'll try to find the reason why they are cut in half upon opening.

Percee:
The problem can be fixed in another way by adding a check in vc-gutter/config.el
```lsp
    (if (fboundp 'fringe-mode) (fringe-mode '4))
;;; BECOMES
    (if (and (fboundp 'fringe-mode)
             (not (featurep! :tools debugger +lsp)))
             (fringe-mode '4))
```
So if a user has debug with lsp enabled, their fringe size is the default (8) to display the whole breakpoint. What do you think?

Also, it seems to me that we want enough width to display breakpoints at all times, not just when the debugguer is running.

hlissner:
Sorry for the _super_ late response: I am closing PRs to prepare to move our modules into a separate repo. That said, the problem this PR tackles is something I want to eventually address, but on-the-fly, when the debugger is activated (and undone when the user quits from it). Because fringe state is global, this needs some more sophisticated machinery. In any case, I'll set it aside for the eventual redesign of the debugger module. Thanks for bringing it to my attention, in any case!

PR #4958, opened 2021-05-01
:input layout: fix calls of +layout-bepo-rotate-keymaps

gagbo:
After the refactor, the main function to call to fix small inconsistencies went from a raw evil-collection-setup-hook
 function to a proper helper function with a more logical signature. 

This fixes the calls to the new function, so that the custom fixes are properly applied


hlissner (approved):

hlissner:
Ah, my bad. Thanks for catching that!

PR #4961, opened 2021-05-01
Feature-gate :tools lsp company-specific hooks

gagbo:
As I’m trying to use [corfu](https://github.com/minad/corfu) for a bit, I found out that this hook prevents lsp-mode from starting if company-mode isn’t used.

### DIFF
--- a/modules/tools/lsp/+lsp.el
+++ b/modules/tools/lsp/+lsp.el
@@ -81,13 +81,14 @@ about it (it will be logged to *Messages* however).")
           (lsp--info "Could not guess project root."))))
     #'+lsp-optimization-mode)
 
-  (add-hook! 'lsp-completion-mode-hook
-    (defun +lsp-init-company-backends-h ()
-      (when lsp-completion-mode
-        (set (make-local-variable 'company-backends)
-             (cons +lsp-company-backends
-                   (remove +lsp-company-backends
-                           (remq 'company-capf company-backends)))))))
+  (when (featurep! :completion company)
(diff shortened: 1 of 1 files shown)


gagbo:
Actually I can't really live without that fix. Even with
```lisp
(when (featurep! :tools lsp)
  (setq lsp-completion-provider :none)
  (remove-hook 'lsp-completion-mode-hook #'+lsp-init-company-backends-h))
```
(wrapped or not inside a `(after! lsp)`), I need to go to the sexp manually and `grab` it in order to have lsp working. Otherwise I can't start lsp-mode at all
```
Error processing message (doom-hook-error lsp-configure-hook #[0 "\205
^@	\205
^@\302 \207" [lsp-auto-configure lsp-completion-enable lsp-completion--enable] 1] (doom-hook-error lsp-completion-mode-hook +lsp-init-company-backends-h (void-variable company-backends))).
```

hlissner:
> `(after! lsp)`

It should be `(after! lsp-mode ...)`. Give that a try.

gagbo:
> It should be `(after! lsp-mode ...)`. Give that a try.

I can confirm it does **not** work :(

hlissner:
Sorry for the long wait. I think a `company-mode` check would probably be better, but I'll sort that out post-merge. Thanks for your help!

gagbo:
> Sorry for the long wait.

No worries :)

I also kind of forgot to make my corfu integration better, that’s a reminder to do so

PR #4962, opened 2021-05-01
Fix making doom-first-input-hook permanent-local

marienz:
We were making the value of doom-first-input-hook permanent-local
instead of doom-first-input-hook itself. That value is normally nil at
this point, which seems to be harmless. Reloading core.el from a script
(while doom-first-input-hook is not empty) results in an error.

Fix it by adding the missing quote present in the other calls.


hlissner (approved):

hlissner:
Good catch, thanks!

PR #4963, opened 2021-05-01
README.org: Add info about vterm in NixOS

futile:
I hit a problem with `epkgs.vterm` from `nixos-20.09` being too old and thus missing `vterm-goto-char` which is present in newer versions of `epkgs.vterm`. Switching to `emacs` & `emacsPackagesNgGen` from `unstable` fixed the issue for me, and I assume this might be a helpful hint in the docs.

I added a note to mention this for future users, and added two other changes/clarifications as well; all changes:

* Changed `epkgs.emacs-libvterm` -> `epkgs.vterm`, as the old name is deprecated.
* Added note that `epkgs.vterm` already contains a `vterm-module.so`
* Added note that the version of `emacs` needs to be recent enough and that otherwise bugs might occur.


hlissner (approved):

hlissner:
Thanks

Issue #4935, opened 2021-04-27
Some keybindings in evil-org-mode not set


pricly-yellow:
Some keybinding not set in evil-org mode. Then i try to use gl gj gk emacs call evil-lion-left, evil-next-line, evil-prev-line accordingly. Rebuild doom not help. This error already mention in [discord](https://discord.com/channels/406534637242810369/695219268358504458/830547972303093870).

**What did you expect to happen?**
gl gj gk buind to evil-org-mode functions.

**What actually happened?**
gl gj gk buind to another evil-mode functions.

**Steps to reproduce:**
1. Enable in init.el evil and org modes.
2. Open any org file.
3. Try to gl gj gk or see bindings in whatis buffer.

**System information:**
https://pastebin.com/8upBfqQS

mmarshall540:
Can confirm.

"gj" and "gk" are shadowed whenever visual-line-mode is enabled. Evil-mode uses them as backup bindings in case you want to use the non-visual line-movement commands when visual-line-mode is on.

And "gl" is shadowed by evil-lion-mode.

We can move these bindings out of our way permanently by putting this in config.el.

```
;;; Unshadow evil-org's bindings, if we expect it to be loaded.
(when (and (featurep! :lang org) (featurep! :editor evil +everywhere))
  ;; Move visual-line-mode bindings from evil-integration.el out of the way.
  (when (and evil-want-integration evil-respect-visual-line-mode)
    (general-define-key
     :definer 'minor-mode
     :states 'motion
     :keymaps 'visual-line-mode
     "gj" nil
     "gk" nil
     "gs M-j" #'evil-next-line
     "gs M-k" #'evil-previous-line))
  ;; Move the evil-lion bindings out of the way too.
  (map!
   :nv "gl" nil
   :nv "gL" nil
   :nv "gzl" #'evil-lion-left
   :nv "gzL" #'evil-lion-right))
```

That fixes it for me.

pricly-yellow:
Your fix also work for me. Maybe you add PR?

mmarshall540:
Glad it works for you too!

I will have to learn how to submit a PR. LOL

PR #4964, opened 2021-05-01
Move certain evil bindings out of evil-org's way.

mmarshall540:
Related to issue #4935. Evil-org's "gj" and "gk" were shadowed whenever visual-line-mode was enabled. And "gl" was shadowed by evil-lion.

Not sure if it's best to unbind and rebind just the "gj" and "gk" keys, or if it's preferable to re-define the whole map. For now, I went with the former, but the original map is in evil-integration.el at line 492 and is only 10 lines long.

### DIFF
--- a/modules/editor/evil/config.el
+++ b/modules/editor/evil/config.el
@@ -113,6 +113,19 @@ directives. By default, this only recognizes C directives.")
         (evil-ex-nohighlight)
         t)))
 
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but I cannot accept it as it is much too intrusive, and I'd rather not move evil-lion's keys. I'll come up a with a better solution soon.

PR #4967, opened 2021-05-02
Prevent okular from stealing focus

RBckmnn:
Pass the "noraise" flag to okular if it is used as a Latex viewer. If an
okular instance is already displaying the Latex compilation result,
subsequent invocations won't make okular steal focus from Emacs.


hlissner (approved):

hlissner:
Makes sense to me. Thanks!

PR #4975, opened 2021-05-03
Set initial state in prodigy to normal

rgrinberg:
This allows using evil-collection's bindings for this mode.


hlissner:
In this case, do we need a `set-evil-initial-state!` call at all?

rgrinberg:
Probably not. I just noticed the initial state being set to `'normal` elsewhere in the code base and cargo culted.

rgrinberg:
Got rid of it.

hlissner (approved):

PR #4976, opened 2021-05-04
update org-mime

nackjicholson:
Hey Sir,

The `org-mime` version we're using is lacking some useful new functions and fixes. I figured I'd update it. I also wanted to ask if you think `org-mime` should actually be included in this module? We don't provide any extra features or configuration for `org-mime` out of the box really, It's minimally configured and not documented or emphasized in our README. My thought is that most people who use doom and notmuch probably don't even know `org-mime` is available to them and installed. We could probably remove it, One less thing to keep updating or changing :)

I'm willing to change this PR to remove `org-mime` from the module and add a section of documentation about setting it in your own  "user space" configs. Let me know if you think that's a reasonable thing to do. Otherwise, please consider this version bump for merge.

Best,
Will

### DIFF
--- a/modules/email/notmuch/packages.el
+++ b/modules/email/notmuch/packages.el
@@ -2,7 +2,7 @@
 ;;; email/notmuch/packages.el
 
 (package! notmuch :pin "45193bab16c728ba892a5d45fc62ef59e2a6ef85")
-(package! org-mime :pin "9bb6351b25c62835c7881fc64096028eb8ef83ef")
+(package! org-mime :pin "eb21c02ba8f97fe69c14dc657a7883b982664649")
 (when (featurep! :completion ivy)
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but i had a larger bump commit in the pipeline, and bump commits must adhere to a strict template in order to remain machine readable for future CI. In any case, as of ed4b1ef, org-mime is now up-to-date, so I'll consider this resolved. Thanks for bringing it to my attention!

nackjicholson:
@hlissner Thanks!

1. How do you do the "bump commits"? Is there anything read up on that? What's the process I should follow if I think a dependency should be updated or modified?
2.  Any thoughts on the other question I had of removing `org-mime` entirely?

PR #4985, opened 2021-05-05
Edit outdated links in README.md

artemiy312:
Edit outdated links to development plan and plugins under review


hlissner:
Thanks for the PR, but I already have a larger commit in store to update all our docs with discourse links. I won't change these links until the Discourse is officially up.

PR #4990, opened 2021-05-06
lang/emacs-lisp: Replace sym with var in +emacs-lisp-log-unsafe-local-variables-a

Wetlize:
Otherwise it will make Emacs to error with `(void-variable sym)` error when checking the variable.


hlissner (approved):

hlissner:
Whoops, that shows me to push at 4am. Thanks for the fix!

PR #5008, opened 2021-05-08
notmuch: remove org-mime

nackjicholson:
Hi Henrik,

I'm not sure about this one. Feel free to decline.

This module is minimally configured and we're not providing anything extra on top with it. There may be many people that use this module and don't know `org-mime` is available to them. This PR removes `org-mime` and adds a section to the readme under a heading "Writing HTML Emails."

If you have ideas for why this module should stay, please share.

Best,
Will

### DIFF
--- a/modules/email/notmuch/README.org
+++ b/modules/email/notmuch/README.org
@@ -18,17 +18,19 @@
   - [[#downloading-your-email][Downloading your email]]
   - [[#customize-notmuch-hello-buffer][Customize =*notmuch-hello*= buffer]]
   - [[#changing-the-notmuch-landing-page][Changing the =notmuch= landing page]]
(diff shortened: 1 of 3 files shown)


hlissner (changes requested):
Rather than remove it, put it behind an `+org` flag, like [the new mu4e module](https://github.com/hlissner/doom-emacs/pull/3971) does. Frankly, it's only there for mu4e parity, but I'd like to stick to that.

nackjicholson:
@hlissner I will make that change. Thanks for the guidance!

nackjicholson:
@hlissner I altered the commit to put org-mime behind a new +org flag. Please consider for merge at your convenience. Thank you!

hlissner:
Thanks for the help!

PR #5013, opened 2021-05-08
Change some `fboundp`'s and `doom-module-p`'s to `featurep!`'s

iyefrat:
Makes the code more readable and searchable (say if we want to figure out all the places where turning on `ivy` effects the codebase), and from testing on my end doesn't seem to break anything.

### DIFF
--- a/core/autoload/help.el
+++ b/core/autoload/help.el
@@ -200,9 +200,9 @@ selection of all minor-modes, active or not."
 (defun doom/help-search (&optional initial-input)
   "Preform a text search on all of Doom's documentation."
   (interactive)
-  (funcall (cond ((fboundp '+ivy-file-search)
+  (funcall (cond ((featurep! :completion ivy)
                   #'+ivy-file-search)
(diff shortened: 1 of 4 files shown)


hlissner:
Thanks for the PR, but I cannot accept it. The avoidance of `featurep!` in these cases are intentional, to tie functionality to the presence of functions or packages, rather than modules. That said, I'm working on a layer of indirection (generics) to replace all these checks, and that is the direction I'd like to take this instead. Thanks for bringing it to my attention in any case.

ymarco:
These current `fboundp` checks are only right if e.g the ivy module is the sole reason for counsel functions to be loaded. 
That is not the case! For example, lispy has a dependency on swiper.
I had issues while testing the selectrum module where things like `+default/yank-pop` still mapped to counsel function because they were loaded by lispy.

iyefrat:
I've added a temporary fix in the linked commit, but presumably there is a better way to go about it.

PR #5023, opened 2021-05-10
ophints: Add hooks for lispyville

Grazfather:
Because lispyville (depending on how you configure it) replaces some
evil-mode mappings with its own function, we lose the indication of our
deletions/yanks/etc. This change adds all lispyville functions to the
evil-goggles--commands list.

### DIFF
--- a/modules/ui/ophints/config.el
+++ b/modules/ui/ophints/config.el
@@ -23,7 +23,45 @@
             '(+eval:region
               :face evil-goggles-yank-face
               :switch evil-goggles-enable-yank
-              :advice evil-goggles--generic-async-advice)))
+              :advice evil-goggles--generic-async-advice))
+  (when (featurep! :editor lispy)
+    (pushnew! evil-goggles--commands
+              '(lispyville-delete
+                :face evil-goggles-delete-face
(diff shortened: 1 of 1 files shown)


modules/ui/ophints/config.el
```diff
@@ -11,19 +11,57 @@
...
-              :advice evil-goggles--generic-async-advice)
-            '(+evil:yank-unindented
-              :face evil-goggles-yank-face
-              :switch evil-goggles-enable-yank
-              :advice evil-goggles--generic-async-advice)
-            '(+eval:region
-              :face evil-goggles-yank-face
-              :switch evil-goggles-enable-yank
-              :advice evil-goggles--generic-async-advice)))
+  (progn (pushnew! evil-goggles--commands
```
hlissner, line 14: `:config`'s body is implicitly wrapped in a progn, so this is redundant. Please remove it.

hlissner (changes requested):

modules/ui/ophints/config.el
Grazfather: Done

hlissner (approved):

hlissner:
Thanks!

PR #5024, opened 2021-05-10
Bump format-all.

0x450x6c:


hlissner:
Thanks for the PR, but I cannot accept it. `format-all`'s pin is intentionally locked because of breaking changes upstream that are incompatible with the module, and because the module is due to be rewritten with [apheleia](https://github.com/raxod502/apheleia).

PR #5037, opened 2021-05-11
Fix +vterm-toggle not changing directory to project root

### LINKED ISSUE #4834: vterm-toggle-and-cd
I really like the behavior of vterm-toggle.  I just wish it would also cd the project dir when it opens everytime, because when I summon it it is the directory I want to work on.
```elisp
(after! vterm
  (defun mog_cd_for_toggle (arg)
    (vterm-send-string (concat "cd " arg) t)
    (vterm-send-return)
    )
  (defun my_cd_on_toggle (oldfun &rest arg)
 (let* ((project-root (or (doom-project-root) default-directory))
         (default-directory
           (if arg
               default-directory
             project-root))
         display-buffer-alist)
     (apply oldfun arg)
    (vterm-send-C-a)
    (vterm-send-C-k)
    (sleep-for 0.01)
    (if
        (vterm-cursor-in-command-buffer-p)
        (mog_cd_for_toggle project-root)
      )
))
  (advice-add #'+vterm/toggle :around #'my_cd_on_toggle)
)
```

with that it mostly works, but now for some reason the split behavior is different than the normal one, doing top down splits instead of the bottom popup window behavior it used to have. it is also destroying the window when it closes messing with my layout. any idea how I could achieve this it would be appreciated.

**System information**
<!--
  Post the output of `M-x doom/info` or `~/.emacs.d/bin/doom info` to
  pastebin.com and paste the link here.
<!--
https://pastebin.com/6LLMbgxW

sggutier:
Fixes #4834 

Before this change, `+vterm/toggle` would toggle a vterm window and change directory to the directory of the current buffer.  Now it oppens the root of the current project when it does't get an universal argument.

The comments of the function indicate this should be the appropriate behavior.  `+vterm/here` behaves in a similar way.

Before:

output

After:

output2

### DIFF
--- a/modules/term/vterm/autoload.el
+++ b/modules/term/vterm/autoload.el
@@ -30,12 +30,17 @@ If prefix ARG is non-nil, recreate vterm buffer in the current project's root."
             (evil-change-to-initial-state))
           (goto-char (point-max)))
       (setenv "PROOT" (or (doom-project-root) default-directory))
-      (let ((buffer (get-buffer-create buffer-name)))
+      (let* ((project-root (or (doom-project-root) default-directory))
+         (default-directory
+           (if arg
+               default-directory
+             project-root)))
+        (let ((buffer (get-buffer-create buffer-name)))
         (with-current-buffer buffer
           (unless (eq major-mode 'vterm-mode)
             (vterm-mode))
           (+vterm--change-directory-if-remote))
-        (pop-to-buffer buffer)))))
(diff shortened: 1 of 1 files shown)


otavioschwanck (approved):

hlissner (approved):

hlissner:
LGTM. I'll look into mirroring this behavior to the other term modules. Anyhow, thanks for the help!

PR #5058, opened 2021-05-16
:ui modeline: Fix missing arg to +modeline-format-icon function

ericdallo:
Fix for #3307 

Because of a wrong conflict solve on #3307, I missed one function to add the new arg


hlissner (approved):

hlissner:
Thanks!

PR #5062, opened 2021-05-17
Rust/Ivy Module Docs

fosskers:

This PR tweaks the Rust installation instructions and fixes a few minor typos.

### DIFF
--- a/modules/completion/ivy/README.org
+++ b/modules/completion/ivy/README.org
@@ -28,7 +28,7 @@
 
 * Description
 This module provides Ivy integration for a variety of Emacs commands, as well as
-a unified interface for project search and replace, powered by ripgrep.
+a unified interface for project search and replace, powered by [[https://github.com/BurntSushi/ripgrep/][ripgrep]].
 
 #+begin_quote
 I prefer ivy over ido for its flexibility. I prefer ivy over helm because it's
@@ -76,7 +76,7 @@ brew install ripgrep
 
 *** Arch Linux
 #+BEGIN_SRC sh :dir /sudo::
-sudo pacman --needed --noconfirm -S ripgrep
+sudo pacman -S ripgrep
 #+END_SRC
 
 *** openSUSE
--- a/modules/lang/rust/README.org
+++ b/modules/lang/rust/README.org
@@ -9,13 +9,17 @@
   - [[#plugins][Plugins]]
   - [[#hacks][Hacks]]
 - [[#prerequisites][Prerequisites]]
+  - [[#install][Install]]
+    - [[#arch-linux][Arch Linux]]
+    - [[#general][General]]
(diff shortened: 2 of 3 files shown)


modules/lang/rust/README.org
```diff
@@ -72,19 +100,19 @@ Enable the [[file:../../../modules/editor/format/README.org][:editor format]] mo
...
+| ~<localleader> b b~ | ~cargo build~      |
+| ~<localleader> b B~ | ~cargo bench~      |
+| ~<localleader> b c~ | ~cargo check~      |
+| ~<localleader> b C~ | ~cargo clippy~     |
+| ~<localleader> b d~ | ~cargo doc~        |
+| ~<localleader> b n~ | ~cargo update~     |
+| ~<localleader> b o~ | ~cargo outdated~   |
+| ~<localleader> b r~ | ~cargo run~        |
+| ~<localleader> t a~ | ~cargo test~       |
+| ~<localleader> t t~ | ~run current test~ |
```
fosskers, line 115: I swore I trashed this change. Fixing...

modules/completion/ivy/README.org
```diff
@@ -76,7 +76,7 @@ brew install ripgrep
 
 *** Arch Linux
 #+BEGIN_SRC sh :dir /sudo::
-sudo pacman --needed --noconfirm -S ripgrep
+sudo pacman -S ripgrep
```
fosskers, line 79: While `--needed` and `--noconfirm` work just fine, it's a bit unusual to use them when installing a single package.

modules/lang/rust/README.org
```diff
@@ -37,14 +41,38 @@ e.g. ~cargo~.
...
-~curl https://sh.rustup.rs -sSf | sh~
+** Install
+
+This module requires ~rust~, which can be acquired through =rustup=.
+
+*** Arch Linux
+
+#+begin_src sh
+sudo pacman -S rustup
+#+end_src
```
gagbo, line 58: Installing a package manager through your package manager is almost always a bad idea. What makes it better with rustup ? Why would that be better than just using the maintained upstream script ?

fosskers: It is general Arch Linux ethos to register as much as possible through the package manager: https://wiki.archlinux.org/title/Rust#Rustup

iyefrat: I don't have experience on the rust side, but trying to install `ghcup` through `pacman` does not exactly lead to good things.

fosskers: We will have to take that into account when the Haskell Foundation does their planned Unified Installer overhaul.

hlissner (approved):

hlissner:
Thanks for the help!

PR #5063, opened 2021-05-17
Bump :ui doom-themes

prashantvithani:
hlissner/emacs-doom-themes@4199e74 -> hlissner/emacs-doom-themes@4d24728


hlissner:
A larger bump of the module was needed, which was done in a4d2e87. Thanks for bringing it to my attention!

PR #5066, opened 2021-05-18
Prevent electric-quotes in org src-blocks.

### LINKED ISSUE #5051: Electric-quote-mode doesn't play nice with org-mode src-blocks and smartparens-mode
### 

- [X] I have searched [the Discourse](https://discourse.doomemacs.org/docs) for any errors and solutions.

After enabling electric-quote-mode, quotes are converted to curly quotes as you type them. A very nice feature for text-derived modes.

Unfortunately, it also does this in org-mode src blocks, without additional configuration.

`(add-hook 'electric-quote-inhibit-functions 'org-in-src-block-p)` _should_ fix it so that electric-quotes work in org-mode buffers but get disabled when typing inside org-mode src-blocks. This is confirmed to work in a vanilla Emacs configuration.

### What actually happened?

But in a fresh Doom installation, it causes odd behavior inside of src blocks. Even though the quotes should remain straight with the added hook, they still get converted, and in an inconsistent manner.

It appears to be caused by smartparens-mode changing the status quo just before electric-quote-mode does its thing when via post-self-insert-hook.

A work-around is to change the depth of the electric-quote-post-self-insert-function, so that it runs _before_ sp--post-self-insert-hook-handler, instead of after. But I'm not sure if that would be an acceptable way to fix it in Doom, or if it's even desirable to fix it, since it's not enabled by default.

If it would be helpful, I'll be glad to submit a pull-request. Seems like "modules/emacs/electric/config.el" would be an appropriate place to put it. But if it's not something you want to support, feel free to close this issue. The work-around seems to get it done.

### Steps to reproduce

Screenshot from 2021-05-14 00-57-03

Start with a fresh Doom install. 
Follow steps shown in the attached screenshot:
Enable electric-quote-mode. 
Type some quotes inside a src block and see that they're being converted, which you don't want.
The solution should be to do `(add-hook 'electric-quote-inhibit-functions 'org-in-src-block-p)`, which works in a vanilla Emacs install.
But it doesn't fix the problem. It alters the behavior slightly, but quotes are still converted inside the src block.
As shown in the screenshot, a workaround is to change the depth of the electric-quote-mode hook so that it loads before the smartparens hook.

### System Information

https://pastebin.com/R029we93

mmarshall540:
Fixes #5051

Added hook to prevent electric-quote-mode from converting quotes when point is inside an org src-block.

In the linked issue, I had tried using "org-in-src-block-p" as the inhibiting function. It worked, but only after rearranging the functions in post-self-insert-hook so the electric-quote function (electric-quote-post-self-insert-function) would run before the smartparens function (sp--post-self-insert-hook-handler).

Looking deeper into the problem, I found that "org-in-src-block-p" isn't always accurate, because it relies on the "src-block" property of the char at point. But that property is getting switched to nil by sp--post-self-insert-hook-handler (specifically, by the call to sp-insert-pair). 

I think this might be caused by the "src-block" text property getting inherited from the quote that smartparens inserts in front of our own typed quote mark. Not sure.

Unlike "org-in-src-block-p", "org-in-block-p" doesn't rely on an abstracted text property. It examines the text surrounding point to determine if it's within a src-block. So it should more reliably prevent electric-quotes from sullying our src-blocks. (I appreciate any suggestions you might have.)

### DIFF
--- a/modules/emacs/electric/config.el
+++ b/modules/emacs/electric/config.el
@@ -15,4 +15,7 @@ current line.")
       (when (and (eolp) +electric-indent-words)
         (save-excursion
           (backward-word)
-          (looking-at-p (concat "\\<" (regexp-opt +electric-indent-words))))))))
+          (looking-at-p (concat "\\<" (regexp-opt +electric-indent-words)))))))
+
+;; Prevent electric-quotes in org src-blocks.
+  (add-hook 'electric-quote-inhibit-functions (lambda () (org-in-block-p '("src")))))


mmarshall540:
I think I could further improve this by having it run only when org-mode is active.

As-submitted, the extra text-processing only occurs with electric-quote-mode enabled and after a ', ", or ` has been typed.

But there's no reason for it to run at all outside of an org-mode buffer.

hlissner:
I'm currently trying to decide between this or the following:

```lisp
;; in lang/org/config.el
(add-hook! 'org-mode-hook
  (add-hook 'electric-quote-inhibit-functions 'org-in-src-block-p nil t))

;; in core/core-editor.el, in `smartparens` use-package :config
;; Ensure sp--post-self-insert-hook-handler always runs last
(remove-hook 'post-self-insert-hook 'sp--post-self-insert-hook-handler)
(add-hook 'post-self-insert-hook 'sp--post-self-insert-hook-handler 100)
```

Trouble is, the numeric DEPTH argument for `add-hook` is a 27.x feature, and we still support 26.x (at least until the end of June).

mmarshall540:
> I'm currently trying to decide between this or the following:
> 
> ```emacs-lisp
> ;; in lang/org/config.el, in `+org-init-hacks-h'
> (add-hook! 'org-mode-hook
>   (add-hook 'electric-quote-inhibit-functions 'org-in-src-block-p nil t))
> 
> ;; in core/core-editor.el, in `smartparens' use-package block (in :config)
> ;; Ensure sp--post-self-insert-hook-handler always runs last
> (remove-hook 'post-self-insert-hook 'sp--post-self-insert-hook-handler)
> (add-hook 'post-self-insert-hook 'sp--post-self-insert-hook-handler 100)
> ```
> 
> Trouble is, the numeric DEPTH argument for `add-hook` is a 27.x feature, and we still support 26.x (at least until the end of June).

I think that's better, because it solves a potential problem going forward (should some other hook relies on a text property that smartparens indirectly changed).

The DEPTH argument replaces APPEND from 26.x which--since 100 is non-nil--would still put it at the end of the hook, no?

hlissner:
> The DEPTH argument replaces APPEND from 26.x which--since 100 is non-nil--would still put it at the end of the hook, no?

They work a little differently. In 26.x, `APPEND` appends to the hook once at insertion. but `DEPTH` maintains order in perpetuity; i.e. `add-hook` resorts the hook on every insertion.

So if electric loads after smartparens, we're back to square 1.

Although... now that I test, `electric` _is_ loaded immediately at startup so this may be a non-issue...

mmarshall540:
I think I finally solved it.  

It must be obvious that I'm still learning git (and GitHub).  2fc0be6 is the one I was trying to submit. (Also, I'm not sure how to back out the first PR, which was a change to a different file.)

This goes at the end of  +org-init-hacks-h in lang/org/config.el.

```
;; Disable electric-quote-mode when inside of org-mode src-blocks.
(add-hook! 'org-mode-hook
    (add-hook
     'electric-quote-inhibit-functions
     (lambda () (unless (eobp)
                  (or
                   (eq (get-char-property (+ (point) 1) 'src-block) t)
                   (eq (get-char-property (- (point) 1) 'src-block) t)))) nil t))
```

It avoids needing to deal with the order of functions on post-self-insert-hook. Unlike org-in-block-p, it doesn't have to search for src-block fences. It does what _I think_ is an inexpensive check of the char-properties at 2 positions.

Smartparens inserts a closing quotation mark and then moves point behind it. The extra quotation mark doesn't possess the src-block property at that moment. But since point moves behind it, org-in-src-block-p would use that char to decide whether it's in a src-block.

Instead, this checks 1 position forward and one position backward, and if we're at the end of the buffer it just returns nil to avoid an out-of-range error.

mmarshall540:
So we can also fix this for in-line src code. But I think before submitting this, I'll study how to update a pull request.

```emacs-lisp
(add-hook! 'org-mode-hook
  (add-hook
   'electric-quote-inhibit-functions
   (lambda () (or
               (unless (eobp)
                 (or
                  (eq (get-char-property (+ (point) 1) 'src-block) t)
                  (eq (get-char-property (- (point) 1) 'src-block) t)))
               (looking-back ".*src_[-+a-zA-Z]*{[^}]*" (line-beginning-position)))) nil t))
```
test-eq screenshot

mmarshall540:
Am I doing this right? I was hoping to consolidate everything into a single commit for your review, but after reviewing the git docs I'm still not sure how. I really **don't** want to waste your time with a bunch of _crufty_ commits.

This last commit inhibits electric-quotes both inside of org-mode src-blocks and within org-mode's in-line source "snippets" (my own phrasing).

It no longer relies on hook ordering, so it should work for all supported Emacs versions.

mmarshall540:
Should we close this? The last couple commits fixed what we discussed, but I'm still not sure why there ended up being so many other commits added to it.

I can resubmit it as a single commit if you want.

PR #5070, opened 2021-05-19
Update suggested clangd flags in lang/cc docs

jpcima:
This adds a missing clangd flag for LSP, needed to make it work correctly with company-mode.
It's from the default value of `lsp-clients-clangd-args`.
This is with `clangd version 11.1.0`.

If not set, I find C++ completion will produce a single space instead of showing up the list, in a case like `myVariable.`, triggering completion after the dot character.


hlissner (approved):

hlissner:
LGTM. Thanks!

PR #5073, opened 2021-05-20
Bump geiser-guile

pricly-yellow:
Bump :lang scheme

emacs-geiser/guile@700ac985 -> emacs-geiser/guile@8dda28f4

fix geiser for guile 3.0.6


hlissner:
The whole module was overdue a bump, rather than `geiser-guile` alone, so it was bumped in e3bcd04. Thanks for bringing it to my attention!

PR #5075, opened 2021-05-20
featuregate ivy dependent autoloads for irc, pass

iyefrat:
previously these functions would get autoloaded even when another
completion framework was active


hlissner (approved):

hlissner:
Thanks!

PR #5077, opened 2021-05-21
Fix typo in the comments

raszi:
Fixed a typo in the comment section so nobody would copy the wrong example over.


hlissner:
This error was present in several other places, which I've corrected in d9dacce. Thanks for bringing them to my attention!

raszi:
Thank you for fixing it!

PR #5079, opened 2021-05-22
define-localleader-key's featurep -> featurep!

jgrey4296:
Replace the use of featurep with featurep! in `define-localleader-key!`

This stops a race condition where a module with `(map! :map a-map :localleader ....` is loaded
before `evil` has been activated, so the macro expands to use the `doom-local-leader-alt-key` prefix.


hlissner (approved):

hlissner:
Thanks!

Issue #5053, opened 2021-05-15
Composed `?.` ligature in the middle of typescript-tsx-mode code


shackra:
### 

- [X] I have searched [the FAQ](https://doomemacs.org/docs/faq.org) and [the documentation](https://doomemacs.org/docs/getting_started.org) for solutions.
- [X] I have searched [the issue tracker](https://github.com/hlissner/doom-emacs/issues) for similar issues (the closed issues too).
- [X] I have searched [the Discourse](https://discourse.doomemacs.org/docs) for any errors and solutions.
- [X] I can reproduce my issue on [the latest commit](https://github.com/hlissner/doom-emacs/commit) of Doom Emacs.
- [X] I have read [How to Debug Issues](https://discourse.doomemacs.org/how2debug) and found no solution.

### What did you expect to happen?

I did not expect any ligature, or, in this case, that the ligature didn't screw readability

### What actually happened?

in typescript-tsx-mode you can see the follow if you write `?.`

![image](https://user-images.githubusercontent.com/1055216/118347845-6214ed00-b503-11eb-83ba-d4e69dc7cf11.png)

### Describe your attempts to resolve the issue

I haven't touch my configuration for these ligatures, so I don't really know how to solve it

### Steps to reproduce

1. open Emacs
2. open a `.tsx` file
3. write `data?.`
4. voila

### System Information

https://pastebin.com/SVA4K7AE

hlissner:
Try modifying [this line](https://github.com/hlissner/doom-emacs/blob/develop/modules/ui/ligatures/config.el#L191) in `~/.emacs.d/modules/ui/ligatures/config.el`:

```diff
       (dolist (char-regexp +ligatures-composition-alist)
         (set-char-table-range
          +ligature--composition-table
-         (car char-regexp) `([,(cdr char-regexp) 0 font-shape-gstring])))
+         (car char-regexp) `([,(concat "." (cdr char-regexp)) 0 font-shape-gstring])))
       (set-char-table-parent +ligature--composition-table composition-function-table))))
```

And see if it fixes your issue.

shackra:
yes, it does fix the issue

hlissner:
As of 5406745 this issue should be resolved. Let me know if that isn't the case and I'll reopen it. Thanks for bringing it to my attention!

Issue #4922, opened 2021-04-25
Ligatures, Org-mode hide leading stars, and Emacs 28


jcf:
**TL;DR** I'm looking into how to remove the triple-asterisk ligature from Org-mode with Emacs 28 because I like leading "stars" to be hidden. I've found the `composition-function-table`, and the module docs on adding ligatures, just not sure if this is the desired behaviour… for now I'm working around the issue by enabling `+pretty`, with the following org-superstar configuration:

``` elisp
(after! org-superstar
  (setq
   org-superstar-headline-bullets-list '("#")
   org-superstar-special-todo-items nil))
```

**What did you expect to happen?**

With `org-hide-leading-stars` set to true, I'd expect leading stars to only show up as a single asterisk.

**What actually happened?**

Multiples of three asterisks appear as a cute ligature that kinda looks like a crown (with [JetBrains Mono](https://www.jetbrains.com/lp/mono/)).

<img width="133" alt="image" src="https://user-images.githubusercontent.com/18374/115988312-fea53880-a5b0-11eb-9fa6-6e0ebb60f6ee.png">

**Additional details:**

This happens with both `ligatures` and `(ligatures +extra)` enabled, on both macOS and Linux, built from a recent native-comp Emacs master.

```
GNU Emacs 28.0.50 (build 1, x86_64-apple-darwin19.6.0, NS appkit-1894.60 Version 10.15.7 (Build 19H114)) of 2021-01-06
```

If I add `+pretty` to my Org module flags, the pretty stars plugin replaces the plain asterisks, and the ligatures are no longer applied.

```
(org +dragndrop +gnuplot +noter +pandoc +pomodoro +present +roam)
```

**System information:**

https://pastebin.com/ZmFY8jeL

gagbo:
You basically won't have anyway of changing this without getting your feet wet, and I (very personal opinion) shouldn't be changed in doom-emacs but left to the tinkerer.

Changing the value of the composition function table alist
https://github.com/hlissner/doom-emacs/blob/f621ff80471e8d08a72e5ece00641c70b121873a/modules/ui/ligatures/config.el#L50-L79
to something that only matches the ligatures that you want, in the modes that you want is going to be the only way to exclude ligatures included in your font from being considered for ligation by Emacs. The "ugly" regexes have comments with the human readable comment that generated them, so you can look for the entry that has `?*` as key, redo the `regex-opt` call with the ligature excluded, and then replace the value in the variable.

[...]

PR #5082, opened 2021-05-23
feat(ligatures): use ligature.el for Emacs28+

gagbo:
As discussed briefly on Discord + Discourse, use https://github.com/mickeynp/ligature.el to handle the composition table for ligatures if using Emacs28+ with Harfbuzz.

This gives an easier API to handle per-user/per-mode configuration (potentially dealing with #4922 ), and all potential fixes can be targeted upstream instead of maintaining the code here (See #5053 ).

All example configuration values is coming from suggestions in upstream package, not much has been changed to enable discussion :

- Should we keep the "all-prog-mode" `defvar` as is ?
- Should extra ligatures go in their `:lang` modules ? (If so, this PR will _not_ make it, better have other contributions to their respective modules)
- Should the `defvar` keep track of all the ligatures available for each font, in a `+fira/+iosevka...` way ? (If so, this PR will _not_ make the per font change, but I'll try to find a minimal set for the `defvar`)
- Should there be a "all-text-mode" `defvar` as well ? (looking at you, `markdown-mode` and `org-mode`)
  Currently the decision is "no", and instead let users customize ligatures in text-modes with `set-ligatures!`
- Should there be a "t" `defvar` as well ?
  Currently done, with a `nil` default value

PS: thanks @mickeynp for that package, it's been a good relief when we found out about it


gagbo:
Added `ligature.el` manipulation in `set-ligatures!` function so that users keep a single entrypoint.

As discussed, it’s pretty hard to handle inheritance of derived modes, so I think the best way to do it is to allow users to override the only configuration that writes to "high priority" modes through `+ligatures-all-modes-list` and `+ligatures-prog-mode-list` and then only use `set-ligatures!` if they want.

gagbo:
Maybe the `defvar` could be a plist or an alist, so that instead of having multiple variable for `t`, `prog-mode`, `text-mode` etc., we’d have a single list on which we could mapcar `ligature-set-ligatures (car elt) (cdr elt)`.

Not sure if that’s the better API, but I guess it unclutters the namespace

gagbo:
(pong ; don't stale me bot)

I don't remember the blocker, it was something around the API of `set-ligatures!` that didn't integrate `ligatures.el` and `prettify-symbols-mode` seamlessly enough, but I don't remember what

VitalyAnkh:
I have been using this branch for a few days. Emacs runs well and some ligature bugs(like ligature characters took more width than they actually need) seem to disappear. Looking forward to this get merged!

hlissner (approved):

modules/ui/ligatures/README.org
```diff
@@ -213,15 +215,14 @@ want to use.
 
 #+begin_src elisp
 (after! ligature
-  (ligature-set-ligatures '(haskell-mode clojure-mode) '(">>=")))
+  (ligature-set-ligatures '(haskell-mode clojure-mode) '(">>=" ">>-")))
```
hlissner, line 218: Please remove `ligature-set-ligatures` from the examples; I consider it an implementation detail. No harm mentioning that `set-font-ligatures!` is a thin wrapper around it though.

modules/ui/ligatures/autoload/ligatures.el
```diff
@@ -6,61 +6,71 @@
...
+
+Note that this will keep all ligatures in `+ligatures-prog-mode-list' active, as
+`emacs-lisp-mode' is derived from `prog-mode'."
+  (declare (indent defun))
+  (if (null ligatures)
+      (dolist (mode (ensure-list modes))
+        (add-to-list 'ligature-ignored-major-modes mode))
+    (with-eval-after-load 'ligature
+              (dolist (mode (ensure-list modes))
+                (setq ligature-ignored-major-modes (delete mode ligature-ignored-major-modes)))
```
hlissner, line 75: Use `delq` here, considering `mode` will always be a symbol.

modules/ui/ligatures/autoload/ligatures.el
```diff
@@ -6,61 +6,71 @@
...
+              (ligature-set-ligatures (ensure-list modes) font-ligatures))))
```
hlissner, line 76: Please reindent this region; it's hanging quite a bit.

modules/ui/ligatures/autoload/ligatures.el
```diff
@@ -6,61 +6,71 @@
...
+
+  (set-font-ligatures! \\='emacs-lisp-mode nil)
+
+Note that this will keep all ligatures in `+ligatures-prog-mode-list' active, as
+`emacs-lisp-mode' is derived from `prog-mode'."
+  (declare (indent defun))
+  (if (null ligatures)
+      (dolist (mode (ensure-list modes))
+        (add-to-list 'ligature-ignored-major-modes mode))
+    (with-eval-after-load 'ligature
```
hlissner, line 73: Use `after!` here, instead of with-eval-after-load (since it does a little magic to avoid eager byte-compilation of the block, as well as to handle the case where `ligature` is disabled by the user (however unlikely that may be).

```diff
-(with-eval-after-load 'ligature
+(after! ligature
```

modules/ui/ligatures/README.org
```diff
@@ -246,8 +247,7 @@ happen in all modes, you can use
   (ligature-set-ligatures 'eww-mode '("ff" "fi" "ffi")))
```
hlissner, line 247: Same as above (re: `ligature-set-ligatures`)

hlissner (changes requested):
Looks good! Just a few minor changes and I think it's good to go.

hlissner (changes requested):
Ah, forgot one other request: could you reword 63cb3d621afb7de4ef267e23548ffcc1c402858a and dd0ebe087f159e7c56d13ffe2cbfe6931ed87c3b to include a `BREAKING CHANGE:` in their body? Mainly to comment on what will break and how to work around it (if at all).

modules/ui/ligatures/autoload/ligatures.el
gagbo: I'll do this manually as I can, parinfer-rust-mode still has this outstanding bug where having parens in a docstring will demolish your code.

gagbo:
> Ah, forgot one other request: could you reword [63cb3d6](https://github.com/doomemacs/doomemacs/commit/63cb3d621afb7de4ef267e23548ffcc1c402858a) and [dd0ebe0](https://github.com/doomemacs/doomemacs/commit/dd0ebe087f159e7c56d13ffe2cbfe6931ed87c3b) to include a `BREAKING CHANGE:` in their body? Mainly to comment on what will break and how to work around it (if at all).

I just squashed everything, I mostly left the commits like that to remember and revert parts of it if necessary. I don't think it makes a lot of sense to let this be multiple commits on master.

hlissner (approved):

hlissner:

PR #5085, opened 2021-05-23
Add ranger documentation to dired README

j3010:
I added a small ranger section to "/modules/emacs/dired/README.org". This section simply explains that if the +ranger flag for dired is set in init.el often a "deer-mode" or minimal ranger buffer will be opened, and that "Z p" can be used to toggle between this minimal mode and the full ranger mode. I added this because I think deer-mode looks deceptively similar to standard dired-mode and it may not be obvious what is occurring and how to switch to full ranger if the +ranger flag was set without reading the ranger.el docs.


hlissner (approved):

hlissner:
Thanks!

PR #5094, opened 2021-05-25
Bump :app irc

iyefrat:
jorgenschaefer/circe@2f70aa2 -> jorgenschaefer/circe@31ac6ba

adds Libera Chat as the default network


hlissner:
Circe had moved on ahead in the time it took me to review this, so I bumped it (and the other `:app` modules in 05d85f5). Thanks for bringing it to my attention!

PR #5098, opened 2021-05-26
Add :not counsel-org-goto to ivy-prescient-sort-commands

daanturo:
Although `counsel-org-goto` is an alias of `counsel-outline`, while the latter is already in the list, the former's output is still sorted.

Github's "Files changed" is too noisy, here is Magit's more friendly diff:
image


hlissner (approved):

hlissner:
Thanks!

PR #5100, opened 2021-05-26
feat(magit): Add orgit-forge package

Emiller88:
The title says it all.

https://github.com/magit/orgit-forge

Gives a nice link for issues to throw into your TODO list.

It might need a `:after org` call as well or instead of.


hlissner:
I've moved the `package!` to `:lang org`, next to `orgit`'s. No extra configuration is needed since [orgit-forge autoloads its entry point](https://github.com/magit/orgit-forge/blob/master/orgit-forge.el#L47-L52). Thanks for bringing it to my attention!

PR #5103, opened 2021-05-27
lang/rust: Binding for `cargo doc`

fosskers:

`cargo doc --open` (the previous binding of `SPC m b d`) always opens a new
browser tab. For "doc-driven development" this isn't so nice, since we'd like to
stay in the same tab and just refresh its contents after editing. That's what
`cargo doc` (without `--open`) does, which is called by
`#'rustic-cargo-build-doc`.

This commit introduces a binding for the latter behaviour, but asigns it to the
old `SPC m b d`, with the "open" variant as `SPC m b D`. This matches Spacemacs
and follows the usual pattern of "the capital letter variant is the one you use
less often".

Of course if you'd rather not the existing binding be flipped, I'm perfectly willing to switch it.

### DIFF
--- a/modules/lang/rust/config.el
+++ b/modules/lang/rust/config.el
@@ -45,16 +45,17 @@
   (map! :map rustic-mode-map
         :localleader
         (:prefix ("b" . "build")
-          :desc "cargo audit"    "a" #'+rust/cargo-audit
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Thanks!

PR #5112, opened 2021-05-30
Add feature to ocaml module to allow using prettifying-symbols while bypassing tuareg's symbol alist

shonfeder:
The current configuration of the `ocaml` module seems to override the support for ligatures coming from Fira code (and Iosevka) via `prettify-symbols-mode`. The current work-around I've found is to add a feature to disable the "hack" (according to [the documentation](https://github.com/hlissner/doom-emacs/blob/develop/modules/lang/ocaml/README.org#hacks)) of using `set-ligatures!` on the tuareg prettify symbols alist. Which this disabled, I am then free to add what I want in my personal config to get the kind of prettification I want[1].

In general, I think some users of the `ocaml` module will appreciate the ability to use a different symbols for prettifying than that supplied by tuareg, since tuareg's do not maintain character width, and the set is relatively limited. 

I don't think the solution here is the **correct** way, but just an illustration of the effect I am after. I expect that the reviewer will be able to instruct me on the preferred way of achieving the desired outcome which staying consistent with the excellent design philosophy and conventions of the doom project.

I have implemented the bypass currently as a "negative" feature, for the sake of maintaining backwards compatibility with current users. Should I just switch this into a "postive" feature? Should I came at this from a very different angle?

---

[1] Currently, this means adding 

```emacs-lisp
(add-hook! prog-mode
  ;; Enable fira-code ligatures
  (fira-code-mode)
  (prettify-symbols-mode-set-explicitly))
``` 

not nice, but it's working.

### DIFF
--- a/modules/lang/ocaml/README.org
+++ b/modules/lang/ocaml/README.org
@@ -26,7 +26,11 @@ This module adds [[https://ocaml.org/][OCaml]] support to Doom Emacs, powered by
 + Dune file format ([[http://dune.build/][dune]])
 
 ** Module Flags
-This module provides no flags.
++ ~-pretty~ Disable =tuareg-mode='s default ligature configuration. This is
+  useful when you would prefer to use the ligature support provided by a font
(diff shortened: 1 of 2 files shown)


shonfeder:
Closing as an overly messy hack :)

PR #5122, opened 2021-06-01
Add evil-keybindings for various sharper commands

RBckmnn:
When using evil the keybinds defined in the sharper package were not available. This PR maps sharper commands in a way that is mostly analogous to the keybinds originally defined in the package. I changed the bindings for various refresh commands to "gr", however, to be in line with similar other facilities in Doom.


hlissner:
This should probably be upstreamed to evil-collection at some point, but it can live here for now. Thanks for the help!

PR #5123, opened 2021-06-01
doctor: Warn if `shfmt` is missing

fosskers:

This PR adds a check to `doom doctor` for a missing `shfmt` executable, but only if `:editor format` is installed. I should probably alter my last PR to account for `format` (and check for other missing ones).


hlissner (approved):

hlissner:
As a follow up to https://github.com/hlissner/doom-emacs/pull/5135#issuecomment-854891532, this needs to be tackled in one place. That said, perhaps we can add shfmt/prettier/etc to the README.org of their particular modules, but the checks themselves should probably live in the `:editor format` module.

fosskers:
Sounds good, thanks.

PR #5135, opened 2021-06-04
doctor: Warn if `prettier` is missing for markdown

### LINKED ISSUE #5130: prettier should show up as markdown mode's dependencies

doom doc should suggest me to install prettier when markdown support is enabled. perhaps doom should know to not try and format the buffer when prettier is missing.

### What actually happened?

doom doc doesn't say I have missing dependency for markdown, but every time I want to save a markdown document an error pop up with 
```
Error (before-save-hook): Error running hook "+format-buffer-h" because: (error You need the "prettier" command. You may be able to install it via "npm install --global prettier @prettier/plugin-php prettier-plugin-solidity".)
```

Installing prettier seems to make doom happy.

### Steps to reproduce

enable markdown support without having prettier in the system

### System Information

https://pastebin.com/UWDyNAkS

fosskers:

Fixes #5130 

In the same vein as #5123, this will warn users who have both `markdown` and `format` active, but are missing the `prettier` executable.


hlissner (approved):

hlissner:
Thanks for the PR, but I cannot accept it. `prettier` is used for a number of modules/modes, not just markdown. I also haven't decided whether these checks should be done in the format module, or in each of the `:lang` modules that use it. I'll look into it further elsewhere.

fosskers:
What about #5123 then?

PR #5140, opened 2021-06-05
Add yasnippet minor mode when beancount is on

ErnestDong:
e.g.
```
# -*- mode: snippet -*-
# name: today's bean
# key: bean
# --
`(format-time-string "%Y-%m-%d" (current-time))` * "" "${1:Things}"
${2:account}
```

Fixes #0000 


hlissner:
Thanks for the PR, but this would error out if the user has `:editor snippets` disabled. In any case, this was indirectly fixed in 60b5953.

PR #5141, opened 2021-06-05
lang/web: Support go & hugo engines in web-mode

sei40kr:
Set an appropriate engine when editing a Go template or Hugo template in web-mode.

### DIFF
--- a/modules/lang/web/+html.el
+++ b/modules/lang/web/+html.el
@@ -36,6 +36,13 @@
       ("-xml" (memq major-mode '(nxml-mode xml-mode))))
     :ok-statuses '(0 1))
 
+  (set-formatter! '+web-go-template
+    #'(format-all--buffer-native 'web-mode
+                                 (lambda ()
+                                   (indent-region (point-min) (point-max))))
+    :modes '((web-mode (or (bound-and-true-p +web-go-template-mode)
+                           (bound-and-true-p +web-hugo-mode)))))
+
   (setq web-mode-enable-html-entities-fontification t
         web-mode-auto-close-style 1)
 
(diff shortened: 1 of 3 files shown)


modules/lang/web/+html.el
```diff
@@ -36,6 +36,13 @@
       ("-xml" (memq major-mode '(nxml-mode xml-mode))))
     :ok-statuses '(0 1))
 
+  (set-formatter! '+web-go-template
+    #'(format-all--buffer-native 'web-mode
+                                 (lambda ()
+                                   (indent-region (point-min) (point-max))))
```
sei40kr, line 42: For now, there's no perfect formatter for Go templates. I tried Prettier & prettier-plugin-go-template, but it often breaks syntax.

modules/lang/web/autoload/html.el
```diff
@@ -132,3 +132,11 @@ snippet, or `emmet-expand-yas'/`emmet-expand-line', depending on whether
               #'yas-expand
             #'emmet-expand-yas))
          (#'emmet-expand-line))))
+
+;;;###autoload
+(defun +web--go-template-format-buffer ()
```
sei40kr, line 137: I tried to implement this as `lambda` in `set-formatter!` but I couldn't make it work. So I implemented an interactive command instead.

sei40kr:
@hlissner Could you review this PR?

hlissner:
I'm closing this for the same reason #3011 was closed; I am not rejecting these changes, but pending changes to `:editor format` (in #6369) need to be accommodated for, and those efforts will happen in the [new module library](https://github.com/doomemacs/modules) repo instead of here. I will revisit it once these things have been sorted out. Thanks again!

PR #5144, opened 2021-06-05
PULL_REQUEST_TEMPLATE: Simplify wording

xeruf:
- remove excess negations, which hamper understanding
- delineate currently inaccessible links
- note that docs shouldn't be contributed to


hlissner:
Thanks for the PR, but I cannot accept these, as our Discourse will be live soon:tm: and these changes would be immediately undone at that point.

xeruf:
Um, @hlissner there are more changes that would not have to be undone. I can revert those you dislike or you could cherrypick them, because some of the current wordings were really hard to comprehend when I tried to open a PR.

xeruf:
Or I could open a PR that reverts these changes. I know that soonTM might still be off more than we want ;)

PR #5149, opened 2021-06-07
lang/org: Add flag +strike

vherrmann:
This push request adds the +strike flag, which enables a strike-through for finished tasks, to the org module.


hlissner:
Thanks for the PR, but the feature is too trivial to warrant its own flag. Users could do this trivially from their config.

PR #5152, opened 2021-06-07
Use org-tree-slide-play-hook to hide blocks

cnr:
For reasons I don't entirely understand, calling `+org-present-hide-blocks-h` during `org-tree-slide-mode-hook` _does not_ 
hide block headers/footers.

Instead, hooking `org-tree-slide-play-hook` hides those headers/footers.


hlissner:
I can't reproduce the issue, but it appears to work either way, so I'll merge it. Thanks for the help!

PR #5154, opened 2021-06-08
lang/latex +fold: fix error when yas-snippet-{beg,end} are nil

ymarco:
This hook sometimes errors out because yasnippet ...forgets? to set `yas-snippet-{beg,end}`.
It occurs for *some* snippets and I think it only started about a month ago. Not sure why


hlissner (approved):

hlissner:
Thanks!

PR #5155, opened 2021-06-08
Fix '+tmux/cd' to send newline

scoiatael:
Hey,

I recently started testing `tmux` module and found out that `+tmux/cd` (and other functions using it: `+tmux/cd-to-here` and `+tmux/cd-to-project`) never sends newline - looks like `arg` argument is somehow obsolete there? I'm not very proficient in elisp, so I may be wrong though :)

### DIFF
--- a/modules/tools/tmux/autoload/tmux.el
+++ b/modules/tools/tmux/autoload/tmux.el
@@ -65,11 +65,11 @@ but do not execute them."
   (apply #'+tmux +tmux-last-command))
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


hlissner:
There was an issue with `+tmux/cd` I corrected in 09226f9, which should indirectly fix this issue. `+tmux/cd` should send a newline, but only if `noreturn` is nil (the default). Let me know if that gives you any trouble. Thanks for bringing it to my attention!

PR #5159, opened 2021-06-09
Remove dependence on company for fish and bash completion in eshell

Zetagon:
Bash and fish completion work without company, e.g. when the user uses
the minibuffer for completion.

I've removed the feature gate behind company for fish and bash-completion. It works without company and there are people who don't use company. Furthermore it seems to cause some errors, see https://discord.com/channels/406534637242810369/406554085794381833/852098110624235550


hlissner:
Makes sense to me. Thanks!

PR #5173, opened 2021-06-13
`SPC TAB N` binding

fosskers:
This PR adds a small wrapper around `+workspace/new` for the primary use of binding to `SPC TAB N` and interactively prompting the user for the name of their new workspace. While both `+workspace/new` and `+workspace-new` can receive a named argument, neither are set up for user prompting in this way.

This binding offers a shortcut to the common pattern of creating a new workspace via `SPC TAB n` and then immediately renaming it via `SPC TAB r`.

### DIFF
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -365,6 +365,7 @@
         :desc "Display workspaces"           "d" #'+workspace/display
         :desc "Rename workspace"             "r" #'+workspace/rename
         :desc "Create workspace"             "c" #'+workspace/new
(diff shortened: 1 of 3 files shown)


hlissner (changes requested):
Could you add a `C` keybinding [config deffor vanilla users](https://github.com/hlissner/doom-emacs/blob/d989bc6592612840e90ffc324bd4a98972836c24/modules/config/default/%2Bemacs-bindings.el#L367) too?

fosskers:
Done!

hlissner:
Thanks!

PR #5174, opened 2021-06-13
[:input layout bepo] Remove "W" convenience remap

gagbo:
`W` was used to emulate `C-w C-w`, which is useful to quickly change windows
most of the time. But this currently conflicts with edebug evil-collection bindings.
(gives `Key sequence W W starts with non-prefix key W` on the first loaded elisp buffer.).

Better to leave the key alone, avoid error messages altogether, and let users redefine
the binding if they want to.


hlissner (approved):

hlissner:
Thanks!

PR #5175, opened 2021-06-13
:lang sh: fix sh formatter

ztlevi:
This PR fixes the sh formatter. The tab-width doesn't work before.

### DIFF
--- a/modules/lang/sh/config.el
+++ b/modules/lang/sh/config.el
@@ -17,9 +17,9 @@
   (set-docsets! 'sh-mode "Bash")
   (set-electric! 'sh-mode :words '("else" "elif" "fi" "done" "then" "do" "esac" ";;"))
   (set-formatter! 'shfmt
-    '("shfmt" "-ci"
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but the logic is incorrect. This evaluates those embedded forms eagerly (i.e. when the `sh-script` package is loaded), which is not what we want. `set-formatter!` will evaluate embedded forms when the formatter is invoked. I've also tested the former setting and it works as expected; properly respecting `tab-width` and the shell type. Please inspect your private config if it's not working for you.

ztlevi:
I'm doing 
```
(setq-hook! 'sh-mode-hook sh-basic-offset 2
  tab-width 2)
```
in my personal config but it's not effective in the formatter. What's wrong with this?

ztlevi:
I'm doing 
```
(setq-hook! 'sh-mode-hook sh-basic-offset 2
  tab-width 2)
```
in my personal config but it's not effective in the formatter. What's wrong with this?

PR #5179, opened 2021-06-14
Fix #5128: No format-on-save by lsp-formatters

### LINKED ISSUE #5128: Formatting on save fails to take formatting by lsp-servers into account
### 

- [X] I have searched [the Discourse](https://doomemacs.org/d/docs) for any errors and solutions.
- [X] I have read [How to Debug Issues](https://doomemacs.org/d/how2debug) and found no solution.

If I have `:editor (format +onsave)` enabled and `+format-with-lsp` set to true, I expect LSP-servers to format buffers on save, if the provide they corresponding feature.

### What actually happened?

In the situation described above, the buffer is not formatted when saving, even though the LSP-server provides document formatting capacities. The `before-save-hook` does not contain a function to invoke LSP-formatting

I tried to debug the issue and found what seems to be two issues with the following function:

```elisp
(defun +format-enable-for-lsp-on-save-maybe-h ()
  "Enable LSP formatter when LSP client is available."
  (remove-hook 'lsp-mode-hook #'+format-enable-for-lsp-on-save-maybe-h 'local)
  (cond ((not +format-with-lsp) nil)
        ((bound-and-true-p lsp-mode)
         (when (lsp-feature? "textDocument/formatting")
           (+format-enable-on-save-h))
         t)
        ((bound-and-true-p eglot--managed-mode)
         (when (eglot--server-capable :documentRangeFormattingProvider)
           (+format-enable-on-save-h))
         t)
        ((bound-and-true-p lsp--buffer-deferred)
         (add-hook 'lsp-mode-hook #'+format-enable-for-lsp-on-save-maybe-h
                   nil 'local)
         t)))
```

1. That function initially runs after a change of major mode. However, the checks for `lsp-mode` or `lsp--buffer-deferred` fail, because `lsp-mode` is usually started in the `local-vars-hook` of the corresponding major mode which runs after after-change-major-mode-hook.
2. At least for `lsp-mode` and if the lsp-server isn't already running, there is a race condition in the sense that `(bound-and-true-p lsp-mode)` might return true but `(lsp-feature? "textDocument/formatting")` returns `nil`, even though the language server provides document formatting. 

### Steps to reproduce

1. Setup Doom with `:editor (format +onsave)` and `+format-with-lsp` set to true
2. Also enable LSP support for csharp in `init.el`
3. Install omnisharp-roselyn LSP-server
4. Open a csharp-file and perform some editing operations like adding gratuitous whitespace
5. Save the buffer
6. The buffer is not formatted before it is saved
7. Manually invoke `lsp-format-buffer` to confirm that the LSP server is able to format the buffer 

### System Information

https://pastebin.com/PyqgweJd

RBckmnn:
Fixes #5128

In `+format--all-buffer-from-hook-a`, a call to `format-all--formatter-executable` with the formatter being equal to `'lsp` or `'eglot` returns `nil`. That's why `funcall` was never executed in those cases. To fix this issue, this PR introduces special checks for those cases.

### DIFF
--- a/modules/editor/format/config.el
+++ b/modules/editor/format/config.el
@@ -74,11 +74,13 @@ This is controlled by `+format-on-save-enabled-modes'."
 (defadvice! +format--all-buffer-from-hook-a (orig-fn &rest args)
   :around #'format-all-buffer--from-hook
   (letf! (defun format-all-buffer--with (formatter mode-result)
-           (and (condition-case-unless-debug e
-                    (format-all--formatter-executable formatter)
-                  (error
-                   (message "Warning: cannot reformat buffer because %S isn't installed"
-                            (gethash formatter format-all--executable-table))
-                   nil))
-                (funcall format-all-buffer--with formatter mode-result)))
+           (when (or (eq formatter 'lsp)
+                     (eq formatter 'eglot)
+                     (condition-case-unless-debug e
+                         (format-all--formatter-executable formatter)
+                       (error
+                        (message "Warning: cannot reformat buffer because %S isn't installed"
+                                 (gethash formatter format-all--executable-table))
+                        nil)))
+             (funcall format-all-buffer--with formatter mode-result)))
     (apply orig-fn args)))


sagittaros:
This works for me!

hlissner (approved):

hlissner:
Looks good to me. Thanks for the help and for looking into this!

PR #5188, opened 2021-06-17
Fixed #5176: centaur-tabs' gt/gT asking for more key input.

### LINKED ISSUE #5176: centaur-tabs' `gt`/`gT` asking for more key input.
### 

- [X] I have searched [the Discourse](https://doomemacs.org/d/docs) for any errors and solutions.
- [X] I have read [How to Debug Issues](https://doomemacs.org/d/how2debug) and found no solution.

When pressing `gt` or `gT`, the tab should be switched immediately.

### What actually happened?

Instead of switching immediately, it stops waiting for another keystroke.

I saw that if you press `SPC` or a letter, it switches tabs, but if you press an arrow key, the message "No replacement character typed" appears. This message seems to come from .local/straight/repos/evil/evil-common.el:609.

I removed my config and tested with the default doom config, and there `gt` is bound to workspace switching and works normally. Enabling tabs the bug appears again. I tried to remove the keybinding for workspace (in case it was conflicting) but that does not solve the issue.

### Steps to reproduce

Install doom and enable the tabs module. Open two tabs and try to switch with `gt`.

### System Information

https://pastebin.com/65trckWw

holyfight6:
Fixes #5176  

I changed the `+tabs:next-or-goto` and `+tabs:previous-or-goto` functions to use `(interactive "P")` so they accept the prefix as tab index. This reduce the extra key strike when you just want next ore previous, and works with a big tab index. Also the current implementation doesn't work with index(it doesn't use the index at all).

### DIFF
--- a/modules/ui/tabs/autoload.el
+++ b/modules/ui/tabs/autoload.el
@@ -4,15 +4,15 @@
 ;;;###autoload (autoload '+tabs:next-or-goto "ui/tabs/autoload" nil t)
 (evil-define-command +tabs:next-or-goto (index)
   "Switch to the next tab, or to INDEXth tab if a count is given."
-  (interactive "<C>")
-  (if current-prefix-arg
-      (centaur-tabs-select-visible-nth-tab current-prefix-arg)
+  (interactive "P")
+  (if index
+      (centaur-tabs-select-visible-nth-tab index)
     (centaur-tabs-forward)))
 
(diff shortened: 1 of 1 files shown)


holyfight6:
Update: changed the interactive code form `"P"` to`"p"`(the index should only accept number).

hlissner:
Ah, my bad, I resolved this in 1a96f5b before I noticed this PR. In any case, it was fixed. Thanks for bringing it to my attention!

PR #5189, opened 2021-06-17
Point FAQ to Github, not Discourse

joshgoebel:
This changes the FAQ link to a link that works for all users.

Is there a reason this currently points to Discourse which (seems to) force one to have a login to even see the content?

I tried to check the "do-not-PR" list but it has the same issue - of evidently being locked behind needing a Discourse account?

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,7 @@ kind!
 [discord]: https://discord.gg/qvGgnVx
 [discourse]: https://discourse.doomemacs.org
 [documentation]: docs/index.org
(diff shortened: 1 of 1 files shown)


hlissner:
It's pointing at Discourse because it will be our primary resource, but it's login-gated until it is ready, which is taking longer than usual because I'm on hiatus atm to recover my health. I'd rather wait until the Discourse goes live than revert all these links. If you'd like to PR to Doom, I encourage you to create an account on the Discourse for now.

PR #5190, opened 2021-06-18
Add terraform-format-buffer to terraform-mode-map

mjburling:
I've added `terraform-format-buffer` to `terraform-mode-map` after a colleague asked why I had _settled_ for calling `terraform-format-buffer` via `SPC` + `:` each time I needed it. `SPC` + `m` + `f` ought to do nicely.

I thought maybe doing a `(cmd! (compile "terraform fmt"))` like the others in this mode-map might have been nice, but the experience of calling `terraform fmt` on the directory left me with a stale buffer. Using the function for updating only the buffer works just fine, though.

### DIFF
--- a/modules/tools/terraform/README.org
+++ b/modules/tools/terraform/README.org
@@ -62,8 +62,9 @@ common Terraform operations (see Keybindings below).
 * Appendix
 ** Keybindings
 *** :localleader
-| key | description           |
-|-----+-----------------------|
-| =i= | Run =terraform init=  |
-| =p= | Run =terraform plan=  |
-| =a= | Run =terraform apply= |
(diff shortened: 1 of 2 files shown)


sei40kr:
@mjburling 
`:editor format` already supports `terraform fmt` (`SPC c f`).

https://github.com/hlissner/doom-emacs/blob/develop/modules/editor/format/README.org
https://github.com/lassik/emacs-format-all-the-code

mjburling:
>  `:editor format` already supports `terraform fmt` (`SPC c f`).

Hey, that's handy! I wasn't aware of `:editor format`. Closing this PR after all. Thanks very much, @sei40kr!

PR #5191, opened 2021-06-19
show all available lsp servers when switching

chaoky:
when invoking `+lsp/switch-client` show all available lsp servers instead of only high priority ones


chaoky:
somewhat related, maybe should report as a bug:
`+lsp/switch-client` sometimes restores the client priority before `lsp-workspace-restart` is finished, making so the server switch doesn't happen, putting the priority restauration on a 1s timer solved the issue for me, but doesn't seem like a good solution

hlissner:
Perhaps resetting the priority in `lsp-after-initialize-hook` would work? Assuming that hook fires when the new workspace is initialized.

chaoky:
`lsp-after-initialize-hook` works :)

hlissner (approved):

hlissner:
Thanks!

PR #5202, opened 2021-06-23
removed unused use in "file-template rust main"

meltinglava:
Now its will be the same as what `cargo new` generates.


hlissner (approved):

hlissner:
Thanks!

PR #5203, opened 2021-06-24
Define sh-mode for bats files

bersace:
bats is a superset of shell script for writing automated tests exected
with TAP. sh-mode works very well with bats files.

This patch simply add `.bats` file suffix to sh-mode. Note that `.bats` is recommend by the bats project and used in for their own test suites.


hlissner (approved):

hlissner:
Makes sense to me. Thanks!

PR #5205, opened 2021-06-24
better zig witticism

### LINKED ISSUE #5022: Improve Zig Flavor Text
**Describe the feature**
Currently, the Zig module is labelled like so:
```zig               ; C, but simpler```
This is, in my humble opinion, **critically lame** compared to many other modules in `init.el`.

It is a pet peeve of mine to see a language described as "simpler" than another, because that doesn't mean anything. There are many different kinds of "simple". Is it simpler in one lang to write FizzBuzz, or is it simpler for writing a multi-threaded regex parser that traverses several thousand files per second? These are often mutually exclusive "simpler"s.

In a lot of ways, Zig is more complex than C. It has more keywords than C, it makes a programmer think more actively about compile-time vs runtime behavior (that's simpler for some programs, but more complex for others), it has more data types, it has standard coroutines, it generally inhibits simply `malloc`ing data, it has optionals rather than null semantics (simpler in some ways, more complex in others), it has pattern matching, it makes mutability more explicit, etc.

Moreover, Zig has several unique characteristics to identify it with. It's the most popular language to provide such explicit delineation of comptime vs runtime behavior, except _arguably_ C++20. It is technically a wrapper around Clang, which you don't see every day. Its marketing has cute lizards everywhere ([this one](https://dev.to/stein/bio-all-your-parentheses-are-belong-to-us-25lo) is my favorite). And it sort of already has a tagline, which it generates in lieu of 'hello world' for initializing projects: `All your code are belong to us`

There's gotta be _something_ in there. Not that I have any ideas.

**System information**
```
C:\Users\wgooc>doom info
←[31mx There was an unexpected error←[0m
  ←[1mMessage:←[0m File is missing
  ←[1mData:←[0m (file-missing . "Searching for program")
  ←[1mBacktrace:←[0m
    (call-process "uname" nil t nil "-msr")
    (apply call-process "uname" nil t nil "-msr")
    (or (apply #'call-process command nil t nil (remq nil args)) -1)
    (cons (or (apply #'call-process command nil t nil (remq nil args)) -1) (st
    (progn (cons (or (apply #'call-process command nil t nil (remq nil args))
    (unwind-protect (progn (cons (or (apply #'call-process command nil t nil (
    (save-current-buffer (set-buffer temp-buffer) (unwind-protect (progn (cons
    (let ((temp-buffer (generate-new-buffer " temp"))) (save-current-buffer
    (doom-call-process "uname" "-msr")
    (apply doom-call-process ("uname" "-msr"))
←[33m! Extended backtrace logged to .local/doom.error.log←[0m
```

iyefrat:
pros: it's dumber
cons: it's dumber

closes #5022
cc @Cons-Cat

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -168,7 +168,7 @@
        ;;terra             ; Earth and Moon in alignment for performance.
        ;;web               ; the tubes
        ;;yaml              ; JSON, but readable
-       ;;zig               ; C, but simpler
+       ;;zig               ; zag
 
        :email
        ;;(mu4e +gmail)


jeetelongname (approved):

hlissner:
Is `zag` somehow significant to the zig language? If not, it doesn't seem particularly witty, or to say anything about the language.

iyefrat:
Nope! I just thought it was funny. No accounting for taste though. My backup idea was "assembly competitive hello world", which is a [true thing](https://drewdevault.com/2020/01/04/Slow.html) about Zig, but less on the joke side. Another riff on that idea is maybe "syscalls? where we're going, we don't need syscalls", but I have no idea how true the syscall minimalism is outside of the hello world example. If none of these are any good I'm sure someone will come up with something better eventually.

Cons-Cat:
Both Doom Emacs and Zig have branding based on video games. Maybe there's some commonality between Doom and Zero Wing?

iyefrat:
Alright well, given the fact that:

a. my incredible sense of humor is not universal
b. in the new docs™ the witticisms will be embedded in the readme, in which this doesn't work nearly as well due to the lack of surrounding context of more sensible wititcisms

I'm gonna go ahead and close this.

PR #4664, opened 2021-02-16
Add Vertico module

Emiller88:
This is a work in progress, to quote the great @mjlbach "be the PR you want to see in the repo".

The only thing I take credit for so far is pinning the packages and copying and pasting @Townk's config ontop of @jethrokuan's PR (#4519) so I'll only take it personally if you rip up the way I pinned the repos.

Please add anything that you find isn't working to `TODO.org` or PR the fix.

This PR used to be named `Add Selectrum Round 2`

PR #5212, opened 2021-06-29
Add org-cite support to :tools/biblio

bdarcus:
Screenshot from 2021-07-16 10-15-10

**A major new feature branch, the org-cite module, has been merged to org-mode master. This PR adds support for it.**

_I am opening this, in part to facilitate alignment with #2888. These will need to be reviewed in tandem, to figure out the best path forward_.

The above screenshot shows an org-cite citation, with [bibtex-actions](https://github.com/bdarcus/bibtex-actions) and Vertico, where `embark-act` is presenting the contextual which-key menu. 

And here's a quick screencast using the same stack.

https://youtu.be/gqehXIZcqU8

## Background

In April of 2020, @brianmcgillion opened #2888, mostly to add `org-ref` support.

That PR has languished in the face of other doom priorities, but is planned for merger for 3.0.

Since then, work on org-cite is complete. 

I fully expect OC to be the future of citations in org. It will obsolete much (perhaps even most) of the functionality in `org-ref`, and the latter may (at this point it's future is unclear) be rewritten in the future to add the other functionality to org-cite, with the result a smaller, more focused, `org-ref`.

## What this adds

Configuration for the new OC functionality:

1. set up the different "processors" that provide citation and bibliography fontification, preview, export, etc
2. install `citeproc-el`, which is the backend required for the `oc-csl` "export processor", and for previews

I am a little out of my depth on the details of `use-package`, but have reviewed how the org module is coded (for example the `ox-*` config), and incorporated that here.

### What to do about #2888?

Edit: _Remove `org-ref`_.

John Kitchin posted this video on July 17 (and subsequently a followup) showing WIP on converting a next-gen of his software to an org-cite processor.

~https://www.youtube.com/watch?v=4ta4J20kpmM~
https://www.youtube.com/watch?v=3TBYg1MZ6iY

Here's what his "follow processor" menu looks like with this module, and bibtex-actions/vertico (note here it's using `pretty-hydra`, which would be good to add to the hydra module in any case, but he's also experimenting with a `transient` menu).

Screenshot from 2021-07-22 09-49-41

Given this, and his rapid progress, I think it's obvious org-ref should be removed from #2888.

He says in the video he is not planning to release something until Org 9.5 is released.

See below section on configuring these processors though. I don't think a flag will be necessary for this, but it could be an option later.

### `use-package!` and loading?

We need to load the oc processors, if org-cite is present (Emacs 9.5), and if org module. There may be some subtleties I'm missing.

Also, @iyefrat and I have been chatting about how best to handle bibtex-actions configuration for org-cite, when the latter is a new feature for Org 9.5.

Currently I have placed the functions and variables for bibtex-actions org-cite support in a separate file, where I require Org 9.5. 

But I can change it upstream to match what we think the `use-package!` statements here should ideally be, if there's a better way to handle this.

See below, and https://github.com/bdarcus/bibtex-actions/pull/180

### User configuration of org-cite processors?
OC is designed around flexible and modular "capabilities"; currently:

- _activate_ (fontification, preview, etc.)
- _follow_ (`org-open-at-point` behavior)
- _insert_
- _export_

One configures these "processors" like so:

``` elisp
(setq org-cite-activate-processor 'basic)
```
So, for example, when the oc version of `org-ref` (now called `oc-bibtex`) is released, one user might set it like so, which is to use that processor to provide all of the features:

``` elisp
(setq org-cite-insert-processor 'bibtex) ; `org-cite-insert` integration; this currently requires ivy
(setq org-cite-follow-processor 'bibtex) ; `org-open-at-point` integration with the hydra noted above
(setq org-cite-activate-processor 'bibtex) ; fontification and previews with at-point keymap functions
```
... but that insert processor won't work in the vertico module, and so another option, using different processors from different packages:

``` elisp
(setq org-cite-insert-processor 'oc-bibtex-actions) ; `bibtex-actions`, but completing-read-based
(setq org-cite-follow-processor 'bibtex) ; `org-open-at-point` integration; stay with the above
(setq org-cite-activate-processor 'org-cite-csl-activate) ; fontification and output previews
```

So it is a highly flexible, modular, system.

The question, then:

> How should these processors be configured in this module?

Because those processors are basically like binary flags, I don't think module flags makes sense. Rather, we should just have a section of the README for this.

### #4664

The vertico PR adds bibtex-actions, which is designed for this stack. This PR is currently assuming that gets merged first.

### Key bindings?

These the entry points for OC:

1. `org-cite-insert`
2. `org-open-at-point`

These should have convenient access from the same place, but it's a bit unclear ATM where.

The bibtex-actions integration noted above, for example, plugs into these commands (the "c" binding in the screenshot at the top is using `org-cite-insert` behind-the-scenes). Bibtex-actions also has similar commands (as do ivy-bibtex and helm-bibtex actions, which use the same backend functions) that are not limited to org, notably: `org-cite-insert` (org) ~= `bibtex-actions-insert-citation` (latex, markdown). Ideally at least these two should be accessible from Doom menus, with the same binding in all three modes: org, latex, markdown.

OTOH, the org leader menu has the "f" binding for inserting a footnote; to be consistent with that, perhaps a "c" binding goes there?

I don't currently see a Doom binding for `org-open-at-point` (though there is the "goto" submenu). Perhaps there should be one?

## Minimal test file

Assuming you have your citations in the test.bib flie:

``` org
#+cite_export: natbib
#+bibliography: /tmp/test.bib

[cite:@test1]

#+print_bibliography:
```


iyefrat:
@bdarcus what alignment do you need to do with the selectrum PR?

iyefrat:
Your best bet to fix the git history is with git rebase (I assume you are using magit):
- when this branch on the status screen, put your cursor on the earliest commit in this branch (https://github.com/hlissner/doom-emacs/pull/5212/commits/0635fe123cb2f2ce435eef82d0618489b283c080)
- press `r i` to start interactive rebase
- move the cursor on each dummy commit and press `d` to drop
- `C-c C-c` to rebase
- force push with `p -f p` from magit status screen

I don't know how familliar you are with magit, but the manual is very good, and so are the git man pages. However note that the man page for a git subcommand `git foo` is under `man git-foo`, so e.g. the man page for rebasing is `git-rebase`.

bdarcus:
> @bdarcus what alignment do you need to do with the selectrum PR?

@iyefrat - now that I think of it, really nothing is necessary. 

But I could include the `bibtex-actions` and `embark` stuff that's there here.

For example, I currently define a `bibtex-actions` "follow processor" here, which can use `embark` to bind actions to citations via `org-open-at-point`.

But that isn't strictly necessary for this PR. Really just depends on how the timeline plays out on these pieces.

bdarcus:
> Your best bet to fix the git history is with git rebase (I assume you are using magit):

While I know I should use magit, I usually just use the command line, including for rebasing.

iyefrat:
In that case you can do `git rebase -i 0635fe1`, the rest is the same except you do `:wq` to save and quit instead of `C-c`, and force pushing is done with `git push --force-with-lease`

bdarcus:
> In that case you can do `git rebase -i 0635fe1`, the rest is the same except you do `:wq` to save and quit instead of `C-c`, and force pushing is done with `git push --force-with-lease`

Done.

I still have the merge commit though.

iyefrat:
yeah my bad, you need to rebase on one commit before that, do `git rebase -i 0635fe1~`

bdarcus:
There we go; thank you @iyefrat!!

bdarcus:
Actually, spoke too soon. There was then a subsequent conflict with the files, that I resolved, but squashing those doesn't appear to be straightforward.

Is that OK do you think?

Or should I just start over?

iyefrat:
try the following: `git rebase -i` and change the merge commit to a fixup commit by pressing `f` on it. this should work but there might be shenanigans do to it being a merge commit

bdarcus:
If I do, for example, `git rebase -i HEAD~2`, I get another long list of individual commits; I don't get the merge commit to edit.

iyefrat:
At this point I think your best bet is to just delete the branch locally, redo it with the one good commit, and force push.

bdarcus:
> At this point I think your best bet is to just delete the branch locally, redo it with the one good commit, and force push.

I thought I did, but it just got worse :-)

I'll take another look tomorrow.

iyefrat:
@bdarcus good! I suggest you mark the troubleshooting comments as off topic since git shenanigans are not really of interest to anyone taking a look at the PR (with maybe this one left for context as to why there are a bunch of off topic comments).

bdarcus:
> @bdarcus good! I suggest you hide the troubleshooting comments since git shenanigans are not really of interest to anyone taking a look at the PR (with maybe this one left for context as to why there are a bunch of hidden comments).

Done, but at the cost of another off-topic comment, I can't hide your's.

bdarcus:
@iyefrat - do you agree with my assessment that I need to condition line 36 on #4664, given it depends on `bibtex-actions` and `completing-read`?

modules/tools/biblio/packages.el
```diff
@@ -1,8 +1,10 @@
...
-(package! bibtex-completion :pin "9f6ea920a49457d85096caa0e61f086a42b2908e")
+(package! bibtex-completion :pin "ca09076c3d6e51cc4ffe208c8176fccf3710fcc6")
+(package! citeproc :pin "0857973409e3ef2ef0238714f2ef7ff724230d1c")
+
 (when (featurep! :completion ivy)
-  (package! ivy-bibtex :pin "9f6ea920a49457d85096caa0e61f086a42b2908e"))
+  (package! ivy-bibtex :pin "ca09076c3d6e51cc4ffe208c8176fccf3710fcc6"))
 (when (featurep! :completion helm)
-  (package! helm-bibtex :pin "9f6ea920a49457d85096caa0e61f086a42b2908e"))
+  (package! helm-bibtex :pin "ca09076c3d6e51cc4ffe208c8176fccf3710fcc6"))
```
iyefrat, line 10: Is bumping these packages necessary to get the functionality in this PR to work? If so, bump them in a separate bump commit following [this format](https://discourse.doomemacs.org/t/how-to-contribute/44#bump-outdated-packages). If this is unclear, look at doom's commit history for any commit starting with "Bump", and note the notice with the pin symbol after section 8.2.

modules/tools/biblio/config.el
```diff
@@ -12,3 +12,38 @@
...
+      (setq org-cite-activate-processor 'csl-activate)
+
+      ;; required for csl-activate
+      (add-hook 'org-mode-hook
+                (lambda ()
+                  (cursor-sensor-mode 1)))
+
+      ;; follow processor for actions tied to org-open-at-point
+      (when (featurep! :completion selectrum)
+        (setq org-cite-follow-processor 'bibtex-actions))
```
iyefrat, line 37: Indeed, this part depends on the selectrum PR.

modules/tools/biblio/config.el
```diff
@@ -12,3 +12,38 @@
...
+;;; Org-Cite configuration
+
+(when (featurep! :lang org)
+      ;; load org-cite and its processors
+      (require 'oc)
+      (require 'oc-basic)
+      (require 'oc-natbib)
+      (require 'oc-biblatex)
+      (require 'oc-csl)
+      (require 'oc-csl-activate)
```
iyefrat, line 25: Generally package specific configuration is done inside of `use-package` statements. I'm unsure as to the exact correct semantics for doing that here (is org-cite provided as a single package that you can load? @tecosaur what is the way to load things here?). [https://github.com/jwiegley/use-package](use-packagel) is a bit daunting, but the readme is very comprehensive. Ask on the discord if you run into trouble

iyefrat (changes requested):

modules/tools/biblio/config.el
bdarcus: If using `use-package` for this, would that not suggest this goes in the `org` module, and then `when` statement would condition on `biblio`? So the reverse of what I have?

bdarcus: I pushed a commit that makes the change to `use-package!`. 

It seems to work, and I checked with the OC author if he had any input, but he doesn't use `use-package`.

bdarcus:
@iyefrat - I ended up adding bibtex-actions to package.el here, though per my note on discord, whether this is actually a bump here or not vis-a-vis the vertico PR, will depend on timing of all of this.

iyefrat:
If you add the package under the `:completion vertico` flag, this PR should probably be merged after the Vertico PR anyway.

I do have a question about this though. How well does `bibtex-actions` work without Vertico and Embark? It might be worth considering adding it independently of `:completion vertico`. It does look like from the readme that in order to play well with `embark` you need to load the package after embark, but maybe that can change upstream. The only other issue I can think of from my brief familiarity with the package is the lack of ability to bind `bibtex-actions-map` before loading the package, which means that you can't bind in onto `<leader> n b`, and only having`bibtex-actions-open-entry` there without the ability to chose alternate actions on it with `embark` is a bit of a usability issue.

bdarcus:
> If you add the package under the `:completion vertico` flag, this PR should probably be merged after the Vertico PR anyway.

Yes, my thinking. I added a note saying as much to the main post.
 
> I do have a question about this though. How well does `bibtex-actions` work without Vertico and Embark? 

The commands should work in any vertical completing-read-compliant interface, including the now built-in icomplete one, so that `M-x bibtex-actions-open` and such should work.

You just don't get the additional contextual actions that come with Embark.

> It might be worth considering adding it independently of `:completion vertico`. 

Given what I say above, WDYT? It probably wouldn't work well with horizontal completion UIs, for example.

> It does look like from the readme that in order to play well with `embark` you need to load the package after embark, but maybe that can change upstream. 

Am not following here. Can you please rephrase?

> The only other issue I can think of from my brief familiarity with the package is the lack of ability to bind `bibtex-actions-map` before loading the package, which means that you can't bind in onto `<leader> n b`, and only having`bibtex-actions-open-entry` there without the ability to chose alternate actions on it with `embark` is a bit of a usability issue.

I'm also not following the last part here. The embark menus work correctly, since embark relies on the keymaps for those bindings. 

So a limitation is we only have the one  `<leader> n b` binding in doom, but doing `embark-act` from there does bring up all of them.

Screenshot from 2021-07-14 09-26-26
Screenshot from 2021-06-28 07-59-02

iyefrat:
> > It does look like from the readme that in order to play well with embark you need to load the package after embark, but maybe that can change upstream.

>Am not following here. Can you please rephrase?

In the `bibtex-actions` readme the following is the suggested use-package:

```lisp
(use-package bibtex-actions
  :bind (("C-c b" . bibtex-actions-insert-citation)
         :map minibuffer-local-map
         ("M-b" . bibtex-actions-insert-preset))
  :after embark
  :config
  :custom
  ;; make sure to set this to ensure open commands work correctly
  (bibtex-completion-additional-search-fields '(doi url))
  (bibtex-completion-bibliography '("~/bib/references.bib")))
```
Which seems to mean that in order for the package to work with embark, it has to be loaded after it. This also was necessary for getting it to work in the Vertico PR, but adding `:after embark` if you aren't using `embark` will cause the `use-package` statement not to get evaluated. There are three ways to work around this:

1. Rework `bibtex-actions` so that it doesn't need to be loaded after embark

bdarcus:
> 1. Rework `bibtex-actions` so that it doesn't need to be loaded after embark (best option if possible)

I'm a bit fuzzy on `use-package` in general, but maybe the issue isn't the package per se, but the embark variables that need to be set there for it work with it? I mean lines like this in your PR:

``` elisp
(add-to-list 'embark-keymap-alist '(bibtex . bibtex-actions-map))
```

BTW, in `bibtex-actions-org-cite` I require Embark and set those variables there. So `bibtex-actions-embark` would be conceptually similar if that made sense.

iyefrat:
> I'm a bit fuzzy on `use-package` in general,

By the way, the secret to understanding `use-package` is using `SPC m m` to expand the macro, that makes what exactly it's doing clearer in ways the keyword documentation under `SPC h f` sometimes doesn't.

> but maybe the issue isn't the package per se, but the embark variables that need to be set there for it work with it? I mean lines like this in your PR:
> 
> ```emacs-lisp
> (add-to-list 'embark-keymap-alist '(bibtex . bibtex-actions-map))
> ```

Yeah. tag me in the PR that tries to implement this if you do that, I'll look it over. I also might try to do it myself in the near future, but I want to submit the Vertico PR first.
   
> In the [`bibtex-actions-org-cite`](https://github.com/bdarcus/bibtex-actions/blob/main/bibtex-actions-org-cite.el) I just added, I require Embark and set those variables there. Is that WYM?

No, although that does mean that you should add `embark` the :after` statement of the `use-package` statement. 

Having:

```lisp
(use-package! bibtex-actions-embark :after embark)
```
In your emacs config means that it'll load that file only after `embark` is loaded. It should actually be:

```lisp
(use-package! bibtex-actions-embark
 :after (bibtex-actions embark))
```
Because it should also be loaded after `bibtex-actions`.
Adding the `(require 'embark)` in the file itself does a different thing, the `use-package` statement in your emacs config tells emacs "after you finish loading `bibtex-actions` and `embark`, whenever that is, load `bibtex-actions-embark`". The `require` statement in your package tells emacs to load the `embark` file during the loading of `bibtex-actions-org-cite`, but that only happens if you load `bibtex-actions-org-cite` somewhere.

modules/tools/biblio/config.el
```diff
@@ -12,3 +12,61 @@
...
+              org-cite-insert-processor 'oc-bibtex-actions)
+      (setq org-cite-follow-processor 'basic
+            org-cite-insert-processor 'basic))
+
+    ;; setup export processor; default csl/citeproc-el, with biblatex for latex
+    (setq org-cite-export-processors
+          '((beamer natbib)
+            (latex biblatex)
+            (t csl)))
+
```
bdarcus, line 46: @tecosaur - when you get a chance, any suggestions on what the settings here should be for TeX?

Maybe just keep it simple and do this?

```suggestion
    (setq org-cite-export-processors
          '((latex natbib)
            (t csl)))

```

Issue #5215, opened 2021-06-29
With LSP and Snippets enabled, company-mode sorts snippets to the bottom


PatrickWulfe:
### What did you expect to happen?

When I have snippets and LSP enabled, company should show matching snippets before LSP results.

### What actually happened?

When I used snippets and lsp in a dart project, company would put the snippets at the end of the list, which is typically hundreds of entries long.  It had the same behavior with python, but it's LSP didn't make lists as long.

### Describe your attempts to resolve the issue

The default 'company-backends' is "((:separate company-capf company-yasnippet) company-yasnippet)", but when I change it to "((:separate company-yasnippet company-capf) company-yasnippet)", it works as expected.

### Steps to reproduce

1. Enable LSP and snippets for a language
2. Open a file and start typing a snippet enough for company to show up

### System Information

https://pastebin.com/Y9M0mGSs

sei40kr:
You can change `+lsp-company-backends` to work around it:

```emacs-lisp
(setq +lsp-company-backends '(:separate company-yasnippet company-capf))
```

hlissner:
As of ba716d6 snippets are interspersed with the normal completion results, instead of simply placing the snippets in front of them (and thus creating a new problem). This should indirectly fix this issue.

PR #5218, opened 2021-06-30
init commit - Fixed company putting snippets at bottom

PatrickWulfe:
[https://github.com/hlissner/doom-emacs/issues/5215](url)
Fixes #0000 

Previously, using company-mode with both +lsp and snippets would cause the snippets to be put in the company list, but all the way at the bottom.  This change puts them at the top with the lsp results after.


PatrickWulfe:
I meant to do this as two separate pull requests, sorry

PR #5220, opened 2021-06-30
Moved snippets above lsp in company list

PatrickWulfe:
Fixes #0000 

Previously using company with both snippets and lsp would cause the lsp results to be listed first, then the snippets at the bottom.  This behavior essentially ruins the snippet integration since you could have to scroll through hundreds of entries to get to them.  This change moves them to the top of the list, similar to how VSCode displays snippets/lsp.


hlissner:
As of ba716d6 snippets are interspersed with the normal completion results, instead of simply placing the snippets in front of them (and thus creating a new problem). This should indirectly fix your issue.

PR #5221, opened 2021-07-01
Bump :lang ruby projectile-rails

otavioschwanck:
asok/projectile-rails -> asok/projectile-rails@6a18ada356

asok/projectile-rails6a18ada356 includes a annoying bug on template


hlissner:
This was bumped in 6be2262 because a larger bump was already in the works. Thanks for bringing it to my attention!

PR #5222, opened 2021-07-01
use format-buffer-h instead of the original format-all-buffer hook

andresilva:
The original hook for formatting on save was being called which doesn't take into account e.g. `+format-with` local variables.

### DIFF
--- a/modules/editor/format/config.el
+++ b/modules/editor/format/config.el
@@ -64,6 +64,7 @@ This is controlled by `+format-on-save-enabled-modes'."
 ;;   2. Applies changes via RCS patch, line by line, to protect buffer markers
 ;;      and avoid any jarring cursor+window scrolling.
 (advice-add #'format-all-buffer--with :override #'+format-buffer-a)
+(advice-add #'format-all-buffer--from-hook :override #'+format-buffer-h)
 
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but this isn't necessary. We're advising `format-all--probe` [here](https://github.com/hlissner/doom-emacs/blob/develop/modules/editor/format/config.el#L60) with `+format-probe-a`, which does take `+format-with` into account. `format-all-buffer--from-hook` calls `format-all--probe` to determine the formatter to use:

```elisp
(defun format-all-buffer--from-hook ()
  "Internal helper function to auto-format current buffer from a hook.

Format-All installs this function into `before-save-hook' to
format buffers on save. This is a lenient version of
`format-all-buffer' that silently succeeds instead of signaling
an error if the current buffer has no formatter."
  (cl-destructuring-bind (formatter mode-result) (format-all--probe)
    (when formatter
      (format-all-buffer--with formatter mode-result))))
```

So replacing this function won't do anything.

andresilva:
`+format-with` wasn't being taken into account for me when formatting on save, this fixed it for me.

andresilva:
TBH I no longer remember the details of this change and I concede that I may be wrong here and this is useless, so feel free to ignore it.

PR #5227, opened 2021-07-03
don't explicitly set `projectile-completion-system`

iyefrat:
The default value is `'auto`, which handles the detection of the active
completion system, see the definition of `projectile-completing-read`.

-------

- [X] It targets the develop branch
- [X] I've searched for similar pull requests and found nothing
- [X] This change is NOT in Doom's do-not-PR list: https://doomemacs.org/d/do-not-pr
- [X] If I've bumped any packages, I've done so according to https://doomemacs.org/d/how2bump
- [X] I've linked any relevant issues and PRs below
- [X] All my commit messages are descriptive and distinct

### DIFF
--- a/modules/completion/helm/config.el
+++ b/modules/completion/helm/config.el
@@ -172,7 +172,6 @@ Can be negative.")
              helm-projectile-switch-project
              helm-projectile-switch-to-buffer)
   :init
-  (setq projectile-completion-system 'helm)
   (defvar helm-projectile-find-file-map (make-sparse-keymap))
   :config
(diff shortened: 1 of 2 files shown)


hlissner (approved):

hlissner:
I recall setting them explicitly due to lazy loading race conditions with ivy/helm, but I _believe_ I fixed the root issue, so I can safely merge this. Will need to be tested, particularly when invoking projectile commands post-startup, before ivy/helm has loaded.

In any case, thanks for the help!

PR #5229, opened 2021-07-03
add multi-account switching in notmuch module

nackjicholson:
Introduces a new config `+notmuch-mail-folders` which holds a list of mailfolders. The `+notmuch-mail-folder` config that already exists is used as the default. Nothing changes with normal usage of `+notmuch/update` for syncing email folders, but it now accepts a prefix argument (`C-u` / `SPC u`). When supplied the prefix argument allows selection of a mail folder.

This is a non breaking change, adding a new feature that helps users with multiple email folders without changing anything for users that do not.


nackjicholson:
@hlissner What can I do on this to make it merge ready?

nackjicholson:
Okay.

PR #5230, opened 2021-07-03
The dart config was setting the 'flutter-run-or-hot-reload' keybind w…

PatrickWulfe:
…ith init, so if you changed the localleader in your settings, it would stay as SPC m rather than what localleader was mapped to

Fixes #0000 

Previously, if you changed the localleader in you config, then the flutter-run-or-hot-reload command would stay mapped to SPC M rather than to whatever the new localleader is set to.

### DIFF
--- a/modules/lang/dart/config.el
+++ b/modules/lang/dart/config.el
@@ -31,7 +31,7 @@
 (use-package! flutter
   :when (featurep! +flutter)
   :defer t
-  :init
+  :config
   (map! :map dart-mode-map
         :localleader
         "r" #'flutter-run-or-hot-reload))
--- a/modules/lang/dart/config.el
+++ b/modules/lang/dart/config.el
@@ -32,8 +32,8 @@
   :when (featurep! +flutter)
   :defer t
   :init
-  :after dart
-  (map! :map dart-mode-map
+  (map! :after dart-mode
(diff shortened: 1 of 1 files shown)


modules/lang/dart/config.el
```diff
@@ -31,7 +31,7 @@
 (use-package! flutter
   :when (featurep! +flutter)
   :defer t
-  :init
+  :config
   (map! :map dart-mode-map
         :localleader
```
hlissner, line 36: Revert `:config` -> `:init` and add `:after dart` above `:map dart-mode-map`.

hlissner (changes requested):

modules/lang/dart/config.el
```diff
@@ -32,6 +32,7 @@
   :when (featurep! +flutter)
   :defer t
   :init
+  :after dart
```
hlissner, line 35: 1. Needs to be `:after dart-mode`.
2. Needs to be inside the `map!`.

```lisp
  (map! :after dart-mode
        :map dart-mode-map
        ...)
```

hlissner (changes requested):

hlissner (approved):

hlissner:
Thanks!

PR #5232, opened 2021-07-04
Prevent extra variable from being passed in macro.

samrjack:
There was an issue recently brought to light on reddit ([this](https://www.reddit.com/r/DoomEmacs/comments/o4b1yz/pdftools_error_after_opening_a_pdf_and_unable_to/) post) about not being able to open files after working with a PDF. I also have been running into an identical error while working with org capture wherein I could not capture anything without this error popping up.

As the error says, it comes from passing too many arguments to the function `abort-if-file-too-large` which only takes 3 arguments but is being passed an extra `nil`. To prevent this, I've changed the pdf module to only pass the 4th parameter when it is non-nil to begin with.

### DIFF
--- a/modules/tools/pdf/config.el
+++ b/modules/tools/pdf/config.el
@@ -126,7 +126,9 @@
   (defadvice! +pdf-suppress-large-file-prompts-a (orig-fn size op-type filename &optional offer-raw)
     :around #'abort-if-file-too-large
     (unless (string-match-p "\\.pdf\\'" filename)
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Ah, `abort-if-file-too-large`'s 4th argument was added in 27.1. The person reporting the issue is likely on Emacs 26.3, which I will be dropping support for next week. I will merge this for now, but will remove this later when 26.x support is dropped.

PR #5233, opened 2021-07-04
Add configuration to ensure that CIDER is used for babel

IGJoshua:
More than a year ago a change was added to ob-clojure that prevented
evaluating a Clojure code block without setting
`org-babel-clojure-backend`. This change sets this to a default value of
CIDER, since that is the main tool used in Doom.

Without this change, manual configuration is required to use Clojure in
org babel documents.


hlissner (approved):

hlissner:
Sorry for the super late response. Thanks for the help!

PR #5234, opened 2021-07-05
Filter directories when purging straight repos

ErnestDong:
in macOS there is always an annoying ".DS_Store" in straight/repos,
which breaks =doom purge=


ErnestDong:
sorry i made a mistake

PR #5235, opened 2021-07-05
Filter directories when purging straight repos

ErnestDong:
in macOS there is always an annoying ".DS_Store" in straight/repos,
which breaks `doom purge`


hlissner (approved):

hlissner:

PR #5241, opened 2021-07-06
Fix command to show git log of current buffer file

thegeorgeous:
Currently <kbd>SPC g L</kbd> runs `magit-log` which opens a context window for git log with an option selected that limits the log to the file open in buffer. The user then has to press <kbd>l</kbd> to actually see the log. The same result can be achieved by calling `magit-log-buffer-file` instead


hlissner (approved):

hlissner (changes requested):
Could you update the same keybinding [for vanilla users](https://github.com/hlissner/doom-emacs/blob/develop/modules/config/default/+emacs-bindings.el#L328) as well?

thegeorgeous:
@hlissner done!

hlissner (approved):

hlissner:
Perfect. Thanks for the help!

PR #5265, opened 2021-07-16
Declare indent of appendq!, pushnew! and prependq!

paulapatience:
This makes it easier to reduce the depth of indentation when using these macros.


hlissner:
Thanks for the PR, but the previous lack of indent declaration was intentional. The indentation of `setq!`, `prependq!` and `appendq!` are meant to match that of `setq` and `append`.

PR #5266, opened 2021-07-16
Start wl-summary-mode in evil-emacs-state

paulapatience:
wl-summary-mode would not start in evil-emacs-state otherwise.


hlissner (approved):

hlissner:
Sorry for the (tremendously) late response. Thanks for the help!

PR #5271, opened 2021-07-18
lang/org: Add `+roam2` flag for org-roam v2

Wetlize:
This PR introduces support in Doom for `org-roam` v2, which just recently was released. It's still WIP (though most of the things are now done) and the following need to be addressed first, before this will be turned into a PR that can be merged:

- [x] Add evil based bindings for `magit-section`/`org-roam-mode`
- [x] Figure out why does Emacs override evil and leader keys whenever the point is positioned at org-roam's `magit-section`s in `org-roam-buffer` and fix that.
- [x] Finalize the keybindings.
- [x] Defer automatic building of `sqlite3` executable as far as possible (this is a blocking operation that can take a while)
- [x] Improve display of completion candidates  in the minibuffer with text properties
- [x] Confirm that the deferred initialization of the database doesn't cause any problems to org-roam and "outside" of it.

### DIFF
--- a/modules/lang/org/contrib/roam2.el
+++ b/modules/lang/org/contrib/roam2.el
@@ -136,27 +136,46 @@ In case of failure, fail gracefully."
           "t" #'org-roam-tag-add
           "T" #'org-roam-tag-remove
           "r" #'org-roam-ref-add
-          "R" #'org-roam-ref-remove))
-        (:map org-roam-mode-map
-         :when (featurep! :editor evil +everywhere)
-         :nv "]"   #'magit-section-forward-sibling
-         :nv "["   #'magit-section-backward-sibling
-         :nv "gj"  #'magit-section-forward-sibling
-         :nv "gk"  #'magit-section-backward-sibling
-         :nv "gr"  #'revert-buffer
-         :nv "gR"  #'revert-buffer
-         :nv "z1"  #'magit-section-show-level-1
-         :nv "z2"  #'magit-section-show-level-2
-         :nv "z3"  #'magit-section-show-level-3
-         :nv "z4"  #'magit-section-show-level-4
-         :nv "za"  #'magit-section-toggle
-         :nv "zc"  #'magit-section-hide
-         :nv "zC"  #'magit-section-hide-children
-         :nv "zo"  #'magit-section-show
-         :nv "zO"  #'magit-section-show-children
-         :nv "zr"  #'magit-section-show-level-4-all
-         :nv "C-j" #'magit-section-forward
-         :nv "C-k" #'magit-section-backward)))
+          "R" #'org-roam-ref-remove)))
+
+  (when (featurep! :editor evil +everywhere)
+    (add-hook! 'org-roam-mode-hook
+      (defun +org-roam-detach-magit-section-mode-map-h ()
+        "Detach `magit-section-mode-map' from `org-roam-mode-map'.
+Inheriting its keymaps introduces a lot of conflicts in
+`org-roam-mode' based buffers, where Evil and leader keybindings
+will become completely overridden. This is because `magit-section'
+uses 'keymap text-property to attach section-unique keymaps, which
+has a higher level of precedence than `emulation-mode-map-alists'.
+
+Note: We do this each time through the hook, because otherwise
+sections seems to ignore the detachment."
+        (set-keymap-parent org-roam-mode-map nil)))
+
+    (map! :map org-roam-mode-map
+          :nv "]"       #'magit-section-forward-sibling
+          :nv "["       #'magit-section-backward-sibling
+          :nv "gj"      #'magit-section-forward-sibling
+          :nv "gk"      #'magit-section-backward-sibling
+          :nv "gr"      #'revert-buffer
+          :nv "gR"      #'revert-buffer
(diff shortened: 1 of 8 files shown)


modules/lang/org/contrib/roam2.el
```diff
@@ -0,0 +1,122 @@
...
+  :preface
+  ;; Set this to nil so we can later detect if the user has set custom values
+  ;; for these variables. If not, default values will be set in the :config
+  ;; section.
+  (defvar org-roam-directory nil)
+  (defvar org-roam-db-location nil)
+
+  :init
+  (doom-load-packages-incrementally
+   '(ansi-color dash f rx seq magit-section))
```
hlissner, line 27: Use `:defer-incrementally (ansi-color dash f rx seq magit-section)` instead

hlissner (changes requested):
Good stuff! Just needs one minor change and this is good to go.

nackjicholson:
Thanks @Wetlize ! Really glad to see this is in the works!

modules/lang/org/contrib/roam2.el
Wetlize: But if I do it, then Doom will add `org-roam` itself to `doom-incremental-packages` before `org` and its dependencies will be added there. AFAIK, because of that, `org-roam` will be loaded before `org` and its dependencies (since `org-roam` `require`s `org`), which in turn cause eager (instead of incremental) loading of `org`. Should I still do it?

modules/lang/org/doctor.el
```diff
@@ -5,7 +5,8 @@
   (unless (executable-find "gnuplot")
     (warn! "Couldn't find gnuplot. org-plot/gnuplot will not work")))
 
-(when (featurep! +roam)
+(when (or (featurep! +roam)
+          (featurep! +roam2))
   (unless (executable-find "sqlite3")
     (warn! "Couldn't find the sqlite3 executable. org-roam will not work."))
```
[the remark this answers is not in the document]
Wetlize: Sorry for bothering, but can you clarify a bit? Isn't `sqlite3` is a hard dependency of the package or I'm just missing something?

modules/lang/org/contrib/roam2.el
hlissner: Nevermind then. I'll have to fix load order for `:defer-incrementally` to address this. I'll fix this post-merge.

modules/lang/org/doctor.el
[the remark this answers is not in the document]
jethrokuan: in v2 we reverted to using `emacsql-sqlite`, which compiles the sqlite binary during installation and doesn't look for it using `executable-find`, so the check is redundant.

Wetlize:
Sorry folks. I've injured my hand today so this might not ship as fast as I was initially planning. Hope this will be completed by the end of this week.

pgronkievitz:
> Sorry folks. I've injured my hand today (nothing serious, but it will slow me down for some time) so this might not ship as fast as I was initially planning. Hope this will be completed by the end of this week or beginning of the next one.

I think it's not a problem to the most of people. Your health is way more important than writing this. Take care of yourself!

e3bc54b2:
> v2 reverted to using `emacsql-sqlite` instead of `emacsql-sqlite3`. It will now try to build the needed `sqlite3` executable itself, using available C compiler, instead of searching for it with `executable-find` and in case of failure, leaving the package in a half corrupt state.

Does this mean I now need gcc/clang in my path instead of sqlite3?

Wetlize:
> Does this mean I now need gcc/clang in my path instead of sqlite3?

Pretty much it. [Here](https://github.com/skeeto/emacsql/blob/9dca5996168c4963eb67e61c7f17fdcb8228e314/emacsql-sqlite.el#L67) you can see the list of compilers that will be tried by `emacs-sqlite`. Org-roam has also [manual entry](https://github.com/org-roam/org-roam/blob/master/doc/org-roam.org#c-compiler) about this. Doom will try to build the binary either, when you run `doom build` or when `org-roam` will try to query to the database the first time, The latter doesn't work ATM, so right now it will try to build it when the package will be loaded.

Wetlize:
This should be good to go for the review. I'm only not sure about 82761c18475f807b0709843c7df78a0fe312075e, which I'm not aware of a better way to handle this problem.

hlissner (changes requested):
Could you reword 94add08102a129f13ae9b4672e58996e50addfd6 to:

```
Bump org-roam

org-roam/org-roam@e997c01 -> org-roam/org-roam@028c95a
```

Bump commits should always follow this format.

Other than this, I think this is ready to merge.

Wetlize:
Sure, I though you'll squash it all (or ask me to do that).

Wetlize:
Should be good to go. I've also cleaned up the history a bit to make it a bit less noisy.

hlissner (approved):

hlissner:
Great work! Thanks again for your help!

memeplex:
Thanks! A reference to the new flag should be added to the list in modules.org.

hlissner:
I'm currently rewriting and redesigning our module documentation. `+roam2` will be added to it when it is done.

PR #5274, opened 2021-07-20
Fix doom/help-packages to correctly print config locations

Yp3rion:
Sometimes the `doom/help-packages` function encounters an error while printing the list of locations where the specified package is configured, if the line of code contains colon characters. This happens because the colon is interpreted as a separator between arguments to `cl-destructuring-bind`, thus causing a "wrong number of arguments" type of error if exactly three arguments are expected (filename, line number and line of code).

With the proposed fix, `cl-destructuring-bind` can accept an arbitrary number of arguments (three or more), thus removing the error without impacting code logic since only the first two are actually used.

### DIFF
--- a/core/autoload/help.el
+++ b/core/autoload/help.el
@@ -594,7 +594,7 @@ If prefix arg is present, refresh the cache."
               (insert "This package is configured in the following locations:")
               (dolist (location configs)
                 (insert "\n" indent)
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Makes sense to me. Thanks!

PR #5279, opened 2021-07-22
Improve multiple-cursors compatibility with evil-multiedit

daanturo:
After exiting `multiple-cursors`, when the previous state is
multiedit{,-insert}, return to normal state.

Use case:
- Mark occurrences with evil-multiedit
- `iedit-switch-to-mc-mode` ("M-M")
- Make some edit
- "C-g" to exit `multiple-cursors`

Expected behaviour: Return to evil's normal-state
Observed behaviour before the patch: Emacs stays on emacs-state


hlissner (approved):

hlissner:
Thanks!

PR #5281, opened 2021-07-22
Add bindings for yanking and pasting entire buffers

deuill:
This commit adds two leader-key bindings, `SPC b Y` and `SPC b P`, that
allow for yanking (copying) from and pasting into/replacing the entire
buffer under focus. These are generally useful when consolidating
changes between files without having to copy between them, or as a
shortcut for copying content into a scratch buffer.

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -349,12 +349,14 @@
        :desc "New empty buffer"            "N"   #'evil-buffer-new
        :desc "Kill other buffers"          "O"   #'doom/kill-other-buffers
        :desc "Previous buffer"             "p"   #'previous-buffer
+       :desc "Paste and replace buffer"    "P"   #'+default/paste-buffer
        :desc "Revert buffer"               "r"   #'revert-buffer
        :desc "Save buffer"                 "s"   #'basic-save-buffer
        :desc "Save all buffers"            "S"   #'evil-write-all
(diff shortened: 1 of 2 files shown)


deuill:
These are found in Spacemacs, and were useful enough for me (at least) to bring over to Doom when I migrated some time ago. I figured they might be useful to others.

I've only added Evil leader key bindings for these, as Emacs leader key bindings don't have a buffer section, and I wasn't sure if these were appropriate under the file section.

Wetlize:
I think this is redundant in Doom, since there's "whole buffer" text object available in Doom (`+evil:whole-buffer-txtobj`) that you can access with <kbd>a/i</kbd> + <kbd>G</kbd> when pending for a text object or motion. So if you need to yank the whole buffer you do <kbd>yaG</kbd> or if you need to clear buffer contents you do <kbd>daG</kbd>, etc.

hlissner:
I'll sit on this for a while. At the moment I think it's redundant with pre-existing functionality Doom provides for evil users: the `g` text object:

+ `yig` yanks the buffer
+ `vig` will select the buffer, then `p` will replace+paste it

And as leader keys are valuable real estate, I avoid these redundancies.

deuill:
Absolutely, and `g` object for the buffer is new to me -- I'd be happy to help document this and close this PR. Thanks for looking at this so promptly!

PR #5287, opened 2021-07-24
editor/format: provide default formatter

Mic92:
Reading the documentation there is no indication that one has to set a formatter per language in order to use it. After reading the upstream's documentation I found out that one has to call  `format-all-ensure-formatter` in order to achieve set.

### DIFF
--- a/modules/editor/format/config.el
+++ b/modules/editor/format/config.el
@@ -51,6 +51,8 @@ This is controlled by `+format-on-save-enabled-modes'."
 (when (featurep! +onsave)
   (add-hook 'after-change-major-mode-hook #'+format-enable-on-save-maybe-h))
 
+;; set default formatter for languages the user has not set any formatters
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but I must refuse it.

1. `format-all-ensure-formatter` was introduced in a future commit of format-all that Doom does not use.
2. The `format-all` package is frozen at its current commit because the module will be rewritten soon, to use a different package.
3. In the current incarnation of the module, the formatter is chosen _at format time_. This change would cause it to happen at buffer load time. This is not desirable because:
+ The formatter may be changed midsession or by other initialization processes, 
+ It eagerly loads the formatter long before it is needed,
+ It breaks our ability to change formatter from dir/file local variables.

Mic92:
Ok. It might be due to my setup that I was using this module instead of the version that doom-emacs provides.

Mic92:
Is there already a plan which other module doom-emacs will switch to?

PR #5290, opened 2021-07-25
Add org-cite to :tools/biblio

bdarcus:
Sorry; because of the massive recent merges, and my shaky git skills, I had to redo #5212 to get the git correct. 

## Summary

This adds support for native org-mode `org-cite` citations. Specifically it:

1. configures the native `org-cite` processors
2. adds the `oc-bibtex-actions` processor for the vertico module, with "follow", "activate", and "insert" capabilities
3. ~adds and configures the new `org-ref` OC rewrite: https://github.com/jkitchin/org-ref-cite~ (removed for now because of unclear future)
3. installs [citeproc-el](https://github.com/andras-simonyi/citeproc-el), which is required for the `oc-csl` export processor, and will be needed also for forthcoming preview functionality

_Should bump org along with this_, as this depends on a recent change in org (namely that now loads oc-basic, while it didn't initially; so I removed that `use-package!` statement).

https://git.savannah.gnu.org/cgit/emacs/org-mode.git/commit/?id=745f0913fc153322f5abb11aba377415106bd97a

### What to do with #2888?

1. Drop `org-ref` from there (can revisit in the future).
2. Merge this.
3. That can rebase on this and just focus on `org-roam-bibtex`.

On `org-ref` specifically:

It's status has evolved quite a bit over the last few months.

Its author initially published a development repo for an org-ref-cite package, which would have fit beautifully here, and which I initially added.

But since then he basically dropped that and in effect reimplemented aspects of org-cite in an incompatible syntax in org-ref v3.

In the process he removed some of the problematic dependencies and such, and it is possible to use org-cite (for citations) with org-ref (for its other functionality).

It's also possible to use org-ref for all of this, and ignore org-cite.

IMO this was a bad move from the broader org POV, and puts developers, users, and projects like Doom in an awkward situation. [An explanation](https://github.com/jkitchin/org-ref#what-about-org-cite) is included in the README, the first part I agree with, the second which I think is just personal opinion that many will not agree with.

I personally don't think we should support it and focus instead on native org citations. But others may have reasonable differences of opinion.

### Keybindings?

This is tricky. 

We need to bind **`org-cite-insert`** and **`org-open-at-point`**, as these are the two entry points for the org-cite module. 

But how, and where? 

For now, I have added a binding for `org-cite-insert`.

But `org-open-at-point` is also more general, so probably should be in more than one place. 

A biblio submenu would also be helpful given the above, and that `bibtex-actions` (for `vertico`) has multiple entry points, and a keymap to access them (which is also used by `embark`).

### DIFF
--- a/modules/tools/biblio/README.org
+++ b/modules/tools/biblio/README.org
@@ -1,0 +1,151 @@
+#+TITLE:   tools/biblio
+#+DATE:    April 11, 2020
+#+SINCE:   3.0
+#+STARTUP: inlineimages
+
+* Table of Contents :TOC_3:noexport:
+- [[#description][Description]]
+  - [[#maintainers][Maintainers]]
+  - [[#module-flags][Module Flags]]
+  - [[#plugins][Plugins]]
+- [[#prerequisites][Prerequisites]]
+  - [[#pdf-viewing][PDF viewing]]
+  - [[#bibtex-completion][Bibtex completion]]
+- [[#features][Features]]
+- [[#configuration][Configuration]]
+  - [[#org-cite][Org-cite]]
+    - [[#processor-configuration][Processor configuration]]
+    - [[#other-configuration-options][Other configuration options]]
+  - [[#path-configuration][Path configuration]]
+  - [[#templates][Templates]]
+- [[#troubleshooting][Troubleshooting]]
+
+* Description
+This module adds tools to help when working with bibliographies and citations.
+Minimal modifications have been made to the packages and the configuration
+details are listed in [[*Configuration][Configuration]] below. Some sensible defaults have been
+selected so it should be possible to use without modifications.
+
+** Maintainers
++ [[https://github.com/bdarcus][bdarcus]]
++ [[https://github.com/brianmcgillion][bmg]]
+
+** Module Flags
+This module provides no flags.
+
+** Plugins
++  [[https://github.com/tmalsburg/helm-bibtex][bibtex-completion]]
++ ~:completion vertico~
+  + [[https://github.com/bdarcus/citar][citar]]
++ ~:completion helm~
+  + [[https://github.com/tmalsburg/helm-bibtex][helm-bibtex]]
++ ~:completion ivy~
+  + [[https://github.com/tmalsburg/helm-bibtex][ivy-bibtex]]
+
+* Prerequisites
+There are no hard dependencies for this module.
+
+** PDF viewing
+An application for opening PDF files is required. By default =DocView= is used
+though it is highly recommended to enable =:tools pdf= in your personal ~init.el~
+file to enable [[https://github.com/politza/pdf-tools][pdf-tools]].
+
+** Bibtex completion
+For vertico, helm, or ivy bibtex completion you should enable =:completion vertico=, =:completion helm=, or
+=:completion ivy= respectively.
+
+* Features
+Both [[https://github.com/tmalsburg/helm-bibtex][helm-bibtex]] (includes helm-bibtex, ivy-bibtex and bibtex-completion code)
+and [[https://github.com/bdarcus/bibtex-actions][citar]] provide an extensive range of features so it is best to check their
+respective sites for a full list of features.
+
+On a high-level you can expect:
++ bibliography management
++ Notes per reference
++ citation support
++ citation lookup
++ org integration for writing literate latex (org-roam)
++ fast indexing and searching of references.
+
+To understand the interaction of these packages this [[https://www.reddit.com/r/emacs/comments/cxu0qg/emacs_org_and_bibtex_as_alternative_to_zotero_and/eyqe4vq/][reddit]] thread will explain
+the unique features and the overlapping functionality if any.
+
+In addition, this module provides support for native Org-mode citations: =org-cite=.
+
+* Configuration
+
+For all these packages it is advisable to use ~(after! package)~ in your
+~config.el~ to override any default settings and tune the packages to your
+needs.
+
+** Org-cite
+
+*** Processor configuration
+
+=Org-cite= provides rich features and flexible configuration options via its "processor" capabilities.
+
+1. /insert/ provides =org-cite-insert= integration for inserting and editing citations.
+2. /activate/ provides fontification, previews, etc.
+3. /follow/ integrates contextual citation actions with =org-open-at-point=.
+4. /export/ for different output targets.
+
+This module makes available the following processors:
+
+1. The core =oc-basic=, =oc-natbib=, =oc-biblatex=, and =oc-csl=.
+2. [[https://github.com/bdarcus/bibtex-actions][citar]] for integration with =vertico= completion.
+
+For user-interface, the module configures these processors as follows for the different completion modules:
+
+| Feature  | Vertico | Ivy      | Helm     |
+|----------+---------+----------+----------|
+| Insert   | citar   | oc-basic | oc-basic |
+| Activate | citar   | oc-basic | oc-basic |
+| Follow   | citar   | oc-basic | oc-basic |
+
+
+*** Other configuration options
+
+If you like, you can also set the =oc-csl= processor to look in a specific
+directory for your CSL styles:
+
+#+BEGIN_SRC emacs-lisp
+(setq org-cite-csl-styles-dir "~/Zotero/styles")
+#+END_SRC
+
+** Path configuration
+
+You must set the path variable for either =citar= (if using =vertico=
+completion) or =bibtex-completion= (if using =ivy= or =helm=); this module will
+in turn set the =org-cite-global-bibliography= path variable to this:
+
+#+BEGIN_src emacs-lisp
+(setq! bibtex-completion-bibliography '("/path/to/references.bib"))
+#+END_src
+
+#+BEGIN_src emacs-lisp
+(setq! citar-bibliography '("/path/to/references.bib"))
+#+END_src
+
+You may also set the respective note and library path variables as well for
+enhanced functionality:
+
+#+BEGIN_src emacs-lisp
+(setq! bibtex-completion-library-path '("/path/to/library/path/")
+       bibtex-completion-notes-path "/path/to/your/notes/")
+#+END_src
+
+#+BEGIN_src emacs-lisp
+(setq! citar-library-paths '("/path/to/library/files/")
+       citar-notes-paths '("/path/to/your/notes/"))
+#+END_src
+
+** Templates
+
+This module provides reasonable default templates for the packages. However, if
+you wish to change these refer to the respective packages for in-depth
+instructions.
+
+* Troubleshooting
+# Common issues and their solution, or places to look for help.
+
+Look to the respective package repositories.
--- a/modules/tools/biblio/config.el
+++ b/modules/tools/biblio/config.el
@@ -1,11 +1,12 @@
 ;;; tools/biblio/config.el -*- lexical-binding: t; -*-
 
 (use-package! bibtex-completion
+  :when (or (featurep! :completion ivy)
+            (featurep! :completion helm))
   :defer t
   :config
   (setq bibtex-completion-additional-search-fields '(keywords)
-        bibtex-completion-pdf-field "file"));; This tell bibtex-completion to look at the File field of the bibtex to figure out which pdf to open
-
+        bibtex-completion-pdf-field "file"));; This tells bibtex-completion to look at the File field of the bibtex to figure out which pdf to open
 
 (use-package! ivy-bibtex
   :when (featurep! :completion ivy)
@@ -14,7 +15,46 @@
   (add-to-list 'ivy-re-builders-alist '(ivy-bibtex . ivy--regex-plus)))
 
 
-(use-package! citar
-  :when (featurep! :completion vertico)
-  :after embark
-  :defer t)
+;;; Org-Cite configuration
+
+(use-package! oc
+  :after org
+  :config
+  (map! :map org-mode-map
+        :localleader
+        :desc "Insert citation" "@" #'org-cite-insert)
+  (setq org-cite-global-bibliography
+        (let ((paths
+               (cond
+                ((boundp 'citar-bibliography) citar-bibliography)
+                ((boundp 'bibtex-completion-bibliography) bibtex-completion-bibliography))))
+          ;; Always return bibliography paths as list for org-cite.
+          (if (stringp paths) (list paths) paths))
+        ;; setup export processor; default csl/citeproc-el, with biblatex for
+        ;; latex
+        org-cite-export-processors
+        '((latex biblatex)
+          (t csl))
+        org-cite-insert-processor 'citar
+        org-cite-follow-processor 'citar
+        org-cite-activate-processor 'citar
+        org-support-shift-select t))
+
+
+  ;;; Org-cite processors
+(use-package! oc-biblatex
+  :after oc)
+
+(use-package! oc-csl
+  :after oc)
+
+(use-package! oc-natbib
+  :after oc)
+
+;;;; Third-party
+
+(use-package! citar-org
+  :when (featurep! :lang org +roam2)
+  :config
+  ;; Include property drawer metadata for 'org-roam' v2.
+  (setq citar-org-note-include '(org-id org-roam-ref)))
--- a/init.example.el
+++ b/init.example.el
@@ -85,6 +85,7 @@
 
        :tools
        ;;ansible
+       ;;biblio            ; Writes a PhD for you (citation needed)
        ;;debugger          ; FIXME stepping through code, to help you add bugs
        ;;direnv
        ;;docker
--- a/modules/tools/biblio/packages.el
+++ b/modules/tools/biblio/packages.el
@@ -1,12 +1,13 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; tools/biblio/packages.el
 
-(package! bibtex-completion :pin "b85662081de98077f13f1a9fac03764702325d28")
 (when (featurep! :completion ivy)
-  (package! ivy-bibtex :pin "b85662081de98077f13f1a9fac03764702325d28"))
+  (package! bibtex-completion :pin "bb47f355b0da8518aa3fb516019120c14c8747c9")
+  (package! ivy-bibtex :pin "bb47f355b0da8518aa3fb516019120c14c8747c9"))
 (when (featurep! :completion helm)
-  (package! helm-bibtex :pin "b85662081de98077f13f1a9fac03764702325d28"))
+  (package! bibtex-completion :pin "bb47f355b0da8518aa3fb516019120c14c8747c9")
+  (package! helm-bibtex :pin "bb47f355b0da8518aa3fb516019120c14c8747c9"))
 (when (featurep! :completion vertico)
-  (package! citar :pin "fd33f5c4f7981036a969b5ca8aaf42380848ab32"))
+  (package! citar :pin "41ec5d4d5d625f7d784b4de20d14b7bceaf1730c"))
 
-(package! citeproc :pin "0857973409e3ef2ef0238714f2ef7ff724230d1c")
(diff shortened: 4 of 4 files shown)


bdarcus:
FYI @iyefrat @tecosaur I closed the other PR and opened this one to get a clean history with the latest develop commits.

bdarcus:
While this is ready for input on the "issues/questions", I'm reverting to draft as, after talking to its author, I'm going to include at least one additional third-party processor: org-ref-cite.

modules/tools/biblio/config.el
```diff
@@ -6,17 +6,71 @@
...
   :config
   (add-to-list 'embark-keymap-alist '(bibtex . bibtex-actions-map)))
+
+(use-package! citeproc
+  :defer t)
+
+;;; Org-Cite configuration
+
+(use-package! oc
+  :after org
```
bdarcus, line 27:
```suggestion
  :after org bibtex-completion
```

roryk:
So psyched for this, thanks for how much work you have put into it.

modules/tools/biblio/config.el
```diff
@@ -6,17 +6,75 @@
...
+  (map! :map org-mode-map
+      :localleader
+      :desc "Insert citation" "@" #'org-cite-insert)
+  (setq org-cite-global-bibliography
+        (let ((paths (or bibtex-actions-bibliography
+                         bibtex-completion-bibliography)))
+          ;; Always return bibliography paths as list for org-cite.
+          (if (stringp paths) (list paths) paths)))
+  ;; setup export processor; default csl/citeproc-el, with biblatex for latex
+  (setq org-cite-export-processors
```
bdarcus, line 39: I'm getting this error, which points to a load order problem here.

```elisp
Debugger entered--Lisp error: (void-function org-export-derived-backend-p)
  org-export-derived-backend-p(nil latex)
```

This happens when opening an `org-roam` note the first time, but goes away on subsequent times.

modules/tools/biblio/config.el
```diff
@@ -6,17 +6,76 @@
...
+
+(use-package! citeproc
+  :defer t)
+
+;;; Org-Cite configuration
+
+(use-package! oc
+  :after org bibtex-completion bibtex-actions
+  :init
+  :config
```
tecosaur, line 29: Why the empty `:init`?

modules/tools/biblio/config.el
```diff
@@ -6,17 +6,76 @@
...
+  (require 'ox)
+  (map! :map org-mode-map
+      :localleader
+      :desc "Insert citation" "@" #'org-cite-insert)
```
tecosaur, line 33: (nit) This indentation doesn't look right, should it be?:
```
  (map! :map org-mode-map
        :localleader
        :desc "Insert citation" "@" #'org-cite-insert)
```

modules/tools/biblio/config.el
bdarcus: Oversight; fixed.

modules/tools/biblio/config.el
bdarcus: Yep; also fixed.

Thanks.

modules/tools/biblio/README.org
```diff
@@ -0,0 +1,146 @@
...
+
+If you like, you can also set the =oc-csl= processor to look in a specific directory for your CSL styles:
+
+#+BEGIN_SRC emacs-lisp
+(setq org-cite-csl-styles-dir "~/Zotero/styles")
+#+END_SRC
+
+** Path configuration
+
+You must set the path variable for either =bibtex-actions= (if using =vertico= completion) or =bibtex-completion= (if using =ivy= or =helm=); this module will in turn set the =org-cite-global-bibliography= path variable to this:
```
yuuyins, line 119: I'm using vertico. If I

1. Set bibtex-actions-bibliography to bib.bib and bibtex-actions-notes-path to ~/org/ref, and set to nil both bibtex-completion-bibliography and bibtex-completion-notes-path.
2. Put cursor in org-cite citation
3. Try org-ref-cite-open-notes

I get `(wrong-type-argument stringp nil)` https://bin.disroot.org/?aaae5f2c5d0f318a#3E715k3UbbNHCWctZANRSge2diwir9u5WU1aLd8JYqhj


org-ref-cite-open-notes only works if I set bibtex-completion-bibliography to bib.bib and bibtex-completion-notes-path to ~/org/ref

bdarcus: Oh, yeah, I guess I need to generalize those instructions.

Why are you using that command though; in what context?

yuuyins: @bdarcus im just using the defaults processors from this module, i didn't change them.
org-cite-processors

bdarcus: Hmm ... I don't understand why you're seeing that hydra menu then.

You should be seeing the embark/which-key menu.

bdarcus:
With https://github.com/hlissner/doom-emacs/pull/5290/commits/ffdad5920571ab444b3fc07766fc2d1c9bc3403c I've reverted addition of `org-ref-cite` because of its unclear future.

The linter is complaining, so I will likely to need to rebase all this.

modules/tools/biblio/README.org
bdarcus: Since I removed `org-ref-cite` (also, updated the README), can you give another try and see how it works now?

yuuyins: just as a quick feedback i needed to manually set org-cite-global-bibliography because the module wasn't setting it from setting `bibtex-actions-bibliograph` alone. And setting `bibtex-actions-notes-path` doesn't work for `C-c C-o` or `bibtex-actions-dwim` in a org-cite citation, so I'm setting `bibtex-actions-notes-paths` directly https://gist.github.com/yuuyins/a7aeadc2e4ac524d5ad565c5e7e23d52 which makes it work to find bibliography notes. (either with or without `after bibtex-actions`). 

Also it seems `bibtex-actions-bibliography` varaible description is incorrect or some bug is happening: the variable says a list, so i initially tried setting it to a list, but then bibtex-actions-insert i get `(wrong-type-argument stringp ("/mnt/pro/org/org/bib/bib.bib")`. So i needed to insert a stringp for  bibtex-actions-insert.

There's an issue, but I'm not sure if it's by design. That menu on `bibtex-actions-dwim` only is available to org-ref citations (`bibtex-actions-insert-citations`), i.e. not available to org-cite citations (`org-cite-insert`). This is kinda an issue for me because now, since org-ref-cite was removed and that hydra menu isn't appearing, I can't create a new bibliographic note from org-cite citations.

https://user-images.githubusercontent.com/86538850/131725036-e2283922-ee14-4f21-93fb-5e1d0fead902.mp4

bdarcus: > just as a quick feedback i needed to manually set org-cite-global-bibliography because the module wasn't setting it from setting `bibtex-actions-bibliograph` alone. And setting `bibtex-actions-notes-path` doesn't work for `C-c C-o` or `bibtex-actions-dwim` in a org-cite citation, so I'm setting `bibtex-actions-notes-paths` directly https://gist.github.com/yuuyins/a7aeadc2e4ac524d5ad565c5e7e23d52 which makes it work to find bibliography notes. (either with or without `after bibtex-actions`).
> 
> Also it seems `bibtex-actions-bibliography` varaible description is incorrect or some bug is happening: the variable says a list, so i initially tried setting it to a list, but then bibtex-actions-insert i get `(wrong-type-argument stringp ("/mnt/pro/org/org/bib/bib.bib")`. So i needed to insert a stringp for bibtex-actions-insert to work.

Hmm ... I've not seen this, but I'll take another look at the config.el file.
 
> There's an issue, but I'm not sure if it's by design. That menu on `bibtex-actions-dwim` only is available to org-ref citations (`bibtex-actions-insert-citations`), i.e. not available to org-cite citations (`org-cite-insert`). This is kinda an issue for me because now, since org-ref-cite was removed and that hydra menu isn't appearing, I can't create a new bibliographic note from org-cite citations.

What about `embark-act`? Is that what you mean?

If yes, this is what I see:

image

yuuyins: I wasn't aware about embark-act, it does what I need. Thank you! However, I couldn't find a way to customize a template for the note as I could with org-ref-cite by customizing `bibtex-completion-notes-template-multiple-files`. So, the only way is to override `bibtex-actions-file-open-notes-default`? I've looked into `bibtex-actions-template`, but it doesn't seem to do what I need.

pgronkievitz:
It's just my suggestion, but I think it's reasonable
Maybe store all of the keybindings in `SPC-m-R` space? it would stand for 'reference' - it's not ideal, but it's the best I can think of since `SPC-m-c`, `SPC-n-r` and `SPC-m-r` are taken
Another option is `SPC-m-z` as it has one less keystroke, but it has no mnemonic (`z` like Zotero maybe?)

yuuyins:
@pgronkievitz what is `SPC-m`?

I personally use the prefix `n b` (like note -> bibliobraphy) which is by default `bibtex-actions-open-entry`, so repurposing that might be an idea as well

```
(map! :leader
      (:prefix "n b"
       (:prefix "a"
        "a" #'bibtex-actions-add-pdf-attachment
        "l" #'bibtex-actions-add-pdf-to-library)
       ("c" #'org-cite-insert)
       ("d" #'bibtex-actions-dwim)
       (:prefix "i"
        "b" #'bibtex-actions-insert-bibtex
        "c" #'bibtex-actions-insert-citation
        "k" #'bibtex-actions-insert-key
        "p" #'bibtex-actions-insert-preset
        "r" #'bibtex-actions-insert-reference)
       ("I" #'org-cite-insert)
       (:prefix "o"
        "o" #'bibtex-actions-open
        "e" #'bibtex-actions-open-entry
        "l" #'bibtex-actions-open-link
        "n" #'bibtex-actions-open-notes
        "p" #'bibtex-actions-open-pdf)
       ("r" #'bibtex-actions-refresh)
       ("R" #'bibtex-actions-read)))
```

pgronkievitz:
> @pgronkievitz what is `SPC-m`?

yeah, I probably should use a bit better wording. What I meant is to use "localleader" instead of "notes" prefix, because "notes" are available globally and those commands are useful exclusively (or at least mostly) inside org file, so in my opinion it doesn't make any sense to have access to e.g. `#'bibtex-actions-insert-bibtex` in dashboard.

bdarcus:
> What I meant is to use "localleader" instead of "notes" prefix, because "notes" are available globally and those commands are useful exclusively (or at least mostly) inside org file, so in my opinion it doesn't make any sense to have access to e.g. `#'bibtex-actions-insert-bibtex` in dashboard.

It's actually more complicated than this though.

- some commands only make sense in org; namely `org-cite-insert`
- some commands (arguably) only make sense in latex or markdown (the bibtex-actions insert commands)
- but it can be useful to browse outside these modes, which is why the original binding for `ivy-bibtex` and `helm-bibtex` is what it is (global)

mbosley:
I've been playing around with this PR, since I'm very interested in moving away the fairly bloated org-ref package. While I've been able to insert citations and open PDFs with org-cite just fine, one issue I've been having is getting bibtex-actions to play nicely with the existing notes in my `roam` folder. I set `bibtex-actions-notes-path` to my roam folder, where I have an existing set of notes for citations of the form `bibtexkey.org`. When I try to access the citation notes through Embark Act, I get the error `find-file-noselect: Wrong type argument: stringp, nil`. Any advice on fixing this?

bdarcus:
> When I try to access the citation notes through Embark Act, I get the error `find-file-noselect: Wrong type argument: stringp, nil`. Any advice on fixing this?

It's likely a bug. Can you report it on the bibtex-actions tracker?

While you do, can you report if `bibtex-actions-open-notes` works correctly? My suspicion is not; that i'ts not an embark issue.

bdarcus:
Sigh ... something went wrong with git.

ErnestDong:
Could you please make another PR to fix the name change of your [package](https://github.com/bdarcus/citar/pull/349) if you have time? it breaks doom-cli since there is no bibtex-actions any more.

modules/tools/biblio/config.el
```diff
@@ -6,17 +6,60 @@
...
+  :defer t)
+
+;;; Org-Cite configuration
+
+(use-package! oc
+  :after org bibtex-completion citar
   :config
-  (add-to-list 'embark-keymap-alist '(bibtex . bibtex-actions-map)))
+  (require 'ox)
+  (map! :map org-mode-map
```
cdlm, line 31: Shouldn't wait until after `oc` is loaded to setup this binding; on the contrary, `org-cite-insert` should be available in org-mode buffers, and should autoload `oc` if needed.

bdarcus: I'm not actually clear on what the suggest change is, not having a good handle on keybindings in doom.

cdlm: Remove that `(map! …)` from `(use-package! oc …)` and add it back out of it:
```emacs-lisp
(map! :after org
      :map org-mode-map
      :localleader
      :desc "Insert citation" "@" #'org-cite-insert)
```
n.b. the `:after org` clause, or wrap the binding in `(after! org …)`

bdarcus: @tecosaur - any input on this binding question?

tecosaur: It looks sensible to me :+1:

modules/tools/biblio/config.el
```diff
@@ -1,22 +1,63 @@
...
+        :localleader
+        :desc "Insert citation" "@" #'org-cite-insert)
+  (setq org-cite-global-bibliography
+        (let ((paths (or citar-bibliography
+                         bibtex-completion-bibliography)))
```
tecosaur, line 34: Shouldn't this be wrapped with `bound-and-true-p` ?

tecosaur (changes requested):

bdarcus:
I think we recent changes, can probably remove that entirely.

tecosaur:
Even better!

bdarcus:
Except, as I think about it, does it make any practical difference?

The command only makes sense in an org buffer. And if one opens an org
buffer, it should load the binding.

Right?

tecosaur:
@bdarcus I'm no longer sure which of my two messages your comments are referring to.

bdarcus:
> @bdarcus I'm no longer sure which of my two messages your comments are referring to.

Let me move it to the thread it belongs, now that i"m a computer.

modules/tools/biblio/config.el
bdarcus: Except, as I think about it, does it make any practical difference?

The command only makes sense in an org buffer. And if one opens an org
buffer, it should load the binding.

Right?

modules/tools/biblio/config.el
bdarcus: On this, you're referring to line 34?

modules/tools/biblio/config.el
tecosaur: So you currently have the line `:after org bibtex-completion citar` after `(use-package! oc`. This seems equivalent if `citar` is loaded at/before `org`. It looks like you currently load `citar` eagerly?

modules/tools/biblio/config.el
tecosaur: It sounds like you may be able to get away with just `(let ((paths citar-bibliography))` then, or just using the body and replacing `paths` with `citar-bibliography`.

modules/tools/biblio/config.el
bdarcus: >  It looks like you currently load citar eagerly?

That's not my intention.

Perhaps I should just remove the `:after` line?

modules/tools/biblio/config.el
```diff
@@ -1,22 +1,65 @@
...
+  (map! :map org-mode-map
+        :localleader
+        :desc "Insert citation" "@" #'org-cite-insert)
+  (setq org-cite-global-bibliography
+        (let ((paths
+               (cond
+                ((boundp 'citar-bibliography) citar-bibliography)
+                ((boundp 'bibtex-completion-bibliography) bibtex-completion-bibliography)))
+          ;; Always return bibliography paths as list for org-cite.
+          (if (stringp paths) (list paths) paths)))
```
cdlm, line 38:
```suggestion
          (if (stringp paths) (list paths) paths))))
```

hlissner (changes requested):

modules/tools/biblio/config.el
```diff
@@ -1,20 +1,65 @@
...
+(use-package! citar-org
+  :no-require
+  :custom
+  (org-cite-insert-processor 'citar)
+  (org-cite-follow-processor 'citar)
+  (org-cite-activate-processor 'citar)
+  (org-support-shift-select t)
+  (when (featurep! :lang org +roam2)
+    ;; Include property drawer metadata for 'org-roam' v2.
+    (citar-org-note-include '(org-id org-roam-ref))))
```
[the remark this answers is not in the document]
bdarcus: I think I resolve all your comments @hlissner.

hlissner:
> We need to bind org-cite-insert and org-open-at-point, as these are the two entry points for the org-cite module.
> 
> But how, and where?

How about adding a `citation` clause [to +org/dwim-at-point](https://github.com/hlissner/doom-emacs/blob/rewrite-docs/modules/lang/org/autoload/org.el#L125) (which is already on `RET` in normal mode and org already binds `C-c C-o` to `org-open-at-point`)?

```diff
       (while (and context (memq type '(verbatim code bold italic underline strike-through subscript superscript)))
         (setq context (org-element-property :parent context)
               type (org-element-type context)))
       (pcase type
+        (`citation
+         (org-cite-follow context arg))
+
         (`headline
          (cond ((memq (bound-and-true-p org-goto-map)
                       (current-active-maps))
                 (org-goto-ret))
```

As for inserting a citation, considering `org-cite-insert` is now built into org, I think we can add a localleader key to the `:lang org` module for it, perhaps under [the `i` prefix](https://github.com/hlissner/doom-emacs/blob/rewrite-docs/modules/lang/org/config.el#L822-L826) with the other insert commands.

That said, I'm to redesign our org module after the CLI rewrite. It's much overdue, so I'm fine with deciding on a temporary keybind for now.

modules/tools/biblio/config.el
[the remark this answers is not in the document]
hlissner: > I removed the require ox line, and I no longer see that problem. Perhaps it was fixed upstream?

It may depend on the entry point. `ox` explicitly loads `oc`, so if you load `oc` directly, you might still see this issue, though I'm not certain in what scenarios this  would happen naturally.

bdarcus:
> > We need to bind org-cite-insert and org-open-at-point, as these are the two entry points for the org-cite module.
> > But how, and where?
> 
> How about adding a `citation` clause [to +org/dwim-at-point](https://github.com/hlissner/doom-emacs/blob/rewrite-docs/modules/lang/org/autoload/org.el#L125) (which is already on `RET` in normal mode and org already binds `C-c C-o` to `org-open-at-point`)?

That seems sensible at first glance (I wasn't aware of that function), though you'd also want it available for `citation-reference`.

With `citar`, I have left-click on the mouse set to run `org-open-at-point`, and `right-click` to run `embark-act`. The `org-cite-activate-processor` attaches a map with those and other bindings to the citations.

> As for inserting a citation, considering `oc` is now built into org, I think we can add a localleader key to the `:lang org` module for `org-cite-insert`, perhaps under [the `i` prefix](https://github.com/hlissner/doom-emacs/blob/rewrite-docs/modules/lang/org/config.el#L822-L826) with the other insert commands.

Makes sense.

> That said, I'm to redesign our org module after the CLI rewrite. It's much overdue, so I'm fine with deciding on a temporary keybind for now.

I have no strong opinion, but lean towards deferring to the rewrite. I think it's fine as is for now.

hlissner (approved):

bdarcus:
Oh sorry @hlissner - I just bumped the module again to incorporate a minor typo.

hlissner (approved):

hlissner:
Sorry for the _incredibly_ long turnaround time, but I'm happy this is finally merged. Thanks again for all your help!

PR #5291, opened 2021-07-25
Don't hard-code preview key for various consult commands

daanturo:
Currently, if users customize `consult-preview-key` themselves, it's overridden by C-{SPC,M-j,M-k} unless they manually customize for all those commands.

For example:
```lisp
(setq consult-preview-key (seq-map #'kbd '("M-RET" "M-<up>" "M-<down>")))
(map! :map vertico-map
      "M-<down>" #'+vertico/next-candidate-preview)
```

Then when using consult-ripgrep (and many other consult commands), "M-\<down\>" wouldn't work unless those customizations are disabled.
Edit: provide example

### DIFF
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -80,20 +80,23 @@ variable overrides `completion-styles' during company completion sessions.")
   (setq consult-project-root-function #'doom-project-root
         consult-narrow-key "<"
         consult-line-numbers-widen t)
-  (consult-customize
-   consult-ripgrep consult-git-grep consult-grep
-   consult-bookmark consult-recent-file
-   +default/search-project +default/search-project-for-symbol-at-point
-   +default/search-other-project +vertico/search-symbol-at-point
-   +default/search-cwd +default/search-other-cwd
-   +default/search-notes-for-symbol-at-point
-   consult--source-file consult--source-project-file consult--source-bookmark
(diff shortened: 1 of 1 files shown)


minad:
@daanturo Why not reset `consult--read-config` in your configuration if you don't want the customizations? Checking `consult-preview-key` to opt-in to the customizations seems to be a bit artifical. cc @iyefrat

daanturo:
> Why not reset `consult--read-config`

Thank you for pointing out, I didn't know about it since it wasn't documented in consult's README.

> Checking consult-preview-key to opt-in to the customizations seems to be a bit artifical

About this,  should customizing like the following be okay?

```lisp
:preview-key `(,@(doom-enlist consult-preview-key) ,@(list (kbd "C-SPC") (kbd "C-M-j") (kbd "C-M-k")))
```

Out of 5 types of `consult-preview-key`:

```lisp
  :type '(choice (const :tag "Any key" any)
                 (list :tag "Debounced" (const :debounce) (float :tag "Seconds" 0.1) (const any))
                 (const :tag "No preview" nil)
                 (key-sequence :tag "Key")
                 (repeat :tag "List of keys" key-sequence))
```

It can satisfy the later 3, but I am not so sure about the first 2.

iyefrat:
>  Why not reset consult--read-config in your configuration

Ooo this should be in the module readme.

PR #5292, opened 2021-07-25
Improve PHP module

elken:
* Add more detailed README
* Add `composer.el`
* Create a minor mode for phpunit/docker tests

-------

Fixes #0000 


modules/lang/php/README.org
```diff
@@ -130,16 +139,95 @@ export PATH="~/.composer/vendor/bin:$PATH"
...
+  (require 'auth-source)
+  (let ((match (car (apply #'auth-source-search params))))
+    (if match
+        (let ((secret (plist-get match :secret)))
+          (if (functionp secret)
+              (funcall secret)
+            secret))
+      (error "Password not found for %S" params))))
+
+(setq lsp-intelephense-license-key (my-fetch-password :user intelephense))
```
hlissner, line 232: Please run `M-x org-fill-paragraph` on each paragraph here (excluding src blocks and tables).

modules/lang/php/config.el
```diff
@@ -126,6 +129,21 @@
...
+        :prefix ("c" . "composer")
+        "c" #'composer
+        "i" #'composer-install
+        "r" #'composer-require
+        "u" #'composer-update
+        "d" #'composer-dump-autoload
+        "s" #'composer-run-script
+        "v" #'composer-run-vendor-bin-command
+        "o" #'composer-find-json-file
+        "l" #'composer-view-lock-file))
```
hlissner, line 146: Why are we loading `composer` after `php-mode`? Couldn't this be lazy loaded on any of its commands? e.g.

```elisp
(use-package! composer
  :defer t
  :init
  (map! :after php-mode
        :localleader
        :map php-mode-map
        :prefix ("c" . "composer")
        ...))
```

This way, `composer` won't be loaded until one of its commands are invoked.

hlissner: Also, please make sure there are _two_ blank lines above and below `composer`'s `use-package!` blocks.

hlissner (changes requested):

modules/lang/php/README.org
elken: My bad, been a while since I've contributed! Will do

modules/lang/php/config.el
elken: Absolutely no reason, thanks!

elken: > Also, please make sure there are _two_ blank lines above and below `composer`'s `use-package!` blocks.

Not a problem, what's the reasoning behind 2?

elken:
Have I understood `def-project-mode!` correctly? The intention is to only set those variables while the mode is active.

modules/lang/php/config.el
```diff
@@ -3,6 +3,9 @@
 (defvar +php--company-backends nil
   "List of company backends to use in `php-mode'.")
 
+(defvar +php/default-docker-container "php-fpm"
```
hlissner, line 6: Rename variable to `+php-default-docker-container`. The `+X/Y` naming convention is for interactive functions.

modules/lang/php/README.org
```diff
@@ -130,16 +140,103 @@ export PATH="~/.composer/vendor/bin:$PATH"
...
+#+END_SRC
```
hlissner, line 242: Lower case `#+BEGIN_SRC` and `#+END_SRC` for consistency.

hlissner (changes requested):

hlissner:
> Have I understood `def-project-mode!` correctly? The intention is to only set those variables while the mode is active.

Yes, that's fine. Though, are you certain `phpunit-args` and `phpunit-executable` are buffer-local variables? If not, use `setq-local` instead of `setq`.

> I also realised I need to create a `+php/default-docker-compose` variable to expose for weird setups. Will the below work?
> 
> ```lisp
> :files (and "phpunit.xml" (or '+php/default-docker-compose "docker-compose.yml"))
> ```

It would work if the variable weren't quoted. e.g. 

```elisp
:files (and "phpunit.xml" (or +php/default-docker-compose "docker-compose.yml"))
```

modules/lang/php/README.org
elken: Will do!

elken:
> Yes, that's fine. Though, are you certain `phpunit-args` and `phpunit-executable` are buffer-local variables? If not, use `setq-local` instead of `setq`.

That's how they're designed to be afaict, correct me otherwise [defined here](https://github.com/nlamirault/phpunit.el/blob/master/phpunit.el#L139-L149).


> It would work if the variable weren't quoted. e.g.
> 
> ```emacs-lisp
> :files (and "phpunit.xml" (or +php/default-docker-compose "docker-compose.yml"))
> ```

I think I got that after I wrote the comment, work and such kept me from updating this. Thanks!

hlissner (changes requested):
The commits in this PR don't conform to our [git conventions](https://gist.github.com/hlissner/4d78e396acb897d9b2d8be07a103a854):

- Rebase the PR onto `develop`, rather than merging it (to remove the merge commit, 014e7702fcee55dbc2fcac17609ac0085c117983).
- Split up `feature: improve PHP module` into three commits:
  - `feat(php): add composer.el`
  - `feat(php): add project mode for phpunit/docker tests`
  - `docs(php): expand readme`
- Squash 36d4072268426cb4f9ab9fcbdc321b7e64e5937d into `feat(php): add composer.el` and `docs(php): expand readme` above.
- Reword cfcdc75f4bdccf4fea84624949fc81ff5b32d17a to `feat(php): add rainbow-delimiter-mode`

PR #5296, opened 2021-07-25
editor/evil: Add gr keybinding to refresh `helpful` buffers

Wetlize:
I think this was just underlooked? Otherwise there's no way to sync to the latest changes in state without <kbd>M-x</kbd>ing or bindings this in your private config.


hlissner (approved):

hlissner:
Thanks!

PR #5298, opened 2021-07-26
+lookup/in-docsets: Use counsel-dash temporarily for vertico

daanturo:
The command is now useless for vertico users, we should use counsel-dash when possible.
counsel-dash is chosen over helm-dash because it's closer to vertico's interface.

Revert me when the time comes!


KSSysLab:
There are 2 devdocs packages in melpa:
1. https://github.com/astoff/devdocs.el
2. https://github.com/blahgeek/emacs-devdocs-browser

Currently, dash is used on macOS and zeal on Linux, directly using the one of the devdocs packages makes doom-emacs have the consistent behavior on all platforms.

hlissner:
I've implemented `consult` support for `dash-docs` in 2d56dd7, which indirectly resolves this issue. Thanks for bringing it to my attention!

PR #5299, opened 2021-07-26
Vertico improvements: theming, bugfixes, new bindings, adaptations to upstream changes

iyefrat:
This PR contains:
- bumps for `:completion vertico` and `:ui doom`
- teco's recent theming improvements for marginalia and orderless
- better keys for `consult-crm`
- a fix for `<leader> c j` since it used `cmd!` instead of `cmd!!`
- binding the new `consult-line-multi` to `<leader s B>` (the prefix argument makes it search all buffers, without it it just searches buffers of the current project) 
- usage of the recently introduced new embark indicator format
- add marginalia annotations to functions that don't have them
- advise the marginalia annotations for the `project-file` category to use projectile
- a bugfix for `consult-recent-file`
- bugfixes for `+vertico/find-file-in`
- replacing an obsolete embark function
cc: @tecosaur @elken @minad

### DIFF
--- a/modules/completion/vertico/autoload/vertico.el
+++ b/modules/completion/vertico/autoload/vertico.el
@@ -168,6 +168,7 @@ Supports exporting consult-grep to wgrep, file to wdeired, and consult-location 
   "Jump to file under DIR (recursive).
 If INITIAL is non-nil, use as initial input."
   (interactive)
+  (require 'consult)
   (let* ((default-directory (or dir default-directory))
          (prompt-dir (consult--directory-prompt "Find" default-directory))
          (cmd (split-string-and-unquote consult-find-command " "))
@@ -181,7 +182,7 @@ If INITIAL is non-nil, use as initial input."
       :require-match t
       :initial (if initial (shell-quote-argument initial))
       :add-history (thing-at-point 'filename)
-      :category '+vertico
+      :category 'file
       :history '(:input +vertico/find-file-in--history)))))
 
 ;;;###autoload
--- a/modules/completion/vertico/autoload/vertico.el
+++ b/modules/completion/vertico/autoload/vertico.el
@@ -4,6 +4,10 @@
 ;;;###autoload
 (defvar orderless-match-faces)
 
+;; To prevent "Defining as dynamic an already lexical var" from +vertico/embark-preview
+;;;###autoload
+(defvar embark-quit-after-action)
+
 ;;;###autoload
 (defadvice! +vertico--company-capf--candidates-a (fn &rest args)
   "Highlight company matches correctly, and try default completion styles before
@@ -146,7 +150,7 @@ Supports exporting consult-grep to wgrep, file to wdeired, and consult-location 
   (unless (bound-and-true-p consult--preview-function)
     (save-selected-window
       (let ((embark-quit-after-action nil))
-        (embark-default-action)))))
+        (embark-dwim)))))
 
 ;;;###autoload
 (defun +vertico/next-candidate-preview (&optional n)
--- a/modules/completion/vertico/autoload/vertico.el
+++ b/modules/completion/vertico/autoload/vertico.el
@@ -14,6 +14,12 @@ orderless."
     (apply fn args)))
 
 ;;;###autoload
+(defadvice! +vertico--consult-recent-file-a (&rest _args)
+  "`consult-recent-file' needs to have `recentf-mode' on to work correctly"
+  :before #'consult-recent-file
+  (recentf-mode +1))
+
+;;;###autoload
 (cl-defun +vertico-file-search (&key query in all-files (recursive t) prompt args)
   "Conduct a file search using ripgrep.
 
--- a/modules/completion/vertico/autoload/vertico.el
+++ b/modules/completion/vertico/autoload/vertico.el
@@ -8,7 +8,7 @@
 (defadvice! +vertico--company-capf--candidates-a (fn &rest args)
   "Highlight company matches correctly, and try default completion styles before
 orderless."
-  :around 'company-capf--candidates
+  :around #'company-capf--candidates
   (let ((orderless-match-faces [completions-common-part])
         (completion-styles +vertico-company-completion-styles))
     (apply fn args)))
--- a/modules/completion/vertico/autoload/vertico.el
+++ b/modules/completion/vertico/autoload/vertico.el
@@ -237,3 +237,20 @@ targets."
                  (if (cdr targets) "…" "")))
        (if prefix (lookup-key keymap prefix) keymap)
        nil nil t))))
+
+;;;###autoload
+(defun +vertico/crm-select ()
+  "Enter candidate in `consult-completing-read-multiple'"
+  (interactive)
+  (let ((idx vertico--index))
+    (unless (get-text-property 0 'consult--crm-selected (nth vertico--index vertico--candidates))
+      (setq idx (1+ idx)))
+  (run-at-time 0 nil (cmd! (vertico--goto idx) (vertico--exhibit))))
+  (vertico-exit))
+
+;;;###autoload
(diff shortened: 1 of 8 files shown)


modules/ui/doom/packages.el
```diff
@@ -1,5 +1,5 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; ui/doom/packages.el
 
-(package! doom-themes :pin "5221b0600d9da16e0e3af332ff1fe2ef624f0af4")
+(package! doom-themes :pin "b7995ac041f8dadb021cd2445e85d29c9bf718ae")
```
hlissner, line 4: Would you mind bumping this again? To include https://github.com/hlissner/emacs-doom-themes/pull/640 and https://github.com/hlissner/emacs-doom-themes/pull/639

modules/completion/vertico/autoload/vertico.el
```diff
@@ -144,3 +144,10 @@ Supports exporting consult-grep to wgrep, file to wdeired, and consult-location
   (interactive)
   (vertico-previous (or n 1))
   (+vertico/embark-preview))
+
+;;;###autoload
+(defun +vertico/crm-select ()
+  "Select or deselect candidate in `consult-completing-read-multiple'"
+  (interactive)
+  (run-at-time 0 nil #'vertico-exit)
+  (funcall #'vertico-exit))
```
hlissner, line 153: Why is it necessary to call this twice and in this manner?

hlissner (changes requested):

modules/ui/doom/packages.el
iyefrat: Yeah the exact bump commits are placeholders for the moment, I'll bump both to the newest versions once everything is ready (baring any noticeable regressions)

modules/completion/vertico/autoload/vertico.el
iyefrat: Oh, this is a mistake, this is the crm exit function and `vertico-exit` is the selecting function. That's because the default `consult-crm` has you press `RET` once to select and twice to exit.

elken: Is one not to run when the minibuffer closes?

iyefrat:
I'm converting this back to draft because there are some cool upstream embark changes that would be a shame to wait until the next bump to get, but require a bit of manual work on my part to fix the which key intergration. Should be ready soonish.

modules/completion/vertico/config.el
```diff
@@ -138,11 +141,7 @@ overrides `completion-styles' during company completion sessions.")
   :config
   (set-popup-rule! "^\\*Embark Export Grep" :size 0.35 :ttl 0 :quit nil)
 
-  (setq embark-action-indicator
-        (lambda (map _target)
-          (which-key--show-keymap "Embark" map nil nil 'no-paging)
-          #'which-key--hide-popup-ignore-command)
-        embark-become-indicator embark-action-indicator)
+  (setq embark-indicator #'+vertico/embark-which-key-indicator)
```
iyefrat, line 144: Note that there is an interesting alternative to the which-key style indicator, the recently introduced verbose indicator (it's what's used by default). it looks like this:
image

bdarcus: Nice!

Hey, you should send that screenshot to Omar, who needs a nice screenshot!

iyefrat: Uh I think that screenshots on the embark readme should probably have things like a nonarbitrary aspect ratio and not cutting off lines of text at the bottom. Also they should probably be taken from a reasonably default config, e.g. not having the doom modeline, or icons.

Although this reminds me, I should probably add a screenshot to `all-the-icons-completion`

bdarcus: What's different about the new which-key one, other than that submenus work?

iyefrat: Nothing that I can notice to be honest

iyefrat:
@hlissner also note 9d2131b, the lazy loading of `recentf-mode` introduced in 97048e2 had the side effect having consult previews on when using `SPC f r`, due to the fact that `consult-customize` works on named commands. This fixes that using advice, hopefully that's fine.

numkem:
Is it possible to ad the icons to `projectile-find-file`?

Thanks!

iyefrat:
@numkem yeah I'll add this to `all-the-icons-completions` and bump it, good catch

modules/completion/vertico/config.el
```diff
@@ -175,10 +174,16 @@ overrides `completion-styles' during company completion sessions.")
...
+            '(doom/find-file-in-private-config . file)
+            '(doom/describe-active-minor-mode . minor-mode)
+            '(flycheck-error-list-set-filter . builtin)
             '(persp-switch-to-buffer . buffer)
             '(projectile-find-file . project-file)
-            '(doom/describe-active-minor-mode . minor-mode)
-            '(flycheck-error-list-set-filter . builtin)))
+            '(projectile-recentf . project-file)
+            '(projectile-switch-to-buffer . buffer)
+            '(projectile-switch-project . project-file)))
```
iyefrat, line 186: Some of the annotations in the `project-file` type commands where you first select a project and then a file are a bit shaky, due to marginalia not yet providing very good support for these types of project commands. For some commands (e.g. `doom/find-file-in-other-project`) both the project list and file list get file annotations, but for others such as `projectile-switch-project`, they don't really get annotations. This might be due to `marginalia` using `project.el` as the backend or something.

modules/completion/vertico/config.el
oantolin: The main point of having a new which-key indicator is that the indicator protocol changed so the old one won't work anymore! But also, the new one displays (in the which-key help message) whether you are using `embark-act` or `embark-become`, and also shows the current target and an ellipsis to indicate there are further targets you can cycle to.

modules/completion/vertico/config.el
```diff
@@ -174,11 +173,19 @@ overrides `completion-styles' during company completion sessions.")
   :config
   (when (featurep! +icons)
     (add-hook 'marginalia-mode-hook #'all-the-icons-completion-marginalia-setup))
+  (advice-add #'marginalia--project-root :override #'projectile-project-root)
```
iyefrat, line 176: should this be `#'doom-project-root`?

modules/config/default/+evil-bindings.el
```diff
@@ -708,7 +708,8 @@
              ((featurep! :completion ivy)       #'swiper)
              ((featurep! :completion helm)      #'swiper))
        :desc "Search all open buffers"      "B"
-       (cond ((featurep! :completion ivy)       #'swiper-all)
+       (cond ((featurep! :completion vertico)   (cmd!! #'consult-line-multi '(4)))
```
minad, line 711: I've seen this (4) pattern a few times, which emulates the interactive prefix argument. I recommend to use this only for commands where (4) actually carries some meaning and if not use something else. For `consult-line-multi` this works:

~~~ elisp
(consult-line-multi 'all-buffer)
(consult-line-multi '(:include ".*\\.el\\'" :exclude "init\\.el\\'"))
~~~

modules/completion/vertico/config.el
hlissner: Yes, please do

hlissner (changes requested):
Looks good! Just one correction and I think this is ready to merge.

modules/config/default/+evil-bindings.el
iyefrat: fixed

modules/completion/vertico/config.el
iyefrat: fixed

hlissner (approved):

hlissner:
Looks good to me. Great work! And thanks again!

PR #5302, opened 2021-07-27
Changed editor/format lsp feature requirement from textDocument/rangeFormatting to textDocument/formatting

holyfight6:
A lot of lsp servers don't support `textDocument/rangeFormatting` but support `textDocument/formatting`(eg: `rust-analyzer`). The fallback formatter has some issues (like in rust-mode it won't respect the project's cargo files). I think it's better to use lsp as the default formatter as long as the lsp server support file formating.


hlissner (approved):

hlissner:
Makes sense to me, thanks!

PR #5303, opened 2021-07-27
Add "SPC p &" binding as projectile-run-async-shell-command-in-root

daanturo:
Unfortunately I had to truncate "Run" from `:desc`, otherwise other 24 lines must be aligned, too (which will produce diff noises).


hlissner (approved):

hlissner:
Thanks!

PR #5308, opened 2021-07-28
vertico: add missing autoload

naistran:
This autoload is required for `embark-act` to work when first opening an `.el` file. Otherwise we get this error:

```
Symbol’s function definition is void: +vertico--embark-target-package
```


hlissner:
This was fixed in https://github.com/hlissner/doom-emacs/commit/551e5adf815936e572af24789044f735195d41a1. Thanks for bringing it to my attention!

PR #5309, opened 2021-07-28
Allow evil users to quickly close the LSP signature buffer

daanturo:
in normal state by ESC.

We should have a way to quickly escape this beside "C-g":
image

### DIFF
--- a/modules/tools/lsp/+lsp.el
+++ b/modules/tools/lsp/+lsp.el
@@ -58,6 +58,13 @@ about it (it will be logged to *Messages* however).")
   ;; REVIEW Remove this once this is fixed upstream.
   (add-to-list 'lsp-client-packages 'lsp-racket)
 
+  (add-hook! 'doom-escape-hook
+    (defun +lsp-signature-stop-maybe-h ()
+      "Close the displayed `lsp-signature'."
+      (when lsp-signature-mode
+        (lsp-signature-stop)
+        t)))
+
   (set-popup-rule! "^\\*lsp-help" :size 0.35 :quit t :select t)
   (set-lookup-handlers! 'lsp-mode
     :definition #'+lsp-lookup-definition-handler


modules/tools/lsp/+lsp.el
```diff
@@ -58,6 +58,9 @@ about it (it will be logged to *Messages* however).")
   ;; REVIEW Remove this once this is fixed upstream.
   (add-to-list 'lsp-client-packages 'lsp-racket)
 
+  ;; Close the `lsp-signature' buffer with ESC ESC
+  (add-hook 'doom-escape-hook 'lsp-signature-stop)
```
hlissner, line 62: Please sharp-quote the function as per our conventions:

```elisp
(add-hook 'doom-escape-hook #'lsp-signature-stop)
```

hlissner (changes requested):

modules/tools/lsp/+lsp.el
hlissner: Also, this is adding a global hook; what happens when `lsp-signatures-stop` is called in a non-LSP buffer?

daanturo: > Also, this is adding a global hook; what happens when `lsp-signatures-stop` is called in a non-LSP buffer?

As I tried, it didn't produce any messages or raise any errors/warnings.
But perhaps that's not guaranteed in the future, so let's make it lsp-mode only.

daanturo:
BTW do you think other hooks should be executed after closing the signature successfully while it's active? 
Currently the hook returns `t` (prevent other hooks) since in `lsp-signature-mode-map` (signature is shown), `C-g` is directly mapped to `lsp-signature-stop`, which doesn't do anything that is non-lsp.

modules/tools/lsp/+lsp.el
hlissner: > Another side note: another Emacs developer has noted that defun's return value is undefined, but I have seen that

I'm aware of this. `add-hook!` gets around this by reading the function name in the `defun` format.

hlissner:
> BTW do you think other hooks should be executed after closing the signature successfully while it's active?

Yes, that makes sense to me.

Quick question: is turning off `lsp-signature-mode` the right way to close the signature? Does LSP reactivate it every time the signature is shown?

daanturo:
> Quick question: is turning off `lsp-signature-mode` the right way to close the signature?

By inspecting LSP's source, I think it's not supposed to be activated directly but should be turn on/off by `lsp-signature-activate` & `lsp-signature-stop`, respectively as there are only bindings for the 2 command (but not the mode itself) and they also do many things before toggling `lsp-signature-more`.

>  Does LSP reactivate it every time the signature is shown?

Yes, `lsp-signature-mode` is temporary, it is deactivated automatically after we move the cursor out of the describe symbol.

hlissner (approved):

hlissner:
Understood, then this looks fine to merge. Thanks again for your help!

PR #5311, opened 2021-07-29
Fix +lookup/in-docsets's vertico section

daanturo:
Convert the let form to let* to fix `Symbol’s value as variable is void: sink`.


hlissner (approved):

hlissner:
Whoops. Good catch, thanks!

PR #5313, opened 2021-07-29
fix vertico /ssh: tramp hostname completion

stfl:
enable tramp hostname completion in vertico find file. according to vertico docs: https://github.com/minad/vertico#tramp-hostname-completion


iyefrat:
I think it would be a better idea to use the completion style method using `basic-remote` that's listed there.

stfl:
I  added the basic-remote completion and it works although I am not sure I can verify if everything else still works.

stfl:
should the overwrite be with orderless (like it was before) or without (like in the vertico doc)?

modules/completion/vertico/config.el
```diff
@@ -50,10 +50,19 @@ overrides `completion-styles' during company completion sessions.")
...
      ((string-suffix-p "~" pattern) `(orderless-flex . ,(substring pattern 0 -1)))))
+  (defun basic-remote-try-completion (string table pred point)
+    (and (vertico--remote-p string)
+         (completion-basic-try-completion string table pred point)))
+  (defun basic-remote-all-completions (string table pred point)
+    (and (vertico--remote-p string)
+         (completion-basic-all-completions string table pred point)))
+  (add-to-list
+   'completion-styles-alist
+   '(basic-remote basic-remote-try-completion basic-remote-all-completions nil))
```
iyefrat, line 61: Please rename these functions to comply with doom's [naming conventions](https://github.com/hlissner/doom-emacs/blob/develop/docs/contributing.org#naming-conventions).

The function definitions should also probably go in the `autoloads/vertico.el` and be autoloaded, but Henrik will have to verify that, I'm a bit fuzzy on that.

iyefrat:
This portion of the readme is currently going through an overhaul (started after you submitted the PR :D), and might also get released as a separate small package for extra completion styles. I suggest this PR be put on hold until that gets sorted. It might end up getting superseded by a solution that requires adding another package.

iyefrat:
Nevermind, the situation is staying as is in the end, so no changes relating to that needed.

iyefrat:
Hey, I've adapted this commit to the current vertico improvements pr (#5353), since the current one has some merge conflicts and doesn't follow the new git commit conventions (and it will also probably get merged faster this way :stuck_out_tongue:). I've added you as co-author as well.

PR #5315, opened 2021-07-29
Fixed the keybind for fold/unfold all

andrewsuperlegit:
The keybinding for fold all and unfold all was wrong (was missing a C-f)

Fixes #0000 


hlissner (approved):

hlissner:
Good catches, thanks!

PR #5316, opened 2021-07-30
Black list XDG_SESSION_TYPE from doom-env

daanturo:
The variable is obsolete once we switch between "x11" and "wayland".
Also Emacs can infer it perfectly without setting it first.


hlissner (approved):

hlissner:
Makes sense to me, thanks!

PR #5322, opened 2021-07-31
Enable rg's smart case for +vertico-file-search

daanturo:
What vanilla Emacs, Ivy and Helm do by default.

It makes searching much easier & friendlier.


hlissner (approved):

daanturo:
IMO `--smart-case` should be upstreamed for `consult`, too.

hlissner (approved):

hlissner:
> IMO `--smart-case` should be upstreamed for `consult`, too.

Agreed, but for now this will do. Thanks!

iyefrat:
Note that doom implements a custom `consult--grep` command for file searching so this fix is needed either way.

PR #5325, opened 2021-08-01
Update Go tool URL in README

snizovtsev:
As Gore repository has been moved we have to update URL because `go get` won't handle redirects.


hlissner (approved):

hlissner:
Thanks!

PR #5326, opened 2021-08-01
Make mixed-pitch-mode respect `doom-variable-pitch-font`'s size.

harryfei:
set `mixed-pitch-set-height` to true. So mixed-pitch-mode will respect `doom-variable-pitch-font`'s size.

### DIFF
--- a/modules/ui/zen/config.el
+++ b/modules/ui/zen/config.el
@@ -72,4 +72,6 @@
             'org-todo-keyword-outd
             'org-todo
             'org-done
-            'font-lock-comment-face))
+            'font-lock-comment-face)
+
+  (setq mixed-pitch-set-height 't))


tecosaur:
From memory there were some caveats associated with this, though I must admit I forget what.

hlissner:
Please test the following with `(setq mixed-pitch-set-height t)` in a buffer with `mixed-pitch-mode` enabled:

- `text-scale-increase` (`C-=`)
- `text-scale-decrease` (`C--`)
- `doom/increase-font-size` (`C-M-=`)
- `doom/decrease-font-size` (`C-M-=`)

iirc, mixed-pitch's hardcoding will prevent one or more of these from working. The latter two commands may be able to get around it by cycling `mixed-pitch-mode` in buffers where it's activated.

Collisions like these are common with face-remap.

harryfei:
`doom/increase-font-size` and `doom/increase-font-size` doesn't work with this patch.

hlissner:
Sorry for the tremendously late response: with the age of this PR and the caveats of the implementation it proposes, I will close it. I'm not against what it's requesting, however, and I've iterated on several solutions in the time this PR has been open, but I've yet to find an acceptable one. The search continues.

PR #5327, opened 2021-08-01
Allow mu4e to be started early

elken:
`mu4e` complains about a missing function if loaded too early. I don't know if
there's a cleaner way to do this, but it's needed AFAICT.

### DIFF
--- a/modules/email/mu4e/autoload/mu-lock.el
+++ b/modules/email/mu4e/autoload/mu-lock.el
@@ -35,6 +35,7 @@ If STRICT only accept an unset lock file."
 (defun +mu4e-lock-file-delete-maybe ()
   "Check `+mu4e-lock-file', and delete it if this process is responsible for it."
   (when (+mu4e-lock-available)
+    (require 'filenotify)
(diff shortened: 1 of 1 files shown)


hlissner (changes requested):
Add these to the top of the file instead:

```elisp
;;;###autoload (autoload 'file-notify-rm-watch "filenotify")
;;;###autoload (autoload 'file-notify-add-watch "filenotify")
```

elken:
Done, why is that preferred?

hlissner (approved):

hlissner:
> Done, why is that preferred?

It's less of a maintenance burden going forward. i.e. Rather than remembering to `(require 'filewatch)` in all future functions that need them.

-----

Anyhow, thanks for the PR!

elken:
> > Done, why is that preferred?
> 
> It's less of a maintenance burden going forward. i.e. Rather than remembering to `(require 'filewatch)` in all future functions that need them.
> 
> -----
> 
> Anyhow, thanks for the PR!

Ah yeah autoload the library as required by the function, makes sense

PR #5328, opened 2021-08-01
Fix lispy `j` collision

eugercek:
With evil mode, `j` is used for both `special-lispy-down` and `evil-escape` which is bounded to `evil-escape-key-sequence`, "jk" by default.

A demo:
https://user-images.githubusercontent.com/58051033/127784994-05af014c-b857-41c1-ad10-865f198c42c8.mp4

### DIFF
--- a/modules/editor/lispy/config.el
+++ b/modules/editor/lispy/config.el
@@ -39,16 +39,6 @@
           additional-insert))
   :config
   (lispyville-set-key-theme)
-
-  (let ((lisp-family '(lisp-mode
-                       emacs-lisp-mode
-                       ielm-mode
-                       scheme-mode
-                       racket-mode
-                       hy-mode
-                       lfe-mode
-                       dune-mode
-                       clojure-mode
-                       fennel-mode)))
-    (setq evil-escape-excluded-major-modes
-          (nconc lisp-family evil-escape-excluded-major-modes))))
+  (add-hook! 'evil-escape-inhibit-functions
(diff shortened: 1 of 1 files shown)


hlissner (changes requested):
Please use this instead.

```elisp
(add-hook! 'evil-escape-inhibit-functions
  (defun +lispy-inhibit-evil-escape-fn ()
    (and lispy-mode (evil-insert-state-p))))
```

hlissner (approved):

hlissner:
Thanks for your help!

grios-stratio:
Hi!

I rather prefer having evil-escape working even with lispy activated. Is there a way to remove this hook in my personal config? (have tried several combinations of `remove-hook` but I don't seem to make it work :sweat:

eugercek:
You can remove the added function with `remove-hook!`.
On my setup this reverts the change added in this PR.
```lisp
(remove-hook! 'evil-escape-inhibit-functions
              #'+lispy-inhibit-evil-escape-fn)
```
But you'll have the bug (https://user-images.githubusercontent.com/58051033/127784994-05af014c-b857-41c1-ad10-865f198c42c8.mp4).

Also you can change the esapce key, to `fj` as an example, with `evil-espace-sequence`
```lisp
(setq evil-escape-key-sequence "fj")
```

grios-stratio:
Thanks! I though I had tried that before without success, but I must have messed something up in the first try :/. Now it is working :)

PR #5329, opened 2021-08-02
Explicitly set xref handlers when eglot is enabled

nhamlh:
Otherwise major mode's handlers might take place. For
example as in go-mode

Fixes #0000 

### DIFF
--- a/modules/tools/lsp/+eglot.el
+++ b/modules/tools/lsp/+eglot.el
@@ -17,9 +17,11 @@
   :config
   (set-popup-rule! "^\\*eglot-help" :size 0.15 :quit t :select t)
   (set-lookup-handlers! 'eglot--managed-mode
+    :definition      #'xref-find-definitions
+    :references      #'xref-find-references
     :implementations #'eglot-find-implementation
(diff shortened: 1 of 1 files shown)


modules/tools/lsp/+eglot.el
```diff
@@ -17,9 +17,11 @@
   :config
   (set-popup-rule! "^\\*eglot-help" :size 0.15 :quit t :select t)
   (set-lookup-handlers! 'eglot--managed-mode
+    :definition      'xref-find-definitions
+    :references      'xref-find-references
```
hlissner, line 21: Please sharp quote these. e.g. `#'xref-find-definitions`, as per our conventions.

hlissner (changes requested):

modules/tools/lsp/+eglot.el
nhamlh: @hlissner thanks for reviewing. I've updated the PR.

hlissner (approved):

hlissner:
Thanks!

PR #5333, opened 2021-08-02
Mu4e hotfixes

tecosaur:
Let me know if you want something better, but the commits here should be pretty descriptive.
This fixes two issues that have been reported on Discord for mac and mu4e 1.6 users.

```
fix(mu4e): account for signature change in v1.6

Mu4e 1.6 changes the signature of `mu4e-draft-open'. Since we don't care
about the args replace funcall with apply to make sure it's happy no
matter how many args it expected.


fix(mu4e): improve mu-lock compatability with mac

On mac process args aren't defined, but comm is. Since comm also works
on Linux, we now check for that instead, accounting for a potential
capitalisation difference.


docs(mu4e): fix a typo (errors, plural)

Someone politely requested I add this to my PR on Discord ... why not.
```

### DIFF
--- a/modules/email/mu4e/README.org
+++ b/modules/email/mu4e/README.org
@@ -260,7 +260,7 @@ If you have completely lost your install then you can use this handy command!
 
 #+BEGIN_SRC  shell
 find / -type d -iname '*mu4e*'
-# I reccomend rerouting all of the error to /dev/null
+# I recommend rerouting all of the errors to /dev/null
 find / -type d -iname '*mu4e*' 2> /dev/null
 #+END_SRC
 
--- a/modules/email/mu4e/autoload/mu-lock.el
+++ b/modules/email/mu4e/autoload/mu-lock.el
@@ -22,7 +22,7 @@
                    (insert-file-contents +mu4e-lock-file)
                    (buffer-string))))
            (process (process-attributes pid)))
-      (if (and process (string-match-p "emacs" (alist-get 'args process)))
+      (if (and process (string-match-p "[Ee]macs" (alist-get 'comm process)))
           (cons pid process)
         (delete-file +mu4e-lock-file) nil))))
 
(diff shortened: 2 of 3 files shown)


modules/email/mu4e/README.org
```diff
@@ -260,7 +260,7 @@ If you have completely lost your install then you can use this handy command!
 
 #+BEGIN_SRC  shell
 find / -type d -iname '*mu4e*'
-# I reccomend rerouting all of the error to /dev/null
+# I reccomend rerouting all of the errors to /dev/null
```
hlissner, line 263: `s/reccomend/recommend`

modules/email/mu4e/autoload/mu-lock.el
```diff
@@ -22,7 +22,7 @@
                    (insert-file-contents +mu4e-lock-file)
                    (buffer-string))))
            (process (process-attributes pid)))
```
hlissner, line 24: Add `(case-fold-search nil)` to ensure case sensitivity in `string-match-p`.

hlissner (changes requested):

modules/email/mu4e/autoload/mu-lock.el
tecosaur: But we don't want it to be case sensitive...

modules/email/mu4e/README.org
tecosaur: Done.

modules/email/mu4e/autoload/mu-lock.el
hlissner: Don't you? Considering you changed the regexp to `"[Ee]macs"`.

tecosaur: I changed it to `[Ee]macs` because since `case-fold-search` isn't bound to anything, this way we're fine either way.

PR #5336, opened 2021-08-03
docs: add how to remove emacs26 in installation guide

mmajewsk:
In some cases, ubuntu may contain a preinstalled version of Emacs26, which cannot be simply removed with "sudo apt remove emacs", and needs additional "sudo apt autoremove" to properly uninstall emacs26. Only this will result with the command "emacs" opening other version than emacs26.

Ref https://unix.stackexchange.com/a/84488

### DIFF
--- a/docs/getting_started.org
+++ b/docs/getting_started.org
@@ -130,6 +130,12 @@ Or through snap:
 snap install emacs --classic
 #+END_SRC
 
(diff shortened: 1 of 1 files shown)


hlissner:
How frustrating. I'm curious what version of Ubuntu introduced that. In any case, thanks for the PR!

PR #5337, opened 2021-08-03
Improve `:lang haskell`

iyefrat:
As discussed on the discord, this PR removes `ghcide` and `dante` support from the haskell module, and rewrites the readme both to reflect this and to generally be up to date regardless.

I'll probably send more improvements down the line as well, but I think this is a good start.

cc: @fosskers


modules/lang/haskell/README.org
```diff
@@ -5,152 +5,64 @@
...
-+ [[http://learnyouahaskell.com/][Learn you a haskell for great good]]
-+ [[http://haskellbook.com/][Haskell Programming from first principles]]
-+ [[https://github.com/krispo/awesome-haskell][Awesome Haskell]]: an extensive list of haskell resources
-+ [[https://docs.haskellstack.org/en/stable/README/][The Haskell Tool Stack docs]]
- 
+Adds Haskell support to Doom Emacs
+
++ If possible, include a brief list of feature highlights here
++ Like code completion, syntax checking or available snippets
++ Include links to packages & external things where possible
```
fosskers, line 20: Leftover from the template?

fosskers (commented):
Generally looks good. Consider a `M-q` on longer lines to auto-format them.

modules/lang/haskell/README.org
iyefrat: Whoops, good catches. I always forget to `M-q` lines, I should figure out how to automate that.

modules/lang/haskell/README.org
```diff
@@ -5,152 +5,68 @@
...
-must set ~haskell-process-type~ manually:
+After installing your preferred formatter, make sure to set
+=lsp-haskell-formatting-provider= to it.
 
-#+BEGIN_SRC emacs-lisp
-(setq haskell-process-type 'cabal-new-repl)
+Make sure to configure the lsp to use your perfered formatter, e.g.:
+#+BEGIN_SRC elisp
+;; ~/.doom.d/config.el
+(setq lsp-haskell-formatting-provider "brittany")
```
jeetelongname, line 71: Should this not be wrapped in an `after! lsp-haskell` block?

iyefrat: It doesn't change anything, `lsp-haskell-formatting-provider` is defined with a `defcustom`, so it doesn't override a `setq`. Not really sure what level of best practices I'm supposed to get into in a readme snippet.

modules/lang/haskell/doctor.el
```diff
@@ -5,8 +5,7 @@
...
 
-(when (featurep! +dante)
-  (unless (executable-find "cabal")
-    (warn! "Couldn't find cabal, haskell-mode may have issues"))
-  (unless (executable-find "hlint")
-    (warn! "Couldn't find hlint. Flycheck may have issues in haskell-mode")))
+(unless (executable-find "cabal")
+  (warn! "Couldn't find cabal, haskell-mode may have issues"))
+(unless (executable-find "hlint")
+  (warn! "Couldn't find hlint. Flycheck may have issues in haskell-mode"))
```
daanturo, line 11: Shouldn't we emit this warning only when `-lsp`?
```suggestion
(unless (or (featurep! +lsp)
            (executable-find "hlint"))
  (warn! "Couldn't find hlint. Flycheck may have issues in haskell-mode. Install it or enable +lsp"))
```

iyefrat: correct me if I'm wrong, but don't you need to install hlint separately in order to get it to work with the lsp? if not, is this also the case for the formatters?

daanturo: I don't know if nix package manager's haskell-language-server has been tweaked, but it can give diagnostics and formatting capabilities without `hlint` and `brittany`.
However there's `hls-hlint-plugin`, it must serve something so please ignore my suggested change above.

iyefrat: I went and checked the code of `hls`. The plugins have `hlint`, `brittany`, etc. as dependencies, so they don't need to be installed. I checked by enabling the ormolu formatter, and it worked despite me not having ormolu installed. I'll update the doctor checks, thank you for bringing this to my attention!

iyefrat:
I know the CI is failing here due to the commit linting, but I think it has an off-by-one error, the longest line is exactly 72 characters

modules/lang/haskell/README.org
hlissner: It should be wrapped in `(after! lsp-haskell ...)` as a rule of thumb, even if it isn't strictly necessary in this case.

modules/lang/haskell/config.el
```diff
@@ -46,7 +46,4 @@
 ;;
 ;;; Backends
 
-(cond ((featurep! +dante)  (load! "+dante"))
-      ((or (featurep! +lsp)
-           (featurep! +ghcide))
-       (load! "+lsp")))
+(when (featurep! +lsp) (load! "+lsp"))
```
hlissner, line 49: Now that this has been simplified could you move the contents of +lsp.el here?

hlissner (changes requested):

modules/lang/haskell/README.org
iyefrat: Done.

modules/lang/haskell/config.el
iyefrat: Done.

hlissner (approved):

hlissner:
Looks good! Thanks again for your help!

fosskers:
Fantastic!

PR #5338, opened 2021-08-03
Bump ob-restclient pin in modules/lang/org/packages.el

betaprior:
Bumping ob-restclient to latest since the upstream implemented a useful org-babel header argument


betaprior:
bump: ob-restclient

alf/ob-restclient.el@0ebfc7c -> alf/ob-restclient.el@bfbc4d8

PR #5348, opened 2021-08-06
Fix an error on mu4e-bookmarks

seokbeomKim:
Hi, 

Adding 'Flagged messages' to mu4e-bookmarks with attributes results in mu4e loading failure. 
Change the line not to use attributes.

```
error in process filter: mu4e-error: [mu4e] Invalid bookmark in mu4e-bookmarks
error in process filter: [mu4e] Invalid bookmark in mu4e-bookmarks
```

### DIFF
--- a/modules/email/mu4e/config.el
+++ b/modules/email/mu4e/config.el
@@ -91,7 +91,7 @@
 
   (plist-put (cdr (assoc :flags mu4e-header-info)) :shortname " Flags") ; default=Flgs
   (add-to-list 'mu4e-bookmarks
-               '(:name "Flagged messages" :query "flag:flagged" :key ?f) t)
+               '("flag:flagged" "Flagged messages" ?f) t)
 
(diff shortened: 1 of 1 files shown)


tecosaur:
This doesn't seem necessary to me, what version of mu4e are you using?

BencePaul:
Changing this line (i.e. the fix in the commit) allowed me to load mu4e. Before that I kept getting invalid bookmark errors

hlissner:
Thanks for your help!

tecosaur:
@eamonno I'd still be interested to know what version of mu4e you saw this issue with, as if I look at the docstring for `mu4e-bookmarks` I see:
```
List of pre-defined queries that are shown on the main screen.

Each of the list elements is a plist with at least:
:name  - the name of the query
:query - the query expression or function
:key   - the shortcut key.
```

PR #5351, opened 2021-08-07
Make ibuffer-visit-buffer workspace-aware

### LINKED ISSUE #5061: Ibuffer should open buffers in their respective workspaces
### Describe your request

Currently ibuffer (analogue to dired for buffers) is a great way of overviewing all buffers. Buffers are grouped by workspaces which is great. Though, when opening a buffer that belongs to a different workspace, the current workspace does not auto switch to that workspace. I could not find a way to implement this myself, but I assume that it should not be difficult.

### Briefly explain its use-case

1. Open ibuffer
2. switch to a buffer
3. expect the workspace to be switched too

petr-tik:
Fixes #5061 

check if the buffer that is being visited belongs to any
existing workspace and then switch to the workspace before opening the
buffer

** 2 review items marked in code comments

Please review and suggest the best ideas and I will address and/or delete comments accordingly. 

** Performance 

I tried it locally and it worked, but I usually run <6 workspace with 10 buffers each. I am not show what kind of performance 
degradation we should expect for power-users with 10 workspace with 100 buffers in each. 

** On even if ibuffer is commented out in init.el

After testing locally, I realised that this patch works even when ibuffer customisations aren't activated as they aren't in my config
https://github.com/petr-tik/dot-doom-emacs/blob/master/init.el#L69

I understand I could move my patch from modules/ui/workspaces/config.el to the sexp here
https://github.com/hlissner/doom-emacs/blob/5b3f52f5fb98cc3af653b043d809254cebe04e6a/modules/emacs/ibuffer/config.el#L42

When I turned ibuffer on in my config, I experienced an unpleasant slowdown in ibuffer generation (i guess it's projectile running slowly on my tramp projects), so i would personally prefer to leave it where it is, but am happy to take guidance and move under ibuffer/config.el

### DIFF
--- a/modules/ui/workspaces/config.el
+++ b/modules/ui/workspaces/config.el
@@ -106,6 +106,18 @@ stored in `persp-save-dir'.")
                (setq uniquify-buffer-name-style +workspace--old-uniquify-style))
              (advice-remove #'doom-buffer-list #'+workspace-buffer-list)))))
 
+  (defun +workspaces/switch-to-buffer (buf)
+    "Switch to buffer's workspace"
+    (when (bound-and-true-p persp-mode)
+      (if-let* ((workspace-name (cl-find-if (lambda (worksp-name)
+                                               (+workspace-contains-buffer-p buf (persp-get-by-name worksp-name)))
+                                             (+workspace-list-names))))
+          (+workspace-switch workspace-name)
+        ;; REVIEW do we need this user-facing message or is defaulting to this
+        ;; behaviour enough?
+        ;; should we maybe create a new workspace?
+        (message "%s doesn't belong to any workspace" buf))))
+
   ;; Per-workspace `winner-mode' history
   (add-to-list 'window-persistent-parameters '(winner-ring . t))
 
--- a/modules/emacs/ibuffer/config.el
+++ b/modules/emacs/ibuffer/config.el
@@ -54,7 +54,17 @@
     (defun +ibuffer/open-for-current-workspace ()
       "Open an ibuffer window for the current workspace"
       (interactive)
-      (+ibuffer-workspace (+workspace-current-name))))
+      (+ibuffer-workspace (+workspace-current-name)))
+
+    (defun +ibuffer/visit-buffer-in-its-workspace ()
+      "Visit buffer, but switch to its workspace if it exists."
+      (interactive)
+      (if-let* ((buf (ibuffer-current-buffer t)))
+          (progn
+            (+workspaces/switch-to-buffer buf)
+            (ibuffer-visit-buffer))))
+
+    (define-key ibuffer-mode-map [remap ibuffer-visit-buffer] #'+ibuffer/visit-buffer-in-its-workspace))
 
   (when (featurep! :completion ivy)
     (defadvice! +ibuffer-use-counsel-maybe-a (_file &optional _wildcards)


modules/ui/workspaces/config.el
```diff
@@ -112,6 +112,22 @@ stored in `persp-save-dir'.")
                (setq uniquify-buffer-name-style +workspace--old-uniquify-style))
              (advice-remove #'doom-buffer-list #'+workspace-buffer-list)))))
 
+  ;; REVIEW can pass the universal arg to this function and then if activated
+  ;; call with the selected buffer
+  ;; `persp-temporarily-display-buffer'
+  (defun +workspace-ibuffer-switch-to-buffers-workspace (&rest _unused)
+    "Switch to the workspace, which the current ibuffer candidate belongs to."
+    (-if-let* ((buf (ibuffer-current-buffer))
+               (workspace-name (-first (lambda (worksp-name)
```
hlissner, line 121: Doom does not use dash.el. Replace `-first` with `cl-find-if`.

modules/ui/workspaces/config.el
```diff
@@ -112,6 +112,22 @@ stored in `persp-save-dir'.")
...
+                                         (+workspace-contains-buffer-p buf (persp-get-by-name worksp-name)))
+                                       (+workspace-list-names))))
+        (+workspace-switch workspace-name)
+      ;; REVIEW do we need this user-facing message or is defaulting to this
+      ;; behaviour enough?
+      (message "%s doesn't belong to any workspace" buf)))
```
hlissner, line 127: This body should be wrapped in `(when (bound-and-true-p persp-mode) ...)` to guard against cases where it has been temporarily (or otherwise) disabled.

modules/ui/workspaces/config.el
```diff
@@ -112,6 +112,22 @@ stored in `persp-save-dir'.")
...
+
+  (advice-add #'ibuffer-visit-buffer :before #'+workspace-ibuffer-switch-to-buffers-workspace)
```
hlissner, line 129: I do agree there needs to be a "switch-to-buffer's-workspace" command, but I don't think this is the correct approach (I.e. it shouldn't be the default behavior of ibuffer -- resorting to advice should be a last resort).

Instead, I suggest:

1. A separate command is written, e.g. `+ibuffer/visit-buffer-in-workspace` (defined in the `:emacs ibuffer` module). There is a need for a generalized `+workspaces/switch-to-buffer` command too, so I'd appreciate if you wrote one (in the `:ui workspaces` module) and used it in `+ibuffer/visit-buffer-in-workspace`.

2. Either remap `ibuffer-visit-buffer` in `ibuffer-mode-map` (I.e. `(define-key ibuffer-mode-map [remap ibuffer-visit-buffer] #'+ibuffer/visit-buffer-in-workspace)`) _or_ bind it to another key. E.g. `C-RET`. What do you think?

hlissner (changes requested):
Sorry for the late response. There are some issues here that prevent me from accepting this PR. Besides the points I've outlined below, please reword your commit message to conform to [our git conventions](https://gist.github.com/hlissner/4d78e396acb897d9b2d8be07a103a854):

```
feat(ibuffer): add workspace-aware visit-buffer command

Check if the buffer that is being visited belongs to any existing
workspace and then switch to the workspace before opening the buffer

Fix: #5061
```

(The reworded commit reflects the changes I've requested)

> When I turned ibuffer on in my config, I experienced an unpleasant slowdown in ibuffer generation (i guess it's projectile running slowly on my tramp projects), so i would personally prefer to leave it where it is, but am happy to take guidance and move under ibuffer/config.el

I'd like to address this. A simple workaround may be to disable `ibuffer-projectile`. e.g.

```elisp
;; add to $DOOMDIR/packages.el
(package! ibuffer-projectile :disable t)
```

Then run `doom sync -u`, restart Emacs, and see if it resolves the issue.

petr-tik:
thanks for the review. Appreciate you must have been busy, i will try to address it this week and ping you again

modules/ui/workspaces/config.el
petr-tik: > What do you think?

I would like to survey and follow the least surprising behaviour/UX for the users. I will implement the functions first and then we can discuss what the defaults should be.

modules/ui/workspaces/config.el
petr-tik: thanks for spotting. Addressed

modules/ui/workspaces/config.el
petr-tik: thanks for helping me out with this. Replaced

modules/ui/workspaces/config.el
petr-tik: I decided to rename the function to `+ibuffer/visit-buffer-in-its-workspace`. What are your thoughts on that? 

Also, what about the case of a non-workspace buffer? Should we maybe create a new workspace or use `persp-temporarily-display-buffer` to pull the buffer into the current workspace? 

I decided to default to remaping `ibuffer-visit-buffer`, since I thought people who have opted into both workspaces and ibuffer would expect/accept a default action that leverages both simultaneously. Happy to change the default if you feel it's too intrusive

modules/ui/workspaces/config.el
iyefrat: `-if-let*` is from dash too. Can you replace it with `if-let*`?

modules/ui/workspaces/config.el
iyefrat: Replace `-if-let*` with `if-let` here as well.

iyefrat:
boop

petr-tik:
are you booping me or Henrik?

iyefrat:
Oh I just wanted to make sure that the PR first get closed cause it got marked as stale. It's still relevant, Henrik has just been a bit busy

hlissner:
Sorry for taking so long to get to this. After some deliberation, I've added this functionality in c309e61, with four differences to what's proposed here: 

1. If a buffer is present in more than one workspace, prompt the user to select which workspace they want to end up in.
2. If the buffer is present in no workspaces, add it to the current one.
3. Isolate the functionality to the ibuffer module. I'm working on generalizing this functionality in the workspaces module.
4. I've named the function `+ibuffer/visit-workspace-buffer` to match `+vertico/switch-workspace-buffer`, which does similar. I'm not 100% on this naming, but that's a yak I'll shave later (as part of a bigger effort to rename a bunch of poorly named functions/variables in Doom's code base).

In any case, this should resolve this and #5061. Let me know if I've missed anything!

PR #5352, opened 2021-08-07
Fix typo in emacs/undo README.md

Doesntmeananything:
This is a small fix for a typo found in `emacs/undo` readme.


hlissner:
Thanks for the PR, but I must refuse it. There is nothing wrong with your PR, but it comes at a bad time: I'm about a week away from pushing a rework of all our documentation, which includes a rework of all our module documentation, including `:emacs undo`'s. Since this indirectly fixes this issue (its hacks section has been rewritten), and it would swallow up your PR anyway, I'll simply wait for that update instead. In any case, thanks for your help!

PR #5353, opened 2021-08-07
Improve `:completion vertico`

iyefrat:
- Fix a minor issue with the new prefix-help-command settings
- add more thorough explanations to the readme
- Bump the module, and adapt to upstream changes
- update `all-the-icons-completion`'s recipe as it got accepted to melpa
- highlight vertico candidate on embark-act
- explain in the readme how the consult async regexp checks work, and add doctor checks for pcre support in grep and ripgrep
- fix `+vertico/find-file-in`not creating new files
- fix `consult-notmuch` bug (#5369)
- remap `xref-find-apropos` to `consult-lsp-symbols`
- unify error listing under `+default/diagnostics` and use `consult-lsp-diagnostics` if applicable
  (sidenote - should the which key listing still be `List Errors`? `List Diagnostics` is more correct, but less obvious in meanings also might require the change of the `:er[ror]` ex command)
- try basic completion style first when completing for remotes (adapts #5313)
- explain symbol annotations in readme
- feature gate embark action that opens file in new workspace
- add magit status embark action
- some whitespace fixes
- `+vertico/jump-list` fix
- embark package target finder and keymap fixes
- file path completion fixes and docs improvements
- consult-eglot support

Note that the second commit closes #5291, since it was deemed to probably be best solved in a readme explanation. This might also be a good time to decide if you want to keep `C-M-j` and `C-M-k` around as move+preview keys, since we're on the subject (If we end up keeping them around, it might be a good idea to add `C-M-n` and `C-M-p` as well. I think I had a reason for not doing this but I don't remember what it was).

### DIFF
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -105,7 +105,7 @@ overrides `completion-styles' during company completion sessions.")
             (format "%s --color=never -i -H -E .git --regex %s"
                     doom-projectile-fd-binary
                     (if IS-WINDOWS "--path-separator=/" ""))
-        consult-find-args))
+          consult-find-args))
 
   (consult-customize
    consult-ripgrep consult-git-grep consult-grep
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -53,11 +53,17 @@ overrides `completion-styles' during company completion sessions.")
      ;; Flex matching
      ((string-prefix-p "~" pattern) `(orderless-flex . ,(substring pattern 1)))
      ((string-suffix-p "~" pattern) `(orderless-flex . ,(substring pattern 0 -1)))))
+  (add-to-list
+   'completion-styles-alist
+   '(+vertico-basic-remote
+     +vertico-basic-remote-try-completion
+     +vertico-basic-remote-all-completions
+     "Use basic completion on remote files only"))
   (setq completion-styles '(orderless)
         completion-category-defaults nil
         ;; note that despite override in the name orderless can still be used in
         ;; find-file etc.
-        completion-category-overrides '((file (styles . (orderless partial-completion))))
+        completion-category-overrides '((file (styles +vertico-basic-remote orderless partial-completion)))
         orderless-style-dispatchers '(+vertico-orderless-dispatch)
         orderless-component-separator "[ &]")
   ;; ...otherwise find-file gets different highlighting than other commands
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -41,6 +41,9 @@ overrides `completion-styles' during company completion sessions.")
      ((string= "!" pattern) `(orderless-literal . ""))
      ;; Without literal
      ((string-prefix-p "!" pattern) `(orderless-without-literal . ,(substring pattern 1)))
+     ;; Character folding
+     ((string-prefix-p "%" pattern) `(char-fold-to-regexp . ,(substring pattern 1)))
+     ((string-suffix-p "%" pattern) `(char-fold-to-regexp . ,(substring pattern 0 -1)))
      ;; Initialism matching
      ((string-prefix-p "`" pattern) `(orderless-initialism . ,(substring pattern 1)))
      ((string-suffix-p "`" pattern) `(orderless-initialism . ,(substring pattern 0 -1)))
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -169,8 +169,9 @@ overrides `completion-styles' during company completion sessions.")
         (nthcdr pos embark-target-finders)))
   (setq embark-package-map (make-sparse-keymap))
   (map! (:map embark-file-map
-         :desc "Open target with sudo" "s"   #'doom/sudo-find-file
-         :desc "Open in new workspace" "TAB" #'+vertico/embark-open-in-new-workspace)
+         :desc "Open target with sudo" "s" #'doom/sudo-find-file
+         (:when (featurep! :ui workspaces)
+          :desc "Open in new workspace" "TAB" #'+vertico/embark-open-in-new-workspace))
         (:map embark-package-map
          "h" #'doom/help-packages
          "b" #'doom/bump-package
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -149,7 +149,7 @@ overrides `completion-styles' during company completion sessions.")
          :desc "Actions" "a" #'embark-act)) ; to be moved to :config default if accepted
   :config
   (set-popup-rule! "^\\*Embark Export Grep" :size 0.35 :ttl 0 :quit nil)
-  (cl-nsubstitute #'+vertico/embark-which-key-indicator #'embark-mixed-indicator embark-indicators)
+  (cl-nsubstitute #'+vertico-embark-which-key-indicator #'embark-mixed-indicator embark-indicators)
   (add-to-list 'embark-indicators #'+vertico-embark-vertico-indicator)
   ;; add the package! target finder before the file target finder,
   ;; so we don't get a false positive match.
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -91,11 +91,12 @@ overrides `completion-styles' during company completion sessions.")
         consult-async-input-throttle 0.2
         consult-async-input-debounce 0.1)
 
-  (when doom-projectile-fd-binary
-    (setq consult-find-command
-          (format "%s -i -H -E .git --regex %s ARG OPTS"
-                  doom-projectile-fd-binary
-                  (if IS-WINDOWS "--path-separator=/" ""))))
+  (setq +vertico-consult-fd-args
+        (if doom-projectile-fd-binary
+            (format "%s --color=never -i -H -E .git --regex %s"
+                    doom-projectile-fd-binary
+                    (if IS-WINDOWS "--path-separator=/" ""))
+        consult-find-args))
 
   (consult-customize
    consult-ripgrep consult-git-grep consult-grep
@@ -145,8 +146,7 @@ overrides `completion-styles' during company completion sessions.")
          :desc "Actions" "a" #'embark-act)) ; to be moved to :config default if accepted
   :config
   (set-popup-rule! "^\\*Embark Export Grep" :size 0.35 :ttl 0 :quit nil)
-
-  (setq embark-indicator #'+vertico/embark-which-key-indicator)
+  (cl-nsubstitute #'+vertico/embark-which-key-indicator #'embark-mixed-indicator embark-indicators)
   ;; add the package! target finder before the file target finder,
   ;; so we don't get a false positive match.
   (let ((pos (or (cl-position
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -167,18 +167,19 @@ overrides `completion-styles' during company completion sessions.")
         cons
         '+vertico-embark-target-package-fn
         (nthcdr pos embark-target-finders)))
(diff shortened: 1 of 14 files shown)


modules/completion/vertico/config.el
```diff
@@ -145,7 +140,7 @@ overrides `completion-styles' during company completion sessions.")
   :config
   (set-popup-rule! "^\\*Embark Export Grep" :size 0.35 :ttl 0 :quit nil)
 
-  (setq embark-indicator #'+vertico/embark-which-key-indicator)
+  (setq embark-indicators '(+vertico/embark-which-key-indicator embark-high))
```
minad, line 143: Typo? Here you may want to remove embark-verbose-indicator and add +vertico/embark-which-key-indicator instead. We may add more useful indicators in the future, see embark-highlight-indicator and embark-isearch-highlight-indicator.

minad:
> Note that in it's current state this PR removes the use of fd in consult-find commands. 

The best way to get a `consult-fd` command now is to write your own, since `consult-find` became too specialized/tied to POSIX/GNU find. See https://github.com/minad/consult/wiki#find-files-using-fd.

modules/completion/vertico/config.el
iyefrat: Oh whoops, that's supposed to be `embark-highlight-indicator`. Not sure what you mean about `+vertico/embark-which-key-indicator` though, this code does set it doesn't it? Just the highlight indicator as well.

minad: I meant that you may not want to overwrite the list, just replace mixed->whichkey.

minad: And maybe even add https://github.com/oantolin/embark/wiki/Additional-Configuration#colorize-the-current-vertico-candidate-differently-when-acting?

iyefrat: > I meant that you may not want to overwrite the list, just replace mixed->whichkey.

Ah, good point.

> And maybe even add https://github.com/oantolin/embark/wiki/Additional-Configuration#colorize-the-current-vertico-candidate-differently-when-acting?

Yeah I was planning on that, but wanted to get the basics sorted first

minad:
@iyefrat I went with a hybrid approach now for the orderless matching.

* For consult-grep with pcre and consult-git-grep you can use as many regular expressions as you want, and the regular expressions can match in any order.
* For consult-ripgrep without pcre I ony enable unordered matching for up to three regular expressions. For more regular expressions, the unordered matching disenganges. Overall this feels like a good compromise.

See https://github.com/minad/consult/commit/f96718419d2ab54b447c0cb0def2d951a8e9e72b.

iyefrat:
@minad Thanks for the update. I'll update the bump to include this after a bit of testing, worst case it'll get in in the next one. This means that now the main addition of using the orderless regexp compiler is the use of style dispatchers, yeah?

minad:
@iyefrat You probably saw this, but there have been quite a few critical bugs in the Consult 0.10 version due to my grep async refactorings. I just tagged Consult 0.11, so you may want to update the pinning.

zauster:
Hey, I tested your PR (based on Doom bf8495b) and i get an error: 
`
Failed to load magit package incrementally, because: (void-variable embark-indicators)
`
I see that a variable `embark-indicator` (without a trailing s) exists, maybe a typo?

modules/completion/vertico/doctor.el
```diff
@@ -0,0 +1,11 @@
+;;; completion/vertico/doctor.el -*- lexical-binding: t; -*-
+
+(autoload #'consult--grep-lookahead-p "consult")
```
hlissner, line 3: Load the package directly Instead of autoloading this call. e.g. `(require 'consult)`

modules/completion/vertico/packages.el
```diff
@@ -4,22 +4,20 @@
...
 (when (featurep! :checkers syntax)
   (package! consult-flycheck :pin "92b259e6a8ebe6439f67d3d7ffa44b7e64b76478"))
 
-(package! embark :pin "1a7e6b556142216fa5f9b897bd5eca73968f3c49")
-(package! embark-consult :pin "1a7e6b556142216fa5f9b897bd5eca73968f3c49")
+(package! embark :pin "19145d5a3367038f0a9a114b7387c8a896037aec")
+(package! embark-consult :pin "19145d5a3367038f0a9a114b7387c8a896037aec")
 
-(package! marginalia :pin "11235445365c6ab119acabe91828e9182097ece7")
+(package! marginalia :pin "c6ca58bea819c3b5e4b3295ad693c5cd0ae5db31")
```
hlissner, line 18: Please bump these one more time; there are a couple more updates waiting for them.

modules/config/default/+evil-bindings.el
```diff
@@ -398,7 +398,10 @@
        (:when (featurep! :tools lsp +eglot)
         :desc "LSP Execute code action" "a" #'eglot-code-actions
         :desc "LSP Rename" "r" #'eglot-rename
-        :desc "LSP Find declaration" "j" #'eglot-find-declaration)
+        (:when (featurep! :completion vertico)
+         :desc "Jump to symbol in current workspace" "j"   #'consult-eglot-symbols)
+        (:when (not (featurep! :completion vertico))
+         :desc "LSP Find declaration"                "j"   #'eglot-find-declaration))
```
hlissner, line 404: Instead of the `(:when TRUE ...) (:when FALSE ...)` approach, use:

```lisp
        :desc "LSP Find declaration"                "j"   #'eglot-find-declaration
        (:when (featurep! :completion vertico)
         :desc "Jump to symbol in current workspace" "j"   #'consult-eglot-symbols))
```

The latter will overwrite the former when applicable.

modules/config/default/autoload/default.el
```diff
@@ -44,3 +44,16 @@ If ARG (universal argument), runs `compile' from the current directory."
...
     (sleep-for 1))
   (server-start))
+
+;;;###autoload
+(defun +default/diagnostics (&rest arg)
+  "List diagnostics for the current buffer/project.
+If the the vertico and lsp modules are active, list lsp diagnostics for the
+current project. Otherwise list them for the current buffer"
+  (interactive)
+  (cond ((and (featurep! :completion vertico) (featurep! :tools lsp))
```
hlissner, line 54: Should this check for `lsp-mode` (or `eglot-managed-mode`)?

modules/config/default/autoload/default.el
```diff
@@ -44,3 +44,16 @@ If ARG (universal argument), runs `compile' from the current directory."
...
+         (consult-lsp-diagnostics arg))
+        ((featurep! :checkers syntax)
```
hlissner, line 56: This should check for `flycheck-mode`.

modules/config/default/autoload/default.el
```diff
@@ -44,3 +44,16 @@ If ARG (universal argument), runs `compile' from the current directory."
...
+         (flycheck-list-errors))
+        (t
```
hlissner, line 58: And this should check for `flymake-mode`.

modules/tools/lsp/+eglot.el
```diff
@@ -48,3 +48,9 @@ server getting expensively restarted when reverting buffers."
                           (+lsp-optimization-mode -1))))
                 server)))
       (funcall fn server))))
+
```
hlissner, line 51: Two blank lines between top-level `use-package!` blocks, please.

modules/tools/lsp/packages.el
```diff
@@ -2,7 +2,10 @@
 ;;; tools/lsp/packages.el
 
 (if (featurep! +eglot)
-    (package! eglot :pin "c17bdf6c98d6bf0f1a85f1175556e1038654402f")
+    (progn
+      (package! eglot :pin "c17bdf6c98d6bf0f1a85f1175556e1038654402f")
```
hlissner, line 6: eglot was bumped in e03f1d19d. Could you rebase your PR onto `develop`?

modules/config/default/+emacs-bindings.el
```diff
@@ -66,7 +64,10 @@
        (:when (featurep! :tools lsp +eglot)
         :desc "LSP Execute code action"              "a" #'eglot-code-actions
         :desc "LSP Rename"                           "r" #'eglot-rename
-        :desc "LSP Find declaration"                 "j" #'eglot-find-declaration))
+        (:when (featurep! :completion vertico)
+         :desc "Jump to symbol in current workspace" "j" #'consult-eglot-symbols)
+        (:when (not (featurep! :completion vertico))
+         :desc "LSP Find declaration"                "j" #'eglot-find-declaration)))
```
hlissner, line 70: See my review for +evil-bindings.el below.

modules/completion/vertico/config.el
```diff
@@ -92,11 +100,12 @@ overrides `completion-styles' during company completion sessions.")
         consult-async-input-throttle 0.2
         consult-async-input-debounce 0.1)
 
-  (when doom-projectile-fd-binary
-    (setq consult-find-command
-          (format "%s -i -H -E .git --regex %s ARG OPTS"
-                  doom-projectile-fd-binary
-                  (if IS-WINDOWS "--path-separator=/" ""))))
+  (setq +vertico-consult-fd-args
```
hlissner, line 103: No `defvar` exists for `+vertico-consult-fd-args`, but should (at the top of config.el).

hlissner (changes requested):

modules/completion/vertico/doctor.el
iyefrat: Done.

modules/completion/vertico/packages.el
iyefrat: Done (from looking over the git commits, there doesn't seem to be any manual intervention required, and no obvious bugs so far).

modules/config/default/+evil-bindings.el
iyefrat: Done.

modules/config/default/autoload/default.el
iyefrat: Yeah you're right, I'll add a `(bound-and-true-p lsp-mode)`.

modules/config/default/autoload/default.el
iyefrat: Done.

modules/config/default/autoload/default.el
iyefrat: Done. Should there be an error message or something in the base case of having nothing on?

modules/tools/lsp/+eglot.el
iyefrat: Done.

modules/tools/lsp/packages.el
iyefrat: I already rebased, this pin is the updated one :D

modules/completion/vertico/config.el
iyefrat: Should I move the `setq` logic to the `defvar` or initialize it as just the string of`doom-projectile-fd-binary` or something?

modules/config/default/autoload/default.el
hlissner: Ideally, yes. Perhaps a `(user-error "No diagnostics backend is enabled; e.g. flycheck or flymake")`

modules/completion/vertico/config.el
hlissner: Initialize it as `nil` at top-level, keep the `setq`, but wrap it in a check. `+vertico-consult-fd-args` shouldn't be changed if its value isn't `nil`.

modules/config/default/autoload/default.el
iyefrat: Done.

hlissner (approved):

hlissner:
Fantastic work! Thanks again for your help!

PR #5355, opened 2021-08-08
fix(lsp): support +pyright with eglot

ErnestDong:
eglot does not work out-of-the-box with pyright, so I add the five lines


hlissner (approved):

hlissner:

ErnestDong:
Take care of yourself. Hope disease will be far away from you

PR #5357, opened 2021-08-08
Add clojure gitignore template

wandersoncferreira:
Adds "clojure" to the `yas-choose-value` list when creating a new `.gitignore`.

The values were taken from leiningen default template: https://github.com/technomancy/leiningen/blob/master/resources/leiningen/new/template/gitignore and included a `.cpcache` for projects using clojure `deps.edn`.


hlissner:

PR #5365, opened 2021-08-11
Remove +org-init-roam-maybe-h function

jstranik:
The hook is not needed anymore. Org-raom-mode is a major mode now, which means
that it cannot be invoked with the +1 -1 parameters.

The hook is also not required because the mode is used in org-roam-buffer
buffer.

Fixes #0000 

### DIFF
--- a/modules/lang/org/contrib/roam.el
+++ b/modules/lang/org/contrib/roam.el
@@ -9,7 +9,6 @@
 ;;; Packages
 
 (use-package! org-roam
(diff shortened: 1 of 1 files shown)


hlissner:
You're modifying the `+roam` feature, which is for roam v1. We have a `+roam2`, which targets v2. Considering v1 is end-of-life, I assume you must be describing changes that occurred in v2, in which case this PR is incorrectly imposing v2 changes on our v1 configuration, so I'll have to refuse it. Thanks for the PR in any case.

PR #5376, opened 2021-08-13
Correct a typo in a person's name in the ligature module readme

kenranunderscore:
Throughout the readme in the `ligatures` module there were two different spellings of the name `Mitsuharu Yamamoto` who maintains [emacs-mac](https://bitbucket.org/mituharu/emacs-mac).


hlissner:
Thanks for the PR, but I must refuse it. There is nothing wrong with your PR, but it comes at a bad time: I'm about a week away from pushing a rework of all our documentation, which also reworks documentation for each module, including this one (which includes the correct spelling of Mitsuharu's name). As this indirectly fixes this issue and would swallow up your PR anyway, I'll simply wait for that update instead. In any case, thanks for your help!

PR #5377, opened 2021-08-13
fix(multiple-cursors): keybind conflict with embark

### LINKED ISSUE #5374: Keybinding conflict between iedit and embark

After switching from ivy to vertico, there should be no conflicts or errors in Doom.

### What actually happened?

The first time I open a file in a new Emacs instance (with `SPC f p` or `SPC h P`, for example, but *not* with `find-file`), I get a `*Warnings*` popup with the following message:

```
Warning (iedit): Iedit default key "C-;" is occupied by embark-act.
```

I'm guessing this message appears when something loads iedit for the first time, and that there are probably many ways to trigger it.

_No response_

### Steps to reproduce

1. Open Emacs
2. Open a file with `SPC f p` or `SPC h P`

You can see my doom directory [here](https://github.com/ajgrf/dotfiles/tree/master/dot_config/doom).

### System Information

https://pastebin.com/ZPfpNTpY

ajgrf:
Fixes #5374 

Disabled iedit's default keybinding because it conflicts with embark in the new `:completion vertico` module, and iedit would pop up an annoying warning buffer about it. Since iedit is only used in Doom as a dependency of evil-multiedit, this should only negatively impact users who use iedit directly without vim keybindings (probably a tiny number of people).

### DIFF
--- a/modules/editor/multiple-cursors/config.el
+++ b/modules/editor/multiple-cursors/config.el
@@ -19,6 +19,14 @@
         "C-p" #'evil-multiedit-prev))
 
 
+(use-package! iedit
+  :when (featurep! :completion vertico)
+  :defer t
+  :init
+  ;; Fix conflict with embark.
+  (setq iedit-toggle-key-default nil))
+
+
 (use-package! evil-mc
   :when (featurep! :editor evil)
   :commands (evil-mc-make-cursor-here


otavioschwanck (approved):

hlissner (approved):

hlissner:
Sorry for the late response. Thanks for the fix!

PR #5389, opened 2021-08-16
Holy hercules! more portal quit lines

tecosaur:
I'm sure Cave Johnson would approve.

> The lab boys just informed me that I should not have mentioned the control group. They're telling me I oughtta stop making these pre-recorded messages. That gave me an idea: make more pre-recorded messages. I pay the bills here, I can talk about the control group all damn day.


hlissner (approved):

hlissner:
Cave John approves. Thanks!

PR #5394, opened 2021-08-18
docs(dired): update homepage of `all-the-icons-dired`

nagy:
According to https://melpa.org/#/all-the-icons-dired


hlissner:
Thanks for the PR, but I must refuse it. There is nothing wrong with your PR, but it comes at a bad time: I'm about a week away from pushing a rework of all our documentation, which changes how we link packages from module documentation. Since this would indirectly fix this issue, and would swallow up your PR anyway, I'll simply wait for that update instead. In any case, thanks for your help!

PR #5398, opened 2021-08-19
Org roam 2

niyumard:
Added org roam 2 as a module to init.el


hlissner:
Thanks for the PR, but I must refuse it. There is nothing wrong with your PR, but it comes at a bad time: I'm about a week away from pushing a rework of all our documentation, which completely changes docs/modules.org (its contents will be auto-generated from that point forward). Since this would indirectly fix this issue, and would swallow up your PR anyway, I'll simply wait for that update instead. In any case, thanks for your help!

Issue #4151, opened 2020-10-25
Tree sitter for syntax highlighting


systemctl603:
**Describe the feature**

From the site:
- Faster, fine-grained code highlighting.
- More flexible code folding.
- Structural editing (like Paredit, or even better) for non-Lisp code.
- More informative indexing for imenu.

- Link: https://github.com/ubolonton/emacs-tree-sitter

**System information**
<details><pre>
```
SYSTEM  type       gnu/linux
        config     x86_64-pc-linux-gnu
        shell      /usr/bin/fish
        uname      Linux 4.19.128-microsoft-standard #1 SMP Tue Jun 23 12:58:10 UTC 2020 x86_64
        path       (~/.local/bin /usr/local/bin /usr/local/sbin /usr/sbin /usr/bin /sbin /bin /usr/games /usr/local/games /usr/lib/emacs/27.1/x86_64-linux-gnu)
EMACS   dir        ~/.emacs.d/
        version    27.1
        build      Sep 19, 2020
[...]

Issue #197, opened 2017-09-09
[Question] How to change doom-theme background color?


babygau:
I see `doom-one` is used as default theme for `emacs` but I'd love just different background color. How could I achieve that? 

```lisp
(custom-set-faces
  '(default ((t (:background "#1d2021")))))
```
The above snippet only work within the buffer, not neotree or modeline....

hlissner:
Doom uses [solaire-mode](https://github.com/hlissner/emacs-solaire-mode) to help differentiate file-visiting buffers from the rest of Emacs (giving them a different background for the buffer, mode-line faces and more).

If you'd like to keep `solaire-mode`, you'll need to modify `solaire-default-face` too (and any other `solaire-*-face` face you want to modify). 

Otherwise, you may disable `solaire-mode` completely by adding this to your private module:

```emacs-lisp
;; private/$USER/init.el
(def-package-hook! solaire-mode :disable)

;; private/$USER/packages.el
(package! solaire-mode :ignore t)
```

babygau:
Thank you @hlissner, the prob solved, I could use `gruvbox theme` for now 👍

sophana:
commands for disabling solaire-mode don't work.
I've changed default and solaire-default-face, but there is still the currentline under the cursor which is not changed. Do you know which font to change?

mreishus:
This worked for me
```
(custom-set-faces
  '(hl-line ((t (:background "#000000")))))
```

PR #5401, opened 2021-08-20
Add `:tools tree-sitter` module

jeetelongname:
This PR adds tree sitter support to doom. 
currently it adds:
- syntax highlighting to the languages that support it.
- structural text objects using [evil-textobj-treesitter](https://github.com/meain/evil-textobj-treesitter)

With plans to add: 
- better code folding using tree sitter [(and powered with hideshow!)](https://github.com/jeetelongname/hideshow-tree-sitter)
- a better tree visualization mode (similar to the [playground on tree sitters website](https://tree-sitter.github.io/tree-sitter/playground) or `TSplayground` in neovim (package needs to be found or written) 
- functions and query's that leverage tree sitter for statistics and other useful purposes. 
- any other package that is deemed useful! 

more information on module development is available on [this discourse post](https://discourse.doomemacs.org/t/treesitter-module-developement/1589)
PR's and patches are welcome as well as pointers and suggestions! feel free to ping me on discord if you have any questions.

 related issues:
 - #4151

### DIFF
--- a/docs/modules.org
+++ b/docs/modules.org
@@ -195,7 +195,7 @@ Small modules that give Emacs access to external tools & services.
 + [[file:../modules/tools/taskrunner/README.org][taskrunner]] - TODO
 + [[file:../modules/tools/terraform/README.org][terraform]] - TODO
 + tmux - TODO
-+ [[file:../modules/tools/tree-sitter/README.org][tree-sitter]]  - TODO
++ [[file:../modules/tools/tree-sitter/README.org][tree-sitter]]  - syntax and parsing, sitting in a tree...
 + [[file:../modules/tools/upload/README.org][upload]] - TODO
 
 * :ui
--- a/modules/editor/fold/autoload/fold.el
+++ b/modules/editor/fold/autoload/fold.el
@@ -169,11 +169,9 @@ Targets `vimmish-fold', `hideshow', `ts-fold' and `outline' folds."
                                        (eq (overlay-get ov 'creator) 'ts-fold))
                                      ;; `overlays-in' does not provide a list that is sorted
                                      ;; (in the way we need it atleast) so we need to sort it based on direction
-                                     (cl-sort (eval `(overlays-in ,@arg-list))
-                                              (lambda (ov1 ov2)
-                                                (funcall comp-fun (overlay-start ov1) (overlay-start ov2)))))))
-                          (if (and ovs (<= count (length ovs)))
-                                   (goto-char (overlay-start (nth (- (abs count) 1) ovs))))))))
+                                     (cl-sort (apply #'overlays-in arg-list) comp-fun :key #'overlay-start))))
+                          (if (and ovs (<= (abs count) (length ovs)))
+                              (goto-char (overlay-start (nth (- (abs count) 1) ovs))))))))
            if (save-excursion (funcall fn))
            collect it into points
            finally do
--- a/modules/editor/fold/autoload/fold.el
+++ b/modules/editor/fold/autoload/fold.el
@@ -154,7 +154,25 @@ Targets `vimmish-fold', `hideshow', `ts-fold' and `outline' folds."
                         (if (> count 0)
                             (evil-vimish-fold/next-fold count)
                           (evil-vimish-fold/previous-fold (- count))))
-                      (if (/= (point) orig-pt) (point))))
+                      (if (/= (point) orig-pt) (point)))
+                    (lambda ()
+                      ;; ts-fold does not define movement functions so we need to do it ourselves
+                      (when (+fold--ts-fold-p)
+                        (let* ((arg-list (if (> count 0) ;; depending on direction we need to change the ranges
+                                             (list (point) (point-max))
+                                           (list (point-min) (point))))
+                               (comp-fun (if (> count 0) ;; also depending on direction we need to change how we sort the list
+                                             #'<
+                                           #'>))
+                               (ovs (seq-filter
+                                     (lambda (ov)
+                                       (eq (overlay-get ov 'creator) 'ts-fold))
+                                     ;; `overlays-in' does not provide a list that is sorted
+                                     ;; (in the way we need it atleast) so we need to sort it based on direction
+                                     (cl-sort (eval `(overlays-in ,@arg-list))
+                                              (lambda (ov1 ov2)
+                                                (funcall comp-fun (overlay-start ov1) (overlay-start ov2)))))))
+                          (if ovs (goto-char (overlay-start (nth (- (abs count) 1) ovs))))))))
            if (save-excursion (funcall fn))
            collect it into points
            finally do
--- a/modules/editor/fold/autoload/fold.el
+++ b/modules/editor/fold/autoload/fold.el
@@ -109,18 +109,20 @@ Targets `vimmish-fold', `hideshow', `ts-fold' and `outline' folds."
   "Open folds at LEVEL (or all folds if LEVEL is nil)."
   (interactive
    (list (if current-prefix-arg (prefix-numeric-value current-prefix-arg))))
-  (when (featurep 'vimish-fold)
-    (vimish-fold-unfold-all))
-  (save-excursion
-    (+fold--ensure-hideshow-mode)
-    (if (integerp level)
-        (progn
-          (outline-hide-sublevels (max 1 (1- level)))
-          (hs-life-goes-on
-           (hs-hide-level-recursive (1- level) (point-min) (point-max))))
-      (hs-show-all)
-      (when (fboundp 'outline-show-all)
-        (outline-show-all)))))
+  (cond ((+fold--ts-fold-p)
+         (ts-fold-open-all))
+        ((featurep 'vimish-fold)
+         (vimish-fold-unfold-all))
+        (t (save-excursion
+             (+fold--ensure-hideshow-mode)
+             (if (integerp level)
+                 (progn
+                   (outline-hide-sublevels (max 1 (1- level)))
+                   (hs-life-goes-on
+                    (hs-hide-level-recursive (1- level) (point-min) (point-max))))
+               (hs-show-all)
+               (when (fboundp 'outline-show-all)
+                 (outline-show-all)))))))
 
 ;;;###autoload
 (defun +fold/close-all (&optional level)
@@ -128,13 +130,15 @@ Targets `vimmish-fold', `hideshow', `ts-fold' and `outline' folds."
   (interactive
    (list (if current-prefix-arg (prefix-numeric-value current-prefix-arg))))
   (save-excursion
-    (when (featurep 'vimish-fold)
-      (vimish-fold-refold-all))
-    (+fold--ensure-hideshow-mode)
-    (hs-life-goes-on
-     (if (integerp level)
-         (hs-hide-level-recursive (1- level) (point-min) (point-max))
-       (hs-hide-all)))))
+    (cond ((+fold--ts-fold-p)
+           (ts-fold-close-all))
+          (t (when (featurep 'vimish-fold)
+               (vimish-fold-refold-all))
+             (+fold--ensure-hideshow-mode)
+             (hs-life-goes-on
+              (if (integerp level)
+                  (hs-hide-level-recursive (1- level) (point-min) (point-max))
+                (hs-hide-all)))))))
 
 ;;;###autoload
 (defun +fold/next (count)
--- a/modules/editor/fold/autoload/fold.el
+++ b/modules/editor/fold/autoload/fold.el
@@ -32,6 +32,10 @@
             (end-of-line)
             (+fold--hideshow-fold-p))))))
 
+;; NOTE: does this need more?
+(defun +fold--ts-fold-p ()
+  (featurep 'ts-fold))
+
 (defun +fold--invisible-points (count)
   (let (points)
     (save-excursion
@@ -62,7 +66,7 @@
 (defun +fold/toggle ()
   "Toggle the fold at point.
 
-Targets `vimmish-fold', `hideshow' and `outline' folds."
+Targets `vimmish-fold', `hideshow', `ts-fold' and `outline' folds."
   (interactive)
   (save-excursion
     (cond ((+fold--vimish-fold-p) (vimish-fold-toggle))
@@ -70,29 +74,32 @@ Targets `vimmish-fold', `hideshow' and `outline' folds."
            (cl-letf (((symbol-function #'outline-hide-subtree)
                       (symbol-function #'outline-hide-entry)))
              (outline-toggle-children)))
+          ((+fold--ts-fold-p) (ts-fold-toggle))
           ((+fold--hideshow-fold-p) (+fold-from-eol (hs-toggle-hiding))))))
 
 ;;;###autoload
 (defun +fold/open ()
   "Open the folded region at point.
 
-Targets `vimmish-fold', `hideshow' and `outline' folds."
+Targets `vimmish-fold', `hideshow', `ts-fold' and `outline' folds."
   (interactive)
   (save-excursion
     (cond ((+fold--vimish-fold-p) (vimish-fold-unfold))
           ((+fold--outline-fold-p)
            (outline-show-children)
            (outline-show-entry))
+          ((+fold--ts-fold-p) (ts-fold-open))
           ((+fold--hideshow-fold-p) (+fold-from-eol (hs-show-block))))))
 
 ;;;###autoload
 (defun +fold/close ()
   "Close the folded region at point.
 
-Targets `vimmish-fold', `hideshow' and `outline' folds."
+Targets `vimmish-fold', `hideshow', `ts-fold' and `outline' folds."
   (interactive)
   (save-excursion
     (cond ((+fold--vimish-fold-p) (vimish-fold-refold))
+          ((+fold--ts-fold-p) (ts-fold-close))
           ((+fold--hideshow-fold-p) (+fold-from-eol (hs-hide-block)))
           ((+fold--outline-fold-p) (outline-hide-subtree)))))
 
--- a/modules/editor/fold/autoload/fold.el
+++ b/modules/editor/fold/autoload/fold.el
@@ -34,7 +34,7 @@
 
 ;; NOTE: does this need more?
 (defun +fold--ts-fold-p ()
-  (and tree-sitter-mode
+  (and (bound-and-true-p tree-sitter-mode)
        (featurep 'ts-fold)))
 
 (defun +fold--invisible-points (count)
--- a/modules/editor/fold/autoload/fold.el
+++ b/modules/editor/fold/autoload/fold.el
@@ -34,7 +34,8 @@
 
 ;; NOTE: does this need more?
 (defun +fold--ts-fold-p ()
-  (featurep 'ts-fold))
+  (and tree-sitter-mode
+       (featurep 'ts-fold)))
 
 (defun +fold--invisible-points (count)
   (let (points)
--- a/modules/editor/fold/autoload/fold.el
+++ b/modules/editor/fold/autoload/fold.el
@@ -172,7 +172,8 @@ Targets `vimmish-fold', `hideshow', `ts-fold' and `outline' folds."
                                      (cl-sort (eval `(overlays-in ,@arg-list))
                                               (lambda (ov1 ov2)
                                                 (funcall comp-fun (overlay-start ov1) (overlay-start ov2)))))))
-                          (if ovs (goto-char (overlay-start (nth (- (abs count) 1) ovs))))))))
+                          (if (and ovs (<= count (length ovs)))
+                                   (goto-char (overlay-start (nth (- (abs count) 1) ovs))))))))
            if (save-excursion (funcall fn))
            collect it into points
            finally do
--- a/modules/editor/fold/autoload/fold.el
+++ b/modules/editor/fold/autoload/fold.el
@@ -113,16 +113,16 @@ Targets `vimmish-fold', `hideshow', `ts-fold' and `outline' folds."
          (ts-fold-open-all))
         ((featurep 'vimish-fold)
          (vimish-fold-unfold-all))
-        (t (save-excursion
-             (+fold--ensure-hideshow-mode)
-             (if (integerp level)
-                 (progn
-                   (outline-hide-sublevels (max 1 (1- level)))
-                   (hs-life-goes-on
-                    (hs-hide-level-recursive (1- level) (point-min) (point-max))))
-               (hs-show-all)
-               (when (fboundp 'outline-show-all)
-                 (outline-show-all)))))))
+        ((save-excursion
+           (+fold--ensure-hideshow-mode)
+           (if (integerp level)
+               (progn
+                 (outline-hide-sublevels (max 1 (1- level)))
+                 (hs-life-goes-on
+                  (hs-hide-level-recursive (1- level) (point-min) (point-max))))
+             (hs-show-all)
+             (when (fboundp 'outline-show-all)
+               (outline-show-all)))))))
 
 ;;;###autoload
 (defun +fold/close-all (&optional level)
@@ -130,15 +130,16 @@ Targets `vimmish-fold', `hideshow', `ts-fold' and `outline' folds."
   (interactive
    (list (if current-prefix-arg (prefix-numeric-value current-prefix-arg))))
   (save-excursion
-    (cond ((+fold--ts-fold-p)
-           (ts-fold-close-all))
-          (t (when (featurep 'vimish-fold)
-               (vimish-fold-refold-all))
-             (+fold--ensure-hideshow-mode)
-             (hs-life-goes-on
-              (if (integerp level)
-                  (hs-hide-level-recursive (1- level) (point-min) (point-max))
-                (hs-hide-all)))))))
+    (if (+fold--ts-fold-p)
+        (ts-fold-close-all)
+      (progn
+        (when (featurep 'vimish-fold)
+          (vimish-fold-refold-all))
+        (+fold--ensure-hideshow-mode)
+        (hs-life-goes-on
+         (if (integerp level)
+             (hs-hide-level-recursive (1- level) (point-min) (point-max))
+           (hs-hide-all)))))))
 
 ;;;###autoload
 (defun +fold/next (count)
@@ -164,7 +165,7 @@ Targets `vimmish-fold', `hideshow', `ts-fold' and `outline' folds."
                                (comp-fun (if (> count 0) ;; also depending on direction we need to change how we sort the list
                                              #'<
                                            #'>))
-                               (ovs (seq-filter
+                               (ovs (cl-remove-if-not
                                      (lambda (ov)
                                        (eq (overlay-get ov 'creator) 'ts-fold))
                                      ;; `overlays-in' does not provide a list that is sorted
--- a/modules/lang/javascript/config.el
+++ b/modules/lang/javascript/config.el
@@ -330,6 +330,6 @@ to tide."
                js3-mode-hook
                typescript-mode-hook
                typescript-tsx-mode-hook
-               rjsx-mode)
+               rjsx-mode-hook)
              #'turn-on-tree-sitter-mode
              #'+tree-sitter-keys-mode))
--- a/modules/lang/javascript/config.el
+++ b/modules/lang/javascript/config.el
@@ -127,7 +127,17 @@
     (define-derived-mode typescript-tsx-mode web-mode "TypeScript-TSX")
     (when (featurep! +lsp)
       (after! lsp-mode
-        (add-to-list 'lsp--formatting-indent-alist '(typescript-tsx-mode . typescript-indent-level)))))
+        (add-to-list 'lsp--formatting-indent-alist '(typescript-tsx-mode . typescript-indent-level))))
+    (when (featurep! +tree-sitter)
+      (after! tree-sitter
+        (pushnew! tree-sitter-major-mode-language-alist '(typescript-tsx-mode . tsx))
+        (pushnew! evil-textobj-tree-sitter-major-mode-language-alist '(typescript-tsx-mode . "tsx"))
+
+        ;; HACK: the tsx grammer doesn't work with the hightlighting provided by
+        ;; font-lock-keywords.
+        ;; https://github.com/emacs-tree-sitter/tree-sitter-langs/issues/23
+        (setq-hook! 'typescript-tsx-mode-hook
+          tree-sitter-hl-use-font-lock-keywords nil))))
 
   (set-docsets! '(typescript-mode typescript-tsx-mode)
     :add "TypeScript" "AngularTS")
@@ -319,6 +329,7 @@ to tide."
                js2-mode-hook
                js3-mode-hook
                typescript-mode-hook
+               typescript-tsx-mode-hook
                rjsx-mode)
              #'turn-on-tree-sitter-mode
              #'+tree-sitter-keys-mode))
--- a/modules/lang/javascript/config.el
+++ b/modules/lang/javascript/config.el
@@ -319,4 +319,6 @@ to tide."
                js2-mode-hook
                js3-mode-hook
                typescript-mode-hook
-               rjsx-mode) #'turn-on-tree-sitter-mode))
+               rjsx-mode)
+             #'turn-on-tree-sitter-mode
+             #'+tree-sitter-keys-mode))
--- a/modules/lang/javascript/config.el
+++ b/modules/lang/javascript/config.el
@@ -134,8 +134,7 @@
         (pushnew! evil-textobj-tree-sitter-major-mode-language-alist '(typescript-tsx-mode . "tsx"))
 
         ;; HACK: the tsx grammer doesn't work with the hightlighting provided by
-        ;; font-lock-keywords.
-        ;; https://github.com/emacs-tree-sitter/tree-sitter-langs/issues/23
+        ;; font-lock-keywords. See emacs-tree-sitter/tree-sitter-langs#23
         (setq-hook! 'typescript-tsx-mode-hook
           tree-sitter-hl-use-font-lock-keywords nil))))
 
--- a/modules/lang/javascript/config.el
+++ b/modules/lang/javascript/config.el
@@ -324,12 +324,9 @@ to tide."
 
 ;; Tree sitter
 (eval-when! (featurep! +tree-sitter)
-  (add-hook! '(javascript-mode-hook
-               js-mode-hook
-               js2-mode-hook
-               js3-mode-hook
-               typescript-mode-hook
-               typescript-tsx-mode-hook
-               rjsx-mode-hook)
-             #'turn-on-tree-sitter-mode
-             #'+tree-sitter-keys-mode))
+  (add-hook! '(js-mode-local-vars-hook
+               js2-mode-local-vars-hook
+               typescript-mode-local-vars-hook
+               typescript-tsx-mode-local-vars-hook
+               rjsx-mode-local-vars-hook)
+             #'tree-sitter!))
--- a/modules/lang/javascript/config.el
+++ b/modules/lang/javascript/config.el
@@ -311,3 +311,12 @@ to tide."
 
 (def-project-mode! +javascript-gulp-mode
   :when (locate-dominating-file default-directory "gulpfile.js"))
+
+;; Tree sitter
+(eval-when! (featurep! +tree-sitter)
+  (add-hook! '(javascript-mode-hook
+               js-mode-hook
+               js2-mode-hook
+               js3-mode-hook
+               typescript-mode-hook
+               rjsx-mode) #'turn-on-tree-sitter-mode))
--- a/modules/lang/sh/config.el
+++ b/modules/lang/sh/config.el
@@ -96,6 +96,4 @@
 
 ;; Tree sitter
 (eval-when! (featurep! +tree-sitter)
-  (add-hook! 'sh-mode-hook
-             #'turn-on-tree-sitter-mode
-             #'+tree-sitter-keys-mode))
+  (add-hook! 'sh-mode-local-vars-hook #'tree-sitter!))
--- a/modules/lang/sh/config.el
+++ b/modules/lang/sh/config.el
@@ -24,18 +24,18 @@
   (set-repl-handler! 'sh-mode #'+sh/open-repl)
   (set-lookup-handlers! 'sh-mode :documentation #'+sh-lookup-documentation-handler)
   (set-ligatures! 'sh-mode
-    ;; Functional
-    :def "function"
-    ;; Types
-    :true "true" :false "false"
-    ;; Flow
-    :not "!"
-    :and "&&" :or "||"
-    :in "in"
-    :for "for"
-    :return "return"
-    ;; Other
-    :dot "." :dot "source")
+                  ;; Functional
+                  :def "function"
+                  ;; Types
+                  :true "true" :false "false"
+                  ;; Flow
+                  :not "!"
+                  :and "&&" :or "||"
+                  :in "in"
+                  :for "for"
+                  :return "return"
+                  ;; Other
+                  :dot "." :dot "source")
 
   (when (featurep! +lsp)
     (add-hook 'sh-mode-local-vars-hook #'lsp! 'append))
@@ -93,3 +93,7 @@
   :config
   (when (featurep! +lsp)
     (add-hook 'powershell-mode-local-vars-hook #'lsp! 'append)))
+
+;; Tree sitter
+(eval-when! (featurep! +tree-sitter)
+  (add-hook! 'sh-mode-hook #'turn-on-tree-sitter-mode))
--- a/modules/lang/sh/config.el
+++ b/modules/lang/sh/config.el
@@ -24,18 +24,18 @@
   (set-repl-handler! 'sh-mode #'+sh/open-repl)
   (set-lookup-handlers! 'sh-mode :documentation #'+sh-lookup-documentation-handler)
   (set-ligatures! 'sh-mode
-                  ;; Functional
-                  :def "function"
-                  ;; Types
-                  :true "true" :false "false"
-                  ;; Flow
-                  :not "!"
-                  :and "&&" :or "||"
-                  :in "in"
-                  :for "for"
-                  :return "return"
-                  ;; Other
-                  :dot "." :dot "source")
+    ;; Functional
+    :def "function"
+    ;; Types
+    :true "true" :false "false"
+    ;; Flow
+    :not "!"
+    :and "&&" :or "||"
+    :in "in"
+    :for "for"
+    :return "return"
+    ;; Other
+    :dot "." :dot "source")
 
   (when (featurep! +lsp)
     (add-hook 'sh-mode-local-vars-hook #'lsp! 'append))
--- a/modules/lang/sh/config.el
+++ b/modules/lang/sh/config.el
@@ -96,4 +96,6 @@
 
 ;; Tree sitter
 (eval-when! (featurep! +tree-sitter)
-  (add-hook! 'sh-mode-hook #'turn-on-tree-sitter-mode))
+  (add-hook! 'sh-mode-hook
+             #'turn-on-tree-sitter-mode
+             #'+tree-sitter-keys-mode))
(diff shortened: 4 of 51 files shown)


jeetelongname:
cc: @iyefrat

ymarco:
Don't forget to add an entry to modules.org

iyefrat:
And a `doctor.el` check for module support

ymarco:
Would be nice to list the new textobject keybinds in README.org too

petr-tik:
a lurker really keen to see this and LSP work together nicely.

Would it make sense to disable lsp semantic highlighting with `(setq lsp-semantic-tokens-enable nil)` when tree-sitter is enabled ?

modules/tools/treesitter/config.el
```diff
@@ -0,0 +1,44 @@
...
+    "Don't break with errors when current major mode lacks tree-sitter support."
+    :around #'tree-sitter-mode
+    (condition-case e
+        (apply orig-fn args)
+      (error
+       (unless (string-match-p (concat "^Cannot find shared library\\|"
+                                       "^No language registered\\|"
+                                       "cannot open shared object file")
+                               (error-message-string e))
+         (signal (car e) (cadr e)))))))
```
daanturo, line 25: This advice isn't needed anymore, we can verify that `turn-on-tree-sitter-mode` & `global-tree-sitter-mode` won't signal errors:
[tree-sitter.el](https://github.com/emacs-tree-sitter/elisp-tree-sitter/blob/c7a1c34549cad41a3618c6f17e0e9dabd3e98fe1/lisp/tree-sitter.el#L209)
I have opened a PR: [Remove doom-tree-sitter-fail-gracefully-a](https://github.com/jeetelongname/doom-emacs/pull/1)

jeetelongname: It would still make sense to hook tree sitter on prog-modes as thats where a lot of the usecases come from. we should try and load it late as possible to improve performace.

jeetelongname:
yeah that makes sense

Regards,
– *Jeetaditya Chatterjee* 
/Sent using my text editor/

jeetelongname:
> a lurker really keen to see this and LSP work together nicely.
> 
> Would it make sense to disable lsp semantic highlighting with `(setq lsp-semantic-tokens-enable nil)` when tree-sitter is enabled ?
Yeah that would make sense. Ill look into it

modules/tools/treesitter/README.org
```diff
@@ -0,0 +1,75 @@
...
+  - [[#language-support][Language support]]
+  - [[#text-objects][Text Objects]]
+- [[#configuration][Configuration]]
+- [[#troubleshooting][Troubleshooting]]
+
+* Description
+Add [[https://tree-sitter.github.io/tree-sitter/][treesitter]] support to doom:
+
+#+begin_quote
+Tree-sitter is a parser generator tool and an incremental parsing library. It
```
Mic92, line 23: For the nix folks, you can use this snippet for testing:

https://github.com/Mic92/dotfiles/blob/07273b7f57b9c78eaa98eddff2fc52335f7a7ed0/nixpkgs-config/modules/emacs/default.nix#L49
combined with:
https://github.com/Mic92/dotfiles/blob/f52d35be19c16cc166040ab5c7cdbb4686f8434f/home/.doom.d/config.el#L187

Mic92: It's not ideal because it installs all grammars rather than the supported by doom, but it pays the bills for me...

modules/tools/treesitter/config.el
jeetelongname: Also could you rebase your commit so that it conforms to the [commit conventions](https://gist.github.com/hlissner/4d78e396acb897d9b2d8be07a103a854)

meain:
@jeetelongname Just wanted to give you a heads up on the name change of the `evil-textobj-treesitter` package to `evil-textobj-tree-sitter`(extra `-` between tree and sitter). This is done so as to keep it consistent with the other repos ([ref](https://github.com/meain/evil-textobj-tree-sitter/issues/3)). The change will be complete as soon as https://github.com/melpa/melpa/pull/7698 is merged in.

jeetelongname:
@meain I have been following that and am thankful for the change! 

thanks for this!

docs/modules.org
```diff
@@ -194,6 +194,7 @@ Small modules that give Emacs access to external tools & services.
 + [[file:../modules/tools/taskrunner/README.org][taskrunner]] - TODO
 + [[file:../modules/tools/terraform/README.org][terraform]] - TODO
 + tmux - TODO
++ [[file:~/.config/emacs.doom/modules/tools/treesitter/README.org][treesitter]]  - TODO
```
Mic92, line 197:
```suggestion
+ [[file:~/.config/emacs.doom/modules/tools/treesitter/README.org][treesitter]]  - Fast & precise incremental parsing library for syntax highlighting
```

jeetelongname:
While its a good tag line its not /witty/ I have come up with “Syntax and Parsing
sitting in a tree…” but I am yet to add it to the init.example and modules.org
as I want that to be one of the last commits a sign off if you will

Regards,
– *Jeetaditya Chatterjee* 
/Sent using my text editor/

jeetelongname:
to add to my lower comment it not only does syntax highlighting but a bunch of
other stuff as well (even if those come from plugins)

Regards,
– *Jeetaditya Chatterjee* 
/Sent using my text editor/

docs/modules.org
marcbowes: This looks like an absolute path while all the other links are relative.

jeetelongname: Yes thank you! Will fix

modules/tools/tree-sitter/README.org
```diff
@@ -0,0 +1,74 @@
...
+objects languages support]]
+Note: only languages with parsers in emacs have text object support currently.
+Currently text objects are bound to:
+| key | text object |
+|-----+-------------|
+| =f=   | function    |
+| =C=   | class       |
+| =c=   | comment     |
+| =i=   | conditional |
+| =l=   | loop        |
```
sei40kr, line 66:
```suggestion
| =f= | function    |
| =C= | class       |
| =c= | comment     |
| =i= | conditional |
| =l= | loop        |
```

modules/tools/tree-sitter/config.el
```diff
@@ -0,0 +1,32 @@
+;;; ui/tree-sitter/config.el -*- lexical-binding: t; -*-
```
sei40kr, line 1:
```suggestion
;;; tools/tree-sitter/config.el -*- lexical-binding: t; -*-
```

modules/tools/tree-sitter/packages.el
```diff
@@ -0,0 +1,12 @@
+;; -*- no-byte-compile: t; -*-
+;;; ui/tree-sitter/packages.el
```
sei40kr, line 2:
```suggestion
;;; tools/tree-sitter/packages.el
```

modules/tools/tree-sitter/packages.el
```diff
@@ -0,0 +1,12 @@
...
+
+(package! tree-sitter
+  :pin "4d9871d23999fe5f8de821e23c9ec576df2b2738")
+
+(package! tree-sitter-langs
+  :pin "fa47b55f7bd11bd2b17ab48deb03ed23000bb974")
+
+(when (featurep! :editor evil +everywhere)
```
sei40kr, line 10: I don't think +everywhere is needed here.

```suggestion
(when (featurep! :editor evil)
```

modules/tools/tree-sitter/README.org
```diff
@@ -0,0 +1,74 @@
...
+** Maintainers
+- @jeetelongname
+
+** Module Flags
+This module provides no flags.
+
+** Plugins
++ [[https://github.com/emacs-tree-sitter/elisp-tree-sitter][tree-sitter]]
++ [[https://github.com/emacs-tree-sitter/tree-sitter-langs][tree-sitter-langs]]
++ [[https://github.com/meain/evil-textobj-tree-sitter][evil-textobj-tree-sitter]]* (=:editor evil +everywhere=)
```
sei40kr, line 42:
```suggestion
+ [[https://github.com/meain/evil-textobj-tree-sitter][evil-textobj-tree-sitter]]* (=:editor evil=)
```

modules/tools/tree-sitter/config.el
```diff
@@ -0,0 +1,32 @@
...
+
+(use-package! tree-sitter
+  :when (bound-and-true-p module-file-suffix)
+  :hook (prog-mode . turn-on-tree-sitter-mode)
+  :hook (tree-sitter-after-on . tree-sitter-hl-mode)
+  :config
+  (require 'tree-sitter-langs)
+
+  ;; This makes every node a link to a section of code
```
sei40kr, line 10:
```suggestion
  ;; This makes every node a link to a section of code
```

modules/tools/tree-sitter/config.el
```diff
@@ -0,0 +1,32 @@
...
+  (setq tree-sitter-debug-jump-buttons t
+        ;; and this highlights the entire sub tree in your code
+        tree-sitter-debug-highlight-jump-region t))
+
```
sei40kr, line 14:
```suggestion

(use-package! tree-sitter-langs
  :after tree-sitter)

```

modules/tools/tree-sitter/README.org
jeetelongname: While this looks like a problem from a plain text perspective when rendered using org mode (or turned into html) it becomes a non issue (or in the case of org it makes the table look wonky)

modules/tools/tree-sitter/config.el
jeetelongname: Is there any advantage to doing it this way? they seem to be doing the same thing except your one makes a second call to `eval-after-load`.

jeetelongname: Is there any advatage to adding a dedicated use-package declaration? the only thing I see it doing is calling `eval-after-load` a second time

modules/tools/tree-sitter/config.el
```diff
@@ -0,0 +1,32 @@
...
+(when (featurep! :editor evil +everywhere)
```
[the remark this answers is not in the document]
jeetelongname: as this is adding evil support to tree-sitter I will say that it does need it. for a similar example look at the `:lang latex` module where `evil-tex` is only installed if `+everywhere` is true

modules/tools/tree-sitter/README.org
jeetelongname: My mistake (had fortifications on ). This should be fixed now

modules/tools/tree-sitter/config.el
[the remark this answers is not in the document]
sei40kr: @jeetelongname 
From my understanding, `evil +everywhere` is a flag to enable evil keybinds in **tool buffers** like Magit, Treemacs.

I don't know why `evil-tex` is only installed where `+everywhere` is true, but I think we don't need `+everywhere` here.

modules/tools/tree-sitter/config.el
sei40kr: @jeetelongname
The advantage to use `use-package!` is the code looks more declarative.
Current implementation seem to be using `use-package!` as much as possible.

modules/tools/tree-sitter/config.el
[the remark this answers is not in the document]
jeetelongname: I will discuss this with Henrik when i get the chance to as he is the final call here. But I did discuss this in the discord server and the consensus I got was that `+everywhere` is needed here

modules/tools/tree-sitter/config.el
jeetelongname: Yeah after looking at the code base more I will tend to agree with you. Its been added. Thanks for your help!

dit7ya:
Hi, thank you so much for this PR - really excited about Emacs tree-sitter ecosystem grow. I am not sure if it is already covered in the commits, but if not, would it be possible to add something like this [eckon/treesitter-current-functions: Quickly search through and jump to function definitions in the current neovim buffer](https://github.com/eckon/treesitter-current-functions) maybe not just for functions but all text-objects via a selection framework?

jeetelongname:
@dit7ya  Currently I have not written anything like that and don't plan to. I use lsp for jumping to definitions (or just regular isearch) but that all being said I would not mind adding it if one exists and people find it useful. 

If you do get around to writing one (or end up finding one) just ping me a link and ill get onto integrating it

jeetelongname:
Yes that was mentioned in the [discord server](https://doomemacs.org/discord). I was actually writing my own when this came around and stole my thunder :sweat_smile:. The only real problem I have is the non idiomatic way to get ranges so I am discussing a refactor with them. 

As for the goto functionality thats honestly sounds awesome! do let me know when that is merged and I will defiantly be making it in!

ckruse:
`typescript-tsx-mode-hook` is derived from `web-mode`, but `tree-sitter` doesn't play well with web-mode: <https://github.com/emacs-tree-sitter/tree-sitter-langs/issues/23>

jeetelongname:
Hey,

Yes so as people have said tsx mode derives from web mode which is known to not
work with tree-sitter. If there is another way to work with tsx and it has
similar features what I can do is make the module swap out the mode if tree
sitter is enabled though this will come with the caviet of losing all the
niceties web mode provides.

Regards,
– *Jeetaditya Chatterjee* 
/Sent using my text editor/

steven-lang:
Any updates on this? Is this module working as is or are there any conflicts with lsp?

jeetelongname:
This module as is is really usable (I use it everyday as well as many other ~~guinea pigs~~ brave pioneers) as for conflicts with lsp there have never been any. they work on very different parts of emacs and the only real conflict that *may* arise is semantic highlighting and even that is disabled by default in doom. (I remember doing some testing but don't remember if it actually caused a conflict (the fact I can't remember suggests that it did not)) If you do encounter any bugs though then do let me know. 

I have a few more things that I need to do (mostly around language support) and I am working on a playground like plugin which is nearly there (ill probably add it into the module in the coming days).

TLDR, the module is very usable, please report bugs

danilevy1212:
Install instructions while this is being developed? @jeetelongname

jeetelongname:
for the most part pull this branch and doom sync, the only dependency is modules support in emacs and that can't be pulled.

jeetelongname:
My Christmas gift is getting this PR ready to be merged! everything is technically correct but that does not mean I have not made mistakes, please continue to report bugs or taze me on my grammar in the docs.

I would like to thank everyone who has helped and tested this module! you have been invaluable!

otavioschwanck:
Hi, thanks for your PR.

Tested on ruby mode and got:

```
Contacting host: github.com:443
Wrote /Users/otavio/.emacs.d/.local/straight/build-27.2/tsc/tsc-dyn.dylib
Eager macro-expansion failure: (file-missing "Cannot open load file" "No such file or directory" "tsc-dyn")
Downloading https://github.com/emacs-tree-sitter/elisp-tree-sitter/releases/download/0.16.0/tsc-dyn.dylib
Wrote /Users/otavio/.emacs.d/.local/straight/build-27.2/tsc/tsc-dyn.dylib
File mode specification error: (doom-hook-error ruby-mode-hook turn-on-tree-sitter-mode (file-missing Cannot open load file No such file or directory tsc-dyn))
```

OS: Mac
Installed gem tree-sitter and tree-sitter on brew.

otavioschwanck (changes requested):
Ruby tree-sitter not working

otavioschwanck (approved):

jeetelongname:
@otavioschwanck I can&rsquo;t repro, I have been testing this module on a ruby project and its been working fine, does the same error appear on other languages?

Have you tried doom syncing again?

also are you using the m1 macbooks as that is known not to work due to the grammars being built for x86 computers.

and finally this module does not depend on any tree sitter executable as all the grammars are prebuilt, Its only needed if you plan on building the grammars for yourself which is only needed if you use m1 macbooks

- Jeet

otavioschwanck:
> @otavioschwanck I can’t repro, I have been testing this module on a ruby project and its been working fine, does the same error appear on other languages?
> 
> Have you tried doom syncing again?
> 
> also are you using the m1 macbooks as that is known not to work due to the grammars being built for x86 computers.
> 
> and finally this module does not depend on any tree sitter executable as all the grammars are prebuilt, Its only needed if you plan on building the grammars for yourself which is only needed if you use m1 macbooks
> 
> * Jeet

i'm using m1.   I hope to have a way to use on m1 any time soon.

jeetelongname:
there is some work to make some self build infrastructure but its not done, I may start working on some simple alternatives in the interum tho as there are a few m1 users, (I should probably add a warning to the readme as well

aisamu:
Just tested it on typescript & json: works _beautifully_, thanks a lot for taking the time to make this happen!

I was also able to use "parameter.inner" on javascript, overwriting the existing `evil-inner-arg`:

```diff
modified   modules/tools/tree-sitter/config.el
@@ -32,2 +32,3 @@
   (map! (:map +tree-sitter-inner-text-objects-map
+         "a" (evil-textobj-tree-sitter-get-textobj "parameter.inner")
          "f" (evil-textobj-tree-sitter-get-textobj "function.inner")
```
Would it possible to enable it on a per-lang basis? [(some languages don't have that implemented)](https://github.com/nvim-treesitter/nvim-treesitter-textobjects#built-in-textobjects)

jeetelongname:
if langs don't implemenet a text object I see that as an excuse to add the queries to said languages, I have been meaning to add more but have not mainly due to fear of conflicts, but I don't mind adding more

amosbird:
Do we have ways to inspect the object under the cursor which tree-sitter faces should be applied? I'm trying to debug why the returning type of some methods in c++ are not rendered properly.

image

meain:
@amosbird This is what I usually do. Call `customize-face` with the cursor on top and that will give me the name of the face. You can also run `tree-sitter-debug-mode` which will open up the parsed tree for the current buffer in a separate buffer which you can use to check what the node is and then if necessary add custom highlighting. https://emacs-tree-sitter.github.io/syntax-highlighting/customization/ (not sure if there is a doom specific way of doing this).

meain:
That message is supposed to happen the second time(toggling it off) you call `tree-sitter-debug-mode`. Make sure there is no buffer getting created in between

https://user-images.githubusercontent.com/14259816/147959228-d17df285-faef-4680-9597-834dd7f4b591.mp4

.

jeetelongname:
@amosbird I use `SPC h '` to inspect the faces, it should tell you what faces are being applied to the character, You can also inspect the tree with `tree-sitter-debug-mode` as meain has mentioned, I have been using it with this module for pretty much its entire development and its worked fine so if this problem persists let me know

yqrashawn:
Hi , I find some minor issues with ts-fold integration.

At this line https://github.com/MetaMask/metamask-extension/blob/88faef7e19228652e2bfb339ffc3392926ee985d/app/scripts/controllers/transactions/tx-gas-utils.js#L34

when tree-sitter-mode is enabled, I can only fold/unfold at the open bracket
image
when tree-sitter-mode is disabled, I can fold/unfold at any point of this line

Another issue is I use `hs-level-hide` a lot since there's no doom `+fold/.*` functions for that. the `(+fold--ts-fold-p)` cond in `+fold/.*` functions brakes them after `hs-level-hide`

jeetelongname:
ts-fold needs some work which I need to get to. It does not bother me because my folding is quite simple but can be easily disabled with a `(package! ts-fold :disable t)`

As for your second point I am not sure I understand. Does it break hs-level-hide? Or does it break one of the fold functions?

yqrashawn:
> As for your second point I am not sure I understand. Does it break hs-level-hide? Or does it break one of the fold functions?

The behavior is folds created with `hs-hide-level` can't unfold by `+fold/toggle`
This might caused by the overlay.

The fold (created with `hs-hide-level`) behave this when I `(setq hs-set-up-overlay 'ignore)`

Screenshot-2022-01-13-10 19 48

jeetelongname:
thats because `+fold/toggle` will just run the first folder that returns true and not fall through, I am not really sure if this is a bug as mixing folders usually does not go all to well and adding that ability to `+fold/toggle` will probably add in more bugs than it solves, Still `hs-toggle-hiding` works if you want to unfold that and if you disable `ts-fold` then `+fold/toggle` will revert back to using hideshow

amosbird:
When tree-sitter is enabled, sometimes `C-s` key gets mapped to an action which leads to `font-lock-fontify-keywords-region: No match 1 in highlight (1 'font-lock-constant-face prepend)` .

jeetelongname:
@amosbird I will need more information, What language are you using, is `C-s` bound to isearch-forward? in what language? a better backtrace would also be helpful. From the little infomation I have gotten I can't reproduce, It may be a user error but I have too little information to speculate

the-moriarty:
About the problem with m1:

On vim, tree-sitter works perfectly.  Maybe it is a problem with emacs package of tree-sitter?

ckruse:
Tree-sitter support for M1 is blocked by elisp-tree-sitter. The code just doesn't fully support m1 yet, you can track it at [issue #197](https://github.com/emacs-tree-sitter/elisp-tree-sitter/issues/197) at the elisp-tree-sitter repo.

jeetelongname:
@the-moriarty thats because nvim compiles all of the grammers on your system while emacs will pull down prebuilt bins, as ckruse mentioned work is being done for this tho and hopefully soon m1 users will get support on this

sei40kr:
@jeetelongname 
I added the following snippet and got TypeScript TSX syntax highlighting working.

```emacs-lisp
(add-to-list 'tree-sitter-major-mode-language-alist '(typescript-tsx-mode . tsx))
```

jeetelongname:
@sei40kr what about text object support? I will need to spin up a dummy tsx project to test it (so a link to one would be nice) if that works Ill add it and consider tsx supported

sei40kr:
@jeetelongname 
> what about text object support?

Sorry I didn't test it. I'm going to test it tomorrow if I have time.

Here is my Typescript (TSX) project:
https://github.com/sei40kr/yong-ju.me

jeetelongname:
@sei40kr Thats cool ill also give it a test tomorrow and if all goes well then ill add support and keep an eye on it. Ill have to see why thats the case and if there are some underlying problems that may be blockers but hopefully all goes well

sei40kr:
@jeetelongname 
None of Treesitter text objects did work...

Here is my diff:

```diff
diff --git a/modules/lang/javascript/config.el b/modules/lang/javascript/config.el
index 35fe81182..54a483dae 100644
--- a/modules/lang/javascript/config.el
+++ b/modules/lang/javascript/config.el
@@ -125,7 +125,16 @@
     (define-derived-mode typescript-tsx-mode web-mode "TypeScript-TSX")
     (when (featurep! +lsp)
       (after! lsp-mode
-        (add-to-list 'lsp--formatting-indent-alist '(typescript-tsx-mode . typescript-indent-level)))))
+        (add-to-list 'lsp--formatting-indent-alist '(typescript-tsx-mode . typescript-indent-level))))
+    (when (featurep! +tree-sitter)
+      (after! tree-sitter
+        (add-to-list 'tree-sitter-major-mode-language-alist '(typescript-tsx-mode . tsx))
+
+        ;; tsx grammer doesn't work with the highlighting provided by
+        ;; font-lock-keywords.
+        ;; https://github.com/emacs-tree-sitter/tree-sitter-langs/issues/23
+        (setq-hook! 'typescript-tsx-mode-hook
+          tree-sitter-hl-use-font-lock-keywords nil))))
 
   (set-docsets! '(typescript-mode typescript-tsx-mode)
     :add "TypeScript" "AngularTS")
@@ -317,6 +326,7 @@ to tide."
                js2-mode-hook
                js3-mode-hook
                typescript-mode-hook
+               typescript-tsx-mode-hook
                rjsx-mode)
              #'turn-on-tree-sitter-mode
              #'+tree-sitter-keys-mode))
```

meain:
@sei40kr I don't have a doom setup, but adding something like below should work for `evil-textobj-tree-sitter`. I can maybe add this to the default list as well.

```emacs-lisp
(add-to-list 'evil-textobj-tree-sitter-major-mode-language-alist '(typescript-tsx-mode . "tsx"))
```

jeetelongname:
I am also not sure where typescript-tsx-mode is coming from. looking into it there is no real associated file or anything

sei40kr:
@jeetelongname 
I stole typescript-tsx-mode from Spacemacs, and it’s actually web-mode.

It’s defined in modules/lang/javascript/config.el.

jeetelongname:
Yep sounds good Ill push the changes for this, Ill also see if web mode can be supported with a similar trick

the-moriarty:
Hi, can you bump tree-sitter to 258eaa98c95b0afce12bfc1a87e6e9d23b13ba8c and remove the M1 mac comments ?  They fixed!

jeetelongname:
Will do!

jeetelongname:
@the-moriarty I have bumped the module, if you could give it a test (it also looks like that the CI is set up for arm) you will need the rust tool chain if you do need to compile the dynamic module. CC @otavioschwanck @elken

the-moriarty:
The text objects are not working in ruby-mode:

image

The text-objects are disabled by default?

the-moriarty:
Some throughs about how its configured:

the `i` text object is very important.   The tree-sitter version of it is very different (conditional), maybe another key for it?

the `a` text object is different on tree-sitter.  On tree-sitter is only for parameters in definition.  The arguments from functions call are excluded.  Here we can three options:

1. Bring back old behaviour
2. On tree-sitter-mode, push a function to detect parameters AND arguments. (that is what `a` does now)
3. Create or own function to do number 2.

meain:
Regarding 2, not sure if that is what you want to do by default, but in `evil-textobj-tree-sitter` you can specify multiple textobjects and we will match the first one that we can find in the buffer.

```emacs-lisp
(define-key evil-outer-text-objects-map "a" (evil-textobj-tree-sitter-get-textobj ("conditional.outer" "loop.outer")))
```

the-moriarty:
> Regarding 2, not sure if that is what you want to do by default, but in `evil-textobj-tree-sitter` you can specify multiple textobjects and we will match the first one that we can find in the buffer.
> 
> ```emacs-lisp
> (define-key evil-outer-text-objects-map "a" (evil-textobj-tree-sitter-get-textobj ("conditional.outer" "loop.outer")))
> ```

the current `a` looks for argument_list and parameters.   In this case it should be each argument item list, that can be anything.  dont know how to do it

jeetelongname:
@the-moriarty if there is another key for it then I am happy to replace it (forgive me as I don't use i) 
@meain I have bound a to parameter but do not mind adding another, at the moment there is also the `F` text object for funcion calls but I can add that to the param object as well (or remove `F`

meain:
I'll leave the keybindings and everything up to you. Just wanted to let you know that there is an option to provide multiple textobjects in the function.

jeetelongname:
I did not know that, it should be helpful in sorting some defaults

the-moriarty:
> @the-moriarty if there is another key for it then I am happy to replace it (forgive me as I don't use i) @meain I have bound a to parameter but do not mind adding another, at the moment there is also the `F` text object for funcion calls but I can add that to the param object as well (or remove `F`

move the parameter from `a` to `A` or make it works like the old `a` (working with argument_list (it would be 10/10)).

move the conditional from `i` to `v`, is free to use.

jeetelongname:
@the-moriarty those text objects hve been rebound now, if you have any more suggestions lemme know

shaunsingh:
Hi, this module errors out if you aren't using the `evil` module, due to the following lines: https://github.com/jeetelongname/doom-emacs/blob/4662722bbb88e1c038c3734db0c5bd1b4f9aa3f1/modules/tools/tree-sitter/config.el#L17-L28

Should this only run if you have `evil` enabled?

jeetelongname:
Thanks! Sorry for that bit of oversight

sei40kr:
FYI I've created a PR that just maps `terraform-mode` to `hcl` syntax in [tree-sitter-langs](https://github.com/emacs-tree-sitter/tree-sitter-langs). ( https://github.com/emacs-tree-sitter/tree-sitter-langs/pull/85 )

peterhoeg:
FYI, `tree-sitter-langs` also has queries for `hcl`, ie the terraform language.

jeetelongname:
I shall add support for that in a bit, I will need to bump the module as well. (there are also some load order issues I need to deal with) 

Thanks for letting me know!

otavioschwanck:
Solidity released a tree-sitter parser, can you add ?

meain:
@otavioschwanck Adding new languages is to be done in https://github.com/emacs-tree-sitter/tree-sitter-langs

jeetelongname:
Yes what meain said, for the moement I am just depending on tree sitter langs but am happy to bump the module when it supports it

modules/editor/fold/README.org
```diff
@@ -20,8 +20,9 @@ marker, indent and syntax-based code folding for as many languages as possible.
 This module provides no flags.
 
 ** Plugins
-+ [[https://github.com/alexmurray/evil-vimish-fold][evil-vimish-fold]]
 + [[https://github.com/matsievskiysv/vimish-fold][vimish-fold]]
++ [[https://github.com/alexmurray/evil-vimish-fold][evil-vimish-fold]]* (=:editor evil=)
++ [[https://github.com/jcs090218/ts-fold][ts-fold]]* (=:tools tree-sitter=)
```
hlissner, line 25: Please move this to the `rewrite-docs` branch (in another PR).

modules/editor/fold/autoload/fold.el
```diff
@@ -32,6 +32,11 @@
             (end-of-line)
             (+fold--hideshow-fold-p))))))
 
+;; NOTE: does this need more?
+(defun +fold--ts-fold-p ()
+  (and tree-sitter-mode
```
hlissner, line 37:
```diff
-(and tree-sitter-mode
+(and (bound-and-true-p tree-sitter-mode)
```

Otherwise, this will throw an error when the tree-sitter module is disabled/before tree-sitter is loaded.

modules/editor/fold/autoload/fold.el
```diff
@@ -101,32 +109,36 @@ Targets `vimmish-fold', `hideshow' and `outline' folds."
...
+        (t (save-excursion
+             (+fold--ensure-hideshow-mode)
+             (if (integerp level)
+                 (progn
+                   (outline-hide-sublevels (max 1 (1- level)))
+                   (hs-life-goes-on
+                    (hs-hide-level-recursive (1- level) (point-min) (point-max))))
+               (hs-show-all)
+               (when (fboundp 'outline-show-all)
+                 (outline-show-all)))))))
```
hlissner, line 125: Get rid of the leading `t`, as it serves no purpose. E.g.

```diff
-        (t (save-excursion
+        ((save-excursion
```

And re-indent the rest of the block.

modules/editor/fold/autoload/fold.el
```diff
@@ -101,32 +109,36 @@ Targets `vimmish-fold', `hideshow' and `outline' folds."
...
-       (hs-hide-all)))))
+    (cond ((+fold--ts-fold-p)
+           (ts-fold-close-all))
+          (t (when (featurep 'vimish-fold)
+               (vimish-fold-refold-all))
+             (+fold--ensure-hideshow-mode)
+             (hs-life-goes-on
+              (if (integerp level)
+                  (hs-hide-level-recursive (1- level) (point-min) (point-max))
+                (hs-hide-all)))))))
```
hlissner, line 141: Use `if` for cases with only two branches. E.g.

```lisp
(if (+fold--ts-fold-p)
    (ts-fold-close-all)
  (when (featurep 'vimish-fold)
    ...)
  ...)
```

modules/editor/fold/autoload/fold.el
```diff
@@ -142,7 +154,24 @@ Targets `vimmish-fold', `hideshow' and `outline' folds."
...
+                    (lambda ()
+                      ;; ts-fold does not define movement functions so we need to do it ourselves
+                      (when (+fold--ts-fold-p)
+                        (let* ((arg-list (if (> count 0) ;; depending on direction we need to change the ranges
+                                             (list (point) (point-max))
+                                           (list (point-min) (point))))
+                               (comp-fun (if (> count 0) ;; also depending on direction we need to change how we sort the list
+                                             #'<
+                                           #'>))
+                               (ovs (seq-filter
```
hlissner, line 167: `seq` is not guaranteed to be loaded in interactive sessions, but `cl-lib` is, so `s/seq-filter/cl-remove-if-not`.

modules/lang/javascript/config.el
```diff
@@ -125,7 +125,17 @@
...
-        (add-to-list 'lsp--formatting-indent-alist '(typescript-tsx-mode . typescript-indent-level)))))
+        (add-to-list 'lsp--formatting-indent-alist '(typescript-tsx-mode . typescript-indent-level))))
+    (when (featurep! +tree-sitter)
+      (after! tree-sitter
+        (pushnew! tree-sitter-major-mode-language-alist '(typescript-tsx-mode . tsx))
+        (pushnew! evil-textobj-tree-sitter-major-mode-language-alist '(typescript-tsx-mode . "tsx"))
+
+        ;; HACK: the tsx grammer doesn't work with the hightlighting provided by
+        ;; font-lock-keywords.
+        ;; https://github.com/emacs-tree-sitter/tree-sitter-langs/issues/23
```
hlissner, line 136: Use short form references:

```diff
        ;; HACK: the tsx grammer doesn't work with the hightlighting provided by
-       ;; font-lock-keywords.
-       ;; https://github.com/emacs-tree-sitter/tree-sitter-langs/issues/23
+       ;; font-lock-keywords. See emacs-tree-sitter/tree-sitter-langs#23
```

modules/lang/javascript/config.el
```diff
@@ -309,3 +319,15 @@ to tide."
 
 (def-project-mode! +javascript-gulp-mode
   :when (locate-dominating-file default-directory "gulpfile.js"))
+
+;; Tree sitter
+(eval-when! (featurep! +tree-sitter)
+  (add-hook! '(javascript-mode-hook
+               js-mode-hook
+               js2-mode-hook
+               js3-mode-hook
```
hlissner, line 328: Remove `js3-mode-hook` and `javascript-mode-hook` from this list. 

- js3-mode is not our jurisdiction (none of our modules install it), so best we don't enforce our opinions on it in case the user installs it.
- javascript-mode is an alias for js-mode.

modules/lang/sh/config.el
```diff
@@ -24,18 +24,18 @@
...
+                  ;; Types
+                  :true "true" :false "false"
+                  ;; Flow
+                  :not "!"
+                  :and "&&" :or "||"
+                  :in "in"
+                  :for "for"
+                  :return "return"
+                  ;; Other
+                  :dot "." :dot "source")
```
hlissner, line 38: Unwarranted formatting. Please revert this.

modules/tools/tree-sitter/config.el
```diff
@@ -0,0 +1,78 @@
...
+      ;; ts goto nodes
+      (evil-define-key 'normal '+tree-sitter-keys-mode
+        "[g" +tree-sitter-goto-previous-map
+        "]g" +tree-sitter-goto-next-map)
+      keymap)
+    "Basic keymap for tree sitter text objects")
+
+  (define-minor-mode +tree-sitter-keys-mode
+    "A minor mode with tree sitter keybinds."
+    :keymap +tree-sitter-keys-mode-map))
```
hlissner, line 33: There's a level of indirection here I'm not sure is necessary. Please bind these onto `tree-sitter-mode-map` instead. If this was done to make it easier to disable, the `:editor evil +everywhere` check is enough for now. Plus, these few keybinds are already easily reversed if not desired.

hlissner, line 33: As these are part of the module's evil configuration, can they be moved into the `:config` of `evil-textobj-tree-sitter`?

hlissner (changes requested):
Looking good! Besides the smaller nitpicks I've pointed out in my review, here are three larger-scale issues I'd like to see resolved:

1. Adding `turn-on-tree-sitter-mode` and `+tree-sitter-keys-mode` to mode hooks is repeated across many modes. Like the `:tools lsp` module, this would be better wrapped up in [an autodef's hook function](https://github.com/hlissner/doom-emacs/blob/master/modules/tools/lsp/autoload/common.el#L4-L10). For example:

   ```lisp
   ;; in modules/tools/tree-sitter/autoload.el

   ;;;###autodef
   (defun tree-sitter! ()
     (turn-on-tree-sitter-mode)
     (+tree-sitter-keys-mode +1))
   ```

   This would give us more freedom to modify/refactor this later.

2. Instead of adding `tree-sitter!` to `MAJOR-MODE-hook`, add it to `MAJOR-MODE-local-vars-hook` instead. This hook won't trigger for trivial cases (e.g. when org exports src blocks, or packages turn on a major mode in a temp buffer for parsing, etc). 

3. Please move changes to module documentation (except for the tree sitter module's README.org and `docs/modules.org`) to the `rewrite-docs` branch, in another PR. Or at the very least, revert them for now, and they can be done later, once that branch has been merged (won't be long now).

modules/tools/tree-sitter/README.org
```diff
@@ -0,0 +1,126 @@
...
+fixed by either adding in a custom query (which would override the current key
+bound.) or [[https://github.com/nvim-treesitter/nvim-treesitter-textobjects/][contributing upstream!]]
+
+** =No 'TEXTOBJ' text object found=
+the main reason for this is the underlying text object using the =#make-range!=
+predicate, which at the moment [[https://github.com/emacs-tree-sitter/elisp-tree-sitter/issues/180][is not implemented in emacs tree sitter]] (see this
+issue on [[https://github.com/meain/evil-textobj-tree-sitter/issues/33][evil-textobj-tree-sitter]]).
+
+the only way around it is to rewrite the query to not use =#make-range!= or to
+implement that predicate the elisp tree sitter core
```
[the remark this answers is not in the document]
jeetelongname: This should actually be redundent now, ill remove it.

jeetelongname:
Now all I need to do is figure out these load order issues and probably some other stuff

jackmac92:
I'm eager to start working with tree-sitter, and planning to merge this PR in my local doom. I understand it's not done released yet, but can provide a backtrace if I run into a bug

let me know if there's any particular area where testing would be helpful!

jeetelongname:
Please do! I am currently working on figuring out how its being eagar loaded (which it should not) so if you do find out whats causing that it would be a great help!

bhankas:
I only skimmed the code so forgive me if I missed something obvious, but at tree-sitter grammar exists for couple of languages that I use and I see it missing from changes namely:

1. Nix: https://github.com/cstrahan/tree-sitter-nix
2. Beancount: https://github.com/polarmutex/tree-sitter-beancount

We already have LSP flags for both of these modules, and tree-sitter will be nice addition. Will it be possible to include them in this PR?

jeetelongname:
At the moment grammers are only installed through tree sitter langs, I have not added in my own grammers tho it is possible, I would recommend adding them to tree sitter langs instead of doom installing them directly.

That being said I will mess with it and write up some instructions on how to do so for people

bhankas:
If you mean this repo: https://github.com/emacs-tree-sitter/tree-sitter-langs/tree/master/repos

It does not have beancount, but does have Nix.

Either way, instructions will be fine too. Thanks for taking time to implement this!

jeetelongname:
if it does and it comes in the next bump ill add nix into the fold. then again I might have missed something so I will check

VitalyAnkh:
There is a [feature/tree-sitter](https://git.savannah.gnu.org/cgit/emacs.git/log/?h=feature/tree-sitter) branch in the emacs repository and Doom could make use of this in the future.

jeetelongname:
I am aware, I am waiting for it to mature and or the ecosystem to begin to move over before I even consider moving over the module

jeetelongname:
That should be the final thing, Lets see, I hope to get this merged in the next couple of days / weeks. thanks again to everyone who has helped and commented. Your awesome!

modules/tools/tree-sitter/config.el
```diff
@@ -0,0 +1,66 @@
...
+         "i" (+tree-sitter-goto-textobj "conditional.outer" t)
+         "l" (+tree-sitter-goto-textobj "loop.outer" t))
+        (:map +tree-sitter-goto-next-map
+         "a" (+tree-sitter-goto-textobj "parameter.outer")
+         "f" (+tree-sitter-goto-textobj "function.outer")
+         "F" (+tree-sitter-goto-textobj "call.outer")
+         "C" (+tree-sitter-goto-textobj "class.outer")
+         "c" (+tree-sitter-goto-textobj "comment.outer")
+         "i" (+tree-sitter-goto-textobj "conditional.outer")
+         "l" (+tree-sitter-goto-textobj "loop.outer")))
```
gagbo, line 59: The navigation bindings aren't mentioned in the README in a nice table like the text-objects. Just marking this quickly not to forget, we can PR it later anyway (especially if you want to change those)

modules/tools/tree-sitter/config.el
```diff
@@ -0,0 +1,66 @@
...
+         "C" (+tree-sitter-get-textobj "class.inner")
+         "v" (+tree-sitter-get-textobj "conditional.inner")
+         "l" (+tree-sitter-get-textobj "loop.inner"))
+        (:map +tree-sitter-outer-text-objects-map
+         "A" (+tree-sitter-get-textobj '("parameter.outer" "call.outer"))
+         "f" (+tree-sitter-get-textobj "function.outer")
+         "F" (+tree-sitter-get-textobj "call.outer")
+         "C" (+tree-sitter-get-textobj "class.outer")
+         "c" (+tree-sitter-get-textobj "comment.outer")
+         "v" (+tree-sitter-get-textobj "conditional.outer")
```
meain, line 41: Just a small nitpick. As of now `conditional.outer` uses `v` as mapping when selecting a textobj and `i` in case of goto. It would be better if they were consistent.

jeetelongname: I was thinking that, I did not change it now mainly because I set it a while ago and did not feel like making a breaking change without some kind of discussion. 

If I were to make a case for keeping it. in practice the usage is so different it would not matter if they have slightly different semantics. I think being closer to mnemonics where I can (I can't use `i` as a text object but can for navigation) takes precedent over consistency but I can also see the other side.

meain: I think it would be a good idea to have a discussion around it since we are in the last stages. Plus as this is not yet merged in, I don't think it would be that bad to change the defaults. The way I see it, a specific key is bound to a textobj, for example `f` for `function` and now any operation that you want to do on a `function` uses that key.

I might be missing something here, but why exactly can't we use `i` for nav?

*In any case I don't use doom, so I'll let the final decision up to you*

jeetelongname: Yes I agree, I am but one man and my defaults may not represent sane defaults.

I can see the way you think, and for the most part I don&rsquo;t want to break that convention, its why I have tried to keep them both mnemonic and consistent, but (and this is my opinion) trying to keep them mnemonic is more important, its why I did use `i` for nav instead of `v`. I can&rsquo;t use `i` for the text object because *apparently* its a very important and useful text object (I have genuinely never used it). I can use it for nav as its a blank slate and as I have mentioned I find that navigation and text object use very different so I am not sure the semantics carry.

That being said I can also see it the other way in so far that you said, all operations with some kind of structure should use the same key.

hpfr: The way I see it, users will have to internalize the anti-mnemonic `v` binding anyway, so using `i` only when available only increases complexity and isn't worth the mnemonic benefit.  If you're choosing between learning two bindings and one of the two, I think you should go with the one, even if it's the more difficult of the two, since you still come out ahead of learning two.

jeetelongname: thats a good point, one final breaking change for the road then I guess.

hlissner (approved):

hlissner:
And done! Great work @jeetelongname. And thank you to everyone for your help shaping this module with your input and reviews (and putting up with my delays). I'm happy with how it turned out, and look forward to putting it through its paces!

jeetelongname:
tsx mode should be supported (and I have not heard any complaints in that regard). The reason I did not enable it for web mode was because while the same method would work, its a worse experience overall imo, both enabling ts in web-mode or defaulting to html-mode that is. It should be trivial to switch over though.

mohammedzeglam-pg:
> * playground on tree sitters website

i said i was use this hack and maybe the experience not perfect because tree-sitter not support injection so you do not get highlight for style  or  script tag but but i love the coloring from tree-sitter and it was the reason number one to use neovim back, but it think about writing new package to support this work

PR #5402, opened 2021-08-20
docs: Fix broken link to org-roam v2 in org readme

erwald:
The link to org-roam v2 in the org readme is broken. This pr fixes that.

### DIFF
--- a/modules/lang/org/README.org
+++ b/modules/lang/org/README.org
@@ -136,7 +136,7 @@ https://www.mfoot.com/blog/2015/11/22/literate-emacs-configuration-with-org-mode
 + =+roam=
   + [[https://github.com/org-roam/org-roam-v1][org-roam]] (v1)
 + =+roam2=
-  - [[https://github.com/org-roam/org-roam/tree/v2][org-roam]] (v2)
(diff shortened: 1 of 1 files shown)


iyefrat:
the workflow won't run because you're a first time contributor (\:D), but you should change your commit to something like:

`docs(org): fix link to org-roam v2 in readme`

erwald:
> the workflow won't run because you're a first time contributor (:D), but you should change your commit to something like:
> 
> `docs(org): fix link to org-roam v2 in readme`

dang, i thought i got it right but the document was pretty complex! should i close this one and open a new pr with a better commit title?

iyefrat:
You don't need to, you can just overwrite the commit and force push. If you're using `magit`, you can do this by pressing `c w` on the magit status screen which will let you edit the commit message of the last commit, and then you can do `p -f p` to force push to `pushRemote` (which is probably this branch for you)

erwald:
done, thanks!

hlissner (approved):

hlissner:
Sorry for the late response. Thanks for the PR!

PR #5406, opened 2021-08-22
Handle active region properly in `+default/search-buffer` when using vertico

saevarb:
If a region is selected, it is not passed to `consult-line` as the documentation of `+default/search-buffer` indicates it should

### DIFF
--- a/modules/config/default/autoload/search.el
+++ b/modules/config/default/autoload/search.el
@@ -31,8 +31,13 @@ If a selection is active, pre-fill the prompt with it."
           (if (region-active-p)
               #'swiper-isearch-thing-at-point
             #'swiper-isearch))
-         ((featurep! :completion vertico) #'consult-line))))
-
+         ((featurep! :completion vertico)
+          (if (region-active-p)
+              (let ((start (region-beginning))
+                    (end   (region-end)))
+                (deactivate-mark)
+                (consult-line (buffer-substring-no-properties start end)))
+            #'consult-line)))))
 ;;;###autoload
 (defun +default/search-project (&optional arg)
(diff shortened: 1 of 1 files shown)


modules/config/default/autoload/search.el
```diff
@@ -31,7 +31,10 @@ If a selection is active, pre-fill the prompt with it."
           (if (region-active-p)
               #'swiper-isearch-thing-at-point
             #'swiper-isearch))
-         ((featurep! :completion vertico) #'consult-line))))
+         ((featurep! :completion vertico)
+          (if (region-active-p)
+              (consult-line (buffer-substring (mark) (point)))
+            #'consult-line)))))
```
iyefrat, line 37:
```suggestion
         ((featurep! :completion vertico)
          (if (region-active-p)
           (let ((start (region-beginning))
                 (end   (region-end)))
                 (deactivate-mark)
              (consult-line (buffer-substring-no-properties start end))
            #'consult-line)))))
```
This fixes two things:
1. If the mark isn't deactivated, the region expands with the cursor when switching candidates.
2. Doesn't copy the text properties of the region, to fit with manually inputted text.

saevarb: Thanks! I figured the way I was handling the region wasn't quite correct. I have included your improvements and force-pushed.

hlissner:
Sorry for the late response. I've  fixed this in 6b5c34a (with credit) because of an error in this PR's implementation: `consult-line` returns a positional integer, which would cause `call-interactively` to throw a `wrong-type-argument` error. In any case, thanks for the help!

PR #5415, opened 2021-08-25
feat(scheme): auto-add guix channels to load-path

gagbo:
Guix is detected with a simple `executable-find`, and this trick assumes
that guix is running on version 3 of Guile (which hopefully won't
change for a while)

ping @flatwhatson in case I'm doing something horrible


hlissner (approved):

hlissner:

PR #5417, opened 2021-08-25
fix(dired): add dired-mode to `evil-snipe-disabled-modes`

### LINKED ISSUE #3359: evil-sniper overrides S / dired-do-symlink
Should be able to symlink in dired using S.

**What actually happened?**
Pressing S enters evil-snipe.

**Additional details:**
I'm not sure if this is intended or not. I think that we should prefer the dired keybinding in this case, as sniping is not so necessary in a dired buffer.

ArtemSmaznov:
Fixes #3359 

Evil binding overwrite S binding in dired (create symlink) with Snipe's search backwards. Arguably, the symlink one is more useful and I don't see much use for snipe in dired buffers.
Open to discussion if others disagree.

### DIFF
--- a/modules/editor/evil/config.el
+++ b/modules/editor/evil/config.el
@@ -356,7 +356,7 @@ directives. By default, this only recognizes C directives.")
         evil-snipe-repeat-scope 'visible
         evil-snipe-char-fold t)
   :config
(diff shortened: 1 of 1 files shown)


hlissner:
Sorry for the tremendously late response. Thanks for the help!

PR #5418, opened 2021-08-26
feat(julia): Adds support for Julia Snail

kiranshila:
[Julia snail](https://github.com/gcv/julia-snail) is an alternative development setup for Julia that takes inspiration from CIDER/SLIME. I find it much easier to use than vanilla `julia-mode` and I don't get hit with too much context switching going from Clojure to Julia.

I've added a `+snail` flag to the julia module that enables `julia-snail-mode` in julia buffers and associated the matching keybinds from CIDER, which replaces `julia-repl`. The module still uses `julia-mode` as provided by [julia-emacs](https://github.com/JuliaEditorSupport/julia-emacs), so I assume I leave in those `use-package!` statements.

The only bits I'm not sure about is how enforce a requirement for `vterm`, as that's what snail uses as it's REPL. I added a check in `doctor.el`. Additionally, I'm not quite sure how to spawn a REPL that is detached from a julia source buffer, so I feature-gated the autoload `+julia/open-repl` - I'm not sure if that was the right approach and would appreciate feedback.

This new feature is of course opt-in, and doesn't change any base julia module functionality.

Unrelated to providing the snail feature, I also added the hint to mark folders with julia's Project.toml as projectile projects.

### DIFF
--- a/modules/lang/julia/README.org
+++ b/modules/lang/julia/README.org
@@ -8,27 +8,34 @@
   - [[#module-flags][Module Flags]]
   - [[#plugins][Plugins]]
 - [[#prerequisites][Prerequisites]]
+  - [[#julia-snail][Julia Snail]]
   - [[#language-server][Language Server]]
     - [[#lsp-julia][~lsp-julia~]]
     - [[#eglot-jl][~eglot-jl~]]
 - [[#features][Features]]
+  - [[#julia-snail-1][Julia Snail]]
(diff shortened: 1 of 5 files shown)


kiranshila:
Looking more into julia-emacs, many of the features are getting added in, especially with https://github.com/wentasah/EmacsVterm.jl
I'll close this for now, but perhaps we should discus adding in doom-friendly keybindings to juila-mode

PR #5423, opened 2021-08-27
fix(evil): +evil--embrace-escaped isn't active for the first org mode buffer

HyunggyuJang:
For the first buffer by which `evil-embrace` is loaded, `+evil--embrace-escaped` isn&rsquo;t applied. That is, surrounding text object, say `blah` with backslash just result to `\blah\`, contrary to waiting the very next character triggered by `+evil--embrace-escaped`.

This is because the entry associated with `\` doesn&rsquo;t exist when it comes to the default value of `embrace--pairs-list`, hence, from the very *next* buffer the tweak made to `embrace--pairs-list` takes effect.

### DIFF
--- a/modules/editor/evil/config.el
+++ b/modules/editor/evil/config.el
@@ -268,6 +268,7 @@ directives. By default, this only recognizes C directives.")
          . +evil-embrace-angle-bracket-modes-hook-h)
   :hook (scala-mode . +evil-embrace-scala-mode-hook-h)
   :init
+  (defvar embrace--pairs-list '((?\\ . nil)))
   (after! evil-surround
     (evil-embrace-enable-evil-surround-integration))
   :config


hlissner:
I can't reproduce this issue. Can you provide some concrete repro steps?

HyunggyuJang:
Oh, sorry about that. 
1. Start new Doom instance
2. Open any *org* file 
3. Try to surround with escape key

Then you'll see what I posed.

I realized that, so far, this kind of behavior only occurs to org mode buffers, not others. 

Seems like due to the syntax of escape character, `\`, is remapped to symbol in org syntax table, hence this happens.

Thanks for your elaboration.

hlissner:
Sorry for the _super_ late response, but this should now be fixed, as of cb6b065. Thanks for bringing it to my attention!

PR #5428, opened 2021-08-28
Revert "Add set -euo pipefail to sh-mode file template"

samhedin:
This reverts commit ce2fee138a7eef2d01e67e555654bba934623297.

As of https://github.com/hlissner/doom-emacs/pull/2906 `set -euo pipefail` is added to any newly created `.sh` script. This snippet made my bash behave strangely after I used doom emacs to create a new `.bashrc` file.
Examples of what happens when `.bashrc` starts with the snippet:
```
cd Doc<tab> -> terminal crash without error message, despite Documents existing.
cat Documents -> terminal crash without error message. This command makes no sense since Documents is a directory, but it normally outputs the error message "cat: Documents: Is a directory"
```
This had me confused for a while and I thought my OS was at fault before I realized what was going on. I'm sure this snippet is reasonable in most cases, but since it's added automatically I would expect it to never cause issues like this.
Thank you for doom emacs!

### DIFF
--- a/modules/editor/file-templates/templates/sh-mode/__
+++ b/modules/editor/file-templates/templates/sh-mode/__
@@ -1,4 +1,3 @@
 #!/usr/bin/env `(if (equal (file-name-extension buffer-file-name) "zsh") "zsh" "bash")`
-set -euo pipefail
 
 $0


Mic92 (approved):

Mic92 (commented):
is there a way to detect `.bashrc`?

samhedin:
> is there a way to detect `.bashrc`? Ignore my approval above, I thought it was about to be added instead of removed.

I assume there is, and the thought did cross my mind.  The tricky thing would be to make an exhaustive list of files where this snippet needs to be ignored. If it happens for `.bashrc` it probably happens in `.zshrc`, and in <many other places that I can't think of>.
Maybe the easiest way is to not include it in any dotfiles? *Maybe* that would cover all cases where it can lead to trouble?

hlissner:
How about changing the template to this?

```snippet
#!/usr/bin/env `(if (equal (file-name-extension buffer-file-name) "zsh") "zsh" "bash")`
`(unless (string-prefix-p "." (file-name-base buffer-file-name))) "set -euo pipefail\n"`
$0
```

samhedin:
Let's try it!

samhedin:
Based on https://lists.gnu.org/archive/html/help-bash/2020-04/msg00049.html from https://github.com/hlissner/doom-emacs/pull/5508 it seems like maybe the previous commit was more reasonable - and that this line should be removed altogether?

hlissner:
Yes, please remove it. I'll create a snippet for it in https://github.com/hlissner/doom-snippets instead.

hlissner:
Thanks for your help!

PR #5429, opened 2021-08-29
fix(snippets): yas-expand error when yas-choose-value

HyunggyuJang:
`+snippets-prompt-private` is being overzealous that intercepts the candidates triggered by `yas-choose-value`.

For example, triggering `jupyter` in org mode, renders
#+begin_src jupyter-Wrong type argument: yas--template, "python" :session  :async yes

#+end_src

### DIFF
--- a/modules/editor/snippets/autoload/snippets.el
+++ b/modules/editor/snippets/autoload/snippets.el
@@ -20,7 +20,9 @@ If there are conflicting keys across the two camps, the built-in ones are
 ignored. This makes it easy to override built-in snippets with private ones."
   (when (eq this-command 'yas-expand)
     (let* ((gc-cons-threshold most-positive-fixnum)
-           (choices (cl-remove-duplicates choices :test #'+snippets--remove-p)))
+           (choices (condition-case _
+                        (cl-remove-duplicates choices :test #'+snippets--remove-p)
+                      (wrong-type-argument choices))))
       (if (cdr choices)
           (cl-loop for fn in (cdr (memq '+snippets-prompt-private yas-prompt-functions))
                    if (funcall fn prompt choices display-fn)


modules/editor/snippets/autoload/snippets.el
```diff
@@ -20,7 +20,9 @@ If there are conflicting keys across the two camps, the built-in ones are
 ignored. This makes it easy to override built-in snippets with private ones."
   (when (eq this-command 'yas-expand)
     (let* ((gc-cons-threshold most-positive-fixnum)
-           (choices (cl-remove-duplicates choices :test #'+snippets--remove-p)))
+           (choices (condition-case nil
```
hlissner, line 23: Replace `nil` with `_`. This is the convention for binding unused variables.

hlissner (changes requested):
Please wrap the body of your commit message to 72 characters (to conform to our [git conventions](https://gist.github.com/hlissner/4d78e396acb897d9b2d8be07a103a854)). e.g.

```
fix(snippets): yas-expand error when yas-choose-value

`+snippets-prompt-private` is being overzealous that intercepts the candidates
triggered by `yas-choose-value`.

For example, triggering `jupyter` in org mode, renders
#+begin_src jupyter-Wrong type argument: yas--template, "python" :session :async yes

#+end_src
```

Don't have to worry about line length for the #+begin_src line however.

modules/editor/snippets/autoload/snippets.el
HyunggyuJang: Thanks for review! I've amended.

hlissner (approved):

hlissner:
Thanks for your help!

PR #5430, opened 2021-08-29
fix(cc): remove double lenses with lsp

gagbo:
On current pins, ccls and lsp code lenses aren't mutually exclusive, and there is actually 2 sets of code lenses appearing

Unsure but might solve [this](https://discord.com/channels/406534637242810369/406554085794381833/881461815370973194)


gagbo:
This seems safe to merge, as more occurrences of double lenses are happenening

floyza:
@hlissner It appears that since #4233, mainline `lsp-lens` has the lenses displayed inline like `ccls-lens-mode`, so dropping `ccls-lens-mode` in this PR is probably correct.

modules/lang/cc/config.el
```diff
@@ -276,7 +276,6 @@ If rtags or rdm aren't available, fail silently instead of throwing a breaking e
 (use-package! ccls
   :when (featurep! +lsp)
   :unless (featurep! :tools lsp +eglot)
-  :hook (lsp-lens-mode . ccls-code-lens-mode)
```
hlissner, line 279: Add `:defer t` to avoid eagerly loading the package.

hlissner (changes requested):

hlissner (approved):

hlissner:
Thanks again for your help!

PR #5431, opened 2021-08-29
feat(vc): use smerge-vc-next-conflict instead of smerge-next

michaelpj:
This is provided by `smerge-mode` in Emacs 27, and functions exactly
like `smerge-next`, except that if there are no more conflicts in the
current file, it will go to the next file with conflicts.

Arguably this could be a different key binding. However, I find that it
is almost strictly superior to `smerge-next`: I can't think of a case
where I *wouldn't* immediately want to go to the next file with
conflicts, and even if you do by accident you can just pop back to the
preceding buffer. So I think it warrants being the default.

-------

I am aware of https://github.com/hlissner/doom-emacs/pull/4788, which also adds a binding for this. However, I hope this PR will be more acceptable: it doesn't add anything new, but instead modifies the hydra keybinding. 

Fixes #0000 

### DIFF
--- a/modules/emacs/vc/autoload/hydra.el
+++ b/modules/emacs/vc/autoload/hydra.el
@@ -20,7 +20,7 @@
 "
     ("g" (progn (goto-char (point-min)) (smerge-next)))
     ("G" (progn (goto-char (point-max)) (smerge-prev)))
-    ("C-j" smerge-next)
+    ("C-j" smerge-vc-next-conflict)
     ("C-k" smerge-prev)
     ("j" next-line)
     ("k" previous-line)


michaelpj:
I still care about this but I'd be happy with https://github.com/hlissner/doom-emacs/pull/5743

iyefrat:
boop

hlissner:
Sorry for taking _so_ long to get to this; my hesitation has to do with hydra, whose future in Doom is a little uncertain at the moment. For the time being I've merged this into bab4c92 -- I understand there may be recentering issues, but I think that resolve itself if/when I change Doom's default value for `scroll-conservatively` (still experimenting with it). In any case, thanks for bringing this to my attention!

michaelpj:
Thanks!

daanturo:
Hello, currently after merging this we have `smerge-vc-next-conflict` bound to both `C-j` and `n` (the latter has a little manual window re-centering) in the hydra. That's a little confusing now.

Issue #4323, opened 2020-11-26
emacs native comp parinfer-mode not work


locez:
**What did you expect to happen?**

```
M-x: parinfer-mode
```

**What actually happened?**
I got this error. The cursor will jump to the first line, and then all input will be on the first line

```
parinfer-ext::smart-tab:mount: Invalid function: evil-define-key
```

**Additional details:**

```
Doom v2.0.9 (HEAD -> develop e4563e9b0 2020-11-25 15:29:45 -0500)
gccemacs latest commit 
```

- Include a link to your private config
- Include screenshots/casts of your issue
![image](https://user-images.githubusercontent.com/9402092/100339265-9d366900-3014-11eb-90ca-47d4dbd6737c.png)

- If you mention key sequences, include what commands they're bound to (use `SPC
  h k KEY` or `C-h h k KEY` to inspect keys).
- Include any warnings or errors logged to \*Messages\* (use `M-x
  view-echo-area-messages` to see it).

<details><pre>
If an error message is involved include a backtrace of it.

How to acquire a backtrace:
https://github.com/hlissner/doom-emacs/blob/develop/docs/getting_started.org#how-to-extract-a-backtrace-from-an-error
</pre></details>

**Steps to reproduce:**
1. Select these example steps,
2. Delete them,
3. And replace them with precise steps to reproduce your issue.
4. Fill in "system information" below.

**System information:**
<details><pre>
Place the output of `M-x doom/info` or `~/.emacs.d/bin/doom info` here.
```
SYSTEM  type       gnu/linux
        config     x86_64-pc-linux-gnu
        shell      /bin/fish
        uname      Linux 5.7.3-gentoo-locez #4 SMP Sat Nov 7 14:50:28 CST 2020 x86_64
        path       (~/.local/bin ~/.cargo/bin/ ~/go/bin ~/.rvm/bin /usr/local/sbin /usr/local/bin /usr/sbin /usr/bin /sbin /bin /opt/bin /usr/lib/llvm/10/bin /usr/libexec/emacs/28.0.50/x86_64-pc-linux-gnu)
EMACS   dir        ~/.emacs.d/
        version    28.0.50
        build      Nov 26, 2020
[...]

PR #5444, opened 2021-09-02
Improve `:editor parinfer`

fosskers:
As brought up initially in the Discord, this PR modernizes the `parinfer` module by:

- [x] Unbreaking the expected binding of `macrostep-expand` to `SPC m m`.
- [x] Removing the deprecated `parinfer-mode` in favour of the Rust variant (current invoked by `+rust`).
- [x] Modernizing the module README

**Note:** This is a fairly breaking change, as it actually assumes that the user's Emacs was compiled with `--with-modules`. However, given that Doom requires Emacs 27 at minimum and that the major distributions likely ship with an Emacs built with `--with-modules` (I haven't checked), should we be concerned?

### DIFF
--- a/modules/editor/parinfer/config.el
+++ b/modules/editor/parinfer/config.el
@@ -22,8 +22,7 @@
         :i "<tab>"     #'parinfer-smart-tab:dwim-right-or-complete
         :i "<backtab>" #'parinfer-smart-tab:dwim-left
         :localleader
-        "m" #'parinfer-toggle-mode))
-
+        "p" #'parinfer-toggle-mode))
 
 (use-package! parinfer-rust-mode
   :when (featurep! +rust)
@@ -45,5 +44,5 @@
   :config
   (map! :map parinfer-rust-mode-map
         :localleader
-        "m" #'parinfer-rust-switch-mode
-        "M" #'parinfer-rust-toggle-disable))
+        "p" #'parinfer-rust-switch-mode
+        "P" #'parinfer-rust-toggle-disable))
--- a/modules/editor/parinfer/config.el
+++ b/modules/editor/parinfer/config.el
@@ -1,31 +1,6 @@
 ;;; editor/parinfer/config.el -*- lexical-binding: t; -*-
 
-(use-package! parinfer
-  :unless (featurep! +rust)
-  :hook ((emacs-lisp-mode
-          clojure-mode
-          scheme-mode
-          lisp-mode
-          racket-mode
-          hy-mode) . parinfer-mode)
-  :init
-  (setq parinfer-extensions
-        '(defaults
-          pretty-parens
-          smart-tab
-          smart-yank))
-  (when (featurep! :editor evil +everywhere)
-    (push 'evil parinfer-extensions))
-  :config
-  (map! :map parinfer-mode-map
-        "\"" nil  ; smartparens handles this
-        :i "<tab>"     #'parinfer-smart-tab:dwim-right-or-complete
-        :i "<backtab>" #'parinfer-smart-tab:dwim-left
-        :localleader
-        "p" #'parinfer-toggle-mode))
-
 (use-package! parinfer-rust-mode
-  :when (featurep! +rust)
   :when (bound-and-true-p module-file-suffix)
   :hook ((emacs-lisp-mode
           clojure-mode
--- a/modules/editor/parinfer/doctor.el
+++ b/modules/editor/parinfer/doctor.el
@@ -3,8 +3,5 @@
 (unless (fboundp 'module-load)
   (warn! "Your emacs wasn't built with dynamic modules support. `parinfer-rust-mode' won't work"))
 (when (and (eq system-type 'berkeley-unix)
-           (not (file-readable-p (concat user-emacs-directory
-                                         ".local/etc/parinfer-rust/libparinfer_rust.so"))))
-  (warn! (concat "Could not read " user-emacs-directory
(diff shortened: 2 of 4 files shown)


modules/editor/parinfer/config.el
```diff
@@ -45,5 +44,5 @@
   :config
   (map! :map parinfer-rust-mode-map
         :localleader
-        "m" #'parinfer-rust-switch-mode
-        "M" #'parinfer-rust-toggle-disable))
+        "p" #'parinfer-rust-switch-mode
+        "P" #'parinfer-rust-toggle-disable))
```
fosskers, line 48: `p` for "parinfer". Currently otherwise unclaimed.

iyefrat:
> `feat(parinfer): Drop support for parinfer-mode`

I think this should be ` refactor!(parinfer): Drop support for parinfer-mode`, since it's a breaking change and a refactor of existing code.

fosskers:
Reasonable, I'll fix the commit message.

iyefrat:
Looks good, a few points of feedback:
1. the commit that adds the README should be a `docs` not a `refactor`
2. The `#+DATE` tag of when the module was first added should stay the same

fosskers:
Done.

modules/editor/parinfer/doctor.el
```diff
@@ -1,11 +1,10 @@
...
-             (not (file-readable-p (concat user-emacs-directory
-                                           ".local/etc/parinfer-rust/libparinfer_rust.so"))))
-    (warn! (concat "Could not read " user-emacs-directory
-                   ".local/etc/parinfer-rust/libparinfer_rust.so. "
-                   "`parinfer-rust-mode' won't work"))))
+(unless (fboundp 'module-load)
+  (warn! "Your emacs wasn't built with dynamic modules support. `parinfer-rust-mode' won't work"))
+(when (and (eq system-type 'berkeley-unix)
+           (not (file-readable-p (concat user-emacs-directory
+                                         ".local/etc/parinfer-rust/libparinfer_rust.so"))))
```
Mic92, line 7: Would be nice to have an environment variable here or any other way to make this configurable and install it with a system package manager.

fosskers: At current it's downloaded for you, and presumable updated when `parinfer-rust` itself updates.

Mic92: That might not work on NixOS if the shared library has dependencies beyond what emacs already has and I also prefer not download random binaries from the internet.

Mic92: It looks like it can be configured: https://github.com/justinbarclay/parinfer-rust-mode#step-2-configure-parinfer-rust-mode
Would this check not be incorrect if a user provide a different path?

Mic92: Just checked, the library should work like this on NixOS/Guix:

```
ldd .emacs.d/.local/etc/parinfer-rust/parinfer-rust-linux.so
ldd: warning: you do not have execution permission for `/home/joerg/.emacs.d/.local/etc/parinfer-rust/parinfer-rust-linux.so'
        linux-vdso.so.1 (0x00007ffff4b2a000)
        libdl.so.2 => /nix/store/9bh3986bpragfjmr32gay8p95k91q4gy-glibc-2.33-47/lib/libdl.so.2 (0x00007fbd41eaa000)
        librt.so.1 => /nix/store/9bh3986bpragfjmr32gay8p95k91q4gy-glibc-2.33-47/lib/librt.so.1 (0x00007fbd41e9f000)
        libpthread.so.0 => /nix/store/9bh3986bpragfjmr32gay8p95k91q4gy-glibc-2.33-47/lib/libpthread.so.0 (0x00007fbd41e7f000)
        libgcc_s.so.1 => /nix/store/9bh3986bpragfjmr32gay8p95k91q4gy-glibc-2.33-47/lib/libgcc_s.so.1 (0x00007fbd41e65000)
        libc.so.6 => /nix/store/9bh3986bpragfjmr32gay8p95k91q4gy-glibc-2.33-47/lib/libc.so.6 (0x00007fbd41ca0000)
        /nix/store/9bh3986bpragfjmr32gay8p95k91q4gy-glibc-2.33-47/lib64/ld-linux-x86-64.so.2 (0x00007fbd4215f000)

```

fosskers: Let me check. Or do you otherwise have a suggested change?

Mic92, line 7: I would say just using, should fix it:

```suggestion
           (not (file-readable-p parinfer-rust-library)))
```

modules/editor/parinfer/doctor.el
```diff
@@ -1,11 +1,10 @@
...
+  (warn! (concat "Could not read " user-emacs-directory
+                 ".local/etc/parinfer-rust/libparinfer_rust.so. "
```
Mic92, line 9:
```suggestion
  (warn! (concat "Could not read " parinfer-rust-library "."
```

fosskers:
Thanks @Mic92 I will review that on Thursday or Friday (busy day Wednesday).

fosskers:
Okay this should be ready to go.

iyefrat:
I think that this pr closes #4323, doesn't it?

fosskers:
I should think so.

iyefrat:
In that case you should probably add `Fix #4323` to the footer of the commit that drops parinfer support.

fosskers:
Updated.

hlissner (approved):

hlissner:

PR #5446, opened 2021-09-03
broken link

EliMor:
{{{ very small change to getting started doc, seemed like the intention was to make a link? }}}


hlissner:
Thanks for the PR, but I must refuse it. There is nothing wrong with your PR, but it comes at a bad time: I'm about a week away from pushing a rework of all our documentation, which will indirectly fix this issue, and would swallow up your PR anyway, I'll simply wait for that update instead. In any case, thanks for your help!

PR #5450, opened 2021-09-04
Fix extra trailing brace after haskell comment

### LINKED ISSUE #5448: haskell: creating a multiline comment adds an extra } at the end

i tried to enter a multiline comment on a new line in a haskell-mode buffer, by entering the opening delimter `{-`. i expected to see the closing delimiter `-}` automatically being filled in, with my cursor placed in between the two.

### What actually happened?

an _extra_ `}` is entered, so that the line has the text `{--}}`. my cursor is correctly placed, however, between the two `-` characters. 

to me it seems like the normal "smartly insert closing paren" logic and the haskell-specific "smartly insert closing multiline comment delimiter `-}`" are both firing in this instance.

### Steps to reproduce
 
1. open a haskell source file
2. hit `o` to enter insert mode on a new line
3. hit `{` 
4. hit `-`

_No response_

### System Information

https://pastebin.com/6Xs6qzXR

celsobonutti:
Fixes  #5448

I've added a condition to the smartparens braces expansion, not including "{" if we're in `haskell-mode`.
I'm not much of an elisp user, so it may be very wrong. I would be very happy to make it better if anyone has a suggestion :-)

### DIFF
--- a/modules/config/default/config.el
+++ b/modules/config/default/config.el
@@ -205,7 +205,7 @@
       (map! :map markdown-mode-map
             :ig "*" (general-predicate-dispatch nil
                       (looking-at-p "\\*\\* *")
-                      (cmd! (forward-char 2)))))))
+                      (cmd! (forward-char 2)))))
 
     ;; Removes haskell-mode trailing braces
 
@@ -216,7 +216,7 @@
         (sp-local-pair "{-@" "@-}" :actions :rem)
         (sp-local-pair "{-" "-")
         (sp-local-pair "{-#" "#-")
-        (sp-local-pair "{-@" "@-")))
+        (sp-local-pair "{-@" "@-")))))
 
 
 ;;
--- a/modules/config/default/config.el
+++ b/modules/config/default/config.el
@@ -207,6 +207,17 @@
                       (looking-at-p "\\*\\* *")
                       (cmd! (forward-char 2)))))))
 
+    ;; Removes haskell-mode trailing braces
+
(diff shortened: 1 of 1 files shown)


iyefrat:
bump

modules/config/default/config.el
```diff
@@ -205,7 +205,18 @@
       (map! :map markdown-mode-map
             :ig "*" (general-predicate-dispatch nil
                       (looking-at-p "\\*\\* *")
-                      (cmd! (forward-char 2)))))))
+                      (cmd! (forward-char 2)))))
+
+    ;; Removes haskell-mode trailing braces
+
```
hlissner, line 211: Remove superfluous blank line.

modules/config/default/config.el
```diff
@@ -205,7 +205,18 @@
...
+    (after! haskell-mode
```
hlissner, line 212: s/haskell-mode/smartparens-haskell

[smartparens-haskell.el](https://github.com/Fuco1/smartparens/blob/master/smartparens-haskell.el) is where the default sp config lives.

hlissner (changes requested):
Please squash your commits into one and reword the commit to conform this PR to [our git conventions](https://gist.github.com/hlissner/4d78e396acb897d9b2d8be07a103a854):

```
fix(default): smartparens rules for haskell

The default haskell-mode rules would result in an extra } appended to
the closing pair (#5448).

Fix: #5448
```

Issue #2688, opened 2020-03-09
How to disable exit prompt


adrianstaniec:
**What did you expect to happen?**
I expected just to quit emacs, 
just kill it directly without nagging,
as I don't have any buffers/files modified/unsaved.

**What actually happened?**
Emacs is asking me in minibuffer : `Really exit emacs?`...

**Additional details:**
The relevant part of my `init.el`:

```
(doom! :input

       :completion
       company           ; the ultimate code completion backend
       ivy               ; a search engine for love and life

       :ui
       ;;deft              ; notational velocity for Emacs
       doom              ; what makes DOOM look the way it does
       doom-dashboard    ; a nifty splash screen for Emacs
       ; doom-quit         ; DOOM quit-message prompts when you quit Emacs
       ...
      )
```

**Steps to reproduce:**
Start doom emacs.
Open a file (not modified it at all).
Press `SPC q q`.

**System information:**
<pre>
<details>

```
emacs   version    26.3
        features   XPM JPEG TIFF GIF PNG SOUND GSETTINGS GLIB NOTIFY GNUTLS FREETYPE XFT ZLIB TOOLKIT_SCROLL_BARS LUCID X11 XDBE XIM THREADS
        build      Nov 12, 2019
        buildopts  (--with-x-toolkit=lucid)
        windowsys  x
        daemonp    server-running
doom    version    2.0.9
        build      HEAD -> develop, origin/develop, origin/HEAD 10623868b 2020-02-19 23:41:39 -0500
        dir        ~/.doom.d/
system  type       gnu/linux
        config     x86_64-pc-linux-gnu
        shell      /bin/bash
        uname      Linux 4.15.0-74-generic #84-Ubuntu SMP Thu Dec 19 08:06:28 UTC 2019 x86_64
        path       (/fast/anaconda3/condabin /fast/anaconda3/bin ~/bin/ ~/.emacs.d/bin /usr/local/cuda/bin ~/bin ~/.local/bin /usr/games /usr/local/sbin /usr/local/bin /usr/sbin /usr/bin /sbin /bin /snap/bin /usr/lib/jvm/java-8-oracle/bin /usr/lib/jvm/java-8-oracle/db/bin /usr/lib/jvm/java-8-oracle/jre/bin ~/.emacs.d/bin /usr/local/libexec/emacs/26.3/x86_64-pc-linux-gnu/)
config  envfile    envvar-file
        elc-files  0
[...]

PR #5456, opened 2021-09-06
doc: add how to disable doom-quit

jman-schief:
Add how to disable `doom-quit` in module documentation

(from https://github.com/hlissner/doom-emacs/issues/2688)


hlissner:
Redundant with the section on disabling modules in docs/getting_started.org and one of the first results (if not the first) for "emacs quit prompt" (<https://www.gnu.org/software/emacs/manual/html_node/emacs/Exiting.html>). As it doesn't apply specifically to the doom-quit module, it doesn't belong to its module. I'll add an entry for this in docs/faq.org in the future, once I'm done with the rewrite of our docs.

PR #5460, opened 2021-09-07
docs(python): setting lsp PATH for python

acmnu:
The commit adds information about setting PATH for lsp.

It is not so clear, especially for beginners that if you install lsp
in the virtualenv you will have a trouble with running it.

The current order of pyenv-activate and lsp starting does not allow user
to use lsp installed to a project venv so the simplest way I found is
a modification of the PATH.

### DIFF
--- a/modules/lang/python/README.org
+++ b/modules/lang/python/README.org
@@ -84,6 +84,11 @@ manager or manually:
 + *mspyls* can be installed by typing =M-x lsp-install-server RET mspyls=.
 + *pyright* can be installed with ~pip install pyright~ or ~npm i -g pyright~.
 
+Take a note that if you had installed LSP in the virtual env, you have to add
+it to your path (use ~config.el~ for this). For example:
+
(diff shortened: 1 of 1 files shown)


sei40kr:
I'm using Python LSP in the Poetry env, and setting it in `.dir-locals.el` is working for me.

```emacs-lisp
((python-mode . ((lsp-pylsp-server-command . ("poetry" "run" "pylsp")))))
```

I think this is more preferable, because it applies only to current project.

vikigenius:
@sei40kr The solution you suggested throws an error saying 
Process pylsp stderr finished

hlissner:
Thanks for the PR but I cannot accept this. Doom doesn't endorse folks adding manipulating `exec-path` directly. Users should be setting their `PATH` in their shell dotfiles, which Emacs will pick up on (you may need to regenerate your envvar file by running `doom env`). This could be better documented in docs/getting_started.org, which is will focus on to address this. In any case, thanks for bringing it to my attention.

PR #5463, opened 2021-09-09
bump!: :emacs vc

lexa:
BREAKING CHANGE: browse-at-remote-remote-type-domains is replaced by
browse-at-remote-remote-type-regexps. If you have a custom domain
configured replace:

    (add-to-list 'browse-at-remote-remote-type-domains
      ("git.example.com" . "github")

with

    (add-to-list 'browse-at-remote-remote-type-regexp
      ("^git\\.example\\.com$" . "github")

Added support for Gittiles.


hlissner:
LGTM. Thanks for your help.

PR #5468, opened 2021-09-11
Improve `:email mu4e` (for Mu4e 1.6)

### LINKED ISSUE #5027: Cannot open attachments from email view in mu4e
### 

- [X] I have searched [the Discourse](https://discourse.doomemacs.org/docs) for solutions.

When I open an email message in mu4e, I used to be able to press `o` to open one of the attachments. 

### What actually happened?

Today after upgrading Doom I can no longer do that, there's only the option to save the attachment, and then I have to open it manually OR I can use the mouse (ugh) to click the name of the attachment. Also, the attachment list is displayed differently than it used to be:

image

I've tried looking for a function in M-x that would reproduce this functionality, but came empty-handed

### Steps to reproduce

1. Open emacs
2. Open mu4e
3. navigate to a single message that contains an attachment
4. press `o`

### System Information

https://pastebin.com/9yyPjqnW

### LINKED ISSUE #5452: mu4e main view shows a medium black circle emoji instead of an asterisk

By default, mu4e shows an asterisk for list entries (such as for each bookmark or maildir). The asterisk is shown in a default face, so that it matches other text elements and isn't hard to discern from the background.

### What actually happened?

Each asterisk in mu4e-main is replaced with a black circle emoji (⚫, U+26AB MEDIUM BLACK CIRCLE), by the functions `mu4e~main-action-str-prettier-a` and `mu4e~main-keyval-str-prettier-a` by the file `modules/email/mu4e/autoload/advice.el`. In a terminal (for example, iTerm 2 on a Macintosh) that has support for emoji, this is hard to see on a dark background, and also doesn't really match the rest of the view.

My Linux desktop, meanwhile, shows a small bullet point instead of the graphical emoji -- I suspect that this is due to it not having the correct fonts installed/configured, and isn't because of anything happening in emacs. When I run `what-cursor-position` over the character, it tells me that it is indeed MEDIUM BLACK CIRCLE.

This feature should, in my view, be configurable so that it can be disabled or easily replaced. If I've missed an obvious way to do that already, my apologies!

I was able to manually change the character shown by changing this function, as well as `mu4e-main-action-str-prettier-a`.

```
(defun +mu4e~main-keyval-str-prettier-a (str)
   "Replace '*' with '•' in STR."
   (replace-regexp-in-string "\t\\*" "\t•" str))
```

### Steps to reproduce

1. Open Emacs
2. Start mu4e
3. Observe the bookmarks and action listings

### System Information

https://pastebin.com/cfY7x7cK

tecosaur:
This is a response to the recent mu4e 1.6 update, and a minor tweak.

Basically, mu4e 1.6 removed some convenient actions that I'm adding back, and making nicer to use.
This should close #5027, and I've added bullet customisation to close #5452 too while I'm at it.

Regarding the change in actions, here's what's changed:

### Before

image

### After

image

\+ a convenience function for opening *attachments* (only some parts are attachments).


modules/email/mu4e/config.el
```diff
@@ -52,12 +52,14 @@
...
-        (if (executable-find "xdg-user-dir")
-            ;; remove trailing newline
-            (substring (shell-command-to-string "xdg-user-dir DOWNLOAD") 0 -1)
-          (expand-file-name (or (getenv "XDG_DOWNLOAD_DIR")
-                                "Downloads")
-                            "~"))
+        (concat
+         (if (executable-find "xdg-user-dir")
+             ;; remove trailing newline
+             (substring (shell-command-to-string "xdg-user-dir DOWNLOAD") 0 -1)
```
hlissner, line 58: Could you use `(doom-call-process "xdg-user-dir" "DOWNLOAD")` instead of `shell-command-to-string`? This sidesteps any complicated shell configuration that may interfere with the command.

hlissner (changes requested):

modules/email/mu4e/config.el
tecosaur: I've just added 083c81e44b7e97661c17100bebfbc66f3207e38e, let me know if you're happy with that :slightly_smiling_face: .

hlissner (approved):

hlissner:

PR #5480, opened 2021-09-15
fix(layout): prevent magit from stealing t in bépo

gagbo:
Just as magit took control of 's' after "key rotation", magit (at least
when unpinned) takes control of 't' and we need to manually unmap the
key so that 't' goes down one line when the cursor is on a magit section
header


hlissner (approved):

hlissner:
LGTM. Thanks!

PR #5482, opened 2021-09-15
fix(zen): robust visual adjustment upon text scaling

HyunggyuJang:
Visual adjustment upon text scale change
-------

Using `advice-add` to `text-scale-mode` only covers text scaling done
by `text-scale-adjust`. Typically users rely more on `C-+` or `C--`,
which doesn't use `text-scale-adjust` function under the hood.

Hooking to `text-scale-mode` renders more reliable, and broad coverage.

### DIFF
--- a/modules/ui/zen/config.el
+++ b/modules/ui/zen/config.el
@@ -46,7 +46,7 @@
         (window-divider-mode +1))))
 
   ;; Adjust margins when text size is changed
-  (advice-add #'text-scale-adjust :after #'visual-fill-column-adjust))
+  (add-hook 'text-scale-mode-hook #'visual-fill-column-adjust))
 
 
 (use-package! mixed-pitch


hlissner:
1. Can you clarify what is bound to `C-+` and `C--`? In Doom, this is bound to `doom/reset-font-size` and `text-scale-decrease` (and `text-scale-increase` is bound to `C-=`). In vanilla Emacs `C-+` is unbound and `C--` is bound to `negative-argument`. I _think_ you're referring to `text-scale-increase` and `text-scale-decrease`, but both of these call `text-scale-adjust`. I am not aware of any other text scaling mechanism in Emacs.
2. This PR adds a hook to `text-scale-mode`, not `text-scale-mode-hook`, therefore it will never run. This seems like a dubious fix.
3. Could you provide some steps to produce the issue, including the commands. I cannot reproduce any issue with the current setup.

HyunggyuJang:
Henrik, I appreciate your tolerance and generosity.


> 1. Can you clarify what is bound to `C-+` and `C--`?

You are absolutely right. I meant `text-scale-increase/decrease`. Was
bound to `C-=/-`. I was in hurry.

>  I _think_ you're referring to `text-scale-increase` and `text-scale-decrease`, but both of these call `text-scale-adjust`.

Want to clarify a bit. In my Emacs version, 27.2, and from [the mirror's
source
code](https://github.com/emacs-mirror/emacs/blob/1bae08dbf07da212b3fce5c9df0bdeb10e4fd7b5/lisp/face-remap.el#L319-L335),
`text-scale-increase` does **not** call `text-scale-adjust` to set face
remapping.

Frankly, `text-scale-adjust` relies on `test-scale-increase` to reflect
given input scale.

> 2. This PR adds a hook to `text-scale-mode`, not `text-scale-mode-hook`,

Yes, was typo as you pointed out. I made pull request from my personal
configuration, in that process, I manually wrote this pull request;
missed to adding `-hook`, or adding `!` to the end of `add-hook` function.

I'll fix it.

> 3. Could you provide some steps to produce the issue, including the commands.

As I said, `text-scale-increase` doesn't rely on `text-scale-adjust`,
rather call directly `text-scale-mode`. Hence, I think this pull request
resolves, if typo is fixed, will actually amend shortcomings from I
noted above.

Think this applies to only Doom Emacs' evil users, since by [this
mapping](https://github.com/hlissner/doom-emacs/blob/ca8892ee2eba3ead7a334ce51c5c153734aad2da/modules/config/default/%2Bevil-bindings.el#L116),
doom calls `text-scale-increase` directly.

It would be helpful why you thought `text-scale-increase` calls
`text-scale-adjust` at the end of the day.

Thanks!

PR #5486, opened 2021-09-16
fix: unicode font initializing early in daemon

repelliuss:
### Issue

When Emacs is started in daemon mode, *some* unicode characters were broken until `doom/reload-font` is called. I believe `doom-unicode-font` were being set early by `after-init-hook`.

My `doom-unicode-font` is "Joypixels".

### To reproduce

1. Insert 🕐
2. Check above unicode char in daemon mode
3. Call `doom/reload-font` to fix it

Also check it without daemon mode to see it doesn't have this problem.

### Fix

This PR sets `doom-unicode-font` using `after-make-frame-functions` transient hook.

### DIFF
--- a/core/core-ui.el
+++ b/core/core-ui.el
@@ -666,6 +666,9 @@ windows, switch to `doom-fallback-buffer'. Otherwise, delegate to original
 
 ;; Apply `doom-font' et co
 (add-hook 'doom-after-init-modules-hook #'doom-init-fonts-h -100)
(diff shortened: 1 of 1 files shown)


hlissner:
Sorry for the _super_ late response. A fix for this will likely come with #7161 (I've also posted a patch to try). Since this PR is rather old and targets a defunct branch, I will close it. Thanks for the help, in any case!

PR #5031, opened 2021-05-11
elixir language module make alchemst a flag.

mogorman:
> - [x] I've targeted the develop branch
> - [x] I've searched for similar pull requests and found nothing
> - [x] This change is NOT in Doom's do-not-PR list: https://doomemacs.org/d/do-not-pr
> - [x] If I've bumped any packages, I've done so according to https://doomemacs.org/d/how2bump
> - [ ] I've linked any relevant issues and PRs below
> - [ ] All my commit messages are descriptive and distinct

Although alchemist use is still wide, it's development has ceased and
many emacs users have been moving towards lsp.  The alchemist jump to
definition conflicts with the lsp one. I changed the init example to
help maintain stability, but I think most people would be better off
with lsp and not alchemist.

PR #5488, opened 2021-09-16
tweak(elixir): guard alchemist.el with feature flag

merrickluo:
I found that https://github.com/hlissner/doom-emacs/pull/5031 is closed due to stale, but I think the idea is still valid.

### DIFF
--- a/modules/lang/elixir/README.org
+++ b/modules/lang/elixir/README.org
@@ -15,13 +15,14 @@
 - [[#features][Features]]
 - [[#appendix][Appendix]]
   - [[#commands][Commands]]
+    - [[#exunit-mode][exunit-mode]]
 
 * Description
 This module provides support for [[https://elixir-lang.org/][Elixir programming language]] via [[https://github.com/tonini/alchemist.el][alchemist.el]]
 or [[https://github.com/elixir-lsp/elixir-ls/][elixir-ls]].
 
 ** Module flags
-+ ~+lsp~ Enable LSP support. Requires [[https://github.com/elixir-lsp/elixir-ls/][elixir-ls]].
++ ~+lsp~ Enable LSP support. Requires [[https://github.com/elixir-lsp/elixir-ls/][elixir-ls]], also disables alchemist.
 
 ** Plugins
 + [[https://github.com/elixir-editors/emacs-elixir][elixir-mode]]
--- a/modules/lang/elixir/config.el
+++ b/modules/lang/elixir/config.el
@@ -54,6 +54,7 @@
 
 
 (use-package! alchemist
+  :unless (featurep! +lsp)
   :hook (elixir-mode . alchemist-mode)
   :config
   (set-lookup-handlers! 'alchemist-mode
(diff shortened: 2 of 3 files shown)


iyefrat:
Since this is a breaking change (alchemist users need to manually add the feature flag to keep their setup working), Can you please reword the type/scope to `refactor!(elixr)`, and document why this is needed in the commit body with
```
BREAKING CHANGE: lorep ipsum etc etc
```
see the [commit conventions](https://gist.github.com/hlissner/4d78e396acb897d9b2d8be07a103a854) for more details.

cwlbraa:
elixir-ls + eglot works so much dramatically better without alchemist, it's almost criminal. Maybe the right thing to do and a way to make this somewhat more backwards-compatible is to make Elixir similar to other lsp-capable language modules and turn off alchemist when lsp is enabled. This way we don't break people who haven't configured anything, but we improve the experience of those who have already setup elixir-ls.

merrickluo:
@cwlbraa Thanks, that also makes sense to me. My complaint is also that I have to disable alchemist manually after enabling lsp. 

@iyefrat, so, I updated the PR to use the `+lsp` flag instead of adding a new one,  I read the commit convention, but still not sure this counts as `refactor` or `tweak` after the change, I'm keeping tweak right now, but I'm happy to change it if it does not fit.

diannasoriel:
I think this PR is very much necessary, alchemist and lsp/eglot don't really work well together, therefore disabling it with LSP is almost required for Elixir development with Doom at this point.

merrickluo:
I think his is still relevant.

merrickluo:
changed target to master

bhankas:
Is this PR still relevant? I just started learning Elixir and don't understand tooling perfectly. But if it results in better experience, maybe good time to rebase on latest master changes and merge? Even if just confirmation, I can take it forward.

merrickluo:
you can achieve the same thing with `(package! alchemist :disable t)` in
your `packages.el`, I also think it's better to just remove alchemist since
it's unmaintained for almost 5 years now.

bhankas ***@***.***> writes:

> Is this PR still relevant? I just started learning Elixir and don't understand tooling perfectly. But if it results in better experience, maybe
> good time to rebase on latest master changes and merge? Even if just confirmation, I can take it forward.
>
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you authored the thread.


--
Thanks.
Merrick Luo

hlissner:
@merrickluo Sorry for taking so long to get back to this, but I'm looking into resolving this now and wanted to know if it still makes sense to remove alchemist altogether? Is there any merit to keeping it? If they are equal, I'd rather rely on lsp-mode.

ckruse:
Alchemist is de facto dead. It should be dropped completely.

hlissner:
Understood. Done as of 221e75e. Thanks for the help!

adrianbrink:
Is there still support for `iex` and `mix` using this package, because from the config files it seems that this was provided by alchemist.

PR #5492, opened 2021-09-17
fix(notmuch): window is not a valid window error on sync end

HyunggyuJang:
Tidy notmuch update process
-------

Notmuch sync process is done asynchronously, but if the user switch to other buffer before update process, it complains like

```
error in process sentinel: #<window 98> is not a valid window
```

This commit simplifies cleanup process after mail syncing, and still maintain its functionality.


hlissner (approved):

hlissner:

PR #5495, opened 2021-09-18
tweak(web): add nxml mode to LSP enabled modes

svenihoney:
Since lsp-mode supports XML (see
https://emacs-lsp.github.io/lsp-mode/page/lsp-xml/), it makes sense to
add nxml-mode to the list of automatically enabled lsp mode clients.


hlissner (approved):

hlissner:

PR #5502, opened 2021-09-20
Mu4e: improve message checks at send

tecosaur:
Just two little QOL improvements here.

- `org-msg` checks for attachments using an overly simply regex, there are too many false positives. So, let's tighten that up a bit.
- Accidentally sending an email without a subject isn't nice. So, we add a hook to confirm that the user really wants to do this, and if not enter their desired subject line.


hlissner (approved):

hlissner:
LGTM, thanks again!

PR #5504, opened 2021-09-21
feat(default): add .emacs.d search on <leader> s e

iyefrat:
As discussed on the discord.

I think this implementation is more readable than using `(+vertico/project-search nil nil user-emacs-directory)` calls, but tell me if you disagree.


hlissner:
Makes sense to me. Thanks!

PR #5505, opened 2021-09-21
refactor!(:term): toggle commands now always hide

### LINKED ISSUE #3374: [REQUEST] provide an option to toggle vterm pop-up window even when displayed

iyefrat:
BREAKING CHANGE: previously, \<leader\> o t commands would only hide the
terminal popup if it was focused. If not, they would move the focus to
the terminal window. This is unintuitive to the "toggle" description,
and arguably less useful, since refocusing to the terminal can be easily
done with regular window refocus commands. Therefore, \<leader\> o t now
just hides the terminal popup.

Fix #3374


hlissner:
Great work as always. Thanks again!

PR #5508, opened 2021-09-21
Remove 'bash strict mode' from sh template.

createyourpersonalaccount:
There are good reasons not to include

    set -euo pipefail

in shell scripts. A summary and more references may be found here,
https://lists.gnu.org/archive/html/help-bash/2020-04/msg00049.html

### DIFF
--- a/modules/editor/file-templates/templates/sh-mode/__
+++ b/modules/editor/file-templates/templates/sh-mode/__
@@ -1,4 +1,3 @@
 #!/usr/bin/env `(if (equal (file-name-extension buffer-file-name) "zsh") "zsh" "bash")`
-set -euo pipefail
 
 $0


createyourpersonalaccount:
I wanted to read the commit rules but it linked me to discourse and I didn't have an account. However, the github linter mentions 
https://docs.doomemacs.org/latest/#/developers/conventions/git-commits
which I was able to view without an account.

It seems that I need to rewrite the commit and add a `revert` here; since I wish to revert to an earlier commit that did not include the `set -euo pipefail` line.

Please let me know what you think of this change and if it is worthwhile I will then do the above.

hlissner:
Duplicate of #5428.

> I wanted to read the commit rules but it linked me to discourse and I didn't have an account. However, the github linter mentions
https://docs.doomemacs.org/latest/#/developers/conventions/git-commits
which I was able to view without an account.

I'm aware of it. The transition was supposed to be smoother, but I'm on "sabbatical" for health issues, so development is slow at the moment.

In any case, #5428 has a better solution for this and conforms to commit conventions (which I hope answers your question).

PR #5511, opened 2021-09-22
docs(config): replace broken wiki reference in default module

joelpet:
There is no wiki available (anymore?), but there is documentation on how to write your own module in the docs inside the repository, so reference them instead.


hlissner:
Thanks for the PR, but I must refuse it. There is nothing wrong with your PR, but it comes at a bad time: I'll be pushing a rework of all our documentation soon, which indirectly fixes this issue (and many others). I'll simply wait for that instead. In any case, thanks for your help!

joelpet:
No problem at all. Thank you for all the excellent work you put into Doom Emacs!

PR #5512, opened 2021-09-23
Replace mentions of `ob-rust` with `ob-rustic`

fosskers:
A quick tweak. I noticed today that Doom doesn't actually use `ob-rust` anymore, although the docs still mentioned it.


hlissner:
Thanks for the PR, but I must refuse it. There is nothing wrong with your PR, but it comes at a bad time: I'll be pushing a rework of all our documentation soon, which indirectly fixes this issue (and many others). I'll simply wait for that instead. In any case, thanks for your help!

PR #5513, opened 2021-09-23
Bump `php-cs-fixer` version in php module

peeley:
In the pinned version of `php-cs-fixer` present in the php module (when the format module is also enabled), custom `php-cs-fixer` binaries specified via the `php-cs-fixer-command` var do not work.

This new commit (merged [here](https://github.com/OVYA/php-cs-fixer/pull/16)) fixes the issue, and allows for users to specify custom binaries or paths for the `php-cs-fixer` program.


hlissner:
Thanks!

PR #5515, opened 2021-09-24
fix(:lang/common-lisp): check inferior-lisp-program like "ros -Q run"

wood9366:
After I set inferior-lisp-program to "ros -Q run", I see "WARNING: Couldn't find `inferior-lisp-program' (ros -Q run)" when load a ros file.

I see fix code at common-lisp/doctor.el, and I think the same check in config.el should be fix too.


hlissner:

Issue #5439, opened 2021-09-01
Outlook 365 (online and mobile apps) display mu4e messages in a narrow column


TS-CUBED:
### What did you expect to happen?

When I send an org-msg email through mu4e the message gets encoded as html fine.

Most clients (including the Outlook desktop app) will display these emails nicely.

Not sure how to provide debugging info here, since the problem seems to be on MS's side, but maybe somebody knows which HTML setting could cause this.

Running Doom Emacs V 3.0.0alpha

### What actually happened?

However, the Outlook online client, and the Outlook mobile app, will display them in a narrow column, rather than the max column width of 90 characters that the HTML code has in the CSS

### Describe your attempts to resolve the issue

This is related to 
https://github.com/jeremy-compostella/org-msg/issues/131
https://github.com/jeremy-compostella/org-msg/issues/59

Outlook interprets the `max-width` of `90ch` in Dooms mu4e/org-msg CSS as `90px` due to a bug in the CSS implementation in the web and mobile outlook apps.

This causes the email to be displayed in a very narrow column. 

### Steps to reproduce

Send an email using the default Doom CSS to a recipient that uses the web or mobile outlook client.

Open email on web or mobile outlook

FIX: change CSS in `modules/mail/mu4e/config.el` to remove the `max-width 90ch`

### System Information

https://pastebin.com/tPcqXb2Y

TS-CUBED:
Setting the max-width to 800px (or whatever is the optimum for the given font size) does make Outlook behave as expected.

eddiegroves:
I fixed this for now by overriding Doom's custom styles. I preferred the ones org-msg provides by default anyway.

```
(setq org-msg-enforce-css org-msg-default-style)
```

tecosaur:
Out of curiosity @eddiegroves, what aspects of the default styling do you prefer?

eddiegroves:
Hi @tecosaur thanks for the question, it prompted me to look a bit closer at my preferences and to be honest, I'm now liking your/Doom's changes more. It is a lot clearer, the font and font size is more sensible.

I might pull the fix above so I can switch over now.

The only preference I have now, and I will look at customising, is the heading color. I think it's too strong as a shade of red.

For context, most email's I'm writing are in the enterprise and being read by folks using various versions of Outlook, O365 and iOS/MacOS Mail.

tecosaur:
[...]

PR #5521, opened 2021-09-26
tweak(mu4e): change org-msg ch units to em

tecosaur:
Apparently `ch` space units in Outlook are a source of pain (see #5439). So, we change to a similar width in `em`.

Because Outlook interprets 90ch as 90px ... nice job Outlook.

I also noticed that `<kbd>` elements were too big and had too much padding, so I've snuck in a change to them too.

#### Before

image

#### After

image


hlissner (approved):

hlissner:
LGTM. Thanks again!

PR #5523, opened 2021-09-27
improve `:app rss` module

jeetelongname:
This PR makes minor improvements to `:app rss` by adding
- the `+rss/copy-link` command and binding it to `g c` 
  This is similar to the `elfeed-search-browse-url` command but copys the link
  instead of sending it to the web browser 
- integrates `elfeed-goodies` which provides a better and more configurable UI

### DIFF
--- a/modules/app/rss/README.org
+++ b/modules/app/rss/README.org
@@ -30,6 +30,7 @@ This module has no dedicated maintainers.
 
 ** Plugins
 + [[https://github.com/skeeto/elfeed][elfeed]]
++ [[https://github.com/algernon/elfeed-goodies][elfeed-goodies]] 
 + =+org=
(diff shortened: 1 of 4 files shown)


jeetelongname:
It may be worth feature gating goodies (though its really easy to disable)

hlissner (approved):

hlissner:
At first glance I don't think feature gating is necessary here. Otherwise, LGTM. Thanks again!

jeetelongname:
Yeah its a simple matter of disabling the package, though if push comes to shove a flag would not be that bad to add

e3bc54b2:
I agree that disabling the package is not much work, but a flag is still nicer and more explicit. It will also allow customizing elfeed-goodies interface for Doom users in future without affecting existing users. I created #5603 for the same.

PR #5526, opened 2021-09-28
Add `:input bidi` module

iyefrat:
This pull requests adds better support for bidirectional and rtl text. It provides:
- a buffer-local `+bidi-mode` that undoes doom's anti-bidi optimizations, and sets the bidi initial alignment algorithm to start from scratch each line, so there is no need for empty lines to separate paragraphs.
- a global variant `+bidi-global-mode` if people just want it on all the time (off for read-only buffers due to fontification shenanigans)
- `+bidi-(hebrew|arabic)-font` (and faces) for easily setting a better font for these languages
- `+bidi-want-smart-fontify` uses for using said fonts on spaces and punctuation after RTL characters, since in hebrew at least there are almost no good monospace fonts
- instructions in the readme on further customization users probably want to make that can't be universalized in the module (such as language specific fonts)
- `+bidi-paragraph-direction` to customize what `bidi-paragraph-direction` is set to when `+bidi-mode` is on. (should not be used with `+bidi-global-mode`).
- There is currently an issue with the way evil handles arrow key movement (which imo should be visual rather than logical, the way Emacs does it). This is something that I'll fix upstream.

since and date tags will be corrected once the module is approved

### DIFF
--- a/modules/input/bidi/config.el
+++ b/modules/input/bidi/config.el
@@ -3,6 +3,20 @@
 (defvar +bidi-mode-map (make-sparse-keymap)
   "Keymap for `+bidi-mode'.")
 
+(defvar +bidi-hebrew-font (font-spec :family "DejaVu Sans")
+  "Overriding font for hebrew script.
+Must be a `font-spec', see `doom-font' for examples.
+
+WARNING: if you specify a size for this font it will hard-lock any usage of this
+font to that size. It's rarely a good idea to do so!")
+
+(defvar +bidi-arabic-font (font-spec :family "DejaVu Sans")
+  "Overriding font for arabic and arabic-derived scripts.
+Must be a `font-spec', see `doom-font' for examples.
+
+WARNING: if you specify a size for this font it will hard-lock any usage of this
+font to that size. It's rarely a good idea to do so!")
+
 ;;;###autoload
 (define-minor-mode +bidi-mode
   "Minor mode for using bidirectional text in a buffer.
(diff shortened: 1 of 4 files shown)


iyefrat:
In case you have gotten here through email, the top post has been edited with all the relevant info.

iyefrat:
I am not sure what changed, but there has been a regression with the whitespace/punctuation font stuff. I'll undraft this once that's solved.

modules/input/bidi/config.el
```diff
@@ -0,0 +1,93 @@
...
+  "Overrideing font for hebrew script.
+Must be a `font-spec', see `doom-font' for examples.
+
+WARNING: if you specify a size for this font it will hard-lock any usage of this
+font to that size. It's rarely a good idea to do so!")
+
+(defface +bidi-hebrew-face `((t :font ,+bidi-hebrew-font)) "")
+
+(defvar +bidi-arabic-font (font-spec :family "DejaVu Sans")
+  "Overrideing font for arabic and arabic-derived scripts.
```
hlissner, line 16: s/Overrideing/Overriding/

hlissner (changes requested):
Great work! Just have to fix one small typo and it'll be merge ready.

modules/input/bidi/config.el
iyefrat: whoops! fixed.

hlissner (approved):

hlissner:

PR #5528, opened 2021-09-28
fix: org-roam.db not in .gitignore

Patrick0308:
### 

add org-roam.db to .gitignore

### DIFF
--- a/modules/tools/lookup/config.el
+++ b/modules/tools/lookup/config.el
@@ -215,6 +215,15 @@ Dictionary.app behind the scenes to get definitions.")
   :unless IS-MAC
   :defer t
   :config
+  ;; REVIEW Temporarily fix abo-abo/define-word#31
+  (defadvice! +lookup--fix-define-word-a (fn &rest args)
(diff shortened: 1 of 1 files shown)


gagbo:
Hello,

(Just my 2 cents)

Your `org-roam.db` file should not live in the root of the repository, it defaults to [.local/etc/org-roam.db](https://github.com/hlissner/doom-emacs/blob/cdb59b0a3d3e20f37df6fdb5a68c7c8090c9c25d/modules/lang/org/contrib/roam.el#L63)

And `.local/` is already in `.gitignore`, so theoretically there is no need for this change.

If you have `org-roam.db` at the root of the repository (so directly in `.emacs.d` or `.config/emacs`), then it’s probably a bug/issue worth opening instead of making this PR.

Have a nice day :)

gagbo:
(Making an issue will also guide you to share your `doom/info`, which is extremely helpful in understanding how your `org-roam.db` might have ended up in a not-ignored location)

PR #5536, opened 2021-09-29
bump: :tools magit

catap:
magit/forge@bae6a527a738 -> magit/forge@22f905c4e8e2

Fix: https://github.com/magit/forge/issues/382


catap:
@hlissner here the last but for magit/forge for a while I hope.

The root cause is: https://github.com/magit/forge/issues/382#issuecomment-930304337

hlissner:
Thanks again!

PR #5540, opened 2021-09-30
feat(notmuch): allow notmuch module without (:ui workspaces)

hugonobrega:
Remove the dependency of (:ui workspaces) from the notmuch module

The notmuch module had a hard-dep on the (:ui workspaces) module which was unnecessary; these (very simple) changes fix that.
(I'm calling it a "feat" since it introduces new behavior (debatable))


hlissner (approved):

hlissner:
LGTM. Thanks again!

PR #5548, opened 2021-10-01
refactor(pass): make duplicate functions obsolete

dschrempf:
The following functions are affected: `+pass/edit-entry`,
`+pass/copy-field`, `+pass/copy-secret`, `+pass/browse-url`. All of
these functions have counterparts in `password-store`. Additionally, the
functions in password store add git commits when files are
changed.

The obsolete functions can be removed at some point.

Questions:
- Should we also deprecate the `ivy` and `consult` versions? The
`password-store` versions work perfectly well here with `vertico` (not
sure about `ivy` though).

### DIFF
--- a/modules/tools/pass/autoload/ivy.el
+++ b/modules/tools/pass/autoload/ivy.el
@@ -3,7 +3,7 @@
 
 ;;;###autoload
 (defun +pass/ivy (arg)
-  "TODO"
+  "Complete and act on password store entries."
   (interactive "P")
   (ivy-read "Pass: " (password-store-list)
             :action (if arg
@@ -15,7 +15,7 @@
   (ivy-add-actions
    '+pass/ivy
    '(("o" password-store-copy "copy password")
-     ("e" +pass/edit-entry "edit entry")
+     ("e" password-store-edit "edit entry")
      ("u" +pass/copy-user "copy username")
-     ("b" +pass/copy-url "open url in browser")
-     ("f" +pass/copy-field "get field"))))
+     ("b" password-store-url "open url in browser")
+     ("f" password-store-copy-field "get field"))))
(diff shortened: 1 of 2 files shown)


dschrempf:
It turns out that the `password-store-xxx` functions actually fire up `pass` while the Doom specific functions use `find-file`. This means that if the `EDITOR` variable is not set to `emacslient` or some derivative, the functions will actually not be equivalent. In my opinion the Doom specific functions can still be made obsolete and removed, but I wanted to make you aware of this.

hlissner (changes requested):
> It turns out that the `password-store-xxx` functions actually fire up `pass` while the Doom specific functions use `find-file`. This means that if the `EDITOR` variable is not set to `emacslient` or some derivative, the functions are not equivalent. In my opinion the Doom specific functions can still be made obsolete and removed, but I wanted to make you aware of this.

That should be fine. The `with-editor` sets EDITOR locally, so the user doesn't have to.

That said, would you mind changing these to aliases using autoloaded `define-obsolete-function-alias`. e.g.

```elisp
;;;###autoload
(define-obsolete-function-alias '+pass/edit-entry #'password-store-edit "21.12")
;; (And remove the definition for `+pass/edit-entry`)
```

And do so for each of them.

dschrempf:
This reduced the file quite a bit!

We could reduce more if we remove the `+pass-get-secret` function. The only difference to the `password-store-copy` function is that the first returns the secret while the second copies the secret to the kill ring (or clipboard). `+pass-get-secret` and `+pass-get-user` are mentioned in a README when settings up IRC. Also, there are tests for these two functions.

dschrempf:
I should also say that `+pass/ivy` is now using the obsolete functions. I will change this now.

dschrempf:
I changed the ivy actions. However, I did not test the code because I am not using `ivy`.

hlissner (approved):

hlissner:
Great work. Thanks for your help (and congrats on your first contribution to Doom)!

> Should we also deprecate the ivy and consult versions?

I may no longer be necessary for consult, but ivy I think still needs it. I'll investigate this soon.

PR #5549, opened 2021-10-02
getting_started.org: Gentoo guide update for emacs-29 live release

mjkalyan:
Updated the install guide for Gentoo to account for the new emacs-29 package in the ::gentoo repo and make more clear about why you need a live ebuild of version 28 or greater for native compilation.


hlissner:
Thanks for the PR, but I must refuse it. There is nothing wrong with your PR, but it comes at a bad time: I'll be pushing a rework of all our documentation soon, which will include this information and resolve many other issues. I'll simply wait for that instead. In any case, thanks for your help!

PR #5553, opened 2021-10-02
Increase width of mu4e date column to fit YYYY-mm-DD HH:MM.

### LINKED ISSUE #5552: The date column in mu4e headers view is not wide enough for the commonly used "YYYY-mm-DD HH:MM" format

When entering the headers view in mu4e I expect the full date and time of the email to be visible in the date column.

### What actually happened?

The width of the column is not wide enough to even show the full date.
2021-10-02-134107_1920x1200_scrot

Set `:human-date` to `16` in modules/email/mu4e/config.el.
 
```emacs-lisp
mu4e-headers-fields
  '((:account-stripe . 1)
    (:human-date . 16)
    (:flags . 6) ; 3 icon flags
    (:from-or-to . 25)
    (:subject)))
```

### Steps to reproduce

See description. Might depend on your locale.

### System Information

https://pastebin.com/2FEHwzyD

johanthoren:
Fixes #5552

I've increased the width of the mu4e headers-view date column to make room for the commonly used YYYY-mm-DD HH:MM format.

### DIFF
--- a/modules/email/mu4e/config.el
+++ b/modules/email/mu4e/config.el
@@ -71,7 +71,7 @@
         ;; remove 'lists' column
         mu4e-headers-fields
         '((:account-stripe . 1)
-          (:human-date . 16)
+          (:human-date . 8)
           (:flags . 6) ; 3 icon flags
           (:from-or-to . 25)
           (:subject)))
--- a/modules/email/mu4e/config.el
+++ b/modules/email/mu4e/config.el
@@ -71,7 +71,7 @@
         ;; remove 'lists' column
         mu4e-headers-fields
         '((:account-stripe . 1)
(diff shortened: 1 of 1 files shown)


tecosaur:
The default date format is only 8 chars wide by default. If you set something wider in your config, it's easy enough to set this too in your config. I don't see the need for this change to be made in Doom.

johanthoren:
@tecosaur I was under the impression that the date format was based on the system locale. Now I see that I've indeed set this manually in my configuration. I apologize for missing that earlier and I agree that it is then reasonable to also set the width of the column in the configuration.

tecosaur:
No trouble. Glad you've got it working for you :slightly_smiling_face:.

PR #5559, opened 2021-10-02
Bump format-all package

jsmestad:
I cannot pull up the convention doc as the site says "Rip and tear..." so I am guessing it is down 🤞 

This updates `format-all` to the latest commit. There has been some fixes to prevent the annoying popup window on syntax error: `Error (before-save-hook): Error running hook "format-all-buffer--from-hook" because: (error Unable to format due to syntax error)` plus added support for other formatters.


hlissner:
`format-all` is one of our frozen packages, and is in our do-not-PR list.

PR #5562, opened 2021-10-02
Update PR template with discourse links

jsmestad:
I had mistakenly proposed #5559 because the URLs in the PR don't appear to show the correct info anymore. I updated them with what I believe are the correct ones so I don't do something foolish again in future contributions 🙃


hlissner:
This is actually just unfortunate timing. I've been migrating doomemacs.org to a new host + config most of today and haven't gotten to restoring all the links yet. They will be back up within the next hour or two. Sorry for the trouble!

PR #5567, opened 2021-10-03
Scoop install instructions: `scoop bucket add` requires git to work

teodorlu:
Following current installation instructions for Windows with scoop gave me this:

    PS C:\Users\teodorlu> scoop bucket add extras
    Git is required for buckets. Run 'scoop install git' and try again.

Installing git before running `scoop bucket add extras` worked fine.

The text in the docs can probably be improved, feel free to make suggestions.

### DIFF
--- a/docs/getting_started.org
+++ b/docs/getting_started.org
@@ -394,10 +394,11 @@ choco install fd llvm
 #+END_SRC
 
 Scoop will work too, but because Emacs is a GUI application you'll need to
-enable the 'extras' Scoop bucket:
+enable the 'extras' Scoop bucket, and =scoop bucket add= requires git:
 #+BEGIN_SRC sh
+scoop install git
 scoop bucket add extras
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but I must refuse it. There is nothing wrong with your PR, but it comes at a bad time: I'll be pushing a rework of all our documentation soon, which will include this information and resolve many other issues. I'll simply wait for that instead. In any case, thanks for your help!

teodorlu:
Understood.

Another note (if we go back to this PR at a later point):

```
git clone https://github.com/hlissner/doom-emacs ~/.emacs.d
```

Does not pull in the `$HOME` environment variable on Windows / Powershell 7. It clones into a folder named `~`, and creates `.emacs.d` inside it. This works as expected, when `$HOME` is set per the instructions:

```
git clone https://github.com/hlissner/doom-emacs $HOME/.emacs.d
```

PR #5568, opened 2021-10-03
feat(org): add binding for `org-clone-subtree-with-time-shift`

ethan-leba:
Adds an additional org binding, useful for repeated tasks with minor variations in time or name.


hlissner (approved):

hlissner:
LGTM. Thanks for your help!

PR #5572, opened 2021-10-04
fix(dired): ranger-travel bug

### LINKED ISSUE #5509: Interactive ranger-travel chooses wrong directory when using Vertico backend

When both the `vertico` and `(dired +ranger)` Doom packages are enabled in `init.el`, invoking the `ranger-travel` command in a `dired` buffer should allow navigating to an arbitrary directory based on the next input.

### What actually happened?

Attempting to use the `ranger-travel` command interactively with `vertico` as the search backend will always fail to select the chosen directory. Instead, the first directory or file alphabetically is always chosen.

https://user-images.githubusercontent.com/6462592/134276136-5a0f7863-52e1-44e1-9cce-c2a00f31bd07.mp4

I reverted back to `ivy` as my search backend, which resolved the issue immediately without requiring further intervention.

### Steps to reproduce

1. Configure `vertico` as the search backend via `init.el`
2. Enable the `(dired +ranger)` plugin via `init.el`
3. Invoke `doom sync` from the command line
4. Start Emacs
5. Open a dired buffer anywhere (`SPC :` => dired => `RETURN`)
6. Invoke `ranger-travel` (`f`)
7. Attempt to fill in the name of any subdirectory or file, except for the first child (alphabetically) of the current directory
8. Press `RETURN` (the chosen subdirectory/file will now be erroneously replaced by the first child)

### System Information

https://pastebin.com/raw/mkEXtSzP

iyefrat:
Overrides ranger-travel to have better mode detection for which
completion framework to use, and default to emacs built-ins instead of
ido.

In theory this should be a temporary fix until ralesi/ranger.el#236 gets
merged, but the maintenance status of ranger.el is not clear.

Fix #5509
Ref ralesi/ranger.el#236

### DIFF
--- a/modules/emacs/dired/config.el
+++ b/modules/emacs/dired/config.el
@@ -102,6 +102,26 @@ we have to clean it up ourselves."
       (with-current-buffer (window-buffer ranger-preview-window)
         (local-unset-key [mouse-1]))))
 
+  (defadvice! +dired--ranger-travel-a ()
+    "Temprorary fix for this function until ralesi/ranger.el#236 gets merged."
+    :override #'ranger-travel
+    (interactive)
+    (let ((prompt "Travel: "))
+      (cond
+       ((bound-and-true-p helm-mode)
+        (ranger-find-file (helm-read-file-name prompt)))
+       ((bound-and-true-p ivy-mode)
+        (ivy-read prompt 'read-file-name-internal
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Thanks again!

PR #5573, opened 2021-10-04
fix(mu4e): make alerts work again

tecosaur:
At some point in the original PR alerts stopped working. I've finally
tracked down why and found three components to the problem:
- I'm calling `start-process` incorrectly
- `truncate-string-to-width` is called on `nil` when no from name is
  given
- mu4e-alert uses :docid to check for duplicates, which is not provided
  in mu 1.6, and so :message-id must be used instead.

### DIFF
--- a/modules/email/mu4e/config.el
+++ b/modules/email/mu4e/config.el
@@ -616,6 +616,19 @@ See `+mu4e-msg-gmail-p' and `mu4e-sent-messages-behavior'.")
   (mu4e-alert-enable-mode-line-display)
   (mu4e-alert-enable-notifications)
 
+  (when (version<= "1.6" mu4e-mu-version)
+    (defadvice! +mu4e-alert-filter-repeated-mails-fixed-a (mails)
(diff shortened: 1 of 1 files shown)


modules/email/mu4e/config.el
```diff
@@ -616,6 +616,19 @@ See `+mu4e-msg-gmail-p' and `mu4e-sent-messages-behavior'.")
   (mu4e-alert-enable-mode-line-display)
   (mu4e-alert-enable-notifications)
 
+  (when (version<= "1.6" mu4e-mu-version)
+    (defun +mu4e-alert-filter-repeated-mails-fixed (mails)
```
hlissner, line 620: Was this supposed to be a `defadvice!`?

tecosaur: very much so.

hlissner (approved):

hlissner:
LGTM. Thanks again!

PR #5576, opened 2021-10-04
nit: init.example.el: wording, sort

dschrempf:
Minor change.

I added `biblio`, because it was missing.

I sorted two other entries.

I changed the wording of the first sentence. Please let me know if you don't wnat this change, but it sounds better to me.

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -1,7 +1,7 @@
 ;;; init.el -*- lexical-binding: t; -*-
 
-;; This file controls what Doom modules are enabled and what order they load
-;; in. Remember to run 'doom sync' after modifying it!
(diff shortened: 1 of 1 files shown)


iyefrat:
Minor point, but the `biblio` part of this pr is taken care of in #5290

dschrempf:
OK great, thank you! I assume #5290 will be merged some time so I remove `biblio` here.

hlissner:
Thanks for the PR, but I cannot accept it. After the upcoming docs rewrite, the init.example.el file will be automatically generated from modules' README's, as such it will no longer be a file that we need to modify ourselves. Your changes will be swallowed up by that so I'd rather wait for that. Sorry for the trouble, but thanks for your help!

PR #5578, opened 2021-10-04
tweak(default): improve smartparens vanilla bindings

### LINKED ISSUE #5575: Improve smartparens emacs bindings

sexp movements bindings should not be more incompatible than necessary with traditional ones, except for a very good reason.

sexp movements should include basic functionality like going up and down a sexp, and moving to the next and previous list sexp item.

### What actually happened?

[Traditionally](https://www.gnu.org/software/emacs/manual/html_node/emacs/Moving-by-Parens.html) there has been a set of list (parenthetical group) movements:

* C-M-d/u: down/up list
* C-M-n/p: forward/backward list

These are very important bindings.

But in doom emacs:

* They aren't assigned to similar smartparens functions.
* C-M-d has been replaced with a comparatively exotic binding.

```
      (:after smartparens
        :map smartparens-mode-map
        "C-M-a"     #'sp-beginning-of-sexp
        "C-M-e"     #'sp-end-of-sexp
        "C-M-f"     #'sp-forward-sexp
        "C-M-b"     #'sp-backward-sexp
        "C-M-d"     #'sp-splice-sexp
        "C-M-k"     #'sp-kill-sexp
        "C-M-t"     #'sp-transpose-sexp)
```

I just rebound everything in the most conservative fashion:

* C-M-d/u: sp-down/up-sexp
* C-M-n/p: sp-next/previous-sexp
* C-M-f/b: sp-forward/backward-sexp
* C-M-a/e: sp-beginning/end-of-sexp
* C-M-k: sp-kill-sexp

I think doom emacs should be as close to these semantics as possible because these are similar to the traditional meanings and they have always made sense (this is not a case of having to correct archaic defaults at all).

Of course doom could add additional functionality if so desired, but please don't override the basic ones.

See also: https://ebzzry.com/en/emacs-pairs/ (but don't use the bindings suggested there, like C-up/down, which require key arrows, are already taken by the macOS desktop and, furthermore, are unnecessarily at odds with tradition).

### Steps to reproduce

I don't think this is necessary.

### System Information

https://pastebin.com/QBtDe97W

memeplex:
Make bindings better conform to traditional sexp navigation ones.

Fix #5575

@hlissner I've chosen `C-M-<backspace>` for the non-standard `sp-splice-sexp`, I believe it's worth keeping a binding for the function but if you believe it's not the best key please let me know, at least it's not yet used and has mnemonic value.

### DIFF
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -613,13 +613,17 @@
       ;;; smartparens
       (:after smartparens
         :map smartparens-mode-map
-        "C-M-a"     #'sp-beginning-of-sexp
-        "C-M-e"     #'sp-end-of-sexp
-        "C-M-f"     #'sp-forward-sexp
-        "C-M-b"     #'sp-backward-sexp
-        "C-M-d"     #'sp-splice-sexp
-        "C-M-k"     #'sp-kill-sexp
-        "C-M-t"     #'sp-transpose-sexp)
+        "C-M-a"           #'sp-beginning-of-sexp
+        "C-M-e"           #'sp-end-of-sexp
+        "C-M-f"           #'sp-forward-sexp
+        "C-M-b"           #'sp-backward-sexp
+        "C-M-n"           #'sp-next-sexp
+        "C-M-p"           #'sp-previous-sexp
+        "C-M-u"           #'sp-up-sexp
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
LGTM. Thanks for your help!

PR #5585, opened 2021-10-05
tweak(latex): add key bindings for compilation

dschrempf:
This is a minor change.

Doom Emacs was missing key bindings to compile LaTeX documents. This PR adds tow bindings in LaTeX derived modes:
- `SPC m c` for `TeX-command-run-all`, and
- `SPC m m` for `TeX-command-master`.


hlissner:
Not a big fan of the code duplication, but I understand why it's necessary. I'll have to look into that sometime. In any case, thanks for the help!

PR #5599, opened 2021-10-08
fix(go): company-go pin after melpa recipe change

marienz:
melpa/melpa@552d033e573f updated the recipe for company-go to pull from
emacsattic instead of mdempsky/gocode, which does not contain the commit
Doom pins company-go to.

Pin to the latest commit in the new repository instead, which is one
commit behind the previous repo (it does not have
mdempsky/gocode@f531cad26218, but it looks like this does not matter for
Doom).


hlissner:

PR #5603, opened 2021-10-10
tweak(elfeed): Move elfeed-goodies behind +goodies flag

bhankas:
Elfeed-goodies package does not provide much functionality besides visual flair. Some people (like myself) find spartan stock elfeed interface more appealing/preferable compared to eye-candy. But, extra visual flair behind an easy flag (+pretty) should be enough to satisfy both camps.

### DIFF
--- a/modules/app/rss/README.org
+++ b/modules/app/rss/README.org
@@ -30,7 +30,8 @@ This module has no dedicated maintainers.
 
 ** Plugins
 + [[https://github.com/skeeto/elfeed][elfeed]]
-+ [[https://github.com/algernon/elfeed-goodies][elfeed-goodies]] 
++ =+goodies=
+  + [[https://github.com/algernon/elfeed-goodies][elfeed-goodies]]
 + =+org=
   + [[https://github.com/remyhonig/elfeed-org][elfeed-org]]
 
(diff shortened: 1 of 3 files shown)


jeetelongname:
We went through a similar thing back over in the mu4e pr. 

While a flag is more explicit its still just one package. If it was a group I would understand but a simple 

```lisp
(package! elfeed-goodies :disable t) 
```

is all thats needed. If and when the package grows it would make more sense but now its not really an issue IMHO

jeetelongname:
That being said if a flag was to be used `+goodies` would make more sense for the name.

and this is sort of a meta comment but as this is introducing new behavior into the code the type `feat` should be used instead. It should be as simple as using an interactive rebase to fix it.

jeetelongname:
did your configuration break after you disabled the package or was that with goodies only? if its the former that should be brought up with the elfeed maintainer

jeetelongname:
Oh I see. goodies needs a bit of work (which I will get around to eventually...) but an extra line of code in my many thousand line config does not really bother me. I do a similar thing with the mu4e module with disabling the notification package. I can see how it will bother people though. again we wait for Henrik's opinion on this matter

PR #5608, opened 2021-10-10
fix: +vc/browse-at-remote-file-or-region -> +vc/browse-at-remote

linkenneth:
+vc/browse-at-remote-file-or-region no longer exists. +vc/browse-at-remote seems like the reasonable replacement.


hlissner:

PR #5614, opened 2021-10-12
feat(eshell): add vertico completion for eshell history

jwillhite:
Switching to vertico results in +eshell/search-history defaulting to eshell-list-history, using completing-read allows us to tie into vertico completions which are much nicer.


minad:
You can use `consult-history` for that, which has more features. cc @iyefrat

hlissner:
Thanks for your help!

PR #5621, opened 2021-10-12
refactor(latex): nitpick, table d'hôte

dschrempf:
This is a huge set of pedantic changes, only few of them affecting
actual code (only `config.el` has some changes in the code).

Mostly, I tried to:
- use consistent names; use LaTeX wherever possible, not latex;
  `LaTeX-mode` is an alias for `latex-mode`, but AUCTeX uses capitalized
  `LaTeX` for all other variables, hooks, and maps. So I think it is
  cleaner to stick with this nomenclature, whenever it is possible. I did
  not change any variables/functions defined by Doom, but that could be
  done in a future PR;
- fix broken symbol links;
- use capitalized comments with full stops.

If you are not interested in changes like these, let me know. I can also
split this up, if you want. In my opinion, however, it does improve the
code base!

### DIFF
--- a/modules/lang/latex/+fontification.el
+++ b/modules/lang/latex/+fontification.el
@@ -1,11 +1,11 @@
 ;;; lang/latex/+fontification.el -*- lexical-binding: t; -*-
 
-;; Fontification taken from https://tex.stackexchange.com/a/86119/81279
+;; Fontification taken from https://tex.stackexchange.com/a/86119/81279.
 (setq font-latex-match-reference-keywords
-      '(;; biblatex
+      '(;; BibLaTeX.
         ("printbibliography" "[{")
         ("addbibresource" "[{")
-        ;; Standard commands
+        ;; Standard commands.
         ("cite" "[{")
         ("citep" "[{")
         ("citet" "[{")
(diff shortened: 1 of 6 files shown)


hlissner:
Thanks for the PR, but I'll have to turn it down. There are two breaking changes:

1. Where you replace `'latex-mode` with `'LaTeX-mode` and 
2. Remove the `latex-mode-map` keybinds). 

I'm aware the two modes are aliases, but the redundancy was intentional. Our `:completion company` module uses the value of `major-mode` to determine what backends to associate with a buffer, and `major-mode` == `'latex-mode` in LaTeX buffers. We have a similar problem with which-key, which is why we repeat those latex-mode-map keybinds. It's not elegant, but they are necessary.

Besides that, there are some superfluous formatting changes:

- Adding periods to comments that aren't full sentences.
- Capitalizing keywords that should be lowercase (amsmath, enumitem, subcaption, etc).
- Adding a colon to `TODO`, which is done nowhere else in Doom's source code
- Changing `:` to `.` on some comment headers.

There are a few things I do like about this PR, like the fixed symbol linkage, and spelling/grammar corrections, but overall, the changes seem too aggressive. If you're feeling up to correcting these, I will happily reopen this, but I'll close this for now. In any case, thanks for your help.

dschrempf:
Hi! Thanks for your feedback!

I did this some time ago and don't remember where I introduced braking changes into the code. I will have a look! Would you accept a PR only changing comments?

hlissner:
> Hi! Thanks for your feedback!
> 
> I did this some time ago and don't remember where I introduced braking changes into the code. I will have a look! Would you accept a PR only changing comments?

Certainly.

PR #5625, opened 2021-10-13
nit(&early-init): typo

dschrempf:
bene -> been


hlissner:
This comment block will be rewritten once I backport early-init.el from my local CLI rewrite branch, so I'll have to turn it down. Thanks for the help in any case.

PR #5626, opened 2021-10-13
fix: when use golang lsp, not use guru and godoc

Patrick0308:
## Issue
When we use golang lsp, use guru and godoc to find references or documentation.

## To reproduce
configure (go +lsp)
press `gD` use guru to find references

## Fix
When we use golang lsp, not use guru and godoc

### DIFF
--- a/modules/lang/go/config.el
+++ b/modules/lang/go/config.el
@@ -6,10 +6,6 @@
 (after! go-mode
   (set-docsets! 'go-mode "Go")
   (set-repl-handler! 'go-mode #'gorepl-run)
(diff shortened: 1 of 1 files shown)


hlissner:
I don't understand why this is necessary. Due to the order of `set-lookup-handlers!` in the `:tools lsp` module (here) and `:lang go` (here), gD should use LSP when lsp-mode (or eglot) are active, and fall back to guru/godoc otherwise. This change should not be necessary, and may indicate an unrelated issue with your config. Can you confirm?

Patrick0308:
I updated doom emacs. And I can not reproduce the issue.

Issue #5580, opened 2021-10-05
doom-big-font-mode increase org-format-latex-options :scale


io12:
### Describe your request

It would be useful if `doom-big-font-mode` increased the size of `org-mode` fragment previews, instead of just increasing the size of text. This will probably require two new variables for the two fragment scale factors. (`doom-org-latex-scale` and `doom-org-latex-scale-big`?)

### Briefly explain its use-case

This could be useful for demonstrations of `org-mode` or for showing people LaTeX renderings more easily.

hlissner:
I'll welcome a PR to modify the `:lang org` module to add a hook to `doom-big-font-mode-hook` that does the necessary scale adjustments on the fly based off of `doom-big-font-increment` (I'd much rather avoid introducing new variables if I don't have to). 

That said, this will be more complicated than simply just modifying `org-format-latex-options`. It will need to iterate over all visible fragments in all open org buffers and regenerate them all, which sounds like a very bad time in a larger file. An extra effort to defer "regeneration" of buried org buffers would be appreciated.

io12:
Thanks for the feedback! I have a patch for this feature mostly working but I can't find a hook that runs when buffers become visible. It seems like `window-configuration-change-hook` would work, but it's being called for all buffers instead of just visible ones.

PR #5627, opened 2021-10-13
Add Org LaTeX fragment resizing for `doom-big-font-mode`

io12:
Fixes https://github.com/hlissner/doom-emacs/issues/5580

### DIFF
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -1323,4 +1323,47 @@ compelling reason, so..."
       (remove-hook 'post-command-hook #'+org-play-all-gifs-h t)
       (pcase +org-startup-with-animated-gifs
         (`at-point (add-hook 'post-command-hook #'+org-play-gif-at-point-h nil t))
-        (`t (add-hook 'post-command-hook #'+org-play-all-gifs-h nil t))))))
+        (`t (add-hook 'post-command-hook #'+org-play-all-gifs-h nil t)))))
+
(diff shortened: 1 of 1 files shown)


io12:
I didn't know about `image-scaling-factor` but I think it is better to regenerate the fragments at the higher resolution so they aren't blurry.

benneti:
Just to chime in, as it is possible to use svg (dvisvgm instead of dvipng in org-preview-latex-process-alist) it might be simpler to only scale and not regenerate but switch the default in doom from png to svg (although I am not sure whether all supported emacs versions have svg support by default already).

benneti:
``` elisp
(defadvice! +org-latex-preview-scale-a (img)
  "Make org latex preview respect `image-scaling-factor'."
  :filter-return #'org--make-preview-overlay
  (plist-put (cdr img) :scale (image-compute-scaling-factor image-scaling-factor)))
(defun org-latex-update-fragment-scale ()
  ;; remove latex previews
  (org-latex-preview '(64))
  ;; rescale compiled images
  (setq image-scaling-factor (expt text-scale-mode-step text-scale-mode-amount))
  ;; recreate latex previews
  (org-latex-preview '(16)))
(add-hook! 'text-scale-mode-hook #'org-latex-update-fragment-scale)
```
This works for me, mind the redisplay of fragments, as well as the expt instead of math-pow (the latter failed for me for negative scales)

hlissner:
This was auto-closed because I deleted the `develop` branch, but I'm still interested in the functionality. I will revisit it soon.

PR #5638, opened 2021-10-14
fix: update to `go install` instead of `go get`

KarthikNayak:
Currently the dependencies listed use `go get`, but this is soon to be deprecated. Replace
the `go get` commands with `go install` command.


iyefrat:
Hey, can you change your commit from `fix:` to `fix(go):`? It's in the guidelines :)
[Here's](https://gist.github.com/hlissner/4d78e396acb897d9b2d8be07a103a854) a version that doesn't require registration btw

hlissner:
Thanks for your help!

PR #5639, opened 2021-10-14
docs: note on Spacemacs' hybrid mode

### LINKED ISSUE #107: Hybrid evil and emacs keybindings?
This is just a query, probably more so regarding evil-mode than doom, in case anyone has suggestions.

Being a long time emacs user, I've enjoyed learning vim style modal editing using spacemacs because of the 'hybrid-mode' keybinding option which maintains emacs keybindings while in insert mode.  I find I am not yet proficient enough to do away with emacs bindings altogether, but don't want to give up evil mode either.  Is this a behavior that I can replicate in doom?  I wasn't sure if this feature comes out of the box with evil, and I just need to learn to configure it, or if spacemacs has added a lot of customization to make this possible.

Thanks for any suggestions.

iNecas:
Getting to Doom after few years in Spacemacs, I really enjoyed the hybrid mode.
I've found old issue about it https://github.com/hlissner/doom-emacs/issues/107,
and the workarounds worked, but in exact way the issue was describing. The
issue itself was already locked, so couldn't add my notes there, so I figured
opening a PR with the notes might save folks some time to figure this config
out.

Fix #107

### DIFF
--- a/docs/getting_started.org
+++ b/docs/getting_started.org
@@ -678,6 +678,13 @@ our Discord server]].
 #+end_quote
 
 ** TODO From Spacemacs
+
+Some notes from and for users migrating from Spacemacs:
+
++ One can achieve behavior similar to [[https://www.spacemacs.org/doc/DOCUMENTATION.html#hybrid][hybrid editing style]] by putting
+  ~(setq evil-disable-insert-state-bindings t)~ into =~/.doom.d/init.el=
+  file (needs to be set early in the init process to take effect).
+
 #+begin_quote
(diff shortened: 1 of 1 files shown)


hlissner:
Our docs are getting a rewrite. We'll have a dedicated migration guide then -- I'll incorporate this PR there, but for the time being I'll have to turn down this PR. In any case, thanks for the help!

iNecas:
np at all, the main reason for my to open the PR was to link it to the archived issue I couldn't comment on, which had some invalid recommendations that others might be mislead with (as I was)

PR #5648, opened 2021-10-15
Let +phpunit-docker-compose-mode be disabled

elken:
Currently no simple way to disable this, this PR aims to make the functionality
opt-in behind a flag


hlissner:
Thanks for your help!

Issue #5649, opened 2021-10-16
LSP session file stored in unexpected location


mikonieminen:
### What did you expect to happen?

When I remove `~/.emacs.d/.local/cache` folder, I would expect old LSP session data to be removed.

### What actually happened?

LSP session data was preserved and it was not obvious. This caused unexpected behavior when debugging LSP using modes (rustic in this case).

### Describe your attempts to resolve the issue

Remove `~/.emacs.d/.local/etc/lsp-session` and old session data got removed as expected.

### Steps to reproduce

1. open Emacs
2. open a project that uses LSP, for example https://github.com/serde-rs/serde
3. check `~/.emacs.d/.local/etc/lsp-session` and you can find this project listed in the file

### System Information

https://pastebin.com/Mi9xVDe6

PR #5650, opened 2021-10-16
Move lsp-session under doom cache dir

mikonieminen:
Fixes hlissner/doom-emacs#5649

Move `lsp-session` under `~/.emacs.d/.local/cache` instead of keeping it under `~/.emacs.d/.local/etc`.


hlissner:
Thanks for your help!

PR #5662, opened 2021-10-20
markdown: evil normal: map <RET> to markdown-do

edrex:
this does two things:

- Maps `<RET>` to markdown-do (same as spacemacs)
- Updates markdown-mode, to get https://github.com/jrblevin/markdown-mode/pull/666 which enables following `[markdown](links)` and `[[wiki|links]]` with markdown-do

### DIFF
--- a/modules/lang/markdown/config.el
+++ b/modules/lang/markdown/config.el
@@ -127,6 +127,7 @@ capture, the end position, and the output buffer.")
   (map! :map evil-markdown-mode-map
         :n "TAB" #'markdown-cycle
         :n [backtab] #'markdown-shifttab
+        :n "<RET>" #'markdown-do
         :i "M-*" #'markdown-insert-list-item
         :i "M-b" #'markdown-insert-bold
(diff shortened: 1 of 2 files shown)


modules/lang/markdown/config.el
```diff
@@ -127,6 +127,7 @@ capture, the end position, and the output buffer.")
   (map! :map evil-markdown-mode-map
         :n "TAB" #'markdown-cycle
         :n [backtab] #'markdown-shifttab
+        :n "<RET>" #'markdown-do
```
hlissner, line 130: Please remove this keybinding. It is already [bound by evil-collection-markdown-mode](https://github.com/emacs-evil/evil-collection/commit/1d296af2ec45e98b93048148d5468b05385b1ff6#) (evil-collection was bumped a short while ago, in https://github.com/hlissner/doom-emacs/commit/864563460805dc2d460661ef59711474cd447e75).

hlissner (changes requested):

PR #5695, opened 2021-10-27
Recall `mu4e--start` interactively inside `+mu4e-lock-start`

### LINKED ISSUE #5689: [mu4e] mu4e sometimes doesn't open even if the lock file doesn't exist

When opening mu4e, the "booting" process first checks if a lock file is available by calling to `+mu4e-lock-start` which further calls to `+mu4e-lock-available`.
~~If a lock file does not exist, `+mu4e-lock-available` should return `nil`.~~ (I got it wrong)

### What actually happened?

`mu4e` does not open.

Turn on the debugger, put a breakpoint at `+mu4e-lock-available`.

The debugger shows the implementation of this function as 
```elisp
(not (let* ((lock-info (and t (+mu4e-lock-pid-info))) 
            (pid (and lock-info (car lock-info)))) 
           (if pid (if (or strict (/= (emacs-pid) pid)) (progn t)))))
```
Even though `+mu4e-lock-pid-info` returns `nil` as expected, the above implementation still returns `t`.

### Steps to reproduce

This behavior happens sometimes (if I restart emacs then everything works just fine).

### System Information

https://pastebin.com/m1na30qF

vqhuy:
This seems to fix #5689

`mu4e--start` should be called interactively, as it is done in `=mu4e`.
It fixes a problem when sometimes `mu4e` does not open its main view.

### DIFF
--- a/modules/email/mu4e/autoload/mu-lock.el
+++ b/modules/email/mu4e/autoload/mu-lock.el
@@ -54,7 +54,7 @@ Else, write to this process' PID to the lock file"
     (write-region (number-to-string (emacs-pid)) nil +mu4e-lock-file)
     (delete-file +mu4e-lock-request-file)
     (call-process "touch" nil nil nil +mu4e-lock-request-file)
-    (funcall orig-fun callback)
+    (funcall-interactively orig-fun callback)
     (setq +mu4e-lock--request-watcher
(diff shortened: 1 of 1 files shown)


vqhuy:
This doesnt really fix my issue, close it for now.

PR #5704, opened 2021-10-29
display dired-icons with colors

teeann:
Module `(dired +icons)` currently display icons in monochrome (i.e black & white). 
This PR aims to provide a better default UI with colored dired icons.

### DIFF
--- a/modules/emacs/dired/config.el
+++ b/modules/emacs/dired/config.el
@@ -119,6 +119,9 @@ we have to clean it up ourselves."
   ;;      wdired-mode.
   (defvar +wdired-icons-enabled -1)
 
+  ;; display icons with colors
+  (setq all-the-icons-dired-monochrome nil)
+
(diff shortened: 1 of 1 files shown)


iyefrat:
Fwiw, none of the other places doom uses icons have them in monochrome (ibuffer, vertico, ivy), so adding this would make it more consistent. Might be worth adding a `+icons-monochrome` variable to globally set this on or off, although that would require upstream support for that (and maybe moving doom's buffer icons to `all-the-icons-ibuffer` instead of adding it locally).

hlissner:
Sorry for the tremendously late response. Though I'm not a fan of the aesthetic, this makes sense to me for consistency. In any case, thanks for the help!

iyefrat:
@hlissner this got merged onto develop

hlissner:
My bad! I've now cherry picked it onto master: https://github.com/hlissner/doom-emacs/commit/8342db4f35c656c6d7572c0c0d2a6ec0f47dad94

PR #5705, opened 2021-10-29
nit: Update emacs version

HaoZeke:
The previous version is no longer valid.


HaoZeke:
Oops, sorry for the noise, didn't realize the discourse was up after logging in.

PR #5708, opened 2021-10-30
Fix commit linter bugs

iyefrat:
dev(ci): fix false negative in commit linter

  The one value allowed per trailer check shouldn't apply to names since
  they often have spaces.

dev(ci): allow periods in names in commit linter

  For initials.


core/cli/ci.el
```diff
@@ -219,7 +219,8 @@ representing the current commit being checked against. See
                        (truncate-string-to-width (string-trim line) 16 nil nil "…")
                        (match-string 1 line))))
             (pcase-dolist (`(,key . ,value) trailers)
-              (if (string-match-p " " value)
+              (if (and (not (member 'name (alist-get key doom-cli-commit-trailer-keys nil nil #'equal)))
```
hlissner, line 222: use `(memq 'name (cdr (assoc key doom-cli-commit-trailer-keys)))` instead of `(member ...)`.

hlissner (changes requested):

core/cli/ci.el
iyefrat: Done!

hlissner (approved):

Issue #5046, opened 2021-05-13
toogle-input-method is broken


ineu:
### What did you expect to happen?

`toggle-input-method` works as before

### What actually happened?

113dcd74cf4f7e064990cda907b7e655bed197b0 breaks toggle-input-method for me. It used to properly toggle between default and 'ru' locale. Now instead of a locale it shows the lowercase 'm' in the status line. I still can explicitly set the method via `set-input-method`, then it works fine. It's only the default (right after emacs is started) that has changed.

For now I have added `(set-input-method "russian-computer")` to my config as a workaround. However, the previous autodetection behavior was more convenient.

### Steps to reproduce

1. Try to toggle the input method via `C-\` right after emacs starts

### System Information

https://pastebin.com/9mV0P8vs

dukebarman:
Same problem on macOS

hlissner:
This issue should now be resolved. Let me know if that isn't the case and I'll reopen it. Thanks for bringing it to my attention!

ineu:
The solution doesn't actually restore the previous behavior for me. Now, when I press `C-\`, it asks for the input method, no auto-detection. That's expected as the patch unsets the default-input-method, but this is not how it used to work.

To get the old behavior, I updated the eary-init.el file, reverting both the recent patch and https://github.com/hlissner/doom-emacs/commit/113dcd74cf4f7e064990cda907b7e655bed197b0:
```diff
diff --git a/early-init.el b/early-init.el
index da91ba904..4bc305f43 100644
--- a/early-init.el
+++ b/early-init.el
@@ -57,10 +57,10 @@
 
 ;; Contrary to what many Emacs users have in their configs, you don't need
 ;; more than this to make UTF-8 the default coding system:
-(set-language-environment "UTF-8")
+;; (set-language-environment "UTF-8")
 
 ;; set-language-enviornment sets default-input-method, which is unwanted
-(setq default-input-method nil)
+;; (setq default-input-method nil)
 
 ;; Ensure Doom is running out of this file's directory
 (setq user-emacs-directory (file-name-directory load-file-name))
```
That makes the auto-detection work. 

Looks like the language environment is set by Emacs itself, before the Doom's initialization starts. That's what I get in `emacs -Q`:
```
current-language-environment is a variable defined in ‘mule-cmds.el’.

Its value is "Russian"
Original value was "English"
```

PR #5709, opened 2021-10-30
fix: unset default-input-method in early-init.el

iyefrat:
set-language-environment sets default-input-method, which is unwanted.

Fix: #5046

-------

@ineu @dukebarman please confirm on your end that this fixes the issue. It unsets the default input method, but I never had the locale autodetection thing, so I'm not sure if that's still a problem.

### DIFF
--- a/early-init.el
+++ b/early-init.el
@@ -59,6 +59,9 @@
 ;; more than this to make UTF-8 the default coding system:
 (set-language-environment "UTF-8")
 
+;; set-language-enviornment sets default-input-method, which is unwanted
+(setq default-input-method nil)
+
 ;; Ensure Doom is running out of this file's directory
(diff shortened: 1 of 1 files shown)


hlissner (approved):

dukebarman:
Sorry for the delay, maybe you're right. It doesn't work for me but I've moved to reverse-im package with this config:
```
 (use-package! reverse-im
  :custom
  (reverse-im-input-methods '("russian-computer"))
  :config
  (reverse-im-mode t)
  (after! evil
    (define-key evil-normal-state-map (kbd "C-х") #'evil-force-normal-state)
    (define-key evil-insert-state-map (kbd "C-х") #'evil-normal-state)
    (define-key evil-visual-state-map (kbd "C-х") #'evil-exit-visual-state)))
```

PR #5722, opened 2021-11-02
fix: wrong number of arguments when calling (dtrt-indent-mode)

daanturo:
`doom--fix-broken-smie-modes-a` forces `dtrt-indent-mode` to expect exactly one argument so an error will be raised when the latter is turned on without arguments programmatically in lisp.

### DIFF
--- a/core/core-editor.el
+++ b/core/core-editor.el
@@ -499,7 +499,7 @@ files, so this replace calls to `pp' with the much faster `prin1'."
   (push '(t tab-width) dtrt-indent-hook-generic-mapping-list)
 
   (defvar dtrt-indent-run-after-smie)
-  (defadvice! doom--fix-broken-smie-modes-a (fn arg)
+  (defadvice! doom--fix-broken-smie-modes-a (fn &optional arg)
     "Some smie modes throw errors when trying to guess their indentation, like
 `nim-mode'. This prevents them from leaving Emacs in a broken state."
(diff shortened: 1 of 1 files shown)


core/core-editor.el
```diff
@@ -499,7 +499,7 @@ files, so this replace calls to `pp' with the much faster `prin1'."
   (push '(t tab-width) dtrt-indent-hook-generic-mapping-list)
 
   (defvar dtrt-indent-run-after-smie)
-  (defadvice! doom--fix-broken-smie-modes-a (fn arg)
+  (defadvice! doom--fix-broken-smie-modes-a (fn &rest args)
```
hlissner, line 502: Please use `&optional arg` instead.

core/core-editor.el
```diff
@@ -512,7 +512,7 @@ files, so this replace calls to `pp' with the much faster `prin1'."
                          (message "[WARNING] Indent detection: %s"
                                   (error-message-string e))
                          (message ""))))) ; warn silently
-        (funcall fn arg)))))
+        (apply fn args)))))
```
hlissner, line 515: And revert back to `(funcall fn arg)`.

hlissner (changes requested):

hlissner (approved):

hlissner:
Thanks for your help!

PR #5724, opened 2021-11-03
fold: Fix folding interactions with markdown-mode

linkenneth:
This PR implements the following:

1. Support `outline-minor-mode` in `+fold/close-all`. It appeared to me that the rest of the functions in this file supported `outline-minor-mode` folding except this one, so I added it. (https://github.com/hlissner/doom-emacs/commit/48a42b2d0829fdaa57b66b772d0e2ef8b46f55e9)
2. Adds `outline-minor-mode` as a hook to `markdown-mode` so that `zm`, `zc` etc. bindings automatically work with markdown headers. This works in a very similar way to how `org-mode` folding works (separate question, is there a reason we need to have dedicated `org-mode` functions for folding? can we not use `outline-minor-mode`?) (https://github.com/hlissner/doom-emacs/commit/bc01467deffea981932dc8408e8d6ce21604f288)
3. Aligns the behavior of `zo` with `zc` for `outline-minor-mode` folds. It appeared to me that `zc` was closing recursively, but `zo` was not. This means that a single `zc` could cause one to need to `zo` multiple times. This is a problem primarily because the `fold` module doesn't differentiate between recursive and non-recursive folding yet. (https://github.com/hlissner/doom-emacs/commit/3d6358ab7df54e9c3cf7bae1e346540feb0ebb78)


hlissner:
> Support outline-minor-mode in +fold/close-all. It appeared to me that the rest of the functions in this file supported outline-minor-mode folding except this one, so I added it. 

I'm on the fence with this one. Outline support was intentionally left out of those commands because outline folding overrides _all_ other folding, which generally isn't very useful (markdown-mode may be the exception, rather than the rule). I'll have to sit on that.

> Adds outline-minor-mode as a hook to markdown-mode so that zm, zc etc. bindings automatically work with markdown headers. This works in a very similar way to how org-mode folding works

Makes sense to me!

> (separate question, is there a reason we need to have dedicated org-mode functions for folding? can we not use outline-minor-mode?)

This is better directed at the org-mode devs. After all, org-mode is derived from outline-mode, yet they wrote their own folding commands on top of outline. I haven't looked too deeply into it; I simply assume they provide more specialized support for org buffers and didn't want more corner cases to deal with down the road (virtually inevitable with any org configuration).

> Aligns the behavior of zo with zc for outline-minor-mode folds. It appeared to me that zc was closing recursively, but zo was not. This means that a single zc could cause one to need to zo multiple times. This is a problem primarily because the fold module doesn't differentiate between recursive and non-recursive folding yet.

I _think_ there was a reason I designed it this way, but it eludes me. I'll give this PR a test run this weekend and look into it.

linkenneth:
> > Support outline-minor-mode in +fold/close-all. It appeared to me that the rest of the functions in this file supported outline-minor-mode folding except this one, so I added it.
> 
> I'm on the fence with this one. Outline support was intentionally left out of those commands because outline folding overrides _all_ other folding, which generally isn't very useful (markdown-mode may be the exception, rather than the rule). I'll have to sit on that.
> 
> > Adds outline-minor-mode as a hook to markdown-mode so that zm, zc etc. bindings automatically work with markdown headers. This works in a very similar way to how org-mode folding works
> 
> Makes sense to me!
> 
> > (separate question, is there a reason we need to have dedicated org-mode functions for folding? can we not use outline-minor-mode?)
> 
> This is better directed at the org-mode devs. After all, org-mode is derived from outline-mode, yet they wrote their own folding commands on top of outline. I haven't looked too deeply into it; I simply assume they provide more specialized support for org buffers and didn't want more corner cases to deal with down the road (virtually inevitable with any org configuration).
> 
> > Aligns the behavior of zo with zc for outline-minor-mode folds. It appeared to me that zc was closing recursively, but zo was not. This means that a single zc could cause one to need to zo multiple times. This is a problem primarily because the fold module doesn't differentiate between recursive and non-recursive folding yet.
> 
> I _think_ there was a reason I designed it this way, but it eludes me. I'll give this PR a test run this weekend and look into it.

Have you had a chance to try this PR out over the weekend? What do you think?

linkenneth:
Closing this as this has been open for too long and I'd rather not deal with merge conflicts. If this is still an interesting idea, I'd be happy to create another PR for it.

PR #5725, opened 2021-11-03
change to gplv3 license

leemandmem:
Fixes #0000 


tecosaur:
> {{{ Summarize what you've changed HERE and why }}}

gagbo:
The very short response is "It's not your code, so it's not your choice to change the license", because of this simple rule (let's call it "intellectual property", or "attribution rules" or "who does, chooses"), this code change makes no sense.

PR #5733, opened 2021-11-04
feat(doom-quit): hide some messages with +friendly

ethan-leba:
I feel that some of the default `doom-quit` messages are a little offputting for those who aren't aware of the references. I think doom emacs is an amazing framework and I wouldn't want any newcomers to be turned off by some of the more aggressive messages when they already may be feeling overwhelmed by learning (doom) emacs. I also understand others may like these messages, so I've come up with the following compromise:

1. Add `+friendly` flag to `doom-quit`. When `+friendly` is enabled, hide some of the quit messages.
2. Add `+friendly` to the default `init.el`.

Current doom users will be unaffected but may opt-in with the flag, while newcomers will have `+friendly` enabled by default and must opt-out.

### DIFF
--- a/docs/modules.org
+++ b/docs/modules.org
@@ -202,7 +202,7 @@ Aesthetic modules that affect the Emacs interface or user experience.
 + [[file:../modules/ui/deft/README.org][deft]] - TODO
 + [[file:../modules/ui/doom/README.org][doom]] - TODO
 + [[file:../modules/ui/doom-dashboard/README.org][doom-dashboard]] - TODO
-+ [[file:../modules/ui/doom-quit/README.org][doom-quit]] - TODO
++ [[file:../modules/ui/doom-quit/README.org][doom-quit]] =+friendly= - TODO
 + [[file:../modules/ui/emoji/README.org][emoji]] =+ascii +github +unicode= - Adds emoji support to Emacs
 + [[file:../modules/ui/hl-todo/README.org][hl-todo]] - TODO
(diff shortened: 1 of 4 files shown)


ethan-leba:
I'm open to suggestions as to what exactly should be hidden by `+friendly` as well, some of the choices were a bit abritrary -- it was the two portal messages that spurred me to make this PR.

jeetelongname:
I am curious as to what you find off putting about those  quotes? Coming from someone who has not interacted with those games as well they did not really bother me and I am not sure if I see how they are unfriendly. (Of course I am but one person so this is anecdotal)

ethan-leba:
Some of these perhaps shouldn't be hidden (open for discussion), but as an example:
> I don't know what you think you are doing, but I don't like it. I want you to stop.

Personally feels unnerving and unpleasant when I'm simply trying to restart my editor or whatnot.

jeetelongname:
Yeah discussing it on the discord server i can see why people would be disuaded by these messages, I guess it would make sense to just disable the module by default as whatever we put behind that flag will be arbitrary in one sense or the other and its not like this module is important

hlissner:
Instead of introducing a flag, I've decided to simply disable the module by default. It doesn't really offer any value to warrant it being a default anyhow. At the very least, those messages shouldn't catch folks by surprise anymore, so I'll consider this resolved. Thanks for bringing it to my attention!

PR #5752, opened 2021-11-09
refactor(notmuch): parameterize delete/spam tags

leezu:
Introduce `+notmuch-delete-tags` and `+notmuch-spam-tags` variables instead of hardcoding the tags. The default values are set to the previously hardcoded values and the change does thus not break any existing users.

### DIFF
--- a/modules/email/notmuch/autoload.el
+++ b/modules/email/notmuch/autoload.el
@@ -76,32 +76,32 @@
 ;;;###autoload
 (defun +notmuch/search-delete ()
   (interactive)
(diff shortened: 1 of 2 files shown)


hlissner (changes requested):
Please reword the commit message to conform to [our git conventions](https://gist.github.com/hlissner/4d78e396acb897d9b2d8be07a103a854):

```
refactor(notmuch): parameterize delete/spam tags
```

- `feat` is not the correct type for this kind of change.
- No scope was provided.

hlissner:
Thanks for your help!

PR #5754, opened 2021-11-10
fix(latex): run AUCTeX folding after font locking

vqhuy:
This guarantees that `TeX-fold-buffer` is run after the style hooks and not before. Otherwise, it looks like they may reset/erase font-lock stuff you have set up.

Currently, `TeX-fold-buffer` sometimes is applied before font locking hooks, causing incomplete folding (usually wrong colorization).

### DIFF
--- a/modules/lang/latex/config.el
+++ b/modules/lang/latex/config.el
@@ -115,9 +115,11 @@ If no viewers are found, `latex-preview-pane' is used.")
 
 (use-package! tex-fold
   :when (featurep! +fold)
-  :hook (TeX-mode . TeX-fold-buffer)
+  :hook (TeX-mode . +latex-TeX-fold-buffer-h)
   :hook (TeX-mode . TeX-fold-mode)
   :config
+  (defun +latex-TeX-fold-buffer-h ()
+    (run-with-idle-timer 0 nil 'TeX-fold-buffer))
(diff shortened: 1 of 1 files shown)


modules/lang/latex/config.el
```diff
@@ -115,9 +115,11 @@ If no viewers are found, `latex-preview-pane' is used.")
 
 (use-package! tex-fold
   :when (featurep! +fold)
-  :hook (TeX-mode . TeX-fold-buffer)
+  :hook (TeX-mode . +delayed-TeX-fold-buffer)
   :hook (TeX-mode . TeX-fold-mode)
   :config
+  (defun +delayed-TeX-fold-buffer ()
```
hlissner, line 121: Rename to `+latex-Tex-fold-buffer-h` to conform to [Doom's naming conventions](https://github.com/hlissner/doom-emacs/blob/develop/docs/contributing.org#naming-conventions).

hlissner (changes requested):

modules/lang/latex/config.el
vqhuy: Done.

hlissner (approved):

hlissner:
Thanks again for your help!

PR #5755, opened 2021-11-10
fix(org): correct renamed variable

ntharim:
`doom-inhibit-local-var-hooks` was renamed to `doom-inhibit-major-mode-post-hooks`.


hlissner:
Ack, sorry about that. The rename of that variable was premature; it snuck into 68d8364. I've reverted the change in 06f1e24, which should indirectly fix your issue. Thanks for bringing it to my attention!

PR #5756, opened 2021-11-10
quick fix(mu4e): incorrect access of from addr slot

tecosaur:
Within +mu4e/capture-msg-to-agenda, the from address is stored in the
cdar not the cadr.

Unfortunately, I think I just made a typo in the original implementation, oh well — it comes up rarely but this should fix it.


hlissner (approved):

hlissner:
LGTM! Thanks!

PR #5764, opened 2021-11-11
fix(org): no roam buffer for capture buffers

vherrmann:
This pr changes the hook `org-roam-find-file`, so that it checks if the buffer was opened by `org-capture`. 

PS: using `(not (window-parameter nil 'window-side))` instead of `(not org-capture-mode)` would probably work as well, but this line was removed in the transition from roam to roamv2, which is why I'm not sure if this line is still wanted


hlissner (approved):

vherrmann:
I already did that…

hlissner (approved):

hlissner:
LGTM! Thanks for your help!

PR #5767, opened 2021-11-11
nit: Update grammar README

junjizhi:
Add a note for macOS users who uses `brew install java languagetool`.


hlissner:
Thanks for the PR but I must turn it down. `brew install java` should install java into your `$PATH`, which Doom should pick up so long as your envvar file is up to date (i.e. run `$ doom env`)

If for some reason java _isn't_ installed onto your `$PATH`, then I will reopen this (but the snippet will need to be updated to reflect that _and_ should be done on the `rewrite-docs` branch -- where I'm working on a rewrite/redesign of all our docs, including the docs for modules).

PR #5771, opened 2021-11-12
 fix(doom): inconsistent doom-variable-pitch-font name

notarock:
This commit fixes doom-variable-font so it is consistent with the actual variable used within doom, which is 'doom-variable-pitch-font'.


hlissner (approved):

hlissner:
Thanks for your help!

PR #5776, opened 2021-11-15
fix(mu4e): only bind `h` when `workspaces` loaded

dschrempf:
Tiny change.

Fixes bug about void function `+workspace/other` when pressing `h` in `mu3e` when workspaces are deactivated.


tecosaur:
Good catch!

tecosaur (approved):

hlissner (approved):

hlissner:
Thanks!

PR #5780, opened 2021-11-16
feat(literate): tangle config files inside custom modules

pakelley:
Allows you to use a `config.org` to configure your private modules (using as part of the `literate` module).
If this seems useful, I'm happy to add some tests and docs.
Also, I didn't set up auto-generation of the other files yet for fear of slowing down the existing module-loading code... Currently, I have a private module working by adding the following to its `cli.el`:
```
(when (featurep! :config literate)
  (add-hook 'doom-sync-pre-hook (apply-partially '+literate-tangle-module-h "lang" "org-patch")))
```
Lastly, this is the first time I've written anything somewhat serious in elisp, so I'm happy to fix anything that seems awful (but I did read through the styleguide and do my best to conform to it)

### DIFF
--- a/modules/config/literate/autoload.el
+++ b/modules/config/literate/autoload.el
@@ -15,16 +15,12 @@ byte-compiled from.")
 ;;;###autoload (add-hook 'org-mode-hook #'+literate-enable-recompile-h)
 
 ;;;###autoload
-(defun +literate-tangle-h ()
-  "Tangles `+literate-config-file' if it has changed."
-  (and (not (getenv "__NOTANGLE"))
+(defun +literate-tangle-fn (target dest cache guard-env)
+  "Tangles TARGET if it has changed"
+  (and (not (getenv guard-env))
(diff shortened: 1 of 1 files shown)


hlissner:
It's an interesting idea, but I'd rather not add support for literate modules to Doom. I also have plans to make literate configs a first-class citizen in Doom once [the CLI rewrite](https://github.com/hlissner/doom-emacs/issues/4273) is done; I may reconsider it then, but for the time being I must refuse it as it introduces too much complexity.

pakelley:
@hlissner gotcha, makes sense! If `+literate-tangle-h` was split into `+literate-tangle-h` and `+literate-tangle-fn` it would still be possible for users (or at least me) to re-use the literate config's org-tangle logic.
If I simplified this PR to only split up that function, would that be simple enough for you to consider merging in?
Either way, thanks for all of the work on Doom 🙂

PR #5782, opened 2021-11-17
fix(dart): remove hardcoded path to flutter SDK

CrimsonKoba:
This is unnecesary since hover search flutter in the path

Emacs says that file does not exists when I try to run any flutter command and I can't override `flutter-sdk-path` on the config file.
Removing these lines everything works fine

### DIFF
--- a/modules/lang/dart/config.el
+++ b/modules/lang/dart/config.el
@@ -4,9 +4,6 @@
   :when (featurep! +lsp)
   :hook (dart-mode-local-vars . lsp!)
   :config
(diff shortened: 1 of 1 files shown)


hlissner (changes requested):
Please reword commit to conform to [our git conventions](https://gist.github.com/hlissner/4d78e396acb897d9b2d8be07a103a854):

```
fix(dart): remove hardcoded path to flutter SDK

Instead, flutter.el looks for the executable in PATH.
```

hlissner (approved):

hlissner:
Thanks for your help!

PR #5786, opened 2021-11-17
fix(popup,undo): make undo-tree visualizer work with our popup manager

### LINKED ISSUE #5317: The first call to `org-clock-in`/`org-resolve-clocks`/`org-files-list` on start loads all org-agenda-files in buffers

When I clock into a task, this should not have an incidence on open buffers. 

### What actually happened?

When I start Emacs and clock into a task, all the files listed in `org-agenda-files` are opened in buffers. 
This only happens at the first call to `org-clock-in`. Subsequent calls do not trigger this behaviour. 

I have managed to narrow the issue down to `org-files-list`. 
Specifically:
- `org-clock-in` calls `org-resolve-clocks`
- `org-resolve-clocks` calls `org-files-list` (which is the only existing reference to `org-files-list`)
- `org-files-list` returns `org-agenda-files` list, plus all open Org files, according to the description

---

I have checked that this is the problem by running the following in the scratch buffer at start:

``` elisp
#+begin_src elisp
(dolist (file (org-files-list))) 
#+end_src
```

This snippet does not return anything, but triggers all org files to be loaded into buffers. 

---

I have not found any mention of this issue anywhere, which makes me think that either it's my config, or no one else is bothered by having all `org-agenda-files` in the buffer list. 

---

The source code for `org-files-list`:
``` elisp
;; Defined in ~/.emacs.d/.local/straight/repos/org/lisp/org.el
(defun org-files-list ()
  "Return `org-agenda-files' list, plus all open Org files.
This is useful for operations that need to scan all of a user's
open and agenda-wise Org files."
  (let ((files (mapcar #'expand-file-name (org-agenda-files))))
    (dolist (buf (buffer-list))
      (with-current-buffer buf
	(when (and (derived-mode-p 'org-mode) (buffer-file-name))
	  (cl-pushnew (expand-file-name (buffer-file-name)) files
		      :test #'equal))))
    files))
```

### Steps to reproduce

1. Start Emacs
2. Clock-in to a task

### System Information

https://pastebin.com/A5LJXHy6

vqhuy:
Fixes #5317

`undo-tree-visualizer-{show/hide}-diff` changes the value of `undo-tree-visualizer-diff`, so even if it is set to `t` (as default in doom), when the user quits the visualizer (by pressing `q`), this variable is set to `nil` (in `undo-tree-visualizer-hide-diff`).
An unexpected behavior is that first the visualizer is shown in a side window, and the second time it is invoked, it is shown in a popup.

This forces `undo-tree` visualizer always show in popup, and makes it respect the value of `undo-tree-visualizer-diff`.

### DIFF
--- a/modules/emacs/undo/config.el
+++ b/modules/emacs/undo/config.el
@@ -80,6 +80,15 @@
       :filter-return #'undo-tree-make-history-save-file-name
       (concat file ".zst")))
 
+  ;; Make these functions respect the value of `undo-tree-visualizer-diff'
+  (defun +keep-undo-tree-visualizer-diff-a (fn &rest args)
+    (let (old-undo-tree-visualizer-dif undo-tree-visualizer-diff)
+      (apply fn args)
+      (setq undo-tree-visualizer-diff old-undo-tree-visualizer-dif)))
+
+  (advice-add #'undo-tree-visualizer-show-diff :around #'+keep-undo-tree-visualizer-diff-a)
+  (advice-add #'undo-tree-visualizer-hide-diff :around #'+keep-undo-tree-visualizer-diff-a)
+
   ;; Strip text properties from undo-tree data to stave off bloat. File size
   ;; isn't the concern here; undo cache files bloat easily, which can cause
   ;; freezing, crashes, GC-induced stuttering or delays when opening files.
--- a/modules/ui/popup/+hacks.el
+++ b/modules/ui/popup/+hacks.el
@@ -362,10 +362,8 @@ Ugh, such an ugly hack."
 (defadvice! +popup--use-popup-window-for-undo-tree-visualizer-a (fn &rest args)
   "TODO"
   :around #'undo-tree-visualize
-  (if undo-tree-visualizer-diff
-      (apply fn args)
-    (letf! ((#'switch-to-buffer-other-window #'pop-to-buffer))
-      (apply fn args))))
+   (letf! ((#'switch-to-buffer-other-window #'pop-to-buffer))
+      (apply fn args)))
 
 ;;;###package wgrep
 (progn


modules/emacs/undo/config.el
```diff
@@ -80,6 +80,15 @@
       :filter-return #'undo-tree-make-history-save-file-name
       (concat file ".zst")))
 
+  ;; Make these functions respect the value of `undo-tree-visualizer-diff'
+  (defun +keep-undo-tree-visualizer-diff-a (fn &rest args)
+    (let (old-undo-tree-visualizer-dif undo-tree-visualizer-diff)
+      (apply fn args)
+      (setq undo-tree-visualizer-diff old-undo-tree-visualizer-dif)))
```
hlissner, line 87: There are a number of issues here. 

1. You're not doing anything with `old-undo-tree-visualizer-dif`, therefore `undo-tree-visualizer-diff` is _always_ set to `nil` whenever `undo-tree-visualizer-{show,hide}-diff` is invoked.

2. The advice does not return the return value of the function they're advising. This may be an issue later and should be rectified.

3. Please use `defadvice!` instead of two `advice-add` calls.

4. That `undo-tree-visualizer-{show,hide}-diff` have side effects should be considered a bug. Please report it upstream so it can be eventually dealt with there.

modules/ui/popup/+hacks.el
```diff
@@ -362,10 +362,8 @@ Ugh, such an ugly hack."
 (defadvice! +popup--use-popup-window-for-undo-tree-visualizer-a (fn &rest args)
   "TODO"
   :around #'undo-tree-visualize
-  (if undo-tree-visualizer-diff
-      (apply fn args)
-    (letf! ((#'switch-to-buffer-other-window #'pop-to-buffer))
-      (apply fn args))))
+   (letf! ((#'switch-to-buffer-other-window #'pop-to-buffer))
+      (apply fn args)))
```
hlissner, line 366: Please revert this, as it breaks https://github.com/hlissner/doom-emacs/commit/db9106f92cf71e6878b08650c45cfe89455f4959

hlissner (changes requested):

vqhuy:
Thanks for taking a look, @hlissner . I close it now as it seems to be an upstream bug.

PR #5787, opened 2021-11-17
fix(popup): show reftex-citation in popup

vqhuy:


hlissner (changes requested):
Please reword commit to:

```
fix(popup): show reftex-citation in popup
```

There is no "reftex" module, so it is not a valid scope.

vqhuy:
Done.

hlissner (approved):

hlissner:
Thanks!

PR #5789, opened 2021-11-17
fix(mu4e): do not set coding-system-for-read globally

### LINKED ISSUE #5788: mu-lock.el sets coding-system-for-read variable globally

According to the variable's docstring, `coding-system-for-read` should not be set globally:
```
It is useful to bind this variable with let, but do not set it globally.
```
It's default value when running `emacs -Q` is `nil`. I expected it to stay set to `nil`.

### What actually happened?

When updating mails in `mu4e`, `coding-system-for-read` gets set to `utf-8` globally. 

This causes issues, for example when opening a zip archive in dired: instead of seeing the contents of the archive in `arc-mode`, the file is shown raw and the following message is logged:
```
File mode specification error: (args-out-of-range elsarticle.zip 1425900 1425904)
```
(for the zip file called "elsaarticle.zip")

I have looked for occurrences where `coding-system-for-read` is being `setq`'ed and found `+mu4e-lock-pid-info` as an offender in `mu-lock.el`, part of the mu4e module. Setting the value within the `let*` statement fixes the issue. PR is in progress.

### Steps to reproduce

1. Open Emacs
2. Start mu4e (C-c M M)
3. Start update (U)
4. Check status of coding-system-for-read

### System Information

https://pastebin.com/Mgtd5LiL

hperrey:
Fixes #5788 

In the function `+mu4e-lock-pid-info` in `mu-lock.el`, the variable `coding-system-for-read` is set globally which is against its documentation's recommendation and can cause issues as described in #5788. Setting the value in the `let*` statement instead through `setq` fixes this issue for me.

### DIFF
--- a/modules/email/mu4e/autoload/mu-lock.el
+++ b/modules/email/mu4e/autoload/mu-lock.el
@@ -16,9 +16,9 @@
   "Get info on the PID refered to in `+mu4e-lock-file' in the form (pid . process-attributes)
  If the file or process do not exist, the lock file is deleted an nil returned."
   (when (file-exists-p +mu4e-lock-file)
-    (let* ((pid (string-to-number
+    (let* ((coding-system-for-read 'utf-8)
+           (pid (string-to-number
                  (with-temp-buffer
-                   (setq coding-system-for-read 'utf-8)
                    (insert-file-contents +mu4e-lock-file)
                    (buffer-string))))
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Thanks for your help!

PR #5790, opened 2021-11-17
fix(notmuch) incorrect type used for window

Descanonge:
`if-let*` returns a window type, not a buffer which would cause a `Wrong type argument: stringp, #<window 3 on *doom*>` when trying to access notmuch via `=notmuch` while it is already open in a workspace.


Descanonge:
sorry I messed up the commit name. should I resubmit a PR with the correct syntax ?

PR #5792, opened 2021-11-18
fix(evil): handle edge cases of embrace

vqhuy:
If the surround key is `\(`, the text should be surrounded as
`\( text \)`. The same for `\[` and `\{`.

Currently, when the surround key is `\(`, `+evil--embrace-escaped` returns `"\( " . "\ )"`, which changes the text to `\( text\ )` (notice a whitespace before `\` and `)`).

### DIFF
--- a/modules/editor/evil/autoload/embrace.el
+++ b/modules/editor/evil/autoload/embrace.el
@@ -17,10 +17,11 @@
   (let ((char (read-char "\\")))
     (if (eq char 27)
         (cons "" "")
-      (let ((pair (+evil--embrace-get-pair (string char)))
-            (text (if (sp-point-in-string) "\\\\%s" "\\%s")))
-        (cons (format text (car pair))
-              (format text (cdr pair)))))))
+      (let* ((pair (+evil--embrace-get-pair (string char)))
+             (escape (if (sp-point-in-string) "\\\\" "\\"))
+             (escape (format "\\1%s" (regexp-quote escape))))
+        (cons (replace-regexp-in-string "^\\( *\\)" escape (car pair))
+              (replace-regexp-in-string "^\\( *\\)" escape (cdr pair)))))))
 
 ;;;###autoload
 (defun +evil--embrace-latex ()


modules/editor/evil/autoload/embrace.el
```diff
@@ -15,12 +15,15 @@
...
-    (if (eq char 27)
-        (cons "" "")
-      (let ((pair (+evil--embrace-get-pair (string char)))
-            (text (if (sp-point-in-string) "\\\\%s" "\\%s")))
-        (cons (format text (car pair))
-              (format text (cdr pair)))))))
+    (cond ((eq char 27) (cons "" ""))
+          ((eq char 40) (cons "\\( " " \\)"))
+          ((eq char 91) (cons "\\[ " " \\]"))
+          ((eq char 123) (cons "\\{ " " \\}"))
```
hlissner, line 21: Could you generalize this into the else clause of the `cond`?

We'd just need to detect if `(cdr pair)` starts with a space, and if so, then prepend a space to `text` (the format string).

hlissner (changes requested):

modules/editor/evil/autoload/embrace.el
vqhuy: but we also need to remove the space between `\` and the `)` returned by `+evil--embrace-escaped`. Is that an elegant solution? Not sure if the user has some weird pairs that include whitespace, so trying to generalize this might break users' setting?

vqhuy: I mean this fix is specified for the default setting, if the user has extra setting then she needs to be careful with her own implementation.

vqhuy: actually we also need to remove that space too. I gave it a try in https://github.com/hlissner/doom-emacs/pull/5792/commits/3bec6ae972272714b5c7275114e7d5cd3d630c12 but not sure how `sp-point-in-string` works here, so I just ignored it.

vqhuy: FYI, putting `(end (if (sp-point-in-string) "%s\\\\" "%s\\"))` causes an error `Invalid use of \\ in replacement text` by `replace-match` inside `replace-regexp-in-string`.

vqhuy:
@hlissner Is there anything I can do to get this done?

hlissner (approved):

modules/editor/evil/autoload/embrace.el
```diff
@@ -18,9 +18,12 @@
...
       (let ((pair (+evil--embrace-get-pair (string char)))
-            (text (if (sp-point-in-string) "\\\\%s" "\\%s")))
-        (cons (format text (car pair))
-              (format text (cdr pair)))))))
+            (start (if (sp-point-in-string) "\\\\%s" "\\%s"))
+            (end "%s\\\\"))
+        (cons (format start (car pair))
+              (replace-regexp-in-string "^[[:space:]]*"
+                                        (lambda(s) (format end s))
+                                        (cdr pair)))))))
```
hlissner, line 26: To fix the issue with in-string escaping, try this:

```diff
-      (let ((pair (+evil--embrace-get-pair (string char)))
-            (start (if (sp-point-in-string) "\\\\%s" "\\%s"))
-            (end "%s\\\\"))
-        (cons (format start (car pair))
-              (replace-regexp-in-string "^[[:space:]]*"
-                                        (lambda(s) (format end s))
-                                        (cdr pair)))))))
+      (let* ((pair (+evil--embrace-get-pair (string char)))
+             (escape (if (sp-point-in-string) "\\\\" "\\"))
+             (escape (format "\\1%s" (regexp-quote escape))))
+        (cons (replace-regexp-in-string "^\\( *\\)" escape (car pair))
+              (replace-regexp-in-string "^\\( *\\)" escape (cdr pair)))))))
```

hlissner (changes requested):

modules/editor/evil/autoload/embrace.el
hlissner: Backslashes need another step of escaping to be used in a regular expression (e.g. filter it through `regexp-quote`). See my recent review for a solution that takes care of this.

vqhuy:
Thanks, I think the PR is good now.

hlissner (approved):

hlissner:
LGTM. Thanks again for your help! (And for putting up with my nitpicks)

PR #5796, opened 2021-11-18
fix: +default/discover-projects not reading depth

sei40kr:
Fix `+default/discover-projects` to read the depth in `projectile-project-search-path`.

### DIFF
--- a/modules/config/default/autoload/files.el
+++ b/modules/config/default/autoload/files.el
@@ -52,7 +52,7 @@ If prefix ARG is non-nil, prompt for the search path."
         (dolist (dir projectile-project-search-path)
           (cl-destructuring-bind (dir . depth) (if (consp dir) dir (cons dir nil))
             (if (not (file-accessible-directory-p dir))
-                (message "%S was inaccessible and couldn't searched" dir)
+                (message "%S was inaccessible and couldn't be searched" dir)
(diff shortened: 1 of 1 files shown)


modules/config/default/autoload/files.el
```diff
@@ -50,9 +50,10 @@ If prefix ARG is non-nil, prompt for the search path."
                  (funcall projectile-add-known-project project-root)
                  (message "Added %S to known project roots" project-root)))
         (dolist (dir projectile-project-search-path)
-          (if (not (file-accessible-directory-p dir))
-              (message "%S was inaccessible and couldn't searched" dir)
-            (projectile-discover-projects-in-directory dir)))))))
+          (cl-destructuring-bind (dir . depth) (if (consp dir) dir (cons dir nil))
+            (if (not (file-accessible-directory-p dir))
+                (message "%S was inaccessible and couldn't searched" dir)
```
hlissner, line 55: s/couldn't searched/couldn't be searched/

hlissner (changes requested):

modules/config/default/autoload/files.el
sei40kr: @hlissner 
Fixed → c220898

hlissner (approved):

hlissner:
Thanks again!

PR #5798, opened 2021-11-19
dev(doctor): add error message for Emacs dev versions

gagbo:
The message simply test for the "not-a-version-yet" patch number and
displays an explanation to prepare users for raised error frequency and
having a backup version of Emacs if they do not want to deal with issues.

(Of course, feel free to discard it if it's too much).

Rendered (yes, I struggled with autofill :( ) :

image

### DIFF
--- a/core/cli/doctor.el
+++ b/core/cli/doctor.el
@@ -50,6 +50,17 @@ in."
   (print! (start "Checking your Emacs version..."))
   (print-group!
    (cond
+    ((string= ".50" (substring emacs-version -3))
+     (error! "Emacs development version detected (%s)" emacs-version)
+     ;; There are 2 newlines between each item to fight against
+     ;; the (fill-region) call in `doom--output-autofill'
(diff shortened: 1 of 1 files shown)


iyefrat:
I feel like this should match `.60` and other pre release versions as well, even if they are less likely to break than `.50` versions.

gagbo:
I hesitated a bit, but I chose to let `.60` versions be caught in the `EMACS28+` cond where the message is just a warning and not as alarming; once emacs branches for release, the feature freeze is in full effect and breaking changes (almost) completely stop

hlissner (approved):

hlissner:
Thanks!

tasmo:
Although the PR is merged already I want to ask if it is really an **error** to use a development version of Emacs. An error implements that something is really nor working which is not the issue in any use case (like mine). I actually use Emacs v`29.0.50` without any issue.

Wouldn't it be better to handle this as a **warning**?

PR #5803, opened 2021-11-21
fix(lookup): xref 1.3.0 removed var xref--marker-ring

### LINKED ISSUE #5737: Jump to definition does not work on typescript files

Pressing `g d` on a typescript file uses tide for resolving the definition.

### What actually happened?

Pressing `g d` on a typescript file uses some other method (xref?) for resolving the definition.

`tide-jump-to-definition` works fine.

_No response_

### Steps to reproduce

1. Open Emacs
2. Open a typescript file
3. Move cursor over a typescript variable/method
4. Press `g d`

### System Information

https://pastebin.com/kPTu2Ckd

juster:
Fixes #5737 

This is not a mode specific issue but is actually caused by changes to _xref_. _Xref_ was changed in the emacs repo about a month ago. I have already created a patch for a mode that needed to adjust to _xref_'s internal changes.

[xref's recent history](https://github.com/emacs-mirror/emacs/commits/master/lisp/progmodes/xref.el)

Xref started many backwards incompatible changes in the name of performance. The marker ring no longer exists as a variable. `+lookup--xref-show` will silently fail every time.

Switches to using `xref-after-jump-hook` to test whether _xref_ actually jumped to a definition. The hook is part of the exposed interface and should hopefully not change in the near future.

### DIFF
--- a/modules/tools/lookup/autoload/lookup.el
+++ b/modules/tools/lookup/autoload/lookup.el
@@ -188,15 +188,16 @@ This can be passed nil as its second argument to unset handlers for MODES. e.g.
                         (xref-find-backend)
                         identifier)))
     (when xrefs
-      (let ((marker-ring (ring-copy xref--marker-ring)))
+      (let* ((jumped nil)
+             (xref-after-jump-hook
+              (cons (lambda () (setq jumped t))
+                    xref-after-jump-hook)))
         (funcall (or show-fn #'xref--show-defs)
                  (lambda () xrefs)
                  nil)
         (if (cdr xrefs)
             'deferred
-          ;; xref will modify its marker stack when it finds a result to jump to.
-          ;; Use that to determine success.
-          (not (equal xref--marker-ring marker-ring)))))))
+          jumped)))))
 
 (defun +lookup-dictionary-definition-backend-fn (identifier)
   "Look up dictionary definition for IDENTIFIER."


modules/tools/lookup/autoload/lookup.el
```diff
@@ -182,21 +182,20 @@ This can be passed nil as its second argument to unset handlers for MODES. e.g.
...
-        (if (cdr xrefs)
-            'deferred
-          ;; xref will modify its marker stack when it finds a result to jump to.
-          ;; Use that to determine success.
-          (not (equal xref--marker-ring marker-ring)))))))
+      (setq +lookup--xref-jumped nil)
+      (funcall (or show-fn #'xref--show-defs) #'(lambda () xrefs) nil)
+      (if (cdr xrefs)
+          'deferred
+        +lookup--xref-jumped))))
```
hlissner, line 198: A let-binding would be a much cleaner approach. See if this works.

```elisp
(let* ((jumped nil)
       (xref-after-jump-hook
        (cons (lambda () (setq jumped t))
              xref-after-jump-hook)))
  (funcall (or show-fn #'xref--show-defs)
           (lambda () xrefs)
           nil)
  (if (cdr xrefs)
      'deferred
    jumped))
```

Then you can remove the `defvar` and `add-hook` lines.

hlissner (changes requested):
To conform to [our git conventions](https://gist.github.com/hlissner/4d78e396acb897d9b2d8be07a103a854) please reword the commit to:

```
fix(lookup): xref 1.3.0 compatibility

Xref started many backwards incompatible changes in the name of
performance.  The marker ring no longer exists as a variable.
+lookup--xref-show will silently fail every time.

Switches to using xref-after-jump-hook to test whether xref actually
jumped to a definition. The hook is part of the exposed interface and
should hopefully not change in the near future.

Fix: #5737
Ref: https://github.com/emacs-mirror/emacs/commits/master/lisp/progmodes/xref.el
```

modules/tools/lookup/autoload/lookup.el
juster: Yes the dynamic binding worked. I force pushed it.

juster:
I unconsciously changed the formatting. I restored the formatting and force pushed again.

hlissner (changes requested):
The commit subject is still missing a valid scope:

```diff
-fix: xref 1.3.0 removed var xref--marker-ring 
+fix(lookup): xref 1.3.0 removed var xref--marker-ring 
```

hlissner (approved):

hlissner:
Thanks for your help!

PR #5805, opened 2021-11-21
feat(notmuch): only tag new messages w/ afew 

tasmo:
Tagging all messages can cause very heavy load and is not necessary.
Just indexing new messages is quicker and lighter.
If one needs to tag all messages again it is still possible to do so manually.

Changes `afew -a -t` to `afew -n -t` which is an equivalent of `afew --new --tag`.

### DIFF
--- a/modules/email/notmuch/config.el
+++ b/modules/email/notmuch/config.el
@@ -19,8 +19,8 @@ More specifically, this accepts one of the following symbols (see
 OR a shell command string such as
 
   \"path/to/some/shell-script.sh\"
-  \"offlineimap && notmuch new && afew -a -t\"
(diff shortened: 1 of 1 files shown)


hlissner (changes requested):
To adhere to [our git conventions](https://gist.github.com/hlissner/4d78e396acb897d9b2d8be07a103a854), please reword the commit to:

```
feat(notmuch): only tag new messages w/ afew

Indexing all messages can be very expensive, so some time can be saved
by only indexing new messages.
```

hlissner (approved):

hlissner:
Thanks for your help!

PR #5808, opened 2021-11-21
tweak!(input): remove remapping of i_ctrl-* keys

gagbo:
BREAKING CHANGE: Ctrl-* keys are back to default meaning in insert mode

Changing the signification of Ctrl-* keys in insert mode seems
unnecessary and ends up confusing because it is not necessarily the same
bindings as in normal mode.

Creating the PR to announce the possible change to @brabalan and ask for input. I'm going to use that for a couple of weeks before committing to a decision


alan-gh:
I’m giving it a try, I’ll report if I find anything troubling.

brabalan:
I’m giving it a try, I’ll report if I find anything troubling.

hlissner (approved):

PR #5810, opened 2021-11-22
fix(biblio): do not load org mode eagerly.

HyunggyuJang:
-------

Due to unconditional loading configuration for `citar-org` package, `org` is also being loaded eagerly.
Especially, causes a warning message from doom, like

``` text
‘org’ was already loaded by the time lang/org loaded, this may cause issues
```

### DIFF
--- a/modules/tools/biblio/config.el
+++ b/modules/tools/biblio/config.el
@@ -54,6 +54,7 @@
 ;;;; Third-party
 
 (use-package! citar-org
+  :no-require
   :when (featurep! :lang org +roam2)
   :config
   ;; Include property drawer metadata for 'org-roam' v2.


HyunggyuJang:
Should also resolve #5811

bdarcus:
Don't you still have to check the completion module in the `when`?

HyunggyuJang:
> Don't you still have to check the completion module in the `when`?

In theory, no need to add conditional clause for the completion engine. However, aesthetically agrees with your opinion it would be better to check based on the completion module explicitly.

HyunggyuJang:
@bdarcus Think it does not have to do with current PR, but I wonder what is the variable `citar-org-note-include` for. Maybe we should replace that with `citar-file-note-org-include`?

HyunggyuJang:
Summing up, will the following be the right thing to do?  

```emacs-lisp
(use-package! citar-org
  :defer t
  :when (and (featurep! :completion vertico)
             (featurep! :lang org +roam2))
  :config
  ;; Include property drawer metadata for 'org-roam' v2.
  (setq citar-file-note-org-include '(org-id org-roam-ref)))
```

hlissner (approved):

bdarcus:
> Summing up, will the following be the right thing to do?

I believe so.

Except, I'm now realizing `citar-file-note-org-include` doesn't really make sense given it's in `citar-org`. But I'll take a closer look at that separately.

hlissner:
I've merged this now to fix the immediate problem. Don't worry about the extra `:completion vertico` check (I'm already working on a refactor that will add it). In any case, thanks for your help!

PR #5820, opened 2021-11-23
docs(vertico): fix #+CREATED and #+SINCE keywords

iyefrat:
They were generated using a subsequent Vertico PR, not the one that
added the module.


hlissner (approved):

hlissner:
Thanks again!

PR #5822, opened 2021-11-24
tweak(company): company-selection-wrap-around = t

teeann:
This PR do the followings:
 - Add `company-files`, which provides completion for file paths, to `company-backends`. This completion is really useful, and doesn't introduce significant CPU overhead.
 - Set value of `company-selection-wrap-around` to `t`. This behavior is the default in most code editors I know, for good reasons.

### DIFF
--- a/modules/completion/company/config.el
+++ b/modules/completion/company/config.el
@@ -10,6 +10,7 @@
   (setq company-minimum-prefix-length 2
         company-tooltip-limit 14
         company-tooltip-align-annotations t
+        company-selection-wrap-around t
         company-require-match 'never
         company-global-modes
         '(not erc-mode


hlissner (changes requested):
While the `company-selection-wrap-around` change is acceptable, I'm not accepting changes to the global default value of `company-backends`.

Evil users have "omni-completion" keybinds on `C-x` in insert mode. `company-files`, specifically, is on `C-x C-f`. Alternatively, you can add `company-files` to a particular mode with `set-company-backend!` (see `C-h f set-company-backend\!` for examples).

Otherwise, you can change company-backends yourself with:

```elisp
(after! company
  (setq company-backends ...))
```

Anyhow, please make the following changes:

1. Remove ab99b5f1f6f47d874f1183a5577e62a7e09a05bc.
2. Reword 5450c9909fd9a1f1a1a1535c772eaac6a3291f2f to conform to [our git conventions](https://gist.github.com/hlissner/4d78e396acb897d9b2d8be07a103a854):

   ```
   tweak(company): company-selection-wrap-around = t
   ```

teeann:
Sure. I'll make those changes.

For the your decline of `company-backends` change, I'm not sure I understand your reasons above. As someone who switched from vanilla Emacs to evil/vim keybinding, I find having `company-files` completion **automatically** invoked much more convenient than having to press `C-x C-f` each time I want completion.

If you're still convinced the change to `company-backends` is not necessary, I suggest adding the configuration of `company-backends` to the documentation of `company` module.

hlissner:
> For the your decline of company-backends change, I'm not sure I understand your reasons above. As someone who switched from vanilla Emacs to evil/vim keybinding, I find having company-files completion automatically invoked much more convenient than having to press C-x C-f each time I want completion.

There's a spectrum to completion in editors. On one end is the manual approach: where you invoke the precise backend you need, and only when you need it (where completion won't bother you otherwise); and on the other end is the automatic approach: inundate the user with all candidates from all backends as they type, all the time, whether or not they need them.

I'm staunched in the manual camp. IDEs tend to be the opposite. Our company module is designed for the former, and I don't intend to change that (at least, as the default), but I may add a feature flag later to support it (then make idle-completion opt-in). It'll be some time until I get around to it though, and until then, I'd rather not change it. What's more: it's trivial enough for users to change themselves.

> If you're still convinced the change to company-backends is not necessary, I suggest adding the configuration of company-backends to the documentation of company module.

I've documented how to use `set-company-backend!` in the module's docs, but I'll look into make it clearer.

PR #5824, opened 2021-11-24
tweak(mu4e): mu4e-update-interval = 300

teeann:
Most standard email clients automatically check for new emails every 5-10 minutes. 
This PR sets the update interval to 5 minutes instead of `nil`


hlissner:
Thanks for the PR, but I must refuse it. "Your system, your rules" is one of Doom's goals; to not have side-effects on anything beyond your `~/.emacs.d` and `~/.doom.d` without your explicit permission (i.e. it's opt-in). This especially includes automated syncing with external services.

PR #5825, opened 2021-11-24
fix(ruby): enable web-mode for erb files

offbyone:
.erb files in bare Doom were being opened in HTML mode, which mangles formatting
on save. This ensures that when ruby and web features are enabled, they open in
the more full-featured web-mode editor.

### DIFF
--- a/modules/lang/ruby/config.el
+++ b/modules/lang/ruby/config.el
@@ -35,6 +35,10 @@
         "{" #'ruby-toggle-block))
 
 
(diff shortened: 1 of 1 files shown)


otavioschwanck:
Here, already opens on web-mode

hlissner:
This is already in effect [in the :lang web module](https://github.com/hlissner/doom-emacs/blob/develop/modules/lang/web/%2Bhtml.el#L6).

PR #5843, opened 2021-11-28
fix(lib): add-hook! fails when passed a function quoted list

HyunggyuJang:
-------

I.e., diverges from what documented at demos.org:
```emacs-lisp
(add-hook! some-mode #'(enable-something and-another))
```
which should be expanded something like
```emacs-lisp
(add-hook 'some-mode-hook #'enable-something)
(add-hook 'some-mode-hook #'and-another)
```
but actually, ends up
```emacs-lisp
(add-hook 'some-mode-hook #'(enable-something and-another))
```

This commit addresses this issue.

### DIFF
--- a/core/core-lib.el
+++ b/core/core-lib.el
@@ -641,9 +641,9 @@ This macro accepts, in order:
     (while rest
       (let* ((next (pop rest))
              (first (car-safe next)))
-        (push (cond ((memq first '(function nil))
+        (push (cond ((null first)
                      next)
-                    ((eq first 'quote)
+                    ((memq first '(function quote))
                      (let ((quoted (cadr next)))
                        (if (atom quoted)
                            next


hlissner:
I'm inclined to say the demos are what's wrong, not this behavior, because sharp quoting is for quoting function symbols, rather than lists. The elisp interpreter allows it; treating the two the same in most contexts, but I'm not convinced `add-hook!` should, just because. I'll sit on this for a day.

HyunggyuJang:
Okay, understood. Well, I just thought it might be better make explicit that the arguments are about function for both Emacs evaluator and human reader since the expected elements for quoted list in this case are all functions.

But, as you said, 

- sharp quote is for function symbols,

HyunggyuJang:
Okay, understood. Well, I just thought it might be better make explicit that the arguments are about function for both Emacs evaluator and human reader since the expected elements for quoted list in this case are all functions.

But, as you said,

- sharp quote is for function symbols,
- `add-hook!` already supports arbitrary number of function symbols as arguments, like `(add-hook! mode #'fun1 #'fun2 ...)`,

yes, might be better not to exploit such corner cases.

hlissner:
Sorry for the seriously reply! I've fixed the sharp-quoting example from the documentation in e8c22c7. Thanks for bringing it to my attention!

PR #5850, opened 2021-11-29
Add instructions to get Spotlight to open Emacs

JasonMFry:
Spotlight doesn't work with symbolic links. The solution I found was to use Automator.

(doomeamcs.org sites are behind a registration page? So I didn't sign up just to check the sites. imo this information should be accessible without registering for something.)


hlissner:
Thanks for the PR, but I must refuse it. Changes to our docs is in [our do-not-PR](https://discourse.doomemacs.org/do-not-pr) list because a rework of all our documentation is in the works.

PR #5851, opened 2021-11-29
fix(evil): correct typesetting quotations in latex

vqhuy:
Use `evil-embrace` for typsetting quotations in latex instead of
`evil-surround`.

Single quotation marks are produced in LaTeX using ` and ' . Double quotation marks are produced by typing `` and ''. 

This should be an upstream fix for `evil-embrace` but looking at the latest commit of https://github.com/cute-jumper/embrace.el, I think it's better to send a PR to doom.


hlissner (approved):

hlissner:

PR #5859, opened 2021-12-02
fix(popup): report-emacs-bug and its saved  files are treated as popups.

daanturo:
Steps to reproduce:
1. `report-emacs-bug`
2. Switch to buffer `*unsent mail to bug-gnu-emacs@gnu.org*`
3. Save the buffer, then kill it
4. Open the saved message

Observed:
- After step 1, the new buffer is poped up briefly, then replaced by `*Bug Help*` buffer and doesn't appear.
- After step 4, the buffer (whose name is `*message*-yyyymmdd-hhmmss`) who holds the saved message file is treated as a popup and at the bottom.


hlissner (approved):

hlissner:

PR #5869, opened 2021-12-06
bump: ivy-posframe

ethan-leba:
tumashu/ivy-posframe@9c83828 -> tumashu/ivy-posframe@5d9420

Notably fixes a bug where the ivy frame disappears after using avy.

Ref: https://github.com/tumashu/ivy-posframe/issues/114

### DIFF
--- a/modules/completion/ivy/packages.el
+++ b/modules/completion/ivy/packages.el
@@ -18,7 +18,7 @@
     (package! flx :pin "647cb2f92f9936c62e277d7a74ad54a80502d255")))
 
 (when (featurep! +childframe)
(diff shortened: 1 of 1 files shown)


ethan-leba:
Thanks for the heads up!

iyefrat:
Seems that you have more issues with the commit hashes (see the falied CI). On recent versions of doom, it runs the commit linter before you commit and warns you to edit it if the message fails. I would recommend updating to a newer version, but in the mean time, you can run `doom ci lint-commits HEAD~1` on your pr branch to check it locally.

hlissner:
Thanks for the PR, but a larger bump was already in the works for the `:completion` category, so I merged it into that effort (in 3d50c36), which should resolve this issue. Thanks for bringing it to my attention!

PR #5875, opened 2021-12-07
feat: change inferior-fsharp-program to dotnet

elken:
Applies from .NET Core 2 onwards, so ideally this would just be hardcoded but it's fine to be gated too.


hlissner (approved):

hlissner:
LGTM. Thanks for the help!

PR #5881, opened 2021-12-09
bump: :core to fix emacs 29 problem with general.el

### LINKED ISSUE #5785: `wrong-number-of-arguments: #<subr general-unbind-non-prefix-key> 5` errors at startup

After the default install I expect the splash screen to load and the default keybindings to work

### What actually happened?

doom loads a warning buffer with the following in and keybindings are broken
```
Error (doom-init-ui-hook): Error running hook "persp-mode" because: (wrong-number-of-arguments #<subr general-unbind-non-prefix-key> 5) Disable showing Disable logging
Error (doom-first-input-hook): Error running hook "evil-snipe-mode" because: (wrong-number-of-arguments #<subr general-unbind-non-prefix-key> 5) Disable showing Disable logging
```

I tried using -debug-init but this didn't seem to help since it left emacs in the same state

### Steps to reproduce

install emacs from source with ./configure --with-native-compilation --with-mailutils --with-imagemagick --with-xwidgets --with-dbus --without-libsystemd --with-x
run doom install 
run emacs

### System Information

https://pastebin.com/gWDuV010

quangIO:
Fixes #5785

This should fix the general-unbind-non-prefix-key issue when using general.el on Emacs 29

### DIFF
--- a/core/packages.el
+++ b/core/packages.el
@@ -48,5 +48,5 @@
 (package! project :pin "a546cce47c434b0b4bf24bf1d5ed6e4623492072")
 
 ;; core-keybinds.el
-(package! general :pin "26f1d4c4e258c40e6b70ce831a04800c914f29b3")
+(package! general :pin "9651024e7f40a8ac5c3f31f8675d3ebe2b667344")
 (package! which-key :pin "4790a14683a2f3e4f72ade197c78e4c0af1cdd4b")


hlissner:
A larger bump was necessary for the rest of Doom's core's packages (and this PR's commit message didn't conform to [our git conventions](https://gist.github.com/hlissner/4d78e396acb897d9b2d8be07a103a854)), so I've bumped general.el (among others) in 5535457, which should resolve this and #5785. Thanks for bringing it to my attention!

PR #5883, opened 2021-12-09
fix(latex): install style/ etc in auctex recipe

mjkramer:
In order for AUCTeX to work properly, various ingredients (such as the `style` subdirectory) must be present in the package's build directory. For reasons I don't understand, after a "fresh" `doom sync` (i.e. after nuking `.local/straight`), the `auctex` build directory contains All The Things from the repo directory (including autotools junk), so that AUCTeX should initially work fine. However, the next time `doom sync` is run, the `auctex` package gets rebuilt, and we end up (per Straight's defaults) with only the top-level `.el` files symlinked into the build directory, breaking AUCTeX.

This PR updates the `auctex` package recipe to explicitly list the "extra" stuff that AUCTeX needs. I've confirmed that the package's build directory now has the expected structure and that AUCTeX works.  The "fresh" `doom sync` no longer installs autoconf junk, and the package is no longer rebuilt on the next `doom sync`.

At https://github.com/raxod502/straight.el/issues/240 there's further discussion on the intricacies of installing AUCTeX with Straight. I don't understand all of the nuances nor the solutions suggested there. The simple fix in this PR seems to be sufficient.


modules/lang/latex/packages.el
```diff
@@ -1,7 +1,9 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; lang/latex/packages.el
 
-(package! auctex :pin "3b0a080ae596c26c17b15ba9c71fc5542eae238b")
+(package! auctex :pin "3b0a080ae596c26c17b15ba9c71fc5542eae238b"
+  :recipe (:files ("*.el" "*.info" "dir"
+                   "doc" "etc" "images" "latex" "style")))
```
hlissner, line 6: Simplify with `:defaults` and reformat to be consistent with other multi-line `package!` statements:

```elisp
(package! auctex
  :recipe (:files (:defaults "etc" "images" "latex" "style"))
  :pin "3b0a080ae596c26c17b15ba9c71fc5542eae238b")
```

hlissner (changes requested):

modules/lang/latex/packages.el
mjkramer: Reformatted. The problem with `:defaults` is that it links the *contents* of `doc/` into the top-level of the build dir, instead of linking the `doc` directory itself. A normal AUCTeX installation contains the `doc` directory as a whole. We could do `(:defaults "doc" ...)`, but that would pollute the top-level with links to the Texinfo files. Not a big deal I guess, but if we want a "clean" installation, it seems we must manually enumerate the `:files`.

mjkramer:
I should note that this fix still doesn't provide a "complete" AUCTeX installation. To generate some of the docs, `make` must currently be manually run inside the repo.  Buried somewhere in raxod502/straight.el#240 is probably a viable way to automate the process. But that's an issue for another day. For now, at least all of AUCTeX's functionality is working.

hlissner:
> To generate some of the docs, make must currently be manually run inside the repo.

Does that make command work in a shallow clone of the project? If so, adding `:post-build ("make")` to the recipe should do it. If not, then I'll worry about that when I have the time; a jury rigged solution will be necessary.

modules/lang/latex/packages.el
hlissner: Ah, I see the problem. It's fine the way it is then.

mjkramer:
OK, so Straight is actually installing AUCTeX from https://github.com/emacs-straight/auctex rather than upstream. This "fork" already contains almost all of the `make`-generated stuff (as listed in GNUmakefile); the only exceptions are `doc/preview-dtxdoc.texi` and `doc/version.texi`. So, a few options from here:

- Use this fix as-is, and declare that the lack of `preview-dtxdoc.texi` and `version.texi` is not the end of the world.
- Raise an issue to see if Radon can get `emacs-straight/auctex` to include these two missing files. There's no dedicated issue tracker for that repo, but I guess I could just add a comment to https://github.com/raxod502/straight.el/issues/240. Not sure if it's worth bothering him.
- Override the built-in recipe to clone from upstream (GNU Savannah) rather than `emacs-straight/auctex`. I tried doing the following:

```emacs-lisp
(package! auctex
  :recipe (:host nil
           :repo "git://git.savannah.gnu.org/auctex.git"
           :files ("*.el" "*.info" "dir"
                   "doc" "etc" "images" "latex" "style")
           :pre-build (("./autogen.sh") ("./configure") ("make")))
  :pin "3b0a080ae596c26c17b15ba9c71fc5542eae238b")
```

But then I got an error during a (fresh) `doom sync`:

```
      > Cloning project...
      - Checked out project: 401e4aa68e5f46fe72ca6f9a1cb07cd84f5ee981
      > Building project...
      > Building project > Cloning xref...
      > Building project > Building xref...
      x There was an unexpected error
        Message: error
        Error: (error ":pre-build command error in \"xref\" recipe in command \"(\\\"./autogen.sh\\\")\"")
        Backtrace:
          (signal error (":pre-build command error in \"xref\" recipe in command \"(...
          (error ":pre-build command error in %S recipe in command %S" "xref" "(\"./...
          (condition-case err (if (cl-every #'stringp command) (if (apply #'straight...
          (let ((command (car --dolist-tail--))) (condition-case err (if (cl-every #...
          (while --dolist-tail-- (let ((command (car --dolist-tail--))) (condition-c...
          (let ((--dolist-tail-- (if (cl-every #'listp commands) commands (list comm...
          (let ((default-directory repo) (commanderror nil)) (let ((--dolist-tail-- ...
          (if repo (let ((default-directory repo) (commanderror nil)) (let ((--dolis...
          (let* ((commands (and t (if post post-build pre-build))) (repo (and comman...
          (let* ((--cl-rest-- recipe) (pre-build (car (cdr (plist-member --cl-rest--...
      ! Extended backtrace logged to ~/apps/doom-emacs/.local/doom.error.log
```

So it looks like the `:pre-build` options are "leaking" from the `auctex` recipe into those for other packages (in this case, `xref`). Not sure if I'm doing it wrong or if this is a bug in Straight or `core-packages`.

Worth noting: Doing a full build from Savannah would probably break the recipe for Windows users who don't have a GNU toolchain installed. Simply omitting the `:pre-build` on Windows wouldn't work, because if we use the upstream repo, we *have* to do the `:pre-build`.

tl;dr The easiest and most conservative option seems to be sticking with `emacs-straight/auctex` and the amended `:files`. Interestingly, based on https://github.com/raxod502/straight.el/blob/b712a2f31f9bc452a2fa55aebc6753d946bfbb34/straight.el#L3223, it looks like we should not need to do this trick with the `:files`. No idea why Straight isn't doing what the code says it should be doing.

hlissner (approved):

hlissner:
This was a messier issue than I had expected. While incurring Auctex's build process would be the most correct, it would also be the least cross-compatible, so I think this `:files` solution is our best bet for now. I'll revisit this after the CLI rewrite to see if I can get straight to behave. Anyhow, thanks for looking into this, and for your help!

PR #5887, opened 2021-12-11
bump: org-contrib

iyefrat:
emacsmirror/org-contrib@0231706232f2 -> emacsmirror/org-contrib@e3183921779e

The previous commit seems to no longer exist upstream


hlissner:
I've bumped org-contrib as part of a larger bump in cc73c2f, which should resolve this issue. Thanks for bringing it to my attention!

PR #5888, opened 2021-12-11
fix(default): only use `woman` on macOS

ethan-leba:
`man -k "^"` is very slow on macOS, which is what `M-x man` uses for `completing-read`: https://github.com/emacs-mirror/emacs/blob/3af9e84ff59811734dcbb5d55e04e1fdb7051e77/lisp/man.el#L932

```
sh-3.2$ time man -k "^" > /dev/null

real	0m1.599s
user	0m0.391s
sys	0m1.209s
```

It's possible that this could just be a me issue, so if anyone else with a mac can fact-check this that would be great!

### DIFF
--- a/modules/config/default/autoload/default.el
+++ b/modules/config/default/autoload/default.el
@@ -18,10 +18,13 @@ If ARG (universal argument), runs `compile' from the current directory."
 
 ;;;###autoload
 (defun +default/man-or-woman ()
-  "Invoke `man' if man is installed, otherwise use `woman'."
+  "Invoke `man' if man is installed and the platform is not MacOS, otherwise use `woman'.
+
+`man -k \"^\"` is very slow on MacOS, which is what `Man-completion-table' uses to
+generate `completing-read' candidates."
   (interactive)
   (call-interactively
-   (if (executable-find "man")
+   (if (and (not IS-MAC) (executable-find "man"))
        #'man
      #'woman)))
 
--- a/modules/config/default/config.el
+++ b/modules/config/default/config.el
@@ -51,6 +51,16 @@
       (setq-local epa-file-encrypt-to (default-value 'epa-file-encrypt-to)))))
 
 
+(after! woman
+  ;; The woman-manpath default value does not necessarily match man. If we have
+  ;; man available but aren't using it for performance reasons, we can extract
+  ;; it's manpath.
+  (when (executable-find "man")
+    (setq woman-manpath
+          (split-string (cdr (doom-call-process "man" "--path"))
+                        path-separator t))))
+
+
 (use-package! drag-stuff
   :defer t
   :init


elken:
image
First 2 are under zsh, second 2 under sh

ethan-leba:
should i add a note explaining the perf. issue to the docstring, or is it fine as is?

hlissner (changes requested):
> should i add a note explaining the perf. issue to the docstring, or is it fine as is?

Yes, please do.

hlissner (approved):

ethan-leba:
Sorry for proposing more last-minute changes, but I've realized that the `woman-manpath` isn't necessarily correct (not sure how this variable is set, really) -- it's missing some more man exotic locations which contain things like `git`.

```
Value
("/usr/local/share/man" "/usr/share/man" "/opt/X11/share/man" "/Library/Apple/usr/share/man" "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/share/man" "/Library/Developer/CommandLineTools/usr/share/man")

Original Value
("/usr/share/man" "/usr/local/share/man" "/usr/X11/man" "/Library/Apple/usr/share/man"
 ("/bin" . "/usr/share/man")
 ("/sbin" . "/usr/share/man")
 ("/usr/bin" . "/usr/share/man")
 ("/usr/sbin" . "/usr/share/man")
 ("/usr/local/bin" . "/usr/local/share/man")
 ("/usr/local/sbin" . "/usr/local/share/man")
 ("/usr/X11/bin" . "/usr/X11/man")
 ("/usr/bin/X11" . "/usr/X11/man")
 ("/usr/bin/mh" . "/usr/share/man"))
```

Adding this snippet to my config was able to resolve this issue:
```emacs-lisp
(after! woman
  (setq woman-manpath (split-string (cdr (doom-call-process "man" "--path")) path-separator t)))
```

and it seems performant enough to run at load time, though it would probably be ideal to store this value:
```emacs-lisp
(defun test-manpath ()
  (split-string (cdr (doom-call-process "man" "--path")) path-separator t))

(benchmark 1000 '(test-manpath))
"elapsed time: 6.061939s"
```

I'm not sure where this would fit in the default config module (if it does at all), do you have any thoughts?

hlissner:
I'm working on a cleaner way for modules to cache data to file during `doom sync`. In the meantime, I suggest caching the manpath on first use (or when given the prefix arg):

```diff
+(defun +default/man-or-woman (&optional nocache)
   "Invoke `man' if man is installed and the platform is not MacOS, otherwise use `woman'.

 `man -k \"^\"` is very slow on MacOS, which is what `Man-completion-table' uses to
 generate `completing-read' candidates."
+  (interactive "P")
+  (when (or nocache (null (get '+default/man-or-woman 'cached)))
+    (setq woman-manpath
+          (split-string (cdr (doom-call-process "man" "--path"))
+                        path-separator t))
+    (put '+default/man-or-woman 'cached t))
   (call-interactively
    (if (and (not IS-MAC) (executable-find "man"))
        #'man
      #'woman)))
```

hlissner:
Actually, scratch that, your `(after! woman ...)` snippet is good enough. I'll accept putting it in `:config default` for now. It's not the ideal placement, but I'll revisit that once my todo list looks less intimidating.

modules/config/default/autoload/default.el
```diff
@@ -16,6 +16,15 @@ If ARG (universal argument), runs `compile' from the current directory."
...
 
+(after! woman
+  ;; The woman-manpath default value does not necessarily match man. If we have
+  ;; man available but aren't using it for performance reasons, we can extract
+  ;; it's manpath.
+  (when (executable-find "man")
+    (setq woman-manpath
+          (split-string (cdr (doom-call-process "man" "--path"))
+                        path-separator t))))
+
```
hlissner, line 27: Can you move this in `config/default/config.el` in the "reasonable defaults" section?

hlissner (changes requested):

hlissner (approved):

hlissner:
LGTM. Thanks for your help! (And for putting up with all my nitpicks!)

PR #5898, opened 2021-12-14
fix: ccls interfering with lsp mode in other modes

rvlo:
When init.el had both (cc +lsp) and (clojure +lsp), documentation
wouldn't show in the echo area, or would flash and be replaced by the
text: Error processing message (wrong-type-argument hash-table-p nil).

ccls-code-lens-mode was being enabled in the clojure buffers and I would
assume the same can happen with other languages when using LSP.

Restricted ccls-code-lens-mode to cc modes.


gagbo:
It’s probably going to be superseded by https://github.com/hlissner/doom-emacs/pull/5430, since ccls got its lenses normalized with the rest of lsp-mode

hlissner:
Duplicate of #5430

PR #5901, opened 2021-12-16
fix(magit): move code-review files to etc dir

fuzzycode:
The database and log file of `code-review` were put directly inside `.emacs.d`causing the repository to appear dirty. This PR moves the files inside the `.local/etc/code-review` folder to comply with the rest of doom functionality. 
The issue was introduced in 2d3a68df496f when switching to `code-review`.


hlissner (approved):

hlissner:

PR #5903, opened 2021-12-16
bump: :emacs dired

mjkramer:
I just had `doom sync` fail during checkout of `dired-git-info`. It looks like the repo's git history has been rewritten, so that the previously-pinned commit no longer exists. I've updated the hash to point to the "new" `master` (which is still timestamped Dec 2019).


hlissner (approved):

hlissner:

PR #5905, opened 2021-12-17
fix(nim): correct nim-mode SHA

cdlm:
Wrong SHA in fd2788c (looks like a copy-paste mishap).

Ref: nim-lang/nim-mode@744e076


hlissner (approved):

hlissner:
Good catch, thanks!

PR #5907, opened 2021-12-17
fix(magit): unbind M-1 etc. for code-review

fuzzycode:
code-review is built on magit-sections so it will have the same default
bindings for M-1, M-2 etc. To be consistent with the rest of doom, those
should be used for switching workspace and z1, z2 etc. should be used for
toggling outlines in a magit-sections buffer such as the code-review buffer.

This only affects users with :editor (evil +everywhere) enabled.


fuzzycode:
Had some issues with my git config email, hence the updates.

hlissner (approved):

hlissner:

PR #5908, opened 2021-12-18
 Add :send-region and :send-buffer options to `set-repl-handler!`

### LINKED ISSUE #4375: Python send to REPL blank lines problem
This is the same issue as kaz-yos/eval-in-repl#15

```python
def test_function(t):
    print(t)

    print(t)

print("ok")

test_function("ok")
```

select code and call `+python/open-ipython-repl`:

Should evaluate without problems just like when code is evaluated **without** sending it to REPL.

**What actually happened?**
 
```
>>> def test_function(t):
...     print(t)
... 
>>>     print(t)
  File "<stdin>", line 1
    print(t)
IndentationError: unexpected indent
>>> 
>>> print("ok")
ok
>>> 
>>> test_function("ok")
ok
>>> 
>>> 
```

Don't know if there's a smart way to resolve this.
This is just how REPL behaves when a blank line is inserted.
If instead of blank line a `TAB` is inserted this evaluates as expected.

ethan-leba:
Implements the `:send-region` and `:send-buffer` handlers as mentioned on the roadmap [here](https://discourse.doomemacs.org/t/development-roadmap/42#v2201-2), and sneaks in a fix for the Python REPL :)

Fixes #4375

### DIFF
--- a/modules/tools/eval/autoload/eval.el
+++ b/modules/tools/eval/autoload/eval.el
@@ -114,7 +114,10 @@
 
 ;;;###autoload
 (defun +eval/buffer-or-region ()
-  "Evaluate the whole buffer."
+  "Evaluate the region if it's active, otherwise evaluate the whole buffer.
+
+If a REPL is open the code will be evaluated in it, otherwise a quickrun
+runner will be used."
   (interactive)
   (call-interactively
    (if (use-region-p)
--- a/modules/tools/eval/autoload/eval.el
+++ b/modules/tools/eval/autoload/eval.el
@@ -73,7 +73,9 @@
                           (buffer-file-name (buffer-base-buffer))))
                         "emacs")
                  (alist-get 'emacs-lisp-mode +eval-runners)))
-        (+eval/region (point-min) (point-max))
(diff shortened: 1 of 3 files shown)


modules/tools/eval/autoload/eval.el
```diff
@@ -73,7 +73,9 @@
                           (buffer-file-name (buffer-base-buffer))))
                         "emacs")
                  (alist-get 'emacs-lisp-mode +eval-runners)))
-        (+eval/region (point-min) (point-max))
+        (if-let ((buffer-handler (plist-get (cdr (alist-get major-mode +eval-repls)) :send-buffer)))
+            (funcall buffer-handler)
+          (+eval/region (point-min) (point-max)))
```
ethan-leba, line 78: I added handlers for both region and buffer since that's what's proposed on the roadmap, but I'm wondering if just a region handler would suffice, and we can simply pass it point-min/max like the default option is here. Thoughts?

hlissner (approved):

modules/tools/eval/autoload/eval.el
hlissner: I think this makes perfect sense, so let's go with that.

hlissner:
This looks great! I look forward to trying it out. Thanks so much for your help!

PR #5926, opened 2021-12-21
fix(org): tab indentation of section breaks in math environment

roiholtzman:
In org mode while using cdlatex to write math, if one 
writes a math expression in a section (i.e. heading), 
then a `tab` indents the section instead of performing 
a `cdlatex-tab`.
This fix takes care of this issue to have the wanted 
behavior: if in math environment and hit `tab` while 
in section, execute `cdlatex-tab`.

### DIFF
--- a/modules/lang/org/autoload/org.el
+++ b/modules/lang/org/autoload/org.el
@@ -375,6 +375,10 @@ Made for `org-tab-first-hook' in evil-mode."
   (cond ((not (and (bound-and-true-p evil-local-mode)
                    (evil-insert-state-p)))
          nil)
+          ((and org-cdlatex-mode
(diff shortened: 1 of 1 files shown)


hlissner:
There were a few issues with this PR (a merge commit, odd whitespace, and 9203885 did not meet [our git conventions](https://gist.github.com/hlissner/4d78e396acb897d9b2d8be07a103a854)), so I've merged your changes in 67a9573. Thanks for bringing them to my attention!

roiholtzman:
@hlissner thanks a lot! I am sorry I didn't do the PR well, it was my first PR, and I saw only the conventions for the commit after I performed the commit, and when I tried to amend it I got into trouble.
Anyway, thanks a lot for including it!

PR #5929, opened 2021-12-23
docs(org): warn about org-roam-v1 status

iyefrat:
- add doctor warning about eventual removal
- clarify the readme notice, since the migration-wizard has stabilized


iyefrat:
oh wait, this modifies the readme. given the fact that this doesn't really make sense as two separate commits, I'll just undraft this after the docs-rewrite is live.

hlissner (approved):

PR #5930, opened 2021-12-23
fix(default): smartparens rules for haskell

iyefrat:
The default haskell-mode rules would result in an extra } appended to
the closing pair (#5448).

Fix: #5448
Close: #5450

-------

I want this to be fixed, but @celsobonutti seems to be busy, so I incorporated the review requests into a new PR. They're still credited as the commit author of course.


hlissner (approved):

hlissner (approved):

hlissner:

PR #5928, opened 2021-12-23
bump: upgrade lsp-mode

aburtsev:
Fixes #5927

Upgrade lsp-mode to the version with fixes of eslint 8.x support

PR #5933, opened 2021-12-24
bump: :tools lsp

### LINKED ISSUE #5927: eslint 8.x doesn't work

lsp-mode supports eslint 8.x

### What actually happened?

After updating eslint to 8.5 the lsp stops showing hints from eslint.

After investigation, it's appeared that problem is in that version 8.x of eslint introduced some breaking changes.

This issue is already [solved in lsp-mode](https://github.com/emacs-lsp/lsp-mode/issues/3272).

I tried to manually change the version in `.local/straight/repos/lsp-mode/clients/lsp-eslint.el` and everything started to work.

image

### Steps to reproduce

Upgrade eslint to 8.x in any project

### System Information

https://pastebin.com/EricrU8r

aisamu:
Fixes #5927

This supersedes https://github.com/hlissner/doom-emacs/pull/5928 (a newer version) and fixes the issues "introduced" by the bump

emacs-lsp/lsp-mode@41173dca4d6a -> emacs-lsp/lsp-mode@2736652be86

https://github.com/emacs-lsp/lsp-mode/commit/20045c1cb607701237beff62b9e7c2c4a69a5e79 introduces a "breaking" change (quoted since private), renaming `lsp--matching-clients?` to `lsp--supports-buffer`

Co-authored-by: Alexandr Burtsev <me@aburtsev.ru>

### DIFF
--- a/modules/tools/lsp/+lsp.el
+++ b/modules/tools/lsp/+lsp.el
@@ -133,7 +133,7 @@ server getting expensively restarted when reverting buffers."
       (require 'lsp-mode)
       (lsp--require-packages)
       (if (or (lsp--filter-clients
-               (-andfn #'lsp--matching-clients?
+               (-andfn #'lsp--supports-buffer?
                        #'lsp--server-binary-present?))
               (not (memq +lsp-prompt-to-install-server '(nil quiet))))
           (apply fn args)
--- a/modules/tools/lsp/autoload/lsp-mode.el
+++ b/modules/tools/lsp/autoload/lsp-mode.el
@@ -28,7 +28,7 @@
      (require 'lsp-mode)
      (list (completing-read
             "Select server: "
-            (or (mapcar #'lsp--client-server-id (lsp--filter-clients (-andfn #'lsp--matching-clients?
+            (or (mapcar #'lsp--client-server-id (lsp--filter-clients (-andfn #'lsp--supports-buffer?
                                                                              #'lsp--server-binary-present?)))
                 (user-error "No available LSP clients for %S" major-mode))))))
   (require 'lsp-mode)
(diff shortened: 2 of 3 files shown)


iyefrat:
Can you add:
```
Close: #5928
```
to the bottom of your commit?

Also, from looking at the newer commits, I couldn't really find any changes relating to what you mentioned. Can you point them out? Just wondering.

aisamu:
Sure!

> couldn't really find any changes relating to what you mentioned.

They're all squashed into the `bump` commit. Would you rather have a separate `fix` commit?

(As for the changes themselves, they're on the +lsp.el and lsp-mode.el files, which are showing up on the diff)

iyefrat:
Ah, I misunderstood, I thought the fix was in the `lsp-mode` commits. It's good that the fix to the changes introduced in the bump commit are in the same commit.

However, your commit message seems to get the CI stuck in a loop. I'm not why this happened instead of it just failing, but it seems to stem from the fact that the second commit in the bump had an 11-hash instead of a 12-hash. Also (sorry for missing this the first time), when you reference external resources, you should use the shortened hash if possible and add a `Ref:` footer at the end. Could you change the message to the following:

```
bump: :tools lsp

emacs-lsp/lsp-mode@41173dca4d6a -> emacs-lsp/lsp-mode@2736652be86f

Updates module to upstream "breaking" change (quoted since private),
renaming `lsp--matching-clients?` to `lsp--supports-buffer`. 

Fix: #5927
Close: #5928
Ref: emacs-lsp/lsp-mode@20045c1cb607
Co-authored-by: Alexandr Burtsev <me@aburtsev.ru>
```

aisamu:
Done! 
I had tried running the CI scripts locally and they indeed hung, but I misattributed the root cause to differences in the environments.
Thanks for the hand-holding!

PR #5942, opened 2021-12-30
tweak(dired): don't auto-revert when remote

floyza:
On slow connections `dired-auto-revert-buffer` causes excessive
slowdowns; this limits its behavior to local files.


hlissner:

Issue #5904, opened 2021-12-17
lsp-mode throws error void-function lsp--matching-clients?


vikigenius:
### What did you expect to happen?

lsp server should start correctly when opening a rust file.

### What actually happened?

Lsp server does not start, instead it throws the following error.

```
Error (rustic-mode-local-vars-hook): Error running hook "rustic-setup-lsp" because: (void-function lsp--matching-clients?)
Error (hack-local-variables-hook): Error running hook "doom-run-local-var-hooks-h" because: (doom-hook-error rustic-mode-local-vars-hook rustic-setup-lsp (void-function lsp--matching-clients?))
```

### Describe your attempts to resolve the issue

I tried removing the unpin lsp

```lisp
;; (unpin! lsp-mode)
;; (unpin! lsp-ui)
```

After removing the unpin it works, but I want to have the latest version of lsp all the time, so having a fix would be good.

### Steps to reproduce

1. Upgrade to latest develop
2. Unpin lsp-mode to use latest commit
3. Open any rust file.
4. lsp server fails to start and produces error message as described above.

### System Information

https://pastebin.com/L2qm2uae

vikigenius:
The `lsp-mode` maintainers confirm that the `lsp--matching-clients?` has been renamed. And say that it is not supposed to be used externally.

thaenalpha:
I know what the new function name is, but I don't know how to make a contribution to fixing this. My code base is on my fork feature branch which never makes a pull request.
Should I switch to the 'develop' branch to make a pull request on this, and then switch it back to my ongoing branch, leave unstaged changes on this until this was merged into upstream?

michaelpj:
As of [recently](https://github.com/emacs-lsp/lsp-mode/pull/3230) there's a variable to control prompting for server installation, so we can just delete that entire `defadvice` and just set `lsp-enable-server-download`.

michaelpj:
Probably it would also be better if we didn't have lots of things that use internal functions from `lsp-mode`.

hlissner:
> After removing the unpin it works, but I want to have the latest version of lsp all the time, so having a fix would be good.

This isn't a Doom bug. Versions are pihned to avoid exactly these errors. I will close this since it isn't a bug.

In any case, I will bump the module (and remove the advice) soon.

fredfortier:
> > After removing the unpin it works, but I want to have the latest version of lsp all the time, so having a fix would be good.
> 
[...]

PR #5945, opened 2022-01-01
fix(lsp): Use variable instead removed function

ericdallo:
Latest lsp-mode is not working with doom-emacs because of a custom advice that uses a removed private function, the advice can be removed if using a recent new variable for the same purposes.

The new variable `lsp-enable-suggest-server-download` should be used to disable automatic server download, this fixes a issue where doom was using a removed private-function on its custom advice

image

Related issue https://github.com/hlissner/doom-emacs/issues/5904


jeetelongname:
It would be better to split this into a bump and a fix commit with a proper bump commit message.

ericdallo:
Done @jeetelongname

aisamu:
Related PR (bumps lsp and fixes the issue without disabling auto-installation altogether) : https://github.com/hlissner/doom-emacs/pull/5933

PR #5949, opened 2022-01-03
docs(lookup): fix typos

l2dy:
`dash-docset-install-docset` does not exist.


hlissner (approved):

hlissner:

PR #5953, opened 2022-01-03
`:completion vertico` improvements

iyefrat:
- a rework of how the `completing-read-multiple` keybindings work (see commit message for details)
- module bump 
- added `C-c C-s` binding for `embark-collect-snapshot`
- all embark buffers are now handled as popups, and don't die after dismissal
- add `+default/search-emacsd` to manual preview list

### DIFF
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -131,6 +131,7 @@ orderless."
    +default/search-project-for-symbol-at-point
    +default/search-cwd +default/search-other-cwd
    +default/search-notes-for-symbol-at-point
+   +default/search-emacsd
    consult--source-file consult--source-project-file consult--source-bookmark
    :preview-key (kbd "C-SPC"))
   (consult-customize
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -172,6 +172,7 @@ orderless."
         (:map minibuffer-local-map
          "C-;"               #'embark-act
          "C-c C-;"           #'embark-export
+         "C-c C-s"           #'embark-collect-snapshot
          :desc "Export to writable buffer" "C-c C-e" #'+vertico/embark-export-write)
         (:leader
          :desc "Actions" "a" #'embark-act)) ; to be moved to :config default if accepted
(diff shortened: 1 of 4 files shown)


iyefrat:
Oh also I forgot, for some reason in the crm commit, the last line (starting with `S-TAB:`) got recognized as a trailer by the CI. this didn't happen with the previous likes even though they have the xxx: format as well. not sure what's up with that, I had to add an extra line to get the ci to pass

hlissner (changes requested):
> I had to add an extra line to get the ci to pass

Better to use list notation to avoid it:

```
refactor!(vertico): crm keybindings behaviour

BREAKING CHANGE: This commit changes the behaviour of the TAB and RET
keys in a consult-completing read multiple session, in order to make
them more intuitive. The behaviour is now:

- TAB: (unchanged) always select or deselect the current candidate, and if
  the candidate is selected, move the index to the next one (this allows
  for pressing TAB repeatedly to select multiple subsequent candidates).
- RET: If no candidates have been selected, select the current candidate
  and exit the completion session. If some have been selected, disregard
  the current candidate and exit.
- S-TAB: (new) like TAB, but the keeps the input.
```

That, or I can force-merge, despite any CI warnings.

Anyhow, I'll look into fixing the linter soon.

iyefrat:
good idea, I switched to list notation.

modules/completion/vertico/config.el
```diff
@@ -171,11 +172,12 @@ orderless."
         (:map minibuffer-local-map
          "C-;"               #'embark-act
          "C-c C-;"           #'embark-export
+         "C-c C-s"           #'embark-collect-snapshot
          :desc "Export to writable buffer" "C-c C-e" #'+vertico/embark-export-write)
         (:leader
          :desc "Actions" "a" #'embark-act)) ; to be moved to :config default if accepted
   :config
-  (set-popup-rule! "^\\*Embark Export Grep" :size 0.35 :ttl 0 :quit nil)
+  (set-popup-rule! "^\\*Embark" :size 0.35 :ttl nil :quit nil)
```
iyefrat, line 180: @hlissner just to make sure this change doesn't slip by. I think it makes sense to handle all embark buffers similarly, and usually you export stuff when you want to keep it around for a while. However doing `SPC b k` on these buffers when they are displayed in a popup window doesn't seem to actually kill them, but killing them after switching to them with `SPC <` does. Any idea why that might be the case?

iyefrat: i'm delaying this change to the next vertico pr

hlissner (approved):

hlissner:
Great work as always! Thanks again!

PR #5954, opened 2022-01-03
feat(default): add better smerge-basic-map binding

iyefrat:
SPC g m is so much nicer than C-c ^, and this is used for the hydra
anyway if :ui hydra is on

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -438,6 +438,7 @@
        :desc "Revert file"                 "R"   #'vc-revert
        :desc "Copy link to remote"         "y"   #'+vc/browse-at-remote-kill
        :desc "Copy link to homepage"       "Y"   #'+vc/browse-at-remote-kill-homepage
+       :desc "SMerge"                      "m"   smerge-basic-map
        (:when (modulep! :ui hydra)
         :desc "SMerge"                    "m"   #'+vc/smerge-hydra/body)
        (:when (modulep! :ui vc-gutter)


hlissner:
Won't this throw `void-variable: smerge-mode-map` if smerge isn't available at startup?

iyefrat:
Whoops, not sure how that slipped by in local testing. I still think it would be good to have though, do you think that adding it in `:emacs vc` in an `(after! smerge-mode)` block would be a good idea? Or does that obfuscate the binding too much?

There have been some recent `smerge-mode` bindings added to `evil-collection`, which are a possible alternative. but they are for the most part under the `g` prefix key, which makes them not very discoverable.

iyefrat:
boop

iyefrat:
Thanks, rebased. However Henrik's comment is still relevant, when checking this branch out and starting Emacs I gut a void-variable on `smerge-mode-map`. Still have no idea how I didn't catch that the first time around, but when Henrik gets to this pr we can discuss if the proposal above is a good fit.

UndeadKernel:
It'd be great if you could also add the binding to the +emacs-bindinds file as well. It should be in a similar place as where the evil bindings are.

panchoh:
The [smerge-mode](https://github.com/emacs-evil/evil-collection/tree/master/modes/smerge-mode) from evil-collection, for reference.

hlissner:
Sorry for the ridiculously late response, but this should be resolved in 682f151; which makes smerge-mode use the localleader prefix when it's active. Thanks for bringing it to my attention!

PR #5955, opened 2022-01-04
docs(haskell): fix docs for alternative formatter

carlosdagos:
Small fixup to use the proper form when setting an alternative formatter using the `after!` macro.

_Small note:_ This PR targets the `rewrite-docs` branch as per [the Discord message here](https://discord.com/channels/406534637242810369/406624667496087572/927300806250754048)

### DIFF
--- a/modules/lang/haskell/README.org
+++ b/modules/lang/haskell/README.org
@@ -76,7 +76,7 @@ After installing your preferred formatter, make sure to set
 Make sure to configure the lsp to use your perfered formatter, e.g.:
 #+begin_src emacs-lisp
 ;; ~/.doom.d/config.el
-(after!
+(after! lsp-haskell
   (setq lsp-haskell-formatting-provider "brittany"))
 #+end_src
 


iyefrat:
Hey, can you edit your commit message so that `!after` is `after!`?

modules/lang/haskell/README.org
```diff
@@ -76,7 +76,7 @@ After installing your preferred formatter, make sure to set
 Make sure to configure the lsp to use your perfered formatter, e.g.:
 #+begin_src emacs-lisp
 ;; ~/.doom.d/config.el
-(after!
+(after! haskell
```
hlissner, line 79: The package is named `haskell-mode`, so this must be `(after! haskell-mode`

hlissner (changes requested):

modules/lang/haskell/README.org
michaelpj: Probably it should actually be after `lsp-haskell`? (`lsp-haskell` is the `lsp-mode` Haskell client, `haskell-mode` is the old Haskell major mode).

carlosdagos: I had it as `(after! haskell ...)` and that worked fine. But I tried as @michaelpj mentioned and works fine as well, actually. And I think it's actually the better approach, as we're `setq`ing the `lsp-haskell-formatting-provider` var after all. I'll change it to that 😄

hlissner:

PR #5960, opened 2022-01-05
tweak(web): detect and set angular in web-mode

mytram:
Hi 

Just send in a small tweak regarding angular and the ionic framework.

Here are the details: 

- Add @angular/core in detecting angular to set the web-angularjs minor
mode

- At the same time set the web-mode-engine to "angular" so that
`format-all` uses prettier to format ionic `*.page.html` files in
addition to `*.component.html` files.

When `format-all` sees the web-mode-engine is set to "angular", it
applies `prettier` as opposed to `html-tidy`, as `tidy` doesn't
recognise angular component tags and hence errors out. The current
web-mode only sets the engine to "angular" for "*.components.html".
However, ionic angular uses file names such as "*.page.html", setting
engine will apply prettier to these html fils.

Many thanks

Marty

### DIFF
--- a/modules/lang/web/config.el
+++ b/modules/lang/web/config.el
@@ -43,7 +43,10 @@
 (when (featurep! :lang javascript)
   (def-project-mode! +web-angularjs-mode
     :modes '(+javascript-npm-mode)
-    :when (+javascript-npm-dep-p 'angular))
+    :when (+javascript-npm-dep-p '(angular @angular/core))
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
LGTM. Thanks for your help!

mytram:
Thank you for your awesome work!

PR #5972, opened 2022-01-08
fix(clojure): Add clojurec-mode repl+eval handlers

zane:
As `clojurec-mode` is a distinct mode from `clojure-mode` when `+eval/open-repl-same-window` or `+eval/open-repl-other-window` is run in a buffer when `clojurec-mode` is the major mode the list of available REPLs will include REPLs for other languages, for example `emacs-lisp (default)`.

What's more, if you then try to select `clojure (default)` from the list you get the following error in the minibuffer:
```
and: Symbol’s function definition is void: cider-current-repl
```
(This may be an unrelated issue.)

> 👋🏻 This is my first pull request against Doom. Please let me know if I got anything wrong, or if you'd like me to submit an issue for the above behavior!

### DIFF
--- a/modules/lang/clojure/config.el
+++ b/modules/lang/clojure/config.el
@@ -34,9 +34,9 @@
   :hook (clojure-mode-local-vars . cider-mode)
   :init
   (after! clojure-mode
-    (set-repl-handler! 'clojure-mode #'+clojure/open-repl :persist t)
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Thanks for your help!

> and: Symbol’s function definition is void: cider-current-repl

This is a known issue. `:tools debugger` needs to be tweaked to autoload repl handlers and their dependencies. I haven't had time to look into it yet.

PR #5976, opened 2022-01-10
fix(javascript): incorrect lsp formatting for tsx

jsravn:
Fixes https://github.com/emacs-lsp/lsp-mode/discussions/3308

lsp-mode doesn't detect the correct indentation, since it treats the derived tsx-mode as a web-mode. This change modifies the lsp-mode internal alist to recognize tsx-mode correctly. Ideally lsp-mode would have a better way to modify its known mode lists but I couldn't see it.

### DIFF
--- a/modules/lang/javascript/config.el
+++ b/modules/lang/javascript/config.el
@@ -122,7 +122,10 @@
                     'jsx-tide)))))
   :config
   (when (fboundp 'web-mode)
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Thanks again for your help!

PR #5985, opened 2022-01-11
Fix type in org README capture example

Bounga:
Fix typo in org module README


jeetelongname:
Can you uncapitalize Fix. This is to get the ci to pass

iyefrat:
Also as per the pr template, pr's that change module readmes should be done against the `rewrite-docs` branch

Bounga:
Ok sure, sorry for the mistakes. I'll change the case and create a new MR against `rewrite-docs`

PR #6002, opened 2022-01-15
docs(mu4e): enable automatic email fetching

teeann:
Add docs to enable automatic email fetching, which is the default behavior of any sane email client.

### DIFF
--- a/modules/email/mu4e/README.org
+++ b/modules/email/mu4e/README.org
@@ -266,6 +266,14 @@ If you don't like this use:
 (package! mu4e-alert :disable t)
 #+end_src
 
+** Enabling automatic email fetching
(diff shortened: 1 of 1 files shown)


jeetelongname:
the snippet should probably be wrapped in an 
```lisp
(after! mu4e
    ...)
```

iyefrat:
it's a `defcustom` so no need, it won't override user configuration.
also cc @tecosaur

tecosaur:
This seems like it's probably worth mentioning.

tecosaur (approved):

hlissner (approved):

hlissner:
Thanks for your help!

PR #6005, opened 2022-01-15
tweak(rss): use asterisk around workspace name

jeetelongname:
This is to make it conform to the convention that all other app
workspaces conform to (plus it was bugging me)


hlissner (approved):

hlissner:
Good catch. Thanks again!

PR #6006, opened 2022-01-15
docs(haskell): add lsp predicate to format check

jeetelongname:
So that if someone follows the advice of the warning, it will then go
away


hlissner (approved):

hlissner:
Thanks for your help!

PR #6017, opened 2022-01-18
Add graphviz/dot check to plantuml/doctor.el

Davoodeh:
As it is stated in the plantuml [quickstart](https://plantuml.com/starting) guide, plantuml has a soft dependency, which is graphviz/dot.
However, for like 90% of its functionality (10% being sequence diagrams) is required. Therefore it is safe to say it must be in the doctor.el of the module.

Also, it depends on libawt and one other libawt_xawt.so not just "java". The point is, it requires a full openjdk-jre installation not a headless version. Headless version does not provide the second file so java throws a long exception.

I am not gonna add that to the doctor.el. But I suggest it is needed. A pseudo-code like this is my mind but since I was not sure I didn't try to add it for I know there are more aware guys here who can do this in a more optimized manner:

```lisp
(when 
 (grep-in-text (doom-process (concat plantuml-java-path plantuml-java-flags plantuml-jar-path))
  :for "libawt_xawt.so")
 (warn!
  (concat
   "Couldn't find libawt_xawt.so! It appears some Java libraries are missing "
   "(try installing a full Java installation rather than a headless version)")))
```

P.S: sorry for the hasty p/r and 2 commits, if it is a headache to merge the commits please let me know and I will make a new and clean p/r.

### DIFF
--- a/modules/lang/plantuml/doctor.el
+++ b/modules/lang/plantuml/doctor.el
@@ -5,6 +5,9 @@
   ;; java
   (unless (executable-find "java")
     (warn! "Couldn't find java. PlantUML preview or syntax checking won't work"))
+  ;; graphviz
+  (unless (executable-find "dot")
+    (warn! "Couldn't fiind dot. PlantUML might not show any output"))
   ;; plantuml.jar
   (unless (file-exists-p plantuml-jar-path)
     (warn! "Couldn't find plantuml.jar. Install it with M-x plantuml-download-jar")))
--- a/modules/lang/plantuml/doctor.el
+++ b/modules/lang/plantuml/doctor.el
@@ -7,7 +7,7 @@
     (warn! "Couldn't find java. PlantUML preview or syntax checking won't work"))
   ;; graphviz
   (unless (executable-find "dot")
-    (warn! "Couldn't fiind dot. PlantUML might not show any output"))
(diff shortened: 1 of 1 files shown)


jeetelongname:
2 things, 
- this can be one commit if you drop the second and edit the first, this would be a lot cleaner, 
- Your commits need to conform to the [commit conventions](https://gist.github.com/hlissner/4d78e396acb897d9b2d8be07a103a854) this is whats causing the failing checks

if you are not sure how to do this then [have a look at this guide](https://hackernoon.com/beginners-guide-to-interactive-rebasing-346a3f9c3a6d) should be a good introduction to interactive rebasing (in magit you can start an interative rebase by placing your cursor on the commit you want to start the rebase from (in this case the first in your pr) and hitting `ri`

Davoodeh:
@jeetelongname
Thank you so much mate I'll take a look at the guide. Didn't know that guide existed.
I use the web interface of github for quick single line patches like this. That's why I did a second commit instead of amending and also why force pushing was not possible.

You put too much effort in explaining that I just cannot neglect. I'll do a rebase and force push and update this.

hlissner:
I squashed and reword the commit so this could be merged. Thanks for the help!

Davoodeh:
Thank you for your extra effort
I will make sure I'll do my best in making better commits

PR #6024, opened 2022-01-21
Fix .gitignore for eaf

Valeriooooh:
Emacs application framework is installed in the site-lisp/ directory, whenever you doom upgrade you have to remove the directory and reinstall it after.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,9 @@
 test/.local*/
 test/result
 
+# possible eaf installation
+site-lisp/
+
 # emacs tempfiles that shouldn't be there
 .dap-breakpoints
 .org-id-locations


hlissner:
I can't find any mention in EAF's source that would create a site-lisp directory. Are you sure this isn't an artifact of your personal config?

Valeriooooh:
It is not mentioned in the source code but the default installation in the readme file:

[https://github.com/emacs-eaf/emacs-application-framework#install](url)

`
git clone --depth=1 -b master https://github.com/emacs-eaf/emacs-application-framework.git ~/.emacs.d/site-lisp/emacs-application-framework/
`

I think it could also be:
`
*/emacs-application-framework/
`
Maybe it is a more general solution

hlissner:
It'd be better to clone it to `~/.doom.d/emacs-application-framework`, then add that to your load-path instead:

```elisp
;; in ~/.doom.d/config.el
(use-package! eaf
  :load-path "emacs-application-framework/"
  ...)
```

(Note: Doom modifies `use-package`'s `:load-path` to resolve from the file it's in, if not given an absolute path)

Valeriooooh:
That's actually a good alternative, thanks

PR #6029, opened 2022-01-23
feat(clojure): add doctor.el for clj-kondo

caiorulli:
I've added a `doctor.el` to the clojure module with a check for the `clj-kondo` binary, required by `flycheck-clj-kondo`.
I end up forgetting to install clj-kondo when setting up a new computer, and `doom doctor` has always been a friend.

Please let me know if there's anything that needs fixing - the warning text or other things - as this is my first contribution here :eyes:

### DIFF
--- a/core/cli/upgrade.el
+++ b/core/cli/upgrade.el
@@ -13,8 +13,23 @@ following shell commands:
     bin/doom clean
     bin/doom sync -u"
   :bare t
-  (let ((doom-auto-discard force-p))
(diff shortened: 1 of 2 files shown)


caiorulli:
I saw on discord we should actually target master right? :eyes: 
Not sure it will work, let me see if just changing the base works...

caiorulli:
Nope, bad idea. Will reopen the PR

PR #6030, opened 2022-01-23
fix(evil): wrong-number-of-arguments when run-hooks window-selection-change-functions

daanturo:
`window-selection-change-functions` expects hooks to have one argument.

Edit: functions such as `treemacs--follow-project` (by `treemacs-project-follow-mode`) will raise errors after calling `evil-window-vsplit`.


hlissner (approved):

hlissner:
Thanks for your help

PR #6040, opened 2022-01-26
feat(clojure): add doctor.el for clj-kondo

caiorulli:
I've added a doctor.el to the clojure module with a check for the clj-kondo binary, required by flycheck-clj-kondo.
I end up forgetting to install clj-kondo when setting up a new computer, and doom doctor has always been a friend.

Please let me know if there's anything that needs fixing - the warning text or other things - as this is my first contribution here :eyes: 

I saw on discord we should actually target master right? This is the correct PR now (I previously targeted develop with https://github.com/hlissner/doom-emacs/pull/6029)

### DIFF
--- a/modules/lang/clojure/doctor.el
+++ b/modules/lang/clojure/doctor.el
@@ -1,0 +1,6 @@
+;; -*- lexical-binding: t; no-byte-compile: t; -*-
+;;; lang/clojure/doctor.el
+
+(when (featurep! :checkers syntax)
+  (unless (executable-find "clj-kondo")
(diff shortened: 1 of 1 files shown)


jeetelongname:
Doctor changes should use the doc type and make it a little more clear in the message that you added a check for clj-kondo instead of an entire Dr.el. thanks for your work!

caiorulli:
Thanks! @jeetelongname in this case amending the commit to fix it would be the best right? Will fix it right away

jeetelongname:
yeah just rewording the commit should be enough

hlissner (approved):

hlissner:
Thanks for your help!

PR #6054, opened 2022-01-29
Fix a regression in 451be94fb8c7

offbyone:
Fix a regression introduced in 451be94fb8c7, which passed an invalid input to
min when refreshing if there were no org-babel result blocks.


hlissner:
This was already fixed in https://github.com/hlissner/doom-emacs/commit/2c06577edf7c0a625723d653aecc128442437419. Thanks though!

offbyone:
Dang it, I tried :)

PR #6056, opened 2022-01-29
docs(lsp): add reference to LSP tuning guide

teeann:
Add reference to LSP tuning guide in docs.

### DIFF
--- a/modules/tools/lsp/README.org
+++ b/modules/tools/lsp/README.org
@@ -129,6 +129,9 @@ your ~doom!~ block in =$DOOMDIR/init.el=):
 +(python +lsp)
 #+end_src
 
+** Tuning performance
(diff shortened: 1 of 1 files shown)


modules/tools/lsp/README.org
```diff
@@ -129,6 +129,9 @@ your ~doom!~ block in =$DOOMDIR/init.el=):
 +(python +lsp)
 #+end_src
 
+** Tuning performance
```
hlissner, line 132: Please rename the heading "LSP is slow". For consistency's sake, headings under "Troubleshooting" should summarize problems or questions.

hlissner (changes requested):

hlissner:
Thanks for your help!

PR #6057, opened 2022-01-29
tweak(rss): use variable-pitch in elfeed-show-mode

teeann:
Add `variable-pitch-mode` to `elfeed-show-mode` for better legibility

### DIFF
--- a/modules/app/rss/config.el
+++ b/modules/app/rss/config.el
@@ -33,6 +33,7 @@
 
   ;; Enhance readability of a post
   (add-hook 'elfeed-show-mode-hook #'+rss|elfeed-wrap)
+  (add-hook 'elfeed-show-mode-hook #'variable-pitch-mode)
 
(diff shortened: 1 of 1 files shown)


jeetelongname:
Two things this is more a feat than a tweak, second you can use the `add-hook!` syntax sugar to turn this into one call which is more idomatoc

hlissner:
Thanks for the PR, but I cannot accept it. There are two issues with this:

1. I'd rather we use `mixed-pitch-mode`, if we _were_ to impose variable-pitch fonts anywhere.
2. Variable-pitch fonts can disturb alignment, and also requires the user properly configure `doom-variable-pitch-font` before the utility is usable. This adds too more friction than good, so I'd rather leave it to the user to enable variable pitch where they want it, rather than activate it anywhere by default.

I might be open to adding an `+everywhere` flag to `:ui zen` (to further solidify a new convention for opting into any module's global changes), but I need to consider it more carefully. Sorry for the trouble, in any case!

PR #6060, opened 2022-01-29
fix(vterm): remove +vterm--change-directory-if-remote

kreisys:
this is no longer necessary as vterm handles this
natively. moreover, it creates a redundant nested
ssh session whenever `SPC o t` is used to re-open an
existing vterm popup.


hlissner (approved):

hlissner:

PR #6062, opened 2022-01-30
fix(vertico): let C-c C-e run existing embark-after-export-hook

daanturo:
Previously, +vertico/embark-export-write doesn't let embark execute other added hooks in embark-after-export-hook.

Edit: Add the edit command at the end of the hook list since other hooks may assume that the buffer isn't yet editable.


hlissner (approved):

hlissner:

PR #6065, opened 2022-01-30
fix(rss): kill-buffer-hook removes +rss-workspace-name

danilevy1212:
Created a new symbol, `+rss-workspace-name`, which holds the value of the name of both the elfeed buffer and elfeed workspace. Then, I changed the `kill-buffer-hook` so it kills the workspace with this value. A previous commit changed the name of the workspace and buffer, but no the target of the `kill-buffer-hook`. This change fixes it avoiding repetition as well.

PR #6066, opened 2022-01-30
fix(rss): unbalanced parens in map!

danilevy1212:
I introduced an umbalanced parens expression in https://github.com/hlissner/doom-emacs/pull/6065, this PR fixes it.

### DIFF
--- a/modules/app/rss/config.el
+++ b/modules/app/rss/config.el
@@ -61,9 +61,9 @@ easier to scroll through.")
       "q" #'elfeed-kill-buffer
       "r" #'elfeed-search-update--force
       (kbd "M-RET") #'elfeed-search-browse-url)
-    (map! (:map elfeed-show-mode-map)
-       :n "gc" nil
-       :n "gc" #'+rss/copy-link)))
+    (map! :map elfeed-show-mode-map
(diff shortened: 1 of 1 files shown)


modules/app/rss/config.el
```diff
@@ -61,9 +61,9 @@ easier to scroll through.")
       "q" #'elfeed-kill-buffer
       "r" #'elfeed-search-update--force
       (kbd "M-RET") #'elfeed-search-browse-url)
-    (map! (:map elfeed-show-mode-map)
+    (map! (:map elfeed-show-mode-map
```
hlissner, line 64: Could you remove the inner parens and align each line with the `:map`? This would be consistent with other uses of this macro. I.e.

```elisp
    (map! :map elfeed-show-mode-map
          :n "gc" nil
          :n "gc" #'+rss/copy-link)))
```

hlissner (changes requested):

modules/app/rss/config.el
danilevy1212: Done :)

hlissner (approved):
Looks good to me!

PR #6077, opened 2022-02-03
module: add org-roam-bibtex to :tools/biblio

brianmcgillion:
This change adds org-roam-bibtex to the :tools/biblio module. org-roam-bibtex provides deep integration of bibliographic tools into the org-roam workflow.

It is a new module in Doom and this patch is a replacement of https://github.com/hlissner/doom-emacs/pull/2888 which had made some choices, mainly about handling the configuration paths, that were contensious. Following work in orf-ref v3 the multiple paths are no longer required so it may not warrent inclusion.

As such this patch focuses on just adding the org-roam integration.

### DIFF
--- a/modules/tools/biblio/config.el
+++ b/modules/tools/biblio/config.el
@@ -49,3 +49,44 @@
   :defer t
   :config
   (add-to-list 'ivy-re-builders-alist '(ivy-bibtex . ivy--regex-plus)))
+
+
+(use-package! org-roam-bibtex
+  :when (featurep! :lang org +roam2)
+  :after org-roam
+  :preface
+  ;; if the user has not set a template mechanism set a reasonable one of them
+  ;; The package already tests for nil itself so we define a dummy tester
+  (defvar orb-preformat-keywords
+    '("title" "url" "file" "author-or-editor" "keywords" "citekey" "pdf"))
+  :hook (org-roam-mode . org-roam-bibtex-mode)
+  :custom
+  (orb-note-actions-interface (cond ((featurep! :completion ivy)  'ivy)
+                                    ((featurep! :completion helm) 'helm)
+                                    ((t                           'default))))
+  :config
+  (setq orb-insert-interface (cond ((featurep! :completion ivy)  'ivy-bibtex)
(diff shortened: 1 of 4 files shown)


brianmcgillion:
Following the work of @bdarcus on org-cite and citar integration, https://github.com/hlissner/doom-emacs/pull/2888 can be abandoned in favor of a more streamlined addition of org-roam-bibtex.

bdarcus:
Aside from the broader discussion on org-ref, some nuances to consider:

With org-cite, ORB is arguably less important, and at least with citar, is not required to integrate with org-roam. Indeed, with the default note function in citar, you get a lighter-weight, still quite functional, experience.

I don't think, then, that it should be installed and configured, as it is now, when using the vertico module. If we want to include this anyway, it might be best behind a flag?

Thinking forward, I'm sure it's just a matter of time before we see org-cite processors for ivy-bibtex and helm-bibtex.

RomeoV:
Just adding my 2ct that I recently switched from the `org-roam-bibtex` stack to `citar` + `embark` + `vertico` and it has been great, to the extend that I removed all my `org-ref` related packages. Amazing job done there!

modules/tools/biblio/config.el
```diff
@@ -49,3 +49,44 @@
   :defer t
   :config
   (add-to-list 'ivy-re-builders-alist '(ivy-bibtex . ivy--regex-plus)))
+
+
+(use-package! org-roam-bibtex
+  :when (featurep! :lang org +roam2)
```
bdarcus, line 55: Per my comment, if adding this, it should have a dedicated flag, given org-roam org-cite support makes this now optional.

https://github.com/hlissner/doom-emacs/pull/6077#issuecomment-1028940175

hlissner:
Sorry for the late response: I've decided to go all in on the citar stack and, eventually, phase out the ivy module (since it and vertico's goals are so similar, and vertico is much better at reaching them), so I must turn down this PR. Thank you for the help and for bringing it to my attention, in any case!

bdarcus:
Coincidentally, we just added a key feature to citar-org-roam this week
that allows users to include arbitrary citar data in their capture
templates.

PR #6084, opened 2022-02-04
fix(org): parity for headlines & items in `+org--insert-item`

vherrmann:
When you have a structure like
``` org
- ThisX
  - Hi 
  - No
```
in org mode and execute `+org-insert-item` (the "X" represents the pointer), with the current behavior the following is created:
``` org
- This
  - X
  - Hi 
  - No
```

This pr will change the behavior, so that its equivalent to the one of headlines:
``` org
- This
  - Hi 
  - No
- X
```

### DIFF
--- a/modules/lang/org/autoload/org.el
+++ b/modules/lang/org/autoload/org.el
@@ -30,20 +30,31 @@
     (pcase (org-element-type context)
       ;; Add a new list item (carrying over checkboxes if necessary)
       ((or `item `plain-list)
-       ;; Position determines where org-insert-todo-heading and org-insert-item
-       ;; insert the new list item.
-       (if (eq direction 'above)
-           (org-beginning-of-item)
-         (org-end-of-item)
-         (backward-char))
-       (org-insert-item (org-element-property :checkbox context))
-       ;; Handle edge case where current item is empty and bottom of list is
-       ;; flush against a new heading.
-       (when (and (eq direction 'below)
-                  (eq (org-element-property :contents-begin context)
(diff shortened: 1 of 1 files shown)


nipuna-perera:
I think the way this works now is better. It is possible to achieve what you are looking for by using `C-RET` instead. Eg:

```
- ThisX
   - Hi
   - No
```

`C-RET` will result in

```
- This
   - Hi
   - No
- X
```
For simple `RET` I would expect the current behavior of creating a new sub-item

vherrmann:
@nipuna-perera As far as i know, C-RET is bound to `+org/insert-item-below`, which uses `+org/insert-item`, while `RET` changes nothing.

nipuna-perera:
> @nipuna-perera As far as i know, C-RET is bound to `+org/insert-item-below`, which uses `+org/insert-item`, while `RET` changes nothing.

You are right, my bad. I forgot I had mine set up according to this - https://kitchingroup.cheme.cmu.edu/blog/2017/04/09/A-better-return-in-org-mode/

hlissner (approved):

hlissner (changes requested):
`develop` is EOL. Could you target the `master` branch instead?

vherrmann:
Of course

hlissner (approved):

hlissner (changes requested):
Could you rebase it onto the top of `master`?

hlissner (approved):

hlissner:
LGTM. Thanks for your help (and putting up with my nits)!

PR #6088, opened 2022-02-06
tweak: don't highlight non-default indentations when read-only

### LINKED ISSUE #6087: Don't use whitespace-mode for read-only buffers
### Describe your request

For read-only buffers, we are not supposed to edit them so turning on `whitespace-mode` only shows distractions when browsing system files.

It's especially annoying when viewing Emacs' code lisp library in  (`.el` files in `/usr/share/emacs/`).

### Briefly explain its use-case

1. Open `/usr/share/emacs/site-lisp/idna.el`
2. See ugly whitespaces and we can't edit them

daanturo:
May fix #6087.

### DIFF
--- a/core/core-ui.el
+++ b/core/core-ui.el
@@ -104,6 +104,7 @@ Does nothing if `whitespace-mode' or `global-whitespace-mode' is already active
 or if the current buffer is read-only or not file-visiting."
   (unless (or (eq major-mode 'fundamental-mode)
               (bound-and-true-p global-whitespace-mode)
+              buffer-read-only
               (null buffer-file-name))
     (require 'whitespace)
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Thanks for the PR (and sorry for the late response), but I have to turn it down because there are instances where a buffer is read-only and non-standard indentation should be highlighted (such as when reading Emacs' built-in source, especially when located in read-only trees -- for example, when Emacs is installed through nix). It might be better to disable it in special-mode derivatives, but I'm convinced a more configurable solution is required. I'll explore that once 3.0 or 3.1 is out. Thanks for bringing this to my attention, anyway!

PR #6096, opened 2022-02-08
fix(org): don't error org-download at top level

lucasvreis:
Org-attach (and thus org-download) can work before the first headline if the user sets `org-attach-auto-tag` to `nil`. But when inserting files in this situation, the `+org--dragndrop-then-display-inline-images-a` advice will display an error message due to using `org-back-to-heading`. So I think it makes more sense to use `org-back-to-heading-or-point-min` in the advice.

### DIFF
--- a/modules/lang/org/contrib/dragndrop.el
+++ b/modules/lang/org/contrib/dragndrop.el
@@ -62,7 +62,7 @@
       (save-excursion
         (org-display-inline-images
          t t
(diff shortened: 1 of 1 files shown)


hlissner (approved):

lucasvreis:
Sorry about the commit body line length, should I fix it?

hlissner (approved):

hlissner:

hlissner:

PR #6105, opened 2022-02-10
nit(elixir): replace comment of Elixir recipe

BartOtten:
Changed the comment of the Elixir recipe. 

The current comment might annoy some Erlang developers and therefor 
was removed from an Elixir forum thread. As Elixir is standing on the shoulders 
of Erlang, let's not offend those who feed it.

### DIFF
--- a/init.example.el
+++ b/init.example.el
@@ -122,7 +122,7 @@
        ;;data              ; config/data formats
        ;;(dart +flutter)   ; paint ui and not much else
        ;;dhall
-       ;;elixir            ; erlang done right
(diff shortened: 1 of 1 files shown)


hlissner:
This has already been changed to `an elegant language for a more civilized age` [in the rewrite-docs branch](https://raw.githubusercontent.com/hlissner/doom-emacs/rewrite-docs/modules/lang/erlang/README.org) (see the #+subtitle: line), so this isn't necessary. 

The lines in init.example.el will be generated from #+subtitle lines in their respective modules' readme once that branch is merged.

PR #6109, opened 2022-02-12
tweak(cc): default to +lsp

l2dy:
Without +lsp, both irony-mode and rtags require extra configuration for
system header path on macOS.

With +lsp, clangd works well out of the box, and it's easier to
configure than two separate daemons.

Ref: https://github.com/Sarcasm/irony-mode/wiki/Mac-OS-X-issues-and-workaround
Ref: Andersbakken/rtags#811


hlissner (approved):

hlissner:

PR #6113, opened 2022-02-14
fix(company): +company--backends minor mode detection

yqrashawn:
the old condition will append any backends added by
`set-company-backend!` once the major-mode in loaded

I see "minor modes" in comments, but the detection also detect major modes.

-------

With config

``` emacs-lisp
(setq +company-backend-alist
  '((prog-mode company-capf)))
(after! js2-mode (set-company-backend! js2-mode 'company-tabnine-capf))
```

1. Open a elisp file. `company-backends` in local buffer is `'(company-capf)`
2. Open a js file. `company-backends` in local buffer is `'(company-tabnine-capf company-capf)`
3. Open another elisp file. `company-backends` in local buffer is `'(company-tabnine-capf company-capf)`

The `company-backends` in step 3 should be same with step 1.

### DIFF
--- a/modules/completion/company/autoload.el
+++ b/modules/completion/company/autoload.el
@@ -56,7 +56,8 @@ Examples:
        (append (cl-loop for (mode . backends) in +company-backend-alist
                         if (or (eq major-mode mode)  ; major modes
                                (and (boundp mode)
-                                    (symbol-value mode))) ; minor modes
+                                    (symbol-value mode)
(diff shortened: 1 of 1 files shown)


yqrashawn:
image
image
On my side it's caused by #5401 [here](https://github.com/hlissner/doom-emacs/pull/5401/files#r806626161)

yqrashawn:
This is fixed in #5401

PR #6121, opened 2022-02-17
change opensource  to freedom license

libererer:
{{{ 

This is greatest text editor. It should have greatest license.

 }}}


hlissner:
No.

PR #6132, opened 2022-02-19
feat(mu4e): more helpful headers mode-line symbols

tecosaur:
Use all-the-icons for nicer symbols, and add helpful minibuffer on-hover descriptions.

image

image

image

### DIFF
--- a/modules/email/mu4e/config.el
+++ b/modules/email/mu4e/config.el
@@ -76,6 +76,26 @@
           (:from-or-to . 25)
           (:subject)))
 
+  ;; Better search symbols
+  (letf! ((defun make-help-button (text help-echo)
+            (with-temp-buffer
+              (insert-text-button text
+                                  'help-echo help-echo
+                                  'mouse-face nil)
+              (buffer-string)))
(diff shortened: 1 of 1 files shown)


modules/email/mu4e/config.el
```diff
@@ -76,6 +76,26 @@
           (:from-or-to . 25)
           (:subject)))
 
+  ;; Better search symbols
+  (cl-flet* ((make-help-button (text help-echo)
```
hlissner, line 80: Could you replace the `cl-flet` call with `letf!`? The lexical-ity of `cl-flet*` is really needed here. I.e.

```lisp
(letf! ((defun make-help-button (text help-echo)
          (with-temp-buffer
            (insert-text-button text
                                'help-echo help-echo
                                'mouse-face nil)
            (buffer-string)))
        (defun make-help-button-cons (text1 text2 help-echo)
          (cons (make-help-button text1 help-echo)
                (make-help-button text2 help-echo))))
```

hlissner (changes requested):

hlissner (approved):

hlissner:

PR #6141, opened 2022-02-22
fix(web): typo

bashu:
web-mode-attribute-transpose should be bind to SPC m a t, instead of SPC m a p


hlissner (approved):

hlissner:

PR #6145, opened 2022-02-23
fix(org): multiple fixes for +org--insert-item

vherrmann:
After using the `+org--insert-item` of #6084 for a decent time, I have found multiple issues with it, e.g. the labels of ordered lists were not updated and it also made a difference if the point is before or after the bullet.

This pr fixes this behavior, but uses a horrible hack. For the edge
case of an empty item, it inserts a no-break space momentarily, so that
`org-insert-item' does the right thing.

### DIFF
--- a/modules/lang/org/autoload/org.el
+++ b/modules/lang/org/autoload/org.el
@@ -30,32 +30,40 @@
     (pcase (org-element-type context)
       ;; Add a new list item (carrying over checkboxes if necessary)
       ((or `item `plain-list)
-       (let* ((item
-               (if (eq 'item (org-element-type context))
-                   context
-                 ;; if the context has type `plain-list', find closest item
-                 (let ((struct (org-element-property :structure context)))
-                   (save-excursion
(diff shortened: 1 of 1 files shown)


vherrmann:
For unambiguity: This pr in fact fixes multiple issues, but fixes them in one go.

HyunggyuJang:
Think this code doesn&rsquo;t cover the following case:  

```org
- ▮
  test
```

where cursor is represented as ▮, which should be transformed with `+org--insert-item` to  

```org
-
  test
- ▮
```

but, with the current modification only results to  

```org
- ▮
-
  test
```

vherrmann:
@HyunggyuJang I just tried it just now and it works for me.

vherrmann:
It should now work.

hlissner (approved):

hlissner:
Sorry for the late response! Thanks for your help!

vherrmann:
np, this pr fixes my own messup…

PR #6153, opened 2022-02-25
bump :lang rust

brotzeit:
brotzeit/rustic@804ebfe -> brotzeit/rustic@eb5ca14

### DIFF
--- a/modules/lang/rust/config.el
+++ b/modules/lang/rust/config.el
@@ -14,8 +14,10 @@
     (defalias 'org-babel-execute:rust #'org-babel-execute:rustic)
     (add-to-list 'org-src-lang-modes '("rust" . rustic)))
   :config
-  (setq rustic-indent-method-chain t)
+  ;; rust-mode
+  (setq rust-prettify-symbols-alist nil)
 
+  (setq rustic-indent-method-chain t)
   (set-docsets! 'rustic-mode "Rust")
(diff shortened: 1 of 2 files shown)


andresilva:
This breaks ligatures on `->` for me. I've bisected rustic and nailed down the breaking change to this commit: https://github.com/brotzeit/rustic/commit/1441f1d4675aefbe7d32b8be3c05fac4ece43461. I guess the next step would be to depend directly on `rust-mode` and bisect it as well, I haven't done this yet.

brotzeit:
@andresilva thanks for testing. The issue is that rustic was a fork for years and we must have applied some changes to either rust-mode or rustic in this time that broke or fixed this issue. If you could take a look at rust-mode would be great, I'll also take a look and try to find the commit that contains the related change.

brotzeit:
Maybe this one  https://github.com/rust-lang/rust-mode/commit/959bfce215c342c83d72650b2b35201de263edfe ?

andresilva:
I was just writing that! Yes, that's the commit that broke it for me, thankfully I can just override `rust-prettify-symbols-alist` to `nil` to use the ligatures I have setup. I looked at `rust-mode.el` before bisecting but somehow missed this, it was obvious in hindsight 😜

brotzeit:
=)

I guess you can't push to my branch so I'll add the fix

PR #6154, opened 2022-02-25
remove obsolete rust config

### LINKED ISSUE #6070: Lsp mode still loaded in Rust file even if +lsp flag has been removed

Doom did not load lsp in rust files.

### What actually happened?

Doom still loads lsp in rust files.

With the Help of Yeet in the discord group

1. Make sure I don't have any lsp config with rust/rustic mode

image

image

```
~
❯ rg rust ~/.doom.d/
/home/user/.doom.d/init.el
151:       rust              ; Fe2O3.unwrap().unwrap().unwrap().unwrap()
```

See, no related lsp config for rust

2. Run `doom sync` and restart Emacs. --> Issue persists.
3. Check `SPC h v rust-mode-hook`

image

No clue

4. Check `SPC h v major-mode`

Oh, It uses `rustic-mode`

5. Check `rustic-mode-hook`

image

🍕 Ouh ouh, we found something interesting!

6. check `rustic-lsp-client`

```lisp
rustic-mode-hook is a variable defined in rustic.el.

Value
(+lookup--init-rustic-mode-handlers-h
 rustic-setup-lsp
 +evil-embrace-angle-bracket-modes-hook-h)

View as literal Set

Documentation
Hook run after entering Rustic mode.
```

7. I need to manually disable it 

```
(after! rustic
  (setq rustic-lsp-client nil))
```

As described here: https://github.com/brotzeit/rustic#lsp

---

Thanks, Yeet (Discord moderator). I learned a lot in Thinking about how to debug an issue in Emacs/Doom.
I thought doom would handle it out of the box.

### Steps to reproduce

1. Install doom
2 Enable +lsp in rust mode
3. Play with it
4. Remove the +lsp in rust mode
5. What? +lsp still loaded!

### System Information

https://paste.sr.ht/~azzamsa/fe39b6ebe08400333096c885a311119fff15a790

brotzeit:
Fixes #6070

and

rustic depends on rust-mode since version 2.0 so the rust-mode hack shouldn't be needed anymore

depends on #6153

### DIFF
--- a/modules/lang/rust/config.el
+++ b/modules/lang/rust/config.el
@@ -9,10 +9,6 @@
 
 (use-package! rustic
   :mode ("\\.rs$" . rustic-mode)
-  :init
-  (after! org-src
-    (defalias 'org-babel-execute:rust #'org-babel-execute:rustic)
-    (add-to-list 'org-src-lang-modes '("rust" . rustic)))
   :config
   (setq rustic-indent-method-chain t)
 
@@ -28,6 +24,7 @@
   (remove-hook 'rustic-mode-hook #'flycheck-mode)
   (remove-hook 'rustic-mode-hook #'flymake-mode-off)
   (unless (featurep! +lsp)
+    (setq rustic-lsp-client nil)
     (after! flycheck
       (add-to-list 'flycheck-checkers 'rustic-clippy)))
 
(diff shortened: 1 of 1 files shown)


hlissner (approved):

modules/lang/rust/config.el
```diff
@@ -9,10 +9,6 @@
 
 (use-package! rustic
   :mode ("\\.rs$" . rustic-mode)
-  :init
-  (after! org-src
-    (defalias 'org-babel-execute:rust #'org-babel-execute:rustic)
-    (add-to-list 'org-src-lang-modes '("rust" . rustic)))
```
hlissner, line 15: This still seems necessary. Neither of these lines are autoloaded in rustic-babel.el and the package isn't eagerly loaded, so org won't use `rustic-mode` on `rust` babel blocks until a rustic-mode buffer is opened elsewhere.

hlissner (changes requested):

modules/lang/rust/config.el
brotzeit: Oh, thanks for the hint. I'll fix it and give you an update.

hlissner:
I've implemented your other changes and then some, to reflect changes upstream, along with a minor refactor of our `:lang rust`. For the time being I've kept the `(after! org-src ...)` block, but I'll keep an eye out for any changes upstream that make them redundant. Anyhow, thanks for bringing it to my attention (and for your great work on rustic)!

PR #6190, opened 2022-03-11
tweak(file-templates): remove default JSON template

isker:
As described at https://www.json.org/json-en.html, JSON has multiple
top-level forms (at least objects and arrays, and potentially all values
as well, depending on who you ask). Of these, I would not say array is a
good default. I frequently find myself deleting this default, generally
to use an object instead.

Because there is no consistent winner, and because the template is so
trivial, it seems best to simply delete it: the cost of the template not
matching the user's intent outweighs any benefit it stands to deliver
when it does match the user's intent.


isker:
I understand this is a pretty nitpicky PR.  Feel free to close it as arbitrarily as I have opened it 🌞.  I personally find this template to be pretty annoying, but I can of course just disable it in my config.

hlissner (changes requested):
Fine by me, but could you remove [its respective entry](https://github.com/hlissner/doom-emacs/blob/master/modules/editor/file-templates/config.el#L63) in `+file-templates-alist` as well?

isker:
Whoops, this is what I get for not testing it.

hlissner (approved):

hlissner:
No worries! And yes, missing templates won't trigger an error, but I will change that in the future. Anyhow, thanks for your help!

PR #6194, opened 2022-03-14
fix(dashboard): better dashboard hints

gagbo:
Prioritize `+doom-dashboard-mode-map` normal mode bindings on the dashboard, and add a
small note about it in the documentation.

### DIFF
--- a/modules/ui/doom-dashboard/README.org
+++ b/modules/ui/doom-dashboard/README.org
@@ -10,6 +10,7 @@
 - [[#configuration][Configuration]]
   - [[#a-custom-banner][A custom banner]]
   - [[#adding-text-to-the-dashboard][Adding text to the dashboard]]
+  - [[#customizing-commands-and-bindings][Customizing commands and bindings]]
   - [[#customizing-faces][Customizing Faces]]
 
 * Description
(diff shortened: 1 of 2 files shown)


gagbo:
Abort mission for now. This fails if the map has no normal mode binding

gagbo:
Got a working version, that is slightly dirtier

MormonJesus69420:
Might be a bit unrelated, but it would be good to fix docstring in  https://github.com/hlissner/doom-emacs/blob/42e5763782fdc1aabb9f2624d468248d6978abe2/modules/ui/doom-dashboard/config.el#L85 

since it seems like `:key STRING` is no longer used, and will throw an error if you try to define a PLIST with `:key` element in it, eg.:
```lisp
("Find File Menu"
     :icon (all-the-icons-octicon "search" :face 'doom-dashboard-menu-title)
     :key "f" 
     :action find-file)
```
will give you error:
```
Error (+doom-dashboard-functions): Error running hook "doom-dashboard-widget-shortmenu" because: (error Keyword argument :key not one of (:icon :action :when :face))
```

hlissner:
Unfortunately, a little more finesse was needed for this, as precedence (with `where-is-internal`) is a real pain to get right, whilest avoiding calling `where-is-internal` more than we have to. I done this in b5093a5.

I've also restored `:key`, so it can now be used to statically define a key shortcut string.

Still wondering if rebinding keys needs documenting at all. Doing so in a hook isn't necessary (and should never be necessary except in extreme cases, \*cough\*eshell). Anyhow, I'll put it in the `rewrite-docs` branch if/when I decide to.

MormonJesus69420:
Thanks a lot for working on the issue, I will change my config accordingly.
Have a nice day

PR #6195, opened 2022-03-14
fix(ibuffer): Wrong number of arguments to ibuffer-find-file with ivy enabled.

p4v4n:
Error message on C-x C-f in `ibuffer` :

```
(wrong-number-of-arguments ((t) (_file &optional _wildcards) "Use `counsel-find-file' instead of `find-file'." (interactive) (counsel-find-file (let ((buf (ibuffer-current-buffer))) (if (buffer-live-p buf) (save-current-buffer (set-buffer buf) default-directory) default-directory)))) 0)
```

### DIFF
--- a/modules/emacs/ibuffer/config.el
+++ b/modules/emacs/ibuffer/config.el
@@ -57,11 +57,11 @@
       (+ibuffer-workspace (+workspace-current-name))))
 
   (when (featurep! :completion ivy)
-    (defadvice! +ibuffer-use-counsel-maybe-a (_file &optional _wildcards)
+    (defadvice! +ibuffer-use-counsel-maybe-a (&optional _file _wildcards)
       "Use `counsel-find-file' instead of `find-file'."
       :override #'ibuffer-find-file
       (interactive)
-      (counsel-find-file
+      (counsel-find-file nil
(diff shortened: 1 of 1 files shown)


modules/emacs/ibuffer/config.el
```diff
@@ -57,11 +57,11 @@
       (+ibuffer-workspace (+workspace-current-name))))
 
   (when (featurep! :completion ivy)
-    (defadvice! +ibuffer-use-counsel-maybe-a (_file &optional _wildcards)
+    (defadvice! +ibuffer-use-counsel-maybe-a (&optional _file _wildcards)
```
hlissner, line 60: Please revert this change. `ibuffer-find-file`'s first argument [is not optional](https://github.com/emacs-mirror/emacs/blob/master/lisp/ibuffer.el#L863).

hlissner (changes requested):

modules/emacs/ibuffer/config.el
p4v4n: @hlissner Made the necessary changes to retain the function signature.

hlissner (approved):

hlissner:
Thanks for your help!

PR #6197, opened 2022-03-15
fix(multiple-cursors): fix keybinds for evil-mc-make-and-goto-next(last)-cursor

blorbx:
The keybind for evil-mc-make-and-goto-last-cursor was set to "C-N". The "map!" macro doesn't account for letter capitalization when using the Control prefix so the bind for evil-mc-make-and-goto-last-cursor overrides the bind for evil-mc-make-and-goto-next-cursor. Same situation with evil-mc-make-and-goto-first-cursor overriding the bind for evil-mc-make-and-goto-prev-cursor. This fixes this by instead using "C-S-n" and "C-S-p" keybinds to make the Shift explicit.

### DIFF
--- a/modules/editor/multiple-cursors/config.el
+++ b/modules/editor/multiple-cursors/config.el
@@ -155,9 +155,9 @@
   (map! :map evil-mc-key-map
         :nv "g." nil
         :nv "C-n" #'evil-mc-make-and-goto-next-cursor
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Thanks for your help!

PR #6202, opened 2022-03-18
fix(org): make `org-download-delete' work

zrquan:
`org-download-delete` gets the file path by searching link like [[file:...]], so modifying `org-download-link-format` make it not work. (seems to be a bug of org-download)

### DIFF
--- a/modules/lang/org/contrib/dragndrop.el
+++ b/modules/lang/org/contrib/dragndrop.el
@@ -33,7 +33,6 @@
                      ((executable-find "gnome-screenshot") "gnome-screenshot -a -f %s"))))
 
         org-download-heading-lvl nil
-        org-download-link-format "[[download:%s]]\n"
         org-download-annotate-function (lambda (_link) "")
         org-download-link-format-function
         (lambda (filename)


hlissner:
Has this been fixed upstream? Why was this closed?

zrquan:
Oh, I thought you didn't notice this pr 🤣  Reopened now.

hlissner:
Sorry for the late response. I was hoping the upstream PR would be merged sooner, but since it looks like org-download is no longer maintained I've merged your upstream fix into Doom (in 1f4b9b7 for the time being). Thanks for bringing it to my attention!

zrquan:
@hlissner The width attributes (attr_org/attr_html) don't work for [[download:...]] links either, I don't understand what is the benefit of modifying the org-download-link-format 🤔

hlissner:
@zrquan It's to limit the scope of its functionality and let me selectively target `download:...` links separately from more generic `attach:` and `file:` links. For example, I specifically want `org-download-delete` to be more discerning about *which* file links I want deleted in a line, since it deletes *all* `file:` links on the current line by default.

> The width attributes (attr_org/attr_html) don't work for [[download:...]] links either,

I'll look into this soon.

zrquan:
Got it, and thanks for your work!

PR #6210, opened 2022-03-22
feat(ivy): use variables for counsel-file-jump fd and rg args

jakebox:
Define variables for the arguments passed to `fd` and `rg` in the `+ivy--counsel-file-jump-use-fd-rg-a` advice, located in Ivy's `config.el`.. This change allows the user to easily customize the arguments passed to the `fd` and `rg` commands. For example, I want to see all non-hidden files (including those not in a git repo): `(setq +ivy--counsel-file-jump-fd-args '("-I" "--color=never"))`.

Without these variables, to achieve default behavior users need to either:

1. Modify the source of `modules/completion/ivy/config.el`.
2. Remove the advice, losing a valuable feature (using `fd`/`rg` in `counsel-file-jump`).

This is a minor but useful change for someone used to `counsel-file-jump` coming from Vanilla Emacs. Thanks for your time and consideration!

Relevant to [#385e60cd26bda9cd5deaeb03f24484b18deea9d7](https://github.com/hlissner/doom-emacs/commit/385e60cd26bda9cd5deaeb03f24484b18deea9d7), `fix(ivy): include hidden files in counsel-file-jump`.

### DIFF
--- a/modules/completion/ivy/config.el
+++ b/modules/completion/ivy/config.el
@@ -14,6 +14,18 @@ When 'everything, also preview virtual buffers")
   "A plist mapping ivy/counsel commands to commands that generate an editable
 results buffer.")
 
+(defvar +ivy--counsel-file-jump-fd-args
+  '("H" "-I" "--color=never" "--type" "file" "--type" "symlink" "--follow")
+  "List of args to be passed to `fd' when used for `counsel-file-jump'.
+
+See `+ivy--counsel-file-jump-use-fd-rg-a'.")
+
+(defvar +ivy--counsel-file-jump-rg-args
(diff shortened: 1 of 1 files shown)


hlissner:
Sorry for the late response: I plan to slowly phase out the ivy module (it's and vertico are so alike, yet vertico is better suited to meeting its goals). I also plan to replace projectile with project. The variables consult already exposes for its use of fd and ripgrep (and more to come) meet the goals of this PR, so I will close it. In any case, thanks for your help and for bringing it to my attention!

jakebox:
Understood, thanks for the note! Interesting re Vertico. I guess I should finally give it a go after my 2+ years with Ivy.

PR #6228, opened 2022-03-31
bump: :ui treemacs

daanturo:
Alexander-Miller/treemacs@1820db42c647 -> Alexander-Miller/treemacs@b18a05b1f620

Includes a fix about compilation for `treemacs-project-follow-mode`.


hlissner:
I already had a larger bump for the :ui category in limbo. I've pushed it in d212f3b, which should resolve this. Thanks for bringing it to my attention!

PR #6231, opened 2022-04-01
feat(modeline): Add header-line support to +light

precompute:
> This adds the +headerline flag, which when used with +light makes use of
the header-line instead of the mode-line. Using +headerline also
disables the mode-line globally.

This PR "copies" the mode-line onto the header-line, and disables the mode-line when both the `+light` and `+headerline` flags are present under `:ui modeline`.  

It involves a `:s/mode-line/header-line/g` / `:s/modeline/headerline/g` and then enabling `(global-hide-mode-line-mode)`.  It works without any issues for me.

2022-04-01_17:50:00_1920x1080

I figured this out yesterday and since it just worked, I decided to PR it.

I did think about using a variable instead of a flag, but because it was big change that wouldn't apply until the module was reloaded, I decided to go with a flag. 

I'm sorry if the commit message is inadeqate / wrong.

I haven't modified the README file for the module (as requested in `do-not-pr.md`) but I can add a line for the new flag if necessary.

### DIFF
--- a/modules/ui/modeline/+light.el
+++ b/modules/ui/modeline/+light.el
@@ -585,29 +585,56 @@ lines are selected, or the NxM dimensions of a block selection.")
 ;;
 ;;; Bootstrap
 
-(defvar +modeline--old-format (default-value 'mode-line-format))
+(if (and (featurep! +light)
+         (not (featurep! +headerline)))
+    (progn
+      (defvar +modeline--old-format (default-value 'mode-line-format))
+      (define-minor-mode +modeline-mode
+        "TODO"
+        :init-value nil
+        :global nil
+        (cond
+         (+modeline-mode
+          (setq mode-line-format
+                (cons
+                 "" '(+modeline-bar
+                      +modeline-format-left
+                      (:eval
+                       (propertize
+                        " "
+                        'display
+                        `((space :align-to (- (+ right right-fringe right-margin)
+                                              ,(string-width
+                                                (format-mode-line '("" +modeline-format-right))))))))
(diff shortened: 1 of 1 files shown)


t-e-r-m:
Should `hide-mode-line-excluded-modes` be set to `nil` when the header-line is active?

t-e-r-m:
Any updates on this?  I've been using this for months without a hitch now.

t-e-r-m:
https://i.imgur.com/pJQA7kV.jpg

Is this what you meant?  This works as-is.  
```
system     Debian GNU/Linux 11 (bullseye) Linux 5.18.0-0.bpo.1-rt-amd64 x86_64
emacs      29.0.50 master 1cdc64cdd ~/.emacs.one/ -> ~/.emacs.one/
doom       3.0.0-dev HEAD -> master 33c5f372 2022-07-09 21:10:40 +0200
```

hlissner:
Thanks for the PR and sorry for the tremendously late response, but I must turn this down. For two reasons:

1. I'm considering deprecating the `:ui modeline`'s `+light` flag, as the feature isn't maintained and that is not likely to change.
2. I think this is better suited to an external package, such as [mode-line-in-header.el](https://github.com/emacsattic/mode-line-in-header/blob/master/mode-line-in-header.el).

precompute:
No problem.  FWIW if anyone wants to implement a header-line like the mode-line, feel free to reference my custom config:

https://github.com/precompute/CleanEmacs/blob/master/modeline-simple.el

PR #6232, opened 2022-04-01
fix(vertico): vertico-resume to vertico-repeat.

egidijusz:
Change vertico-resume-* to vertico-repeat-*.

I think it was a typo, since vertico-resume-* does not exist at https://github.com/minad/vertico/blob/main/extensions/vertico-repeat.el


hlissner (approved):

hlissner:
Good catch! That was indeed a typo. Thanks for the help!

PR #6237, opened 2022-04-03
perf(lookup): defer consult-xref

daanturo:
`consult-xref` is autoloaded anyway. There are some packages that
require `xref` (such as lispy through requiring etags) so without
deferring, loading those packages may cause `consult.el` to be loaded.


hlissner (approved):

hlissner (approved):

hlissner:
Good catch! Thanks again for your help!

PR #6239, opened 2022-04-03
tweak(&org-roam2): map zm to close all folds

dlukes:
When I found out that `zr` works in Org-Roam buffers, I automatically assumed `zm` also does. It currently doesn't, but maybe it would be a good idea to add it?


hlissner (approved):

hlissner:
Sounds sensible to me! Thanks for your help!

Issue #3921, opened 2020-09-10
Formatter function for terragrunt


abeluck:
The terraform module works very well.

There's a wrapper around terraform, [terragrunt](https://terragrunt.gruntwork.io/), that has its [own format command](https://terragrunt.gruntwork.io/docs/reference/cli-options/#hclfmt) for it's `.hcl` files.

It should be relatively simple to write a formatter function for this, as the formatter takes 0 arguments. But I've tried and failed :( 

* command: `terragrunt hclfmt` from the pwd of the file
* applies to `*.hcl` files

**System information**
<details><pre>
((system
  (type . gnu/linux)
  (config . "x86_64-pc-linux-gnu")
  (shell . "/usr/bin/zsh")
  (uname . "Linux 4.19.132-1.pvops.qubes.x86_64 #1 SMP Tue Jul 14 03:42:21 UTC 2020 x86_64")
  (path "~/.local/bin" "/usr/local/bin" "/usr/bin" "/bin" "/usr/local/games" "/usr/games" "/usr/lib/emacs/26.1/x86_64-linux-gnu"))
 (emacs
  (dir . "~/.emacs.d/")
  (version . "26.1")
  (build . "Sep 23, 2019")
[...]

PR #6247, opened 2022-04-05
feat(terraform): support terragrunt

peterhoeg:
Introduce a variable for the terraform executable and as terragrunt supports everything terraform does, use that instead of terraform if found on the path. 

We should probably also set `compile-command` - comments?

This doesn't touch the formatter in any way (ref https://github.com/hlissner/doom-emacs/issues/3921)

### DIFF
--- a/modules/tools/terraform/config.el
+++ b/modules/tools/terraform/config.el
@@ -1,18 +1,21 @@
 ;;; tools/terraform/config.el -*- lexical-binding: t; -*-
 
+(defvar +terraform-runner (if (executable-find "terragrunt") "terragrunt" "terraform")
+  "The default runner - terraform or terragrunt")
+
 (when (featurep! +lsp)
   (add-hook 'terraform-mode-local-vars-hook #'lsp! 'append))
-
 
 (after! terraform-mode
   (set-docsets! 'terraform-mode "Terraform")
 
+  (setq-hook! 'terraform-mode-hook compile-command +terraform-runner)
+
   (map! :map terraform-mode-map
         :localleader
-        :desc "terraform apply" "a" (cmd! (compile "terraform apply" t))
-        :desc "terraform init"  "i" (cmd! (compile "terraform init"))
-        :desc "terraform plan"  "p" (cmd! (compile "terraform plan"))))
-
+        :desc "apply" "a" (cmd! (compile (format "%s apply" +terraform-runner) t))
+        :desc "init"  "i" (cmd! (compile (format "%s init" +terraform-runner)))
+        :desc "plan"  "p" (cmd! (compile (format "%s plan" +terraform-runner)))))
 
 (use-package! company-terraform
   :when (featurep! :completion company)


peterhoeg:
This should be relatively safe to merge. Any concerns?

modules/tools/terraform/config.el
```diff
@@ -1,18 +1,19 @@
...
 
   (map! :map terraform-mode-map
         :localleader
-        :desc "terraform apply" "a" (cmd! (compile "terraform apply" t))
-        :desc "terraform init"  "i" (cmd! (compile "terraform init"))
-        :desc "terraform plan"  "p" (cmd! (compile "terraform plan"))))
-
+        :desc (format "%s apply" +terraform-runner) "a" (cmd! (compile (format "%s apply" +terraform-runner) t))
+        :desc (format "%s init"  +terraform-runner) "i" (cmd! (compile (format "%s init" +terraform-runner)))
+        :desc (format "%s plan"  +terraform-runner) "p" (cmd! (compile (format "%s plan" +terraform-runner)))))
```
hlissner, line 16: I'd recommend removing the prefix for these keys' labels (e.g. `:desc "apply"`, `:desc "init"`, etc), since it introduces a load order constraint for folks who change `+terraform-runner`. I.e. if they set the variable like in an `(after! terraform-mode ...)` block in their `$DOOMDIR/config.el` (which I recommend across the board), it'll be too late for these keybinds to see the change. They'd have to set it at top-level or in `$DOOMDIR/init.el`.

As shame we can't pass which-key a lambda to generate the label when they're read.

hlissner (changes requested):
Sorry for the delay; I've just been preoccupied. There's only one issue I have with this (mentioned below), otherwise it looks good to me.

modules/tools/terraform/config.el
peterhoeg: That's a very good point. If we set `compile-command`, we will have the same problem (it not being reset if +terraform-runner is changed) - what's the proper way to deal with that?

hlissner (changes requested):
Some commit history snuck into this PR. Would you mind rebasing it back onto master?

peterhoeg:
> Some commit history snuck into this PR. Would you mind rebasing it back onto master?

My bad - it’s the branch I’m running off. I will need to create a separate branch to avoid all the noise. Apologies.

modules/tools/terraform/config.el
```diff
@@ -1,18 +1,21 @@
...
+  "The default runner - terraform or terragrunt")
+
 (when (featurep! +lsp)
   (add-hook 'terraform-mode-local-vars-hook #'lsp! 'append))
 
-
 (after! terraform-mode
   (set-docsets! 'terraform-mode "Terraform")
 
+  (setq-local compile-command +terraform-runner)
```
hlissner, line 12: `setq-local` sets `compile-command` buffer-locally, but there's no telling what buffer will be "current" when this line gets evaluated. Please replace it with:

`(setq-hook! 'terraform-mode-hook compile-command +terraform-runner)`

hlissner (changes requested):

modules/tools/terraform/config.el
```diff
@@ -1,18 +1,21 @@
...
+  "The default runner - terraform or terragrunt")
+
 (when (featurep! +lsp)
   (add-hook 'terraform-mode-local-vars-hook #'lsp! 'append))
 
-
 (after! terraform-mode
   (set-docsets! 'terraform-mode "Terraform")
 
+  (setq-hook! terraform-mode-hook compile-command +terraform-runner)
```
hlissner, line 12: `terraform-mode-hook` needs to be quoted:

```diff
-  (setq-hook! terraform-mode-hook compile-command +terraform-runner)
+  (setq-hook! 'terraform-mode-hook compile-command +terraform-runner)
```

hlissner (changes requested):

modules/tools/terraform/config.el
peterhoeg: Fixed

hlissner (approved):

hlissner:

PR #6254, opened 2022-04-06
Remove *Messages* spam

elken:
The faces were defined wrong, so every time a buffer came into focus, it
would spam `*Messages*`.

I had it during testing but I assumed it was leftovers from my previous failures.


hlissner (approved):

hlissner:
LGTM. Thanks!

PR #6256, opened 2022-04-06
fix(org): #+ constructs hidden forever

quaestor:
Leaving org-tree-slide-mode did not unhide #+ constructs.
Now +org-present-hide-blocks-h is triggered when leaving the mode as
well, fixing the behaviour.


hlissner:
Thanks for your help!

PR #6263, opened 2022-04-06
feat(undo): add vundo

danilevy1212:
Added vundo recipe for undo-fu users, so they can't stop being jealous of undo-trees visualizer! Changes are gated behind opting out of the `+tree` flag in the `:emacs undo` module.

### DIFF
--- a/modules/emacs/undo/config.el
+++ b/modules/emacs/undo/config.el
@@ -51,6 +51,18 @@
         filename))))
 
 
+(use-package! vundo
+  :unless (modulep! :emacs undo +tree)
+  :custom
+  (vundo-glyph-alist vundo-unicode-symbols)
+  (vundo-compact-display t)
+  :config
+  (when (featurep! :editor evil)
+    (map! :map vundo-mode-map
+          [remap doom/escape] #'vundo-quit))
+  :defer t)
+
+
 (use-package! undo-tree
   :when (featurep! +tree)
   ;; Branching & persistent undo
--- a/modules/emacs/undo/packages.el
+++ b/modules/emacs/undo/packages.el
@@ -4,4 +4,5 @@
 (if (featurep! +tree)
     (package! undo-tree :pin "e326c6135e62f5fe8536528d3acd5e798f847407")
   (package! undo-fu :pin "ab8bc10e424bccc847800c31ab41888db789d55d")
-  (package! undo-fu-session :pin "3e810c7c9ab75d2b6f92c7c876290abbc164e750"))
+  (package! undo-fu-session :pin "48544cb102fd3d761acf92598076b20bbb4075f9")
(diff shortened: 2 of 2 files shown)


danilevy1212:
I feel like there should be a mention of this package on the module `readme.org`. Happy to add theme to rewrite-docs if need be and this PR gets approved :)

elken:
FYI Vundo is now [on ELPA](https://elpa.gnu.org/packages/vundo.html)

danilevy1212:
@elken Thank you, updated the recipe :+1:

hpfr:
For when Henrik gets around to this, my vote would be to put `vundo` on `SPC b u` and move `doom/sudo-save-buffer` to `SPC b U`. It would be a breaking binding change, but the replacement is read-only unless the user presses more keys in quick succession, which isn’t that likely if they were assuming the binding would save the buffer. I think it’s worth considering because `vundo` is likely to be much more commonly used than saving the buffer as root.

hpfr:
[evil-collection now supports vundo.](https://github.com/emacs-evil/evil-collection/commit/535ad8dc52137f502bfa584925705583209aa093)

@danilevy1212 It looks like you can remove all the remaps except `doom/escape`.

hpfr:
If you like, you can checkout this branch, download the attached, and run `git am 0001-tweak-undo-remove-bindings-in-evil-collection.txt` (or `M-x magit-am-apply-patches` :wink:) to apply a commit with the fix, and then push.

[0001-tweak-undo-remove-bindings-in-evil-collection.txt](https://github.com/doomemacs/doomemacs/files/8929846/0001-tweak-undo-remove-bindings-in-evil-collection.txt)

danilevy1212:
Thanks @hpfr, did the changes :)

danilevy1212:
@hpfr I added an `+EMACS28` flag detector to prevent problems from emacs 27 users

modules/emacs/undo/config.el
```diff
@@ -51,6 +51,18 @@
         filename))))
 
 
+(use-package! vundo
+  :unless (featurep! +tree)
+  :custom
+  (vundo-glyph-alist     vundo-unicode-symbols)
```
tecosaur, line 57: Minor style nit, I don't think the extra space does any good here.

danilevy1212: Removed the space :)

tecosaur: Thanks :)

BenediktBroich:
LGTM!

hlissner:
Sorry for taking so long to get to this! I've merged this in 9350b44 after some modifications, and included a rudimentary keybind. Thanks for your help!

> For when Henrik gets around to this, my vote would be to put vundo on SPC b u and move doom/sudo-save-buffer to SPC b U. 

I agree, but will temporarily shelve it until I'm done moving modules out into their [own repo](https://github.com/doomemacs/modules).

danilevy1212:
Thank you, Henrik!

PR #6266, opened 2022-04-07
docs: add texlive-latexextra dependency for pdflatex exports

noisycomputation:
Exporting an org file to pdf via LaTeX (org-export-dispatch or 'SPC m
e', then 'l p') requires texlive-latexextra (Arch package name) to be
installed on the system.

For reference, the error raised when texlive-latexextra was missing was
that the file 'wrapfig.sty' was not found. The specific error likely
depends on the contents of the exported file. Internet searches reveal
that pdflatex may also complain about 'utf8x.def' being missing; this
file is also provided by texlive-latexextra.

### DIFF
--- a/modules/lang/org/README.org
+++ b/modules/lang/org/README.org
@@ -182,7 +182,7 @@ brew install gnuplot
 
 ** Arch Linux
 #+begin_src sh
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:

PR #6269, opened 2022-04-08
lang/rust: update readme

Schievel1:
Added some more instructions to make it easier for new users to get started with Rust in Doom.


hlissner:
Thanks for the PR, but I'm afraid I must turn it down. It expands on things that will be adequately explained elsewhere (and linked to) after the 3.0 release, or aren't particularly relevant to the module (e.g. the tidbit about Rustlings). That said, there were some small clarifications that I'll work into the module docs later, when I revisit it soon. 

In any case, thanks for bringing these issues to my attention!

PR #6271, opened 2022-04-09
feat(default): on evil bind rename-buffer to <leader> b R

iyefrat:
I use this function a lot, and the native `C-x x r` binding isn't really doom's style


hlissner (approved):

hlissner:

PR #6281, opened 2022-04-11
feat(literate): add async config tangling

tecosaur:
Does what it says on the tin. I think it's great, and some users have mentioned huge improvements.


hlissner (approved):

hlissner:
I think it can be refactored a bit, but besides that it looks good to me. Thanks again for your help!

PR #6287, opened 2022-04-12
fix(default): +default/yank-buffer-path echos the file name as nil on Emacs 29 PGTK

daanturo:
Emacs 29 with PGTK's kill-new doesn't return its STRING argument (with the default settings) anymore, so we explicitly pass the file path to prevent "Copied path to clipboard: nil", (the command still correctly copies the path without this patch, though).

### DIFF
--- a/modules/config/default/autoload/text.el
+++ b/modules/config/default/autoload/text.el
@@ -38,11 +38,14 @@
   (interactive)
   (if-let (filename (or (buffer-file-name (buffer-base-buffer))
                         (bound-and-true-p list-buffers-directory)))
-      (message "Copied path to clipboard: %s"
-               (kill-new (abbreviate-file-name
-                          (if root
-                              (file-relative-name filename root)
-                            filename))))
+      (let ((path (abbreviate-file-name
+                   (if root
+                       (file-relative-name filename root)
+                     filename))))
+        (kill-new path)
+        (if (string= path (car kill-ring))
+            (message "Copied path: %s" path)
+          (user-error "Couldn't copy filename in current buffer")))
     (error "Couldn't find filename in current buffer")))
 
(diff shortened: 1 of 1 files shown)


agzam:
Confirming:

GNU Emacs 29.0.50 (build 1, x86_64-apple-darwin21.5.0, NS appkit-2113.50 Version 12.4 (Build 21F79)) of 2022-07-25

@hlissner Can you merge this? There's zero risk of breaking anything. It's a bit annoying to see "Copied path to clipboard: nil" all the time. 
I was about to submit a fix, but then I found this PR. 

@daanturo, maybe it makes sense to trim the message, removing "to clipboard", where else would it be copied?

daanturo:
> maybe it makes sense to trim the message, removing "to clipboard"

I agree, "Copied" implicitly mentions "clipboard" anyway, but it's up to @hlissner 

Also, I think it's safer if we do a little verification after copying & before emitting the message, like this:
```elisp
        (kill-new path)
        (if (string= path (car kill-ring))
            (message "Copied path: %s" path)
          (user-error "Couldn't copy filename in current buffer"))
```

hlissner (changes requested):
I agree with @daanturo's suggestion. Could you make that change?

EDIT: Would you also mind rewording the commit to use `fix` intead of `tweak`, since it is repairing broken functionality.

hlissner (approved):

hlissner:
LGTM. Thanks again for your help! (And sorry for taking so long to get around to this)

agzam:
> And sorry for taking so long to get around to this

Not at all. If I learned anything about programming - 80% of problems resolve themselves if you just wait long enough. Also, 95% of statistical statements are totally made up. Including this one.

PR #6288, opened 2022-04-12
fix(eval): lookup RET binding in insert if evil

ethan-leba:
Otherwise, if the REPL was not in insert mode the send-to-buffer would
fail.


hlissner (approved):

hlissner:

PR #6298, opened 2022-04-15
feat(clojure): use lsp features when +lsp flag is enabled

ericdallo:
As discussed on Discord #contributing, it's pretty common to users manually disable cider and clj-refactor features even if using +lsp flag, this PR just check if +lsp is enabled and avoiding enable those features, using lsp ones


hlissner (approved):

hlissner:

PR #6300, opened 2022-04-16
feat(org): add +notify option

teeann:
Add `org-wild-notifier` to support notification functionality for org-mode TODOs.
Let me know if there's any feedback, thanks!

### DIFF
--- a/modules/lang/org/contrib/notify.el
+++ b/modules/lang/org/contrib/notify.el
@@ -1,0 +1,7 @@
+;;; lang/org/contrib/notify.el -*- lexical-binding: t; -*-
+;;;###if (featurep! +notify)
+
+(use-package! org-wild-notifier
+  :hook (org-load . org-wild-notifier-mode)
+  :config
+  (setq org-wild-notifier-alert-time '(60 30)))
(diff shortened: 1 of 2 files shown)


peterhoeg:
While I haven't been using this PR, I am using org-wild-notifier with doom and it works great, so :+1: from here.

hpfr:
`org-load-hook` is obsolete; a better option:
```elisp
(after! org
  (org-wild-notifier-mode)
  (setq org-wild-notifier-alert-time '(60 30)))
```

The configuration here is trivial enough that I don’t think it warrants a flag, especially since the alert time is highly variable according to personal preference. Doom is generally about sane defaults and fixing interoperability issues across packages with complex functionality. I think this one is best left to user configs, but of course, it’s up to Henrik.

peterhoeg:
We should have a "user contributed modules/config snippets" repository where smaller things like this can go for inspiration if they don't meet the complexity threshold for inclusion into upstream proper.

iyefrat:
> We should have a "user contributed modules/config snippets" repository where smaller things like this can go for inspiration if they don't meet the complexity threshold for inclusion into upstream proper.

A good place for these types of random snippets to go is on [the Doom discourse](https://discourse.doomemacs.org). Once Doom 3.0 is out, it will be much easier to integrate third part modules, and there will also be a `doomemacs-contrib` repo.

hlissner:
Thanks for the PR, but I must turn it down, for the reasons hpfr mentioned: the implementation for this flag is trivial enough for folks to set up themselves, and the alert timing is too opinionated for us to set, so I'd rather leave it to users to set up. Thanks for bringing it to my attention, in any case!

Issue #5694, opened 2021-10-26
`getting_started.org` missing MSYS2 based Windows installation method


edgimar:
### Describe your request

The `getting_started.org` file contains a few ways of installing emacs, but one relatively straightforward way is via an [MSYS2](https://www.msys2.org) environment.  This provides a native Windows executable, and also a package-management system that allows needed dependencies to be easily installed.

The "feature request" is simply to add this as a listed option in the getting started document.

### Briefly explain its use-case

N/A

samtux:
Here is a personal documentation to follow up the installation:

https://github.com/samtux/doom-emacs/blob/getstarted_msys2/docs/getting_started.org#with-msys2-and-mingw64

samtux:
Here is a personal documentation to follow up the installation:

https://github.com/samtux/doom-emacs/blob/getstarted_msys2/docs/getting_started.org#with-msys2-and-mingw64

PR #6305, opened 2022-04-17
Add method install in Windows with MSYS2 and MingW64 #5694

samtux:
Fixes #0000 


hlissner:
https://gist.github.com/hlissner/bb6365626d825aeaf5e857b1c03c9837

Issue #6144, opened 2022-02-23
Rust files opened via `magit-find-file` seem to be sent to the underlying server


Patryk27:
### What did you expect to happen?

I'd expect that opening an older revision of a Rust file via `M-x magit-find-file` will show me a buffer with the older contents, but won't actually send those older contents to the underlying LSP server.

(as far as I've checked, this works correctly for C++, so it might be specific either to rustic-mode or Doom's integration with rustic-mode.)

### What actually happened?

Opening an older revision of a Rust file sends that older revision to the underlying LSP server, which then thinks the file got replaced with new contents; this, in turn, messes up inlay hints etc. for the original buffer.

### Describe your attempts to resolve the issue

I've tried debugging the issue, but there seem to be too many moving parts for me to grasp it with my current elisp knowledge 😅 

I've also tried reporting the issue to lsp-mode - https://github.com/emacs-lsp/lsp-mode/issues/3233 - and it was eventually doomed that lsp-mode does the correct thing, leaving either rustic-mode or doom-emacs at fault.

### Steps to reproduce

1. Open a Rust code repository,
1. Open some Rust file,
1. Use `M-x magit-find-file` to preview the older revision of that file,
1. See how annotations (e.g. type hints) in the original buffer end up in wrong, seemingly random places.

### System Information

https://pastebin.com/MPTK5x72

PR #6309, opened 2022-04-17
fix(lsp): add a bypass for magit-find-file

Patryk27:
Fixes https://github.com/hlissner/doom-emacs/issues/6144.

# The Bug

`magit-find-file` is a handy function that creates a new buffer containing an older revision of the currently viewed file inside of it.

For that newly-opened buffer to have syntax coloring and stuff, Magit briefly [sets `buffer-file-name`](https://github.com/magit/magit/blob/3cfc8458e14c705afdfbeb7dcd3d3f43d7479344/lisp/magit-files.el#L181) and performs [`(normal-mode t)`](https://github.com/magit/magit/blob/3cfc8458e14c705afdfbeb7dcd3d3f43d7479344/lisp/magit-files.el#L185). This, in turn, triggers related major-mode hooks, such as those which activate lsp -- and you can probably guess where this is going.

Becuase Magit doesn't know about lsp (and vice versa), after the buffer gets opened, lsp - being unaware that the buffer is merely a "temporary preview" - thinks the entire file got suddenly replaced with some new content; among others, this misplaces inlay hints and can even make the original buffer unnavigable (since the original buffer's contents are not actually replaced, but lsp thinks otherwise).

# The Fix

When you use `magit-find-file`, that newly-opened buffer gets created a local variable called `magit-buffer-revision`, which contains a Git revision used to read that file; the fix creates an advice around lsp's entry point and exploits that fact, skipping lsp initialization when that variable is present.

# Drawbacks

Taking a brief look through Magit's code suggests that `magit-buffer-revision` _is_ the correct variable to lay on, but I ain't no Magit hacker - it might be the case that sometimes it's set for regular, file-backed buffers (a bit of testing shows otherwise for the time being, though).

# Alternatives

## `magit-preview-mode`

Magit could implement a minor mode called `magit-preview-mode` and activate it for all "preview-like" actions; the lsp advice could then exploit the existence of `magit-preview-mode` instead of hacking around `magit-buffer-revision`.

## `lsp-deferred`

This whole issue is caused by the fact that Magit has to set `buffer-file-name`, at least for a nanosecond for `(normal-mode 1)` to figure out which major mode to activate - in the next tick, `buffer-file-name` gets set to `nil` anyway.

So if we delayed initializing lsp by a single tick - e.g. by calling/hooking `lsp-deferred` instead of `lsp` - the issue would be solved, too (since by the time lsp is initialized then, `buffer-file-name` would be empty, and lsp itself already has a check for that).

Unfortunately, even a single-tick delay introduces nasty UI artifacts - e.g. it makes lsp-headerline appear after that next tick, and this Drives Me Mad ™️  (yes, I've checked it).

### DIFF
--- a/modules/tools/lsp/+lsp.el
+++ b/modules/tools/lsp/+lsp.el
@@ -67,6 +67,13 @@ Can be a list of backends; accepts any value `company-backends' accepts.")
     :implementations '(lsp-find-implementation :async t)
     :type-definition #'lsp-find-type-definition)
 
+  (defadvice! +lsp--skip-during-magit-preview (fn &rest args)
+    "Ensure that `lsp' is not initialized when user opens the buffer through
+`magit-find-file' and friends."
+    :around #'lsp
+    (unless (bound-and-true-p magit-buffer-revision)
+      (apply fn args)))
+
   (defadvice! +lsp--respect-user-defined-checkers-a (fn &rest args)
     "Ensure user-defined `flycheck-checker' isn't overwritten by `lsp'."
     :around #'lsp-diagnostics-flycheck-enable


elken:
This looks fine to me, might be better served in lsp-mode though

elken:
@Patryk27 if you unpin magit locally and `doom sync -u` (note this will update any other packages too) is this still an issue?

Patryk27:
Hi, I've just checked and unfortunately it remains an issue 😢

elken:
I'm struggling to replacate, are you sure you have the latest magit changes?

Patryk27:
Yeah, I've ensured by locating the changed Magit's function through `helpful-callable` and reading its body; which language you're checking it on - Rust?

Patryk27:
Sure - a pretty minimal reproduction is:

`config.el`:
(empty)

`init.el`:
```
(doom! :completion
       (company +childframe)
       vertico

       :ui
       doom
       doom-dashboard

       :editor
       (evil +everywhere)

       :emacs
       vc

       :tools
       lsp
       magit

       :lang
       (rust +lsp)

       :config
       (default +bindings +smartparens))
```

`packages.el`:
```
(package! magit
  :pin "125a8d5e417dda4438ce41d71a821d8a936fa5ea")
```

With those, the scenario goes:
- Open a Rust project that's a Git repository,
- Open some Rust file,
- Execute `magit-find-file` to preview an older revision of this file (from a different commit or different branch etc.),
- On that newly-opened buffer, execute `describe-mode` and see that  `Lsp` is mentioned in `Minor modes enabled in this buffer:`.

elken:
@Patryk27 can you try [this](https://github.com/elken/magit/tree/fix/find-file-buffer-hooks) branch? Have included a sample recipe below:

```lisp
(package! magit
 :recipe (:host github :repo "elken/magit" :branch "fix/find-file-buffer-hooks"))
```

Patryk27:
Yeah, looks like it works - thanks! 🙂

elken:
I will try, though I don't know how it'll be recieved :p

Patryk27:
Okie, I've managed to make it work by exploiting the fact that GitHub stores forks together with their parent repository, so pinning by commit works (even though the commit doesn't really belong to the original Magit's repo):
```
(package! magit
  :pin "2b9937c0a2e7f293ae44dc8af670b4a962c8653d")
```
... but opening a previous revision seems to fail (picrel)
image

elken:
That's not an error, it's intended. Might have to wrap it in `ignore-errors` to shut it up

hlissner (changes requested):
> Magit could implement a minor mode called magit-preview-mode and activate it for all "preview-like" actions; the lsp advice could then exploit the existence of magit-preview-mode instead of hacking around magit-buffer-revision.

Could we use `magit-setup-buffer-hook` instead? E.g.

```elisp
(add-hook! 'magit-setup-buffer-hook (setq-local doom-inhibit-local-var-hooks t))
```

Doom places all its expensive server init hooks on `MAJOR-MODE-local-vars-hook`, so this has the benefit of addressing more only the lsp-mode use-case.

elken:
Hm that's a decent idea, I was hoping for an upstreamable fix but that might work

elken:
Just been pinged by Tarsius; are we happy with this PR?

Patryk27:
Hi, I've just checked and this issue doesn't happen anymore on recent Doom Emacs & Magit, so it seems that fixes on Magit's side were sufficient 🙂

tarsius:
I intend to revert the "fix" for this: https://github.com/magit/magit/commit/f331092df4d4dfc0a2a7424d929a9c845088d57f. As reported in https://github.com/magit/magit/issues/4986 (but I have run into this myself too), that change prevents hook functions from being run, that we actually want to run.

> For that newly-opened buffer to have syntax coloring and stuff, Magit briefly [sets buffer-file-name](https://github.com/magit/magit/blob/3cfc8458e14c705afdfbeb7dcd3d3f43d7479344/lisp/magit-files.el#L181) and performs [(normal-mode t)](https://github.com/magit/magit/blob/3cfc8458e14c705afdfbeb7dcd3d3f43d7479344/lisp/magit-files.el#L185).

"And stuff" includes running arbitrary functions on major-mode hooks, and I want to bring that back.

If I understand the above correctly, the problem is that lsp runs some thing asynchronously and because of what magit does it runs it at the wrong time and/or twice. Could someone please check if https://github.com/doomemacs/doomemacs/pull/6309/commits/458758bd15276829e84d44d9d1a0079797e6ef00 (or similar) still works. If so, would someone volunteer to upstream that to lsp?

elken:
I can try to verify it & merge it at some soon yeah

PR #6310, opened 2022-04-18
fix(python): usage of obsolete alias letf

panchoh:
Doom doctor croaks:

  > Checking Doom Emacs...
      .config/emacs/modules/lang/python/config.el: Warning: ‘letf’ is an obsolete alias (as of 27.1); use ‘cl-letf’ instead.

### DIFF
--- a/modules/lang/python/config.el
+++ b/modules/lang/python/config.el
@@ -325,7 +325,7 @@
   (defadvice! +python--inhibit-pip-requirements-fetch-packages-a (fn &rest args)
     "No-op `pip-requirements-fetch-packages', which can be expensive."
     :around #'pip-requirements-mode
-    (letf ((#'pip-requirements-fetch-packages #'ignore))
+    (cl-letf ((#'pip-requirements-fetch-packages #'ignore))
       (apply fn args))))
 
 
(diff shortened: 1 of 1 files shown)


modules/lang/python/config.el
```diff
@@ -325,7 +325,7 @@
   (defadvice! +python--inhibit-pip-requirements-fetch-packages-a (fn &rest args)
     "No-op `pip-requirements-fetch-packages', which can be expensive."
     :around #'pip-requirements-mode
-    (letf ((#'pip-requirements-fetch-packages #'ignore))
+    (cl-letf ((#'pip-requirements-fetch-packages #'ignore))
```
hlissner, line 328: This should actually be Doom's `letf!` macro, as this won't work with `cl-letf`.

hlissner (changes requested):

modules/lang/python/config.el
panchoh: Ooops!  Fixing...

panchoh: (Feel free to squash before merging, ofc).

hlissner (approved):

hlissner:
LGTM. Thanks for your help!

PR #6312, opened 2022-04-18
bump: :editor format

dhruvinsh:
Update `format-all` module with latest changes. It has been quite a stable for me and also support new formatter as well for example lua with stylua.


hlissner:
This is in our do-not-PR list. https://discourse.doomemacs.org/t/do-not-pr/

PR #6315, opened 2022-04-18
bump: :lang org +roam2 

fosskers:
This PR bumps the `org-roam` pin to include a fix for the error spam when saving in a narrowed region.

See also https://github.com/org-roam/org-roam/pull/2159

### DIFF
--- a/modules/lang/org/packages.el
+++ b/modules/lang/org/packages.el
@@ -93,7 +93,7 @@
     ;; FIXME A :recipe isn't strictly necessary, but without it, our package
     ;;       bumper fails to distinguish between org-roam v1 and v2.
     :recipe (:host github :repo "org-roam/org-roam")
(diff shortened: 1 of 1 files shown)


iyefrat (changes requested):
Can you change the commit  to `bump: org-roam`? The commit format doesn't distinguish packages that are in feature flags. The linter just doesn't check the `bump` scope yet (see [here](https://github.com/doomemacs/doomemacs/blob/80cd7557e147ccd0abef05ff4eafb93fe9545ad4/core/cli/ci.el#L265-L267))

iyefrat (approved):

hlissner:
The `:lang org` module was overdue a larger bump, so I merged this into 3733861. Thanks for bringing it to my attention!

PR #6318, opened 2022-04-20
feat(macos): add open in new iTerm window

brorbw:
By temporarily enabling the `OpenFileInNewWindows` option for iTerm it allows us to open in a new window rather than a new tab. Personally I use windows more often than tabs and thought that others might feel the same way. This is however quite the hack, and might not fit the project so feel free to ignore and dismiss.

### DIFF
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -261,7 +261,8 @@
         :desc "Send project to Transmit"   "U" #'+macos/send-project-to-transmit
         :desc "Send to Launchbar"          "l" #'+macos/send-to-launchbar
         :desc "Send project to Launchbar"  "L" #'+macos/send-project-to-launchbar
-        :desc "Open in iTerm"              "i" #'+macos/open-in-iterm)
+        :desc "Open in iTerm"              "i" #'+macos/open-in-iterm
+        :desc "Open in new iTerm window"   "I" #'+macos/open-in-iterm-new-window)
        (:when (featurep! :tools docker)
         :desc "Docker" "D" #'docker)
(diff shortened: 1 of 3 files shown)


acristoffers:
I spotted one problem in this code: it does not save and restore whatever value the configuration may already have. What if the user has set it? You are simply overwriting it permanently, therefore possibly messing with whatever preference the user has set.

brorbw:
You are completely right. Adding a check and restoring the original value should be fairly trivial though. I'll take a look at it.

brorbw:
Something like this?

brorbw:
@acristoffers something like this?

acristoffers:
Close. `shell-command` returns the exit code of the application, not its output. And the return code of the command `defaults` is not linked in any way to the value of the property read.

It should be something like this:

```lisp
(let ((command "defaults read com.googlecode.iterm2 OpenFileInNewWindows"))
    (message "Running: %s" command)
    (let ((value (shell-command-to-string command)))
      (if (string= (string-trim value) "1") t nil)))
```

This captures `stdout` and `stderr` of the command into value, and then we simply compare strings to see what we got.

hlissner:
Sorry for the late response. I've thrown together a modified version of this PR in ec2062f, which should fulfill what it was trying to do. Thanks for bringing it to my attention!

acristoffers:
I believe there is an error there. I tried to use the function today and it did not work and looking at the definition, I believe that in line 39 it should read

`(if ,newwindow? (not (equal (read-newwindows) "1")))))`

instead of 

`(if newwindow? (not (equal (read-newwindows) "1")))))`,

because I'm getting `let: Symbol’s value as variable is void: newwindow\?` right now.

PR #6333, opened 2022-04-25
fix(fortran): account for spaces in filenames

fosskers:
This fixes a bug reported to me over Discord. Previously, files that were the direct target for compilation would fail the process if they had a space somewhere in their full filepath.


hlissner (changes requested):
Could you use `%S` instead of `\"%s\"`. It's a slightly more elegant alternative.

fosskers:
Done!

hlissner (approved):

hlissner:

PR #6340, opened 2022-04-27
tweak(vertico): use setq-default to set completion-in-region-function (for Corfu)

daanturo:
Corfu makes completion-in-region-function a local variable in buffers
where it is enabled, so when this form is evaluated in one of those said
buffers (such as opening a file with Emacs before accessing the
minibuffer), completion-in-region-function will just be set locally
there.

This is currently not a problem for `company` users, but ones who try `corfu` will notice that where `corfu` isn't used (such as `M-:`), `vertico` should step in but the `*Completions*` buffer is presented instead.

### DIFF
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -18,13 +18,13 @@ overrides `completion-styles' during company completion sessions.")
   :config
   (setq vertico-resize nil
         vertico-count 17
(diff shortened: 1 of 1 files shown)


iyefrat (approved):

hlissner (approved):

hlissner:

Issue #5715, opened 2021-11-01
Python environment set in envrc isn't picked up


beojan:
### What did you expect to happen?

I setup a Python environment using `.envrc` and the `direnv` module. I expected that environment to be picked up (correct version in the modeline and the installed packages recognized by mspyls).

### What actually happened?

The system Python environment is picked up instead.

### Describe your attempts to resolve the issue

I think this is the same as https://github.com/hlissner/doom-emacs/issues/4655, but that was closed by the reporter.

### Steps to reproduce

1. Enable the `direnv` module
2. Create and `direnv allow` a `.envrc` file setting up a Python environment
3. Open a Python buffer in that directory

### System Information

https://pastebin.com/XAEcVNpK

offbyone:
I can confirm this, or something like this; when I run mspyls with Trace debugging enabled, I see that it's picking up the global asdf Python, instead of the one that direnv installed:

```
Found the following clients for /Users/offby1/projects/voicemail/.direnv/python-3.9.6/lib/python3.9/site-packages/twilio/twiml/voice_response.py: (server-id mspyls, priority 1)
The following clients were selected based on priority: (server-id mspyls, priority 1)
Analysis cache path: /Users/offby1/Library/Caches/Microsoft/Python Language Server
Microsoft Python Language Server version 0.5.59.0
Workspace root: /Users/offby1/projects/voicemail
GetCurrentSearchPaths /Users/offby1/.asdf/installs/python/3.9.6/bin/python 
Interpreter search paths:
    /Users/offby1/.asdf/installs/python/3.9.6/lib/python3.9
    /Users/offby1/.asdf/installs/python/3.9.6/lib/python3.9/lib-dynload
    /Users/offby1/.asdf/installs/python/3.9.6/lib/python3.9/site-packages
User search paths:
    /Users/offby1/projects/voicemail/src
Initializing for /Users/offby1/.asdf/installs/python/3.9.6/bin/python
Analysis caching mode: None.
```

I have this value for `exec-path`:

```
Value in #<buffer voice_response.py>
("/Users/offby1/projects/voicemail/.direnv/python-3.9.6/bin/" "/Users/offby1/.asdf/installs/python/3.9.6/bin/" ... many other paths...)
```

bersace:
This is chicken and egg. Doom can't rely on `VIRTUAL_ENV` because activating a virtualenv sets this variable.

Good sources for selecting virtualenv are :

[...]

PR #6359, opened 2022-05-04
feat(python): Activate local .venv virtualenv

bersace:
Along pyenv, poetry and pipenv, a regular way of creating virtualenv is
to use plain `python -m venv .venv` or sometimes `python3.9 -m
.venv-py3.9`. This patch activate virtualenv managed this way.

.venv can be either a regular virtualenv directory, a symlink pointing
to a virtualenv, a file containing a path to a virtualenv. If no .venv
entry is found, Doom will globs for .venv-* directories.

If multiple .venv-* are found, the latest is preferred. This way
.venv-py3.9 is preferred over .venvpy2.7. Use .venv symlink to choose
another version.

Spacemacs has this behaviour implemented, but does not support `.venv-*` glob, only .venv.


gagbo:
I'm not sure I want this, because I'm already using direnv and `:tools envrc` to do my venv shenanigans in Python projects, and I don't imagine this hook and direnv working well together (especially if I want a different venv active). Opting out of this venv detection can only be done by manipulating hooks which, while perfectly doable, isn't really an easy to use API (especially if you want to deactivate it for some reason in a single project)

elken:
Personally I agree with gagbo, direnv handles this _much_ better, maybe gate this behind a flag and check for `(not (featurep! :tools envrc))` or something?

bersace:
> I'm not sure I want this, because I'm already using direnv and `:tools envrc` to do my venv shenanigans in Python projects, and I don't imagine this hook and direnv working well together (especially if I want a different venv active). Opting out of this venv detection can only be done by manipulating hooks which, while perfectly doable, isn't really an easy to use API (especially if you want to deactivate it for some reason in a single project)

Ok. I may add a feature flag for this or just let this in my config.

How do you bridge envrc and doom without hitting #5715 ?

My experience is that `pyvenv-activate` overrides `VIRTUAL_ENV`, even if already set by envrc. pyvenv can't determine whether `VIRTUAL_ENV` commes from previous `pyvenv-activate` or `direnv`. I need help to understand all the intrications of virtualenv in emacs.

gagbo:
I've never had issues with a bare `layout python3` in my `.envrc`, if I wanted to test different python versions I'd probably install other python versions from source and then set `layout python /usr/local/bin/python-X.Y` in the `.envrc`

Should I need differently named virtualenvs (not the [default one](https://direnv.net/man/direnv-stdlib.1.html#codelayout-python-ltpythonexegtcode)), I'd write my own wrapper in `.config/direnv/direnvrc`. Currently this file has a stolen version of `layout anaconda` that I don't remember the origin (I don't use conda anymore), but I suppose the same script can more or less be done with venvs:

```bash
# Usage: layout local_anaconda <path/to/env> <path/to/environment.yml>
#
# Loads a conda virtual environment under "<path/to/env>" (defaults to ".condaenv" in the current directory).
# If the environment doesn't exist,
# create it using the description in "<path/to/environment.yml>" (defaults to "environment.yml" in the directory of the conda virtual environment)
#
layout_local_anaconda() {
  local env_path=${PWD}/.condaenv
  if [ -n "$1" ]; then
    env_path=$1
  fi

  if [ -d ${env_path} ]; then
    source activate ${env_path}
  else
    local env_parent=$(dirname ${env_path})
    local env_yml=${env_parent}/environment.yml
    if [ -n "$2" ]; then
      env_yml=$2
    fi
    conda env create -p ${env_path} -f ${env_yml}
    source activate ${env_path}
  fi
}
```

bersace:
Thanks for the sharing. I didn't know `layout` and `use` feature of direnv. That's neat.

Testing this, I didn't get to reproduce #5715 reliabily :-( .

Still, how could this PR breaks envrc ?

bersace:
@gagbo @elken If we rely on direnv, loading of `.python-version` should be removed from doom too. The entire autoload/pyenv.el file could be removed.

hlissner:
Sorry for the late response: for the time being I'm going to shelve this PR, because I think we need either:

- A more systemic solution to handle conda, venv, poetry, pyenvs, pipenv, et co without (or with fewer) per-tool cludges. And to support the scenario where one project relies on multiple env managers at once (truly the stuff of nightmares).
- Or to go all-in on recommending direnv to users. 

Frankly, I believe the latter is vastly superior (and already works out of the box with the `:tools direnv` module), but I do want to eventually offer this option later, if I can find the time to untangle python's madness.

There are Emacs plugins to handle each case (conda.el, poetry.el, pyenv.el, etc), but they affect global state and often step on each other's toes (and direnv's on top of that), so they're not quite up to code. Still, users can safely install one or two them and use them independently, if they wish. 

So TLDR: this will be handled later, and I will refer back to this when I do. For the time being I will close this, as I am prepping to move modules out into their own repos for the 3.0 release. Thanks for bringing it to my attention, in any case!

PR #6364, opened 2022-05-05
feat(dart): improve keybindings for better Flutter control

victorblq:
Improving the keybinds for Flutter control on a dart file buffer.

------
- [x] I've searched the issue tracker to ensure this was not PRed before.
- [x] My PR is not on [the do-not-PR list](https://doomemacs.org/d/do-not-pr) for this project.

### DIFF
--- a/modules/lang/dart/config.el
+++ b/modules/lang/dart/config.el
@@ -32,16 +32,12 @@
   :init
   (map! :after dart-mode
         :map dart-mode-map
-        (:localleader
-         (:prefix ("f" . "Flutter")
-          :desc "Flutter run"
+        :localleader
+         (:prefix ("f" . "flutter")
           "f" #'flutter-run
-          :desc "Flutter quit"
           "q" #'flutter-quit
-          :desc "Flutter hot reload"
           "r" #'flutter-hot-reload
-          :desc "Flutter hot restart"
-          "R" #'flutter-hot-restart))))
+          "R" #'flutter-hot-restart)))
 
 
 (use-package! lsp-dart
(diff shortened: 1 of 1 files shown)


ericdallo (commented):
LGTM

modules/lang/dart/config.el
```diff
@@ -32,8 +32,16 @@
...
+        (:localleader
+         (:prefix ("f" . "Flutter")
+          :desc "Flutter run"
+          "f" #'flutter-run
+          :desc "Flutter quit"
+          "q" #'flutter-quit
+          :desc "Flutter hot reload"
+          "r" #'flutter-hot-reload
+          :desc "Flutter hot restart"
+          "R" #'flutter-hot-restart))))
```
hlissner, line 44: It's a mote busier than it needs to be, and is inconsistent with other uses of `map!` across Doom's codebase. Could you inline these and remove a level of parentheses?

I.e.
```lisp
  (map! :after dart-mode
        :map dart-mode-map
        :localleader
        (:prefix ("f" . "Flutter")
         :desc "Flutter run"         "f" #'flutter-run
         :desc "Flutter quit"        "q" #'flutter-quit
         :desc "Flutter hot reload"  "r" #'flutter-hot-reload
         :desc "Flutter hot restart" "R" #'flutter-hot-restart)))
```

(Side note: I'm not sure we _really_ need these descriptions, considering they (practically) match their command's names, but I'll revisit that when I get around to rethinking how Doom uses which-key)

hlissner (changes requested):

modules/lang/dart/config.el
ericdallo: Agreed, the command names are pretty clear and easier if user wants to debug a command

hlissner (approved):

hlissner:
Thanks for your help!

PR #6369, opened 2022-05-05
refactor(format): replace with apheleia

elken:
Replace format-all with apheleia to grealy simplify the formatting setup.

For the most part, this is a low degree of ready the only thing missing is documentation and more granular testing on each module.

I've also had a recent thought on how to handle LSP formatters, previously I ignored them in favour of apheleia but I might forward-port the same logic here for the new module.

### DIFF
--- a/modules/editor/format/autoload/settings.el
+++ b/modules/editor/format/autoload/settings.el
@@ -1,7 +1,7 @@
 ;;; editor/format/autoload/settings.el -*- lexical-binding: t; -*-
 
 ;;;###autodef
-(cl-defun set-formatter! (name args &key modes filter)
+(cl-defun set-formatter! (name args &key modes)
   "Define (or modify) a formatter named NAME.
 
 Supported keywords: :modes :filter
--- a/modules/editor/format/autoload/settings.el
+++ b/modules/editor/format/autoload/settings.el
@@ -71,7 +71,7 @@ Advanced examples:
 "
   (declare (indent defun))
   (cl-check-type name symbol)
-  (after! apheleia
+  (after! apheleia-core
     (if (null args)
         (progn
           (setq apheleia-formatters
--- a/modules/editor/format/config.el
+++ b/modules/editor/format/config.el
@@ -1,5 +1,17 @@
 ;;; editor/format/config.el -*- lexical-binding: t; -*-
 
+(defvar +format-on-save-disabled-modes
+  '(sql-mode           ; sqlformat is currently broken
+    tex-mode           ; latexindent is broken
+    latex-mode
+    org-msg-edit-mode) ; doesn't need a formatter
+  "A list of major modes in which to reformat the buffer upon saving.
+If this list begins with `not', then it negates the list.
+If it is `t', it is enabled in all modes.
+If nil, it is disabled in all modes, the same as if the +onsave flag wasn't
+  used at all.
+Irrelevant if you do not have the +onsave flag enabled for this module.")
+
 (defvar +format-preserve-indentation t
   "If non-nil, the leading indentation is preserved when formatting the whole
 buffer. This is particularly useful for partials.
@@ -22,6 +34,15 @@ select buffers.")
 (when (modulep! +onsave)
   (add-hook 'doom-first-file-hook #'apheleia-global-mode))
 
+(defun +format-enable-on-save-maybe-h ()
+  "Enable formatting on save in certain major modes.
+This is controlled by `+format-on-save-disabled-modes'."
+  (setq-local apheleia-inhibit (or (eq major-mode 'fundamental-mode)
+                                   (string-empty-p (string-trim (buffer-name)))
+                                   (not (null (memq major-mode +format-on-save-disabled-modes))))))
+
+(when (modulep! +onsave)
+  (add-hook 'after-change-major-mode-hook #'+format-enable-on-save-maybe-h))
 
 ;;
 ;;; Hacks
--- a/modules/editor/format/config.el
+++ b/modules/editor/format/config.el
@@ -41,8 +41,9 @@ This is controlled by `+format-on-save-disabled-modes'."
       (string-empty-p (string-trim (buffer-name)))
       (not (null (memq major-mode +format-on-save-disabled-modes)))))
 
-(when (modulep! +onsave)
-  (add-to-list 'apheleia-inhibit-functions #'+format-inhibit-maybe-h))
+(after! apheleia
+  (when (modulep! +onsave)
+    (add-to-list 'apheleia-inhibit-functions #'+format-inhibit-maybe-h)))
 
 ;;
 ;;; Hacks
--- a/modules/editor/format/config.el
+++ b/modules/editor/format/config.el
@@ -33,17 +33,16 @@ select buffers.")
     (funcall orig-fn)))
 
 (add-hook! 'apheleia-post-format-hook
-  ;; HACK `web-mode' doesn't update syntax highlighting after arbitrary buffer
-  ;;      modifications, so we must trigger refontification manually.
-  (defun +format--fix-web-mode-fontification-h ()
-    (when (eq major-mode 'web-mode)
-      (setq web-mode-fontification-off nil)
-      (when (and web-mode-scan-beg web-mode-scan-end global-font-lock-mode)
-        (save-excursion
-          (font-lock-fontify-region web-mode-scan-beg web-mode-scan-end)))))
-  (defun +format--refresh-git-gutter-h ()
-    (when (bound-and-true-p git-gutter-mode)
-      (git-gutter))))
+           ;; HACK `web-mode' doesn't update syntax highlighting after arbitrary buffer
+           ;;      modifications, so we must trigger refontification manually.
+           (defun +format--fix-web-mode-fontification-h ()
+             (when (eq major-mode 'web-mode)
+               (setq web-mode-fontification-off nil)
+               (when (and web-mode-scan-beg web-mode-scan-end global-font-lock-mode)
+                 (save-excursion
+                   (font-lock-fontify-region web-mode-scan-beg web-mode-scan-end)))))
+           (defun +format--refresh-git-gutter-h ()
+             (+vc-gutter-init-maybe-h)))
 
 
 ;;
--- a/modules/editor/format/config.el
+++ b/modules/editor/format/config.el
@@ -41,7 +41,7 @@ This is controlled by `+format-on-save-disabled-modes'."
       (string-empty-p (string-trim (buffer-name)))
       (not (null (memq major-mode +format-on-save-disabled-modes)))))
 
-(after! apheleia
+(after! apheleia-core
   (when (modulep! +onsave)
     (add-to-list 'apheleia-inhibit-functions #'+format-inhibit-maybe-h)))
 
@@ -70,7 +70,7 @@ This is controlled by `+format-on-save-disabled-modes'."
 ;;
 ;;; Additional formatters
 
-(after! apheleia
+(after! apheleia-core
   ;; TODO html-tidy
   (cl-defun apheleia--indent-lisp-buffer
       (&key buffer scratch callback &allow-other-keys)
--- a/modules/editor/format/config.el
+++ b/modules/editor/format/config.el
@@ -65,26 +65,3 @@ This is controlled by `+format-on-save-disabled-modes'."
                    (font-lock-fontify-region web-mode-scan-beg web-mode-scan-end)))))
            (defun +format--refresh-git-gutter-h ()
              (+vc-gutter-init-maybe-h)))
-
-
-;;
-;;; Additional formatters
-
-(after! apheleia-core
-  ;; TODO html-tidy
-  (cl-defun apheleia--indent-lisp-buffer
-      (&key buffer scratch callback &allow-other-keys)
-    "Format a Lisp BUFFER. Use SCRATCH as a temporary buffer and CALLBACK to
-apply the transformation. For more implementation detail, see
-`apheleia--run-formatter-function'."
-    (with-current-buffer scratch
-      (setq-local indent-line-function
-                  (buffer-local-value 'indent-line-function buffer))
-      (setq-local lisp-indent-function
-                  (buffer-local-value 'lisp-indent-function buffer))
-      (funcall (with-current-buffer buffer major-mode))
-      (goto-char (point-min))
-      (let ((inhibit-message t)
-            (message-log-max nil))
-        (indent-region (point-min) (point-max)))
-      (funcall callback))))
--- a/modules/editor/format/config.el
+++ b/modules/editor/format/config.el
@@ -34,15 +34,15 @@ select buffers.")
 (when (modulep! +onsave)
   (add-hook 'doom-first-file-hook #'apheleia-global-mode))
 
-(defun +format-enable-on-save-maybe-h ()
+(defun +format-inhibit-maybe-h ()
   "Enable formatting on save in certain major modes.
 This is controlled by `+format-on-save-disabled-modes'."
-  (setq-local apheleia-inhibit (or (eq major-mode 'fundamental-mode)
-                                   (string-empty-p (string-trim (buffer-name)))
-                                   (not (null (memq major-mode +format-on-save-disabled-modes))))))
+  (or (eq major-mode 'fundamental-mode)
+      (string-empty-p (string-trim (buffer-name)))
+      (not (null (memq major-mode +format-on-save-disabled-modes)))))
 
 (when (modulep! +onsave)
-  (add-hook 'after-change-major-mode-hook #'+format-enable-on-save-maybe-h))
+  (add-to-list 'apheleia-inhibit-functions #'+format-inhibit-maybe-h))
 
 ;;
 ;;; Hacks
--- a/modules/editor/format/config.el
+++ b/modules/editor/format/config.el
@@ -41,8 +41,9 @@ This is controlled by `+format-on-save-disabled-modes'."
       (string-empty-p (string-trim (buffer-name)))
       (not (null (memq major-mode +format-on-save-disabled-modes)))))
 
-(after! apheleia-core
-  (when (modulep! +onsave)
+
+(when (modulep! +onsave)
+  (after! apheleia-core
     (add-to-list 'apheleia-inhibit-functions #'+format-inhibit-maybe-h)))
 
 ;;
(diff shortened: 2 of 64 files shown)


iyefrat:
psst, your last commit is missing a scope

elken:
> psst, your last commit is missing a scope

shoo you

VitalyAnkh:
Could this module have an "only format changed region" function? Like the `clang-format+-context` in [clang-format-plust](https://github.com/SavchenkoValeriy/emacs-clang-format-plus) offers.

elken:
> Could this module have an "only format changed region" function? Like the `clang-format+-context` in [clang-format-plust](https://github.com/SavchenkoValeriy/emacs-clang-format-plus) offers.

The discourse post is the place for discussion

modules/lang/cc/config.el
```diff
@@ -48,6 +48,18 @@ This is ignored by ccls.")
   (set-docsets! 'c-mode "C")
   (set-docsets! 'c++-mode "C++" "Boost")
   (set-electric! '(c-mode c++-mode objc-mode java-mode) :chars '(?\n ?\} ?\{))
+  (when (executable-find "clang-format")
```
offbyone, line 51: Does this mean that if I install `clang-format` while Emacs is running, I'll need to re-start it in order to enable formatting?

offbyone (commented):
It looks like a whole bunch of these formatters depend on startup-time detection of formatting tools. At least for my workflow, I usually note the absence of the formatter, and install it, without restarting Emacs.

modules/lang/cc/config.el
elken: Maybe, but I'll just remove these now since apheleia just ignores if the exe is missing

modules/editor/format/config.el
```diff
@@ -21,69 +18,72 @@ buffer. This is particularly useful for partials.
...
-            ((and (booleanp +format-on-save-enabled-modes)
-                  (not +format-on-save-enabled-modes)))
-            ((and (listp +format-on-save-enabled-modes)
-                  (if (eq (car +format-on-save-enabled-modes) 'not)
-                      (memq major-mode (cdr +format-on-save-enabled-modes))
-                    (not (memq major-mode +format-on-save-enabled-modes)))))
-            ((not (require 'format-all nil t))))
-      (format-all-mode +1)))
+This is controlled by `+format-on-save-disabled-modes'."
+  (setq-local apheleia-inhibit (or (eq major-mode 'fundamental-mode)
```
daanturo, line 40: This may endup reverting `apheleia-inhibit` to nil if it was set to a non-nil value somewhere before.

We may do something like this instead?

```lisp

(defun +format-inhibit-maybe-h ()
  "Enable formatting on save in certain major modes.
This is controlled by `+format-on-save-disabled-modes'."
  (or (eq major-mode 'fundamental-mode)
      (string-empty-p (string-trim (buffer-name)))
      (not (null (memq major-mode +format-on-save-disabled-modes)))))

(when (modulep! +onsave)
  (add-hook 'apheleia-inhibit-functions #'+format-inhibit-maybe-h))

```

elken: Only if the user has specified to do so, so probably not.

modules/lang/fortran/config.el
```diff
@@ -33,6 +33,9 @@
         :desc "build" "b" #'+fortran/build
         :desc "run"   "r" #'+fortran/run)
 
+  (when (executable-find "fprettify")
+    (set-formatter! 'fprettify '("fprettify" "-") :modes '(f90-mode fortran-mode)))
```
fosskers: Since this module defines two `after!` blocks, I suspect this `set-formatter!` should be split into two as well.

elken: Okay, also reminded me to remove the now-redundant executable-find check

modules/editor/format/config.el
daanturo: > Only if the user has specified to do so, so probably not.

If anyone sets `apheleia-inhibit` to t in their `.dir-locals.el`
`.dir-locals-2.el` to disable auto-formatting for some reasons, then this will
surely revert it to nil when the conditions are met. I don't think we should
trigger surprising unwanted formats against the user's expected setting here.

As an alternative to using `apheleia-inhibit-functions`, we can check
`apheleia-inhibit` first

```emacs-lisp
(setq-local apheleia-inhibit (or  (bound-and-true-p apheleia-inhibit) ;; HERE, don't revert it from t to nil
                                  (eq major-mode 'fundamental-mode)
                                  (string-empty-p (string-trim (buffer-name)))
                                  (not (null (memq major-mode +format-on-save-disabled-modes)))))
```

But of course utilizing the above `-functions` hook just looks simpler, and even
when compared the original code.

elken: Still think this is quite an edge case but I've made the change anyway. Thanks :D

peterhoeg:
How far from completion is this? The existing format-all based module isn't that great and I would much prefer not having to work around it's warts locally since this will all be an non-issue in our new, glorious apheleia future.

modules/editor/format/autoload/settings.el
```diff
@@ -165,39 +64,25 @@ Advanced examples:
...
-                         ok-statuses
-                         error-regexp)))
-        (cl-check-type filter (or function null))
-        (+format--set name
-          :function
-          `(lambda (executable mode-result)
-             ,(if filter `(apply #',filter ,fn) fn))
-          :modes modes)
-        name))))
+  (after! apheleia
```
peterhoeg, line 74: I couldn't get this to trigger without changing `(after! apheleia` to `(after! apheleia-core`

peterhoeg:
Maybe I'm reading the documentation the wrong way but simply passing a string with the command like this 

```elisp
(set-formatter! 'nixpkgs-fmt "nixpkgs-fmt" :modes '(nix-mode))
```
fails with `apheleia--run-formatters: Formatter must be a shell command or a Lisp function: nixpkgs-fmt`.

Whereas passing a list works:

```elisp
(set-formatter! 'nixpkgs-fmt '("nixpkgs-fmt") :modes '(nix-mode))
```

modules/editor/format/autoload/settings.el
elken: Looks like this has recently been changed

modules/editor/format/config.el
```diff
@@ -21,69 +18,50 @@ buffer. This is particularly useful for partials.
...
+           ;; HACK `web-mode' doesn't update syntax highlighting after arbitrary buffer
+           ;;      modifications, so we must trigger refontification manually.
+           (defun +format--fix-web-mode-fontification-h ()
+             (when (eq major-mode 'web-mode)
+               (setq web-mode-fontification-off nil)
+               (when (and web-mode-scan-beg web-mode-scan-end global-font-lock-mode)
+                 (save-excursion
+                   (font-lock-fontify-region web-mode-scan-beg web-mode-scan-end)))))
+           (defun +format--refresh-git-gutter-h ()
+             (+vc-gutter-init-maybe-h)))
```
peterhoeg, line 67: This function is not defined if `(modulep! :ui vc-gutter +diff-hl)`

peterhoeg: My workaround:
```elisp
(after! apheleia
  (unless (fboundp '+vc-gutter-init-maybe-h)
    (defsubst +vc-gutter-init-maybe-h () "Dummy to avoid apheleia dumping errors" nil)))
```

modules/editor/format/packages.el
```diff
@@ -1,4 +1,5 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; editor/format/packages.el
 
-(package! format-all :pin "47d862d40a088ca089c92cd393c6dca4628f87d3")
+;; TODO Pin when this is close to finish
+(package! apheleia)
```
hlissner, line 5: Would you mind pinning this now?

hlissner (changes requested):
Is this ready to merge? There are a couple small issues, mainly to do with formatting and conventions, but rather than delay this any longer (this module has been the source of a lot of issues for quite some time now), I'd like to merge this now and polish it going forward.

elken:
> Is this ready to merge? 

Looks like there's some merge conflicts to resolve too but yeah it can be

> There are a couple small issues, mainly to do with formatting and conventions

Yeah, these should be standardised so that the elisp formatter can also accommodate them thus the project formats itself :stuck_out_tongue: 

> I'd like to merge this now and polish it going forward.

If you're sure :)

modules/editor/format/packages.el
elken: Yeah I'll add this in too

offbyone:
Where is this change sitting? It'd be great to get a path forward on formatting, as the existing format-all-the-code model is not serving my needs for Ruby development.

peterhoeg:
I've added this PR to my local fork and am very happily using it. Until there is official progress, that might be a workaround for you too.

modules/editor/format/config.el
hlissner, line 67: 1. Please fix the indentation (should be indented by 2 spaces).
2. Like @peterhoeg mentioned, this puts a hard dependency on git-gutter. Add a `(fboundp '+vc-gutter-init-maybe-h)` check to `+format--refresh-git-gutter-h`.

modules/editor/format/config.el
```diff
@@ -21,69 +18,50 @@ buffer. This is particularly useful for partials.
...
-                    (not (memq major-mode +format-on-save-enabled-modes)))))
-            ((not (require 'format-all nil t))))
-      (format-all-mode +1)))
-
-(when (modulep! +onsave)
-  (add-hook 'after-change-major-mode-hook #'+format-enable-on-save-maybe-h))
-
+(after! apheleia-core
+  (when (modulep! +onsave)
+    (add-to-list 'apheleia-inhibit-functions #'+format-inhibit-maybe-h)))
```
hlissner, line 46: Move the `modulep!` check outside the `after!` block.

modules/editor/format/autoload/settings.el
```diff
@@ -1,112 +1,28 @@
...
-          (puthash m (assq-delete-all name (gethash key format-all-mode-table))
-                   format-all-mode-table)
-        (format-all--pushhash
-         m (cons name (if probe `(lambda () ,probe)))
-         format-all--mode-table)))))
-
 ;;;###autodef
-(cl-defun set-formatter!
-    (name formatter &key modes filter ok-statuses error-regexp)
+(cl-defun set-formatter! (name args &key modes filter)
```
hlissner, line 4: Remove the `filter` argument, as it is unused.

hlissner (changes requested):

modules/ui/vc-gutter/config.el
```diff
@@ -116,6 +116,11 @@ is deferred until the file is saved. Respects `git-gutter:disabled-modes'."
   ;; UX: update git-gutter on focus (in case I was using git externally)
   (add-hook 'focus-in-hook #'git-gutter:update-all-windows)
 
+  ;; Stop git-gutter doing things when we don't want
+  (remove-hook 'post-command-hook #'git-gutter:post-command-hook)
+  (advice-remove #'quit-window #'git-gutter:quit-window)
+  (advice-remove #'switch-to-buffer #'git-gutter:switch-to-buffer)
```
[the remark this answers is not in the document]
elken: No there is a conflict between apheleia and git-gutter [here](https://github.com/radian-software/apheleia/issues/115) so I also copied that from radon's config

modules/ui/vc-gutter/config.el
```diff
@@ -130,6 +135,9 @@ is deferred until the file is saved. Respects `git-gutter:disabled-modes'."
   (advice-add #'magit-stage-file   :after #'+vc-gutter-update-h)
   (advice-add #'magit-unstage-file :after #'+vc-gutter-update-h)
 
+  ;; UX: update git-gutter after reverting a buffer
+  (add-hook 'after-revert-hook #'+vc-gutter-update-h)
```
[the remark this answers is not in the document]
elken: See above

modules/lang/common-lisp/config.el
```diff
@@ -26,10 +26,14 @@
   (after! lisp-mode
     (set-repl-handler! 'lisp-mode #'+lisp/open-repl)
     (set-eval-handler! 'lisp-mode #'sly-eval-region)
+    (set-formatter! 'lisp-indent #'apheleia--indent-lisp-buffer :modes '(lisp-mode))
     (set-lookup-handlers! 'lisp-mode
       :definition #'sly-edit-definition
       :documentation #'sly-describe-symbol))
 
+  (add-hook! 'lisp-mode
+    (after! sly (sly-lisp-indent-compatibility-mode)))
```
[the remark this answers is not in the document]
elken: I don't remember writing this, but yeah I agree it's not good

modules/lang/clojure/config.el
```diff
@@ -16,6 +16,9 @@
 (use-package! clojure-mode
   :hook (clojure-mode . rainbow-delimiters-mode)
   :config
+  (when (executable-find "cljfmt")
+    (set-formatter! 'cljfmt '("cljfmt" "fix" "-") :modes '(clojure-mode clojurec-mode clojurescript-mode)))
```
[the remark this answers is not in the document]
elken: I definitely already have, must have snuck back in during rebasing

modules/editor/format/autoload/settings.el
elken: I left it in in case you wanted me to refactor it to add it in, but happy to just ignore it then

modules/editor/format/config.el
elken: Sure

modules/editor/format/config.el
elken: Will do

modules/lang/zig/doctor.el
```diff
@@ -10,7 +10,10 @@
          "This module requires (:tools tree-sitter)")
 
 (unless (executable-find "zig")
-  (warn! "Couldn't find zig binary"))
+  (warn! "Couldn't find zig binary")
+
+  (when (modulep! :editor format)
+    (warn! "Formatting will be disabled")))
```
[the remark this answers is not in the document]
elken: I imagine I wanted `unless` here yeah

modules/lang/sh/config.el
```diff
@@ -87,8 +87,7 @@
 
 (use-package! fish-mode
   :when (modulep! +fish)
-  :defer t
-  :config (set-formatter! 'fish-mode #'fish_indent))
+  :defer t)
```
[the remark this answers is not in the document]
elken: Will it still be loaded correctly?

modules/lang/dhall/config.el
```diff
@@ -4,7 +4,9 @@
   :defer t
   :config
   (set-repl-handler! 'dhall-mode #'dhall-repl-show)
-  (setq dhall-format-at-save (modulep! :editor format +onsave))
+  (setq dhall-format-at-save (modulep! +dhall-format))
+  (unless (and (modulep! +dhall-format))
+    (set-formatter! 'dhall-format '("dhall" "lint" "-") :modes '(dhall-mode)))
```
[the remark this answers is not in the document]
elken: I wasn't sure how best to handle this, but okay I'll revert it to how it was for now

modules/lang/sh/config.el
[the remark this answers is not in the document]
hlissner: Yes. `fish-mode` [autoloads its entry points](https://github.com/wwwjfy/emacs-fish/blob/master/fish-mode.el#L703-L708).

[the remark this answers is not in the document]
elken: Cool, I'll rebase that commit now then

hlissner (dismissed):

hlissner (approved):

hlissner:
Sorry for taking so long to get to this; this has been overdue a merge for a *long* time. Thanks again for your hard work and help!

yufengwang:
I got this error when using format/region, `Symbol’s function definition is void: apheleia--get-formatter-command`

fosskers:
@yufengwang As mentioned in the new documentation for the `format` module, region formatting is not currently supported by Apheleia.

yufengwang:
okay, is there any workaround? because i highly depend on ` format/region` on a very old codebase

liaowang11:
> okay, is there any workaround? because i highly depend on ` format/region` on a very old codebase @fosskers

With this tweaks, it seems to be working for me.
```diff
@@ -8,7 +8,7 @@
 
 (defun +format-region (start end &optional callback)
   "Format from START to END with `apheleia'."
-  (when-let* ((command (apheleia--get-formatter-command
+  (when-let* ((command (apheleia--get-formatters
                         (if current-prefix-arg
                             'prompt
                           'interactive)))
@@ -23,7 +23,7 @@
       ;; Ensure this temp buffer seems as much like the origin buffer as
       ;; possible, in case the formatter is an elisp function, like `gofmt'.
       (cl-loop for (var . val)
-               in (cl-remove-if-not #'listp (buffer-local-variables origin-buffer))
+               in (cl-remove-if-not #'listp (buffer-local-variables cur-buffer))
                ;; Making enable-multibyte-characters buffer-local causes an
                ;; error.
                unless (eq var 'enable-multibyte-characters)
@@ -45,7 +45,9 @@
              ;; restore indentation without affecting new
              ;; indentation
              (indent-rigidly (point-min) (point-max)
-                             (max 0 (- indent (+format--current-indentation))))))
+                             (max 0 (- indent (+format--current-indentation)))))
+           (set-buffer-modified-p nil)
+           )
          (with-current-buffer cur-buffer
            (delete-region start end)
            (insert-buffer-substring-no-properties formatted-buffer)
```

elken:
In future, please use issues when reporting issues.

As for this change, I've tested it locally and it seems to "kind of" work but since as pointed out it's not supported upstream, if you're happy with this fix I'll PR it in.

offbyone:
@elken thank you for the work on this; it's working well for me, and I'm glad to have it in main.

yufengwang:
@elken  thanks for your hard work

johalun:
@yufengwang 

Hey! Please tell me how you got region formatting working. If I don't modify any settings `SPC c f` formats incorrectly (not using my .clang-format).

If I add `(setq +format-with-lsp nil)` I get the error `Symbol’s function definition is void: apheleia--get-formatter-command` for region formatting but full buffer formatting works.

I'm on latest master and the doom code does contain the changes above.

elken:
a) if you're having issues, report them through the proper channels
b) if you were using the latest master, you'd have #7427 which resolves that error so please update to the latest master

yufengwang:
@johalun I have submit a new issue, it still has bugs, https://github.com/doomemacs/doomemacs/issues/7438

modules/editor/format/autoload/format.el
```diff
@@ -1,276 +1,85 @@
...
+           (delete-region start end)
+           (insert-buffer-substring-no-properties formatted-buffer)
+           (when callback (funcall callback))
+           (kill-buffer formatted-buffer)))))))
 
 
 ;;
 ;;; Commands
 
-(defun +format--org-region (beg end)
```
lroolle, line 228: @elken You may have removed the `+format--org-region` for some reason, but this is quite handy and, as far as I am concerned, I use it every day. Can i add it back in this module?

Issue #6352, opened 2022-05-02
Doom's consult-assisting +vertico/crm-* functions reduce functionality 


t-e-r-m:
### What did you expect to happen?

When selecting an option, there should also be an option to edit the current selection, before pressing enter to confirm the selection(s).  In the default non-multiple completion, this occurs as follows:

- `[tab]` select entry
- move up and edit entry
- press `RET` to accept, and exit completion.

For `consult-completing-read-multiple`, the options are as follows (assume the command has just been called):

1. When cursor is on a pre-fetched candidate (aka not the first, editable line):
  1.1. `RET` to add to "selected candidates" list
  1.2. `[tab]` to add it to the editable line, and have the option to edit it.  Press `RET` to add it to the "selected candidates" list.

I expected the same, default behavior.  `[tab]` to add to the "selected candidates" list is much more intuitive, but restricts 1.2.

### What actually happened?

1. When cursor is on a pre-fetched candidate (aka not the first, editable line):
  1.1. `[tab]` to add to "selected candidates" list
  1.2. `RET` to add to "selected candidates" list and immediately exit completion.
  
`[backtab]`, as far as I can tell, does nothing useful.  One can enter something in the editable first line and then select a candidate with `[backtab]`, but then there's no way to see the actual list of candidates.

The functions that are bound below are responsible:

https://github.com/doomemacs/doomemacs/blob/3b789752a30419535b328c605957b6b43a9d47fa/modules/completion/vertico/config.el#L149-L152

### Describe your attempts to resolve the issue

Since the only entries were the ones Doom populated it with:

`(setq consult-crm-map nil)`

### Steps to reproduce

Eval the following:
```lisp
(progn
  (end-of-buffer)
  (let ((choices (completing-read-multiple "Enter a name: " '("Tom" "Jerry" "Spike" "Tyke"))))
    (while (> (length choices) 0 )
      (insert (concat (pop choices) " ")))))
```

Then eval it again after `(setq consult-crm-map nil)` to witness the difference.

### System Information

https://pastebin.com/7PdNdjV2

iyefrat:
Huh, good point. This never really came up in my usage of `crm`. Since the linked PR bumps the `vertico` module and remove Doom's and Consult's `crm` changes, this issue should get "solved" once that's merged.

PR #6379, opened 2022-05-06
`:completion vertico` updates

iyefrat:
bump: :completion vertico
(... -> ...)
- Remove everything related to `consult-completing-read-multiple` since
the function has been deprecated upstream due to implementation issues

Ref: minad/consult#567
Close: #6352
#
tweak(vertico): improve org consult source

- only add if :lang org is on
- have a version that works before org is loaded
#
feat(vertico): add workarounds for problematic commands
#
fix(vertico): consult tofu regex range

### DIFF
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -29,7 +29,14 @@ overrides `completion-styles' during company completion sessions.")
   ;; cleans ~/foo/bar/// to /, and ~/foo/bar/~/ to ~/.
   (add-hook 'rfn-eshadow-update-overlay-hook #'vertico-directory-tidy)
   (add-hook 'minibuffer-setup-hook #'vertico-repeat-save)
-  (map! :map vertico-map "DEL" #'vertico-directory-delete-char))
+  (map! :map vertico-map "DEL" #'vertico-directory-delete-char)
+
+  ;; These commands are problematic and automatically show the *Completions* buffer
+  (advice-add #'tmm-add-prompt :after #'minibuffer-hide-completions)
+  (defadvice! +vertico--suppress-completion-help-a (fn &rest args)
+    :around #'ffap-menu-ask
+    (letf! ((#'minibuffer-completion-help #'ignore))
+      (apply fn args))))
 
 
 (use-package! orderless
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -101,7 +101,6 @@ orderless."
     [remap switch-to-buffer-other-frame]  #'consult-buffer-other-frame
     [remap yank-pop]                      #'consult-yank-pop
     [remap persp-switch-to-buffer]        #'+vertico/switch-workspace-buffer)
-  (advice-add #'completing-read-multiple :override #'consult-completing-read-multiple)
   (advice-add #'multi-occur :override #'consult-multi-occur)
   :config
   (defadvice! +vertico--consult-recent-file-a (&rest _args)
@@ -145,11 +144,7 @@ orderless."
         :category buffer
         :state    ,#'consult--buffer-state
         :items    ,(lambda () (mapcar #'buffer-name (org-buffer-list)))))
-    (add-to-list 'consult-buffer-sources '+vertico--consult-org-source 'append))
-  (map! :map consult-crm-map
-        :desc "Select candidate" [tab] #'+vertico/crm-select
-        :desc "Select candidate and keep input" [backtab] #'+vertico/crm-select-keep-input
-        :desc "Enter candidates" "RET" #'+vertico/crm-exit))
+    (add-to-list 'consult-buffer-sources '+vertico--consult-org-source 'append)))
 
 
 (use-package! consult-dir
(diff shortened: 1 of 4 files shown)


modules/completion/vertico/config.el
```diff
@@ -44,26 +51,31 @@ orderless."
...
-     ((string-suffix-p "~" pattern) `(orderless-flex . ,(substring pattern 0 -1)))))
+    (let ((tofu-range-regexp (concat "["
+                                     (char-to-string consult--tofu-char)
+                                     "-"
+                                     (char-to-string (+ consult--tofu-char consult--tofu-range))
+                                     "]")))
+      (cond
+       ;; Ensure $ works with Consult commands, which add disambiguation suffixes
+       ((string-suffix-p "$" pattern)
+        `(orderless-regexp . ,(concat (substring pattern 0 -1) (concat tofu-range-regexp "*$"))))
```
iyefrat, line 62: @minad do you think that this should be added to the consult wiki orderless dispatch? I just noticed that the tofu character range changed, apparently this has been broken for a while. Perhaps it makes sense to add the regex range as a consult variable as well, I'm not really sure how/if it's possible to convert `consult--tofu-p` to a regex automatically.

minad: I updated the range in the wiki. It is really unlikely that this will change again, so I would say we keep the hardcoded numbers.

minad: I don't like to load consult for the orderless dispatcher. I always try to defer loading dependencies.

iyefrat: Sure thing, thanks!

iyefrat: > I don't like to load consult for the orderless dispatcher. I always try to defer loading dependencies.

good point, I hadn't thought of that.

minad:
Btw, @iyefrat did you consider switching the `completion-styles` default to `(orderless basic)`? This is the setting which I recommend now. See https://github.com/oantolin/orderless/issues/111 for details. Essentially the difference is that `(orderless basic)` fixes some dynamic completion tables. Example:

~~~ elisp
(completing-read "Dynamic:"
                 (completion-table-dynamic
                  (lambda (input)
                    (list (format "%s1" input)
                          (format "%s2" input)
                          (format "%s3" input)))))
~~~

You can try this with `(orderless)` only and also with `(orderless basic)`. But it is definitely an edge case.

iyefrat:
Yeah I saw that when looking at the wiki, I just wanted to look into it a bit more before adding it to Doom to understand what it fixes. I expect I'll add it after reading the orderless issue.

modules/completion/vertico/config.el
```diff
@@ -15,6 +15,15 @@ overrides `completion-styles' during company completion sessions.")
...
   :hook (doom-first-input . vertico-mode)
+  :init
+  (defun +vertico-crm-indicator (args)
+    (cons (format "[CRM%s] %s"
+                  (replace-regexp-in-string
+                   "\\`\\[.*?]\\*\\|\\[.*?]\\*\\'" ""
+                   crm-separator)
+                  (car args))
+          (cdr args)))
+  (advice-add #'completing-read-multiple :filter-args #'+vertico-crm-indicator)
```
hlissner, line 26: Please use defadvice! (instead of `defun`+`advice-add`) and ensure the function name follows our conventions:

```elisp
(defadvice! +vertico-crm-indicator-a (args)
  :filter-args #'completing-read-multiple
  (cons (format "[CRM%s] %s"
                (replace-regexp-in-string
                 "\\`\\[.*?]\\*\\|\\[.*?]\\*\\'" ""
                 crm-separator)
                (car args))
        (cdr args)))
```

modules/completion/vertico/config.el
```diff
@@ -29,7 +38,14 @@ overrides `completion-styles' during company completion sessions.")
...
   (add-hook 'minibuffer-setup-hook #'vertico-repeat-save)
-  (map! :map vertico-map "DEL" #'vertico-directory-delete-char))
+  (map! :map vertico-map "DEL" #'vertico-directory-delete-char)
+
+  ;; These commands are problematic and automatically show the *Completions* buffer
+  (advice-add #'tmm-add-prompt :after #'minibuffer-hide-completions)
+  (advice-add #'ffap-menu-ask :around (lambda (&rest args)
+                                        (cl-letf (((symbol-function #'minibuffer-completion-help)
+                                                   #'ignore))
+                                          (apply args)))))
```
hlissner, line 48: Use `defadvice!`, a named advice, and `letf!`:

```elisp
(defadvice! +vertico--suppress-completion-help-a (fn &rest args)
  :around #'ffap-menu-ask
  (letf! ((#'minibuffer-completion-help #'ignore))
    (apply fn args)))
```

modules/completion/vertico/config.el
iyefrat: Done

modules/completion/vertico/config.el
iyefrat: Done

hlissner (approved):

PR #6384, opened 2022-05-10
More convenient tangling for literate module

ngharrison:
This PR is meant to make literate configs more convenient:
- Tangling at a more relevant time -- when closing emacs rather than on every save
- Variable for more control over which files trigger re-tangling
- Functions that can be called interactively to re-tangle and reload a private config

The previous pieces are left intact in case the tangle-on-save behavior is still wanted. I'm not super experienced with the function naming conventions and module organization, so those could be things to look at. For now I added all changes into the `autoload.el` file to fit with the previous parts.

A couple final notes. I thought it would be good for the function to be hooked into `doom-init-modules-hook` instead of `kill-emacs-hook`, but I'm guessing org isn't loaded properly yet at that point. Also just found #6281. These could probably be used together.

-----

### DIFF
--- a/modules/config/literate/README.org
+++ b/modules/config/literate/README.org
@@ -12,10 +12,11 @@
 - [[#features][Features]]
 - [[#configuration][Configuration]]
   - [[#change-the-location-of-configorg][Change the location of config.org]]
+  - [[#change-which-files-trigger-tangling][Change which files trigger tangling]]
+  - [[#change-to-tangle-on-save][Change to tangle-on-save]]
   - [[#change-where-src-blocks-are-tangled-or-prevent-it-entirely][Change where src blocks are tangled or prevent it entirely]]
 - [[#troubleshooting][Troubleshooting]]
   - [[#how-to-tangle-to-doomdirinitel][How to tangle to =DOOMDIR/init.el=]]
-  - [[#how-to-disable-tangle-on-save][How to disable tangle-on-save]]
 
 * Description
 This module enables support for a literate config.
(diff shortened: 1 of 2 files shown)


ngharrison:
> Given #6281, I must say that I can' see the point in adding the "tangle on exit not save" functionality.

Sure, with your addition, the tangle-on-save isn't as much of a problem anymore. It's a super great addition. The remaining benefit of this PR is compute efficiency and giving more control to the user. I'll have to just speak from my own experience. I almost only ever need my config tangled when I'm restarting Emacs so all the times tangling before that are a waste. If I could get it hooked into startup instead of closing Emacs, I think that would be even better. For the times I want to reload right away without a full fresh start I can call the function manually.

> I don't see much of a need for a dedicated "tangle + reload" function instead of just waiting for the tangle, and then running `doom/reload`.

I often find I want to reload just my config.el file rather than a full `doom/reload`. Maybe there's already a better way to do this?

tecosaur:
I think the docs additions in this PR are fine, but with the `.el` file I don't see why anything more than a tweak to `+literate-recompile-maybe-h` is needed. For instance, `+literate-recompile-maybe` is nice and fancy, it just feels like overkill to me.

ngharrison:
What would you tweak in `+literate-recompile-maybe-h`? I had originally completely replaced that function with `+literate-recompile-maybe`, but restored it to keep the possibility of tangling on save for other people. The two functions are closely related. The main difference is that `+literate-recompile-maybe-h` already knows a file has been changed and needs to check if it is one of the watched files while `+literate-recompile-maybe` needs to check for both.

tecosaur:
I think `+literate-recompile-maybe-h` could just check the file restriction, and the rest should be left to people's configs.

hlissner:
Thanks for the PR, but I must turn it down (for now). I intend to move `:config literate` into Doom core for the 3.1 release (for first-class support for literate configs), then I'll work on making runtime tangling more configurable. Thanks for bringing it to my attention, in any case!

PR #6387, opened 2022-05-11
fix(magit): only revert if buffer file exists

elken:


modules/tools/magit/autoload.el
```diff
@@ -86,7 +86,8 @@ window that already exists in that direction. It will split otherwise."
 (defun +magit--revert-buffer (buffer)
   (with-current-buffer buffer
     (kill-local-variable '+magit--stale-p)
-    (when buffer-file-name
+    (when (and buffer-file-name
+           (file-exists-p buffer-file-name))
```
hlissner, line 90: Could you join these two lines? To fix the odd indentation and because it's not too long a line anyway.

hlissner (changes requested):

modules/tools/magit/autoload.el
elken: Oh yeah there is an extra space, can do I normally put multiple where clauses on different lines but can merge :D

hlissner (approved):

hlissner:
LGTM. Thanks again!

PR #6393, opened 2022-05-12
fix(go): exactly match test function name

Spadi0:
When specifying a specific test to run using `go test`, the function name is accepted as a regex, not an exact string match. Previously, the `+go/test-single` and `+go/bench-single` functions specified the function name as a raw strings; this meant that these functions could sometimes test/benchmark other functions that contained the name of the intended test.

-----


hlissner (approved):

hlissner:
LGTM. Thanks for your help!

PR #6400, opened 2022-05-14
tweak(default): remove M-{-,=,+} and C-- keybinds

LemonBreezes:
The keybindings `M--`, `M-=`, and `C--` override the built-in commands `negative-argument` and `count-words-region`. Moreover, the `C--` command, `er/contract-region`, is already bound transiently to `expand-region-contract-fast-key` while expanding regions.


hlissner (approved):

hlissner:
Makes sense to me. Thanks again for your help!

PR #6412, opened 2022-05-22
tweak(lua): make fennel's outline-regexp configurable

daanturo:
Like +emacs-lisp-outline-regexp

-----

### DIFF
--- a/modules/lang/lua/config.el
+++ b/modules/lang/lua/config.el
@@ -57,6 +57,9 @@ lua-language-server.")
     (require 'flycheck-moonscript nil t)))
 
 
+(defvar +lua-fennel-outline-regexp "[ \t]*;;;;* [^ \t\n]"
+  "Regexp to use for `outline-regexp' in `fennel-mode'.")
+
 (use-package! fennel-mode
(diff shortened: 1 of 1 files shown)


daanturo:
Alternatively, since the two regexps are the same, we may instead a shared
```lisp
```

hlissner:
Thanks for the PR! I don't mind this idea, but I don't think it's needed or should be addressed in Doom (perhaps a `fennel-outline-regexp` variable should be pushed upstream?). 

The new regexp isn't different enough from [fennel's default](https://git.sr.ht/~technomancy/fennel-mode/tree/main/fennel-mode.el#L323) to warrant the change, and fennel doesn't have autoload cookies or similar (as far as I know) for us to be conscious of (one of the primary reasons `+emacs-lisp-outline-regexp` exists).

Plus, users can already directly configure these variables with `setq-hook!`, like so:

```elisp
(setq-hook! 'fennel-mode-hook outline-regexp "...")
```

I may change my mind on this later, when I introduce shared configuration (between modules), but for now I'll leave this be. In any case, thanks for bringing it to my attention!

PR #6417, opened 2022-05-24
feat(fortran): name executable after the file name

fosskers:
Previously the output was always the default `a.out`. Note that this is only for the single-file non-fpm case, where the user just wants to compile and run a one-off Fortran program.


hlissner (approved):

hlissner:
LGTM. Thanks again!

PR #6444, opened 2022-06-08
fix(org): make last arg of '+org--follow-search-string-a' optional

NightMachinery:
Fixes a regression that disabled org ID links.

-----

### DIFF
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -552,7 +552,7 @@ relative to `org-directory', unless it is an absolute path."
   (+org-define-basic-link "doom-modules" 'doom-modules-dir)
 
   ;; TODO PR this upstream
-  (defadvice! +org--follow-search-string-a (fn link arg)
+  (defadvice! +org--follow-search-string-a (fn link &optional arg)
     "Support ::SEARCH syntax for id: links."
     :around #'org-id-open
     :around #'org-roam-id-open


jeetelongname:
Please add the org scope to the commit message in accordance [with the git conventions](https://discourse.doomemacs.org/t/git-conventions/2407)

jeetelongname:
```
fix(org): made the last arg of '+org--follow-search-string-a' optional
```

The point is not for git but for us, a clean git history is one that allows the project to work more efficently and using conventional commits allows us to auto generate change logs for modules allow people to see what is being touched (scope does not always correspond to module) and provide semantic meaning. it allows us to move fast orderly in a sense

hlissner:
> Fixes a regression that disabled org ID links.

Can you explain this? Neither `org-id-open` or `org-roam-id-open` have an optional second argument, so why would this change make any difference?

NightMachinery:
[**@hlissner**](https://github.com/hlissner) commented on [Jun 17, 2022, 9:59 PM GMT+4:30](https://github.com/doomemacs/doomemacs/pull/6444#issuecomment-1159094566 "2022-06-17T17:29:35Z - Replied by Github Reply Comments"):
> > Fixes a regression that disabled org ID links.
> 
> Can you explain this? Neither `org-id-open` or `org-roam-id-open` have an optional second argument, so why would this change make any difference?

They do in the latest versions:
```
org-id-open is a byte-compiled function defined in C source code.

Signature
(org-id-open ID _)
```

```
org-roam-id-open is a byte-compiled function defined in C source code.

Signature
(org-roam-id-open ID-OR-MARKER &optional STRICT)
```

NightMachinery:
@jeetelongname I amended the commit name.

hlissner:
> ```
> org-roam-id-open is a byte-compiled function defined in C source code.
> 
> Signature
> (org-roam-id-open ID-OR-MARKER &optional STRICT)
> ```

I'm confused. One, because it claims to be defined "in C source code" (which isn't true), and two, because I can't find this signature [in the org-roam repo](https://github.com/org-roam/org-roam/blob/main/org-roam-id.el#L66). Am I missing something?

> I don't understand what _ means in the signature of org-id-open.

It's a convention to indicate that an variable isn't used.

hlissner:
Oh! My bad. I found it [in v1](https://github.com/org-roam/org-roam-v1/blob/946a879a4a18756a0508afba1e0b0fe070c6a8b4/org-roam.el#L1140). Still odd about the C part, but I'm good to accept this in that case.

hlissner (approved):

aikrahguzar:
>  Still odd about the C part, but I'm good to accept this in that case.

Since, this has confused me in the past, this happens when the function is advised. I don't know why but in that case helpful will claim it is part of the C code. This is pretty annoying since it means that I can't find where a function is defined without removing advice or searching through all references which is slow.

Kind of related, if I eval a function definition using `g r a f` helpful will claim that it is a function without a source file while `eval-defun` handle it just fine and link to the correct source file.

hlissner:
Thanks for your help!

> Since, this has confused me in the past, this happens when the function is advised. I don't know why but in that case helpful will claim it is part of the C code.

@aikrahguzar Very strange. I cannot reproduce this. I'll look into this and see if it can be addressed (or at least reported).

> Kind of related, if I eval a function definition using g r a f helpful will claim that it is a function without a source file while eval-defun handle it just fine and link to the correct source file.

I just pushed [a fix for this](https://github.com/doomemacs/doomemacs/commit/6c5537b487b5854051c6a8e26eaf3ef01beca744).

aikrahguzar:
> @aikrahguzar Very strange. I cannot reproduce this. I'll look into this and see if it can be addressed (or at least reported).

I hadn't looked at the helpful repo before now but it seems like it is a known issue with advise to native compiled functions. See https://github.com/Wilfred/helpful/issues/277 and https://github.com/Wilfred/helpful/pull/295

> I just pushed [a fix for this](https://github.com/doomemacs/doomemacs/commit/6c5537b487b5854051c6a8e26eaf3ef01beca744).

Thanks a lot!

PR #6519, opened 2022-06-29
fix(macos): void-variable newwindow? error

acristoffers:
The symbol `newwindow?` in 39 should be a macro's symbol, not a function's one.
This fixes a bug introduced by #6318.


hlissner (approved):

hlissner:

PR #6535, opened 2022-07-05
fix: strip './' from projectile file list produced by fd

### LINKED ISSUE #1: Just wanted to say something.
Hi Henrik, 

Just only saying that I found your .emacs-configuration very inspirational, next to [this one](http://doc.rix.si/org/fsem.html).

However, I noticed you don't use the server-feature of Emacs. You should definitely give it a try (on Windows it works fine too), it's so convenient. And I've just started reading a few random sections of your config, there's already so many cool snippets I'm definitely studying/stealing.

Thanks for the inspiration, it's appreciated!

### LINKED ISSUE #6504: Projectile ignored directories does not work on Mac

This configuration:
```
(after! projectile
  (add-to-list 'projectile-globally-ignored-directories '"vendor"))
```
is used to globally ignore `vendor` folders in every projects.

### What actually happened?

This works as expected on my machine Linux, but not in Mac OS.

The `projectile-globally-ignored-directories` is correctly set with `"vendor"` appended at the beginning of the list on both OS.

1. Ran `projectile-invalidate-cache`
2. Reinstall `.emacs.d` to make sure there is no cache whatsoever
3. Made sure to have `rg` and ran `doom env` and `doom sync`

### Steps to reproduce

1. Add `(after! projectile
  (add-to-list 'projectile-globally-ignored-directories '"vendor"))` to your config

2. Create project directory, with a `vendor` directory and a `test` file inside it.
3. Start Emacs, go to the project root
4. `SPC SPC` shows the `vendor/test` file in it's search result 

### System Information

https://pastebin.com/UgWhgSfK

acristoffers:
The referenced commit incorporates https://github.com/bbatsov/projectile/pull/1784 to fix #6504.

-----

### DIFF
--- a/core/core-projects.el
+++ b/core/core-projects.el
@@ -174,7 +174,7 @@ And if it's a function, evaluate it."
                          (cl-find-if (doom-rpartial #'executable-find t)
                                      (list "fdfind" "fd"))
                        doom-projectile-fd-binary))
-              (concat (format "%s . -0 -H --color=never --type file --type symlink --follow --exclude .git"
+              (concat (format "%s . -0 -H --color=never --type file --type symlink --follow --exclude .git --strip-cwd-prefix"
                               bin)
                       (if IS-WINDOWS " --path-separator=/"))))
            ;; Otherwise, resort to ripgrep, which is also faster than find


hlissner (changes requested):
Bumping projectile won't actually fix the issue because Doom has [its own default for projectile-generic-command](https://github.com/doomemacs/doomemacs/blob/master/core/core-projects.el#L166-L184).

Would you mind rebasing this by replacing your bump commit with a bugfix that adds `--strip-cwd-prefix` to [this line?](https://github.com/doomemacs/doomemacs/blob/master/core/core-projects.el#L177)

Also, please change the commit message to the following (to conform to [our commit conventions](https://discourse.doomemacs.org/t/git-conventions/2407)):

```
fix: strip './' from projectile file list produced by fd

Fix: #6504
```

Side note: I'd rather it be a bugfix than a bump because I have a larger bump commit in the pipeline that would render this bump redundant.

acristoffers:
When squashing with GitHub Desktop App I have two fields for a commit message, so I split your message into two, hoping that GitHub actually merges them into one -m option. Is that correct or do I have to amend the commit message?

hlissner (approved):

hlissner:
> Is that correct or do I have to amend the commit message?

It seems to be correct. Thanks for your help!

PR #6543, opened 2022-07-07
feat(default): bind 'SPC b I' to open ibuffer for workspace

teeann:
As `SPC b i` is already bound to `ibuffer`, this new binding is really convenient and consistent with Doom's current default binding.

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -349,7 +349,8 @@
        :desc "Next buffer"                 "]"   #'next-buffer
        (:when (featurep! :ui workspaces)
         :desc "Switch workspace buffer" "b" #'persp-switch-to-buffer
-        :desc "Switch buffer"           "B" #'switch-to-buffer)
+        :desc "Switch buffer"           "B" #'switch-to-buffer
+        :desc "ibuffer workspace"       "i" #'+ibuffer/open-for-current-workspace)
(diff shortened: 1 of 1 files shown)


modules/config/default/+evil-bindings.el
```diff
@@ -349,7 +349,8 @@
        :desc "Next buffer"                 "]"   #'next-buffer
        (:when (featurep! :ui workspaces)
         :desc "Switch workspace buffer" "b" #'persp-switch-to-buffer
-        :desc "Switch buffer"           "B" #'switch-to-buffer)
+        :desc "Switch buffer"           "B" #'switch-to-buffer
+        :desc "ibuffer workspace"       "i" #'+ibuffer/open-for-current-workspace)
```
hlissner, line 353: This binds `+ibuffer/open-for-current-workspace` to `i` (lowercase). I think you meant to bind it to `I` (uppercase). If not, then the later `i` keybinding (to `ibuffer`) will always override this. If so, please change it.

hlissner (changes requested):

modules/config/default/+evil-bindings.el
teeann: Yes, I've fixed the typo binding as you mentioned

hlissner (approved):

hlissner:
Thanks for your help!

PR #6549, opened 2022-07-09
fix(mu4e): Update to mu 1.8

aikrahguzar:
Fixes #6511 
-----
- [x] I searched the issue tracker and this hasn't been PRed before.
- [x] My changes are not on [the do-not-PR list](https://doomemacs.org/d/do-not-pr) for this project.
- [x] My commits conform to [the git conventions](https://doomemacs.org/d/git-conventions).
- [ ] My changes are visual; I've included before and after screenshots.
- [ ] I am blindly checking these off.
- [x] Any relevant issues or PRs have been linked to.
- [ ] This a draft PR; I need more time to finish it.

Attempts to fix mu4e module so that it can work with mu 1.8. This
mostly involves using the replacements for functions that were
removed, i.e changing `mu4e~stop` to `mu4e--stop`. Backward
compatibility is preserved by defining these new functions as
aliases of old ones.

`=mu4e` is fixed by changing a call to `mu4e~start with` `mu4e`.
`mu4e--start` doesn't work due because it should be preceded by
a call to `mu4e--init-handlers` which is new in 1.8. At least at
first glance `mu4e` handles both situations.

The module is now working for me but there might be further problems
especially with `org-msg` integration and notifications since I don't use these.
And someone still on 1.6 should check that it works too.

PR #6553, opened 2022-07-10
fix(mu4e): Support mu 1.8

### LINKED ISSUE #6511: Doom not working with mu 1.8

Support for **mu version 1.8**

### What actually happened?

[**mu** version 1.8 has been released](https://github.com/djcb/mu/blob/release/1.8/NEWS.org). It seems that Doom do not like this version, **mu4e** cannot be launched after update. The 1.8 has some major changes ([see the release note](https://github.com/djcb/mu/blob/release/1.8/NEWS.org)).

I downgraded my **mu** installation to 1.6 until this gets fixed.

### Steps to reproduce

1. Update **mu** to version 1.8
2. Launch Doom Emacs
3. `SPC a m`

### System Information

https://pastebin.com/MrP8GSp4

tecosaur:
This is basically an iteration on #6549 that should improve compatibility (backwards and forwards).

I've tested this and it seems to work well on 1.8. Ideally, someone could test 1.6 and confirm that this doesn't break anything.

Fixes #6511
Replaces #6549

-----

### DIFF
--- a/modules/email/mu4e/config.el
+++ b/modules/email/mu4e/config.el
@@ -14,7 +14,8 @@
   :commands mu4e mu4e-compose-new
   :init
   (provide 'html2text) ; disable obsolete package
-  (when (or (not (require 'mu4e-meta nil t))
+  (when (or (not (or (require 'mu4e-config nil t)
+                     (require 'mu4e-meta nil t)))
             (version< mu4e-mu-version "1.4"))
     (setq mu4e-maildir "~/.mail"
           mu4e-user-mail-address-list nil))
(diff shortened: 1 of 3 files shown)


modules/email/mu4e/config.el
```diff
@@ -14,14 +14,58 @@
...
+      (defalias 'mu4e-search-rerun #'mu4e-headers-rerun-search
+        "Alias to provide the API of mu4e 1.8.")
+      (defun mu4e (&optional background)
+        "If mu4e is not running yet, start it.
+Then, show the main window, unless BACKGROUND (prefix-argument)
+is non-nil."
+        (interactive "P")
+        (mu4e--start (and (not background) #'mu4e--main-view))))
+    (setq mu4e-view-show-addresses t
+          mu4e-view-show-images t
```
aikrahguzar, line 65: What is the need to provide all these aliases? It should be enough to provide only those that are actually used in the module and that seems like a better option so as not to unnecessarily clutter the namespace. These are supposedly private functions though the weird convention of using tilde obscures that. But if people are using them in their private config I think it is better to correct it there.

And I don't know when `mu4e` was introduced but it is present in 1.6.0.

aikrahguzar:
There are also a couple of flags `list` and `personal` that don't have icons for them. I am not sure if they are new or not but they appeared in my headers view after the update. I think there should either be icons for them or they should be removed from `mu4e-headers-visible-flags` otherwise they look inconsistent and out of place.

tecosaur:
These may well be "private" functions, but unfortunately, a lot of them are also potentially useful. We can try to just use the minimal number based on our config, however by mapping the full lot the chance of accidentally introducing an incompatibility in future is reduced. I basically see this as erring on the side of caution.

tecosaur:
Oh, and I see the issue with flags, I just haven't had time for it yet. At least that's more of a visual niggle than broken functionality.

hlissner (approved):

dschrempf:
FYI, I can use `mu4e` perfectly fine with this PR, thank you! Some notes:

- Sometimes, I get (seemingly random?) the following warning:
  
  ```
  ‘mu4e-headers-search-bookmark’ is an obsolete command (as of 1.7.0); use ‘mu4e-search-bookmark’ instead.
  ```

  I couldn't find references to this function in the Doom configuration, and also not in my private configuration.

- I do have an item (P) for personal mails, but it is unnecessary and ugly. I think we could turn that off using `mu4e-headers-visible-flags`.

- The flag items are a bit too bold (at least there is too much space between them).

Thanks for brining `mu4e` back to Doom Emacs!

peterhoeg:
Apologies for hijacking this PR, but have you noticed extremely poor performance with mu4e compared to regular mu when using doom? I opened an issue with upstream (see https://github.com/djcb/mu/issues/2089#issuecomment-1182842545 for all the gory details), but it turns out that mu4e with `emacs -Q` absolutely flies. I'm still seeing the poor performance with plain doom.

peterhoeg:
I'm adding some additional info in case someone else sees this. The problem is the dashboard (or I guess mu4e-main-mode is the correct term).

When launching mu4e it overwhelms mu server with something I haven't figured out causing mu to max out one CPU. If I wait for it to settle down and then do a search, it's instantaneous. Hitting q to go back to the main screen now maxes out the CPU again so if I try to do another search, I need to wait for mu to die down and then it's executed (quickly).

peterhoeg:
I'm so sorry about the noise. Turns out it was the unread search for a search with lots of regexps that caused it. Adding `:hide-unread t` made everything snappy again.

apisandipas:
Sorry but Whats the status of this PR? mu4e been sorely missed as part of my workflow. I'm wondering if it makes sense to temporarily use tecosaur's forked branch or wait for a release.

seanfarley:
> Sorry but Whats the status of this PR? mu4e been sorely missed as part of my workflow. I'm wondering if it makes sense to temporarily use tecosaur's forked branch or wait for a release.

I think Henrik is currently deep into his sprint now. I believe the PR is ready but, for now, you'll just have to apply the patches yourself. Personally, I merged this PR locally but didn't make a commit.

modules/email/mu4e/config.el
```diff
@@ -14,14 +14,58 @@
...
+      (defalias (intern (concat "mu4e--" transferable-suffix))
+        (intern (concat "mu4e~" transferable-suffix))
+        "Alias to provide the API of mu4e 1.8.")
+      (dolist (transferable-proc-suffixes
+               '("add" "compose" "contacts" "eat-sexp-from-buf" "filter"
+                 "find" "index" "kill" "mkdir" "move" "ping" "remove"
+                 "sent" "sentinel" "start" "view"))
+        (defalias (intern (concat "mu4e--server-" transferable-suffix))
+          (intern (concat "mu4e~proc-" transferable-suffix))
+          "Alias to provide the API of mu4e 1.8."))
```
necaris, line 55: Found a small bug when testing this out on Mu 1.6 -- `transferable-suffix` is reused in the `defalias` here when it should be `transferable-proc-suffix`. I might also suggest adding this detail to the documentation string, as below:
```elisp
      (dolist (transferable-proc-suffix
               '("add" "compose" "contacts" "eat-sexp-from-buf" "filter"
                 "find" "index" "kill" "mkdir" "move" "ping" "remove"
                 "sent" "sentinel" "start" "view"))
        (defalias (intern (concat "mu4e--server-" transferable-proc-suffix))
          (intern (concat "mu4e~proc-" transferable-proc-suffix))
          "Alias to provide the API of mu4e 1.8 (`proc' -> `server')."))
```

hlissner (changes requested):
Looks good to me! Just one minor edit to the commit message's metadata, please:

```difff
-Close: https://github.com/doomemacs/doomemacs/issues/6511
+Fix: #6511
+Close: #6549
 Co-authored-by: Rahguzar <aikrahguzar@gmail.com>
```

tecosaur:
Done. FYI GitHub is doing the urlification of `#6511`.

hlissner (approved):

hlissner:
Great work, as always! Thanks again for your help!

JonathanReeve:
> * I do have an item (P) for personal mails, but it is unnecessary and ugly. I think we could turn that off using `mu4e-headers-visible-flags`.
> * The flag items are a bit too bold (or at least there is too little space between them).

@dschrempf and others: I'm trying to fix the flag icon issue in #6610. Let me know if that fixes it for you, too.

PR #6560, opened 2022-07-12
tweak(default): consult-flymake in a branch of +default-diagnostics

daanturo:
when available instead of plain flymake-show-diagnostics-buffer.

### DIFF
--- a/modules/config/default/autoload/default.el
+++ b/modules/config/default/autoload/default.el
@@ -62,7 +62,9 @@ current project. Otherwise list them for the current buffer"
               (bound-and-true-p flycheck-mode))
          (flycheck-list-errors))
         ((bound-and-true-p flymake-mode)
-         (flymake-show-diagnostics-buffer))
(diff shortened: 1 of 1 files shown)


modules/config/default/autoload/default.el
```diff
@@ -62,7 +62,10 @@ current project. Otherwise list them for the current buffer"
               (bound-and-true-p flycheck-mode))
          (flycheck-list-errors))
         ((bound-and-true-p flymake-mode)
-         (flymake-show-diagnostics-buffer))
+         (cond ((featurep! :completion vertico)
```
jeetelongname, line 65: Why not use `if` here? 
as there are only two cases it makes sense and would be more readable as well

jeetelongname, line 65: Why not use `if` here? as there are only 2 cases it is more idiomatic

hlissner (approved):

hlissner:
Thanks again for your help!

PR #6564, opened 2022-07-15
tweak(lookup): add new values to `+lookup-provider-url-alist`

smallwat3r:
This PR appends to `+lookup-provider-url-alist`, necessary values to search from:
1. [Sourcegraph](https://sourcegraph.com/search), which provides code search accross millions of open source projects.
1. the [Python documentation](https://docs.python.org/3/search.html) (if the `python` module is enabled).

---


hlissner:
Sorry for the tremendously late response: I've been on the fence about this for a while. I want to avoid having to maintain a list of language/toolchain documentation providers; this PR sets a precedent for including more. Additionally, it's a little redundant with the module's `+docsets` feature. I added a devdocs.io provider at some point as a middleground, but I understand if folks would rather not have the middleman.

Sourcegraph, on the other hand, is an excellent addition, so I pushed that in 0eff40c, but I'll have to turn down docs.python.org (and, eventually, doc.rust-lang.org will follow).

I may change my mind and revisit this later, but I'm prepping to move our modules into [a separate repo](https://github.com/doomemacs/modules), so I'll close this for now and shelve these decisions for later. Thanks for bringing it to my attention in any case!

PR #6566, opened 2022-07-17
fix(dired): fix +dired/quit-all for Dirvish.

LemonBreezes:
This PR makes `+dired/quit-all` also kill Dirvish buffers.

### DIFF
--- a/modules/emacs/dired/autoload.el
+++ b/modules/emacs/dired/autoload.el
@@ -4,7 +4,7 @@
 (defun +dired/quit-all ()
   "Kill all `dired-mode' buffers."
   (interactive)
-  (mapc #'kill-buffer (doom-buffers-in-mode 'dired-mode))
+  (mapc #'kill-buffer (doom-buffers-in-mode '(dired-mode dirvish-mode)))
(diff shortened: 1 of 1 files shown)


LemonBreezes:
I've closed this for now because I also want to handle the case that we use the split pane view for Dirvish. In that scenario, I want `q` to restore the previous window configuration.

LemonBreezes:
Currently I'm testing this
```
(defvar +dirvish-previous-window-configuration nil)

(advice-add #'dirvish :before
            (defun +dirvish-a (_)
              (setq +dirvish-previous-window-configuration
                    (plist-put +dirvish-previous-window-configuration
                               (+workspace-current-name)
                               (current-window-configuration)
                               #'string=))))

(advice-add #'+dired/quit-all :after
            (defun +dired/quit-all-a ()
              (set-window-configuration
               (plist-get +dirvish-previous-window-configuration
                          (+workspace-current-name)
                          #'string=))))
```

LemonBreezes:
I'm reopening this because I redid the solution I had previously.

PR #6567, opened 2022-07-17
fix(dired): rebind +dired/quit-all in Dirvish.

LemonBreezes:
Add a command for quitting the current Dirvish session.


LemonBreezes:
I think the Dirvish module needs a little bit of work with its keybindings and that should be an entire PR I think.

PR #6579, opened 2022-07-20
notmuch: improve update

agzam:
Currently, update buffer creates an additional window that stays on. I
think it's better to delete the window and bury the buffer. That way
user still can open the buffer later to check the update status.

### DIFF
--- a/modules/email/notmuch/autoload.el
+++ b/modules/email/notmuch/autoload.el
@@ -60,18 +60,20 @@
 (defun +notmuch/update ()
   "Sync notmuch emails with server."
   (interactive)
-  (with-current-buffer (compile (+notmuch-get-sync-command))
-    (add-hook
-     'compilation-finish-functions
-     (lambda (buf status)
-       (if (equal status "finished\n")
-           (progn
-             (kill-buffer buf)
-             (notmuch-refresh-all-buffers)
-             (message "Notmuch sync successful"))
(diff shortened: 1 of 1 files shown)


modules/email/notmuch/autoload.el
```diff
@@ -60,18 +60,20 @@
...
-     (lambda (buf status)
-       (if (equal status "finished\n")
-           (progn
-             (kill-buffer buf)
-             (notmuch-refresh-all-buffers)
-             (message "Notmuch sync successful"))
-         (user-error "Failed to sync notmuch data")))
-     nil
-     'local)))
+  (let ((compilation-buffer-name-function (lambda (_) (format "* notmuch update *"))))
```
hlissner, line 63: Name the buffer `*notmuch update*`, as the extra whitespace is unconventional.

hlissner (changes requested):
Please reword the commit message to conform to [our git conventions](https://discourse.doomemacs.org/t/git-conventions/2407), and make it more descriptive:

```
tweak(notmuch): bury update buffer instead of killing it

+notmuch/update would create an additional window that outlasts the
update process. Instead, delete the window and bury the buffer, so that
it may be read later.
```

hlissner (approved):

hlissner:
Thanks for your help!

Issue #6581, opened 2022-07-21
doomscript fails under Chemacs2


jeffkowalski:
### What did you expect to happen?

executing ```doomscript``` at the commandline, should invoke the named script

### What actually happened?

```
$ doomscript
```
(note that you don't even need a script argument to reproduce the error)
It yields the following failure:

```
Debugger entered--Lisp error: (file-missing "Cannot open load file" "No such file or directory" "core-lib")
  require(core-lib)
  eval-buffer(#<buffer  *load*-987115> nil "/home/jeff/.emacs.d-doom/core/core.el" nil t)  ; Reading at buffer position 10779
  load-with-code-conversion("/home/jeff/.emacs.d-doom/core/core.el" "/home/jeff/.emacs.d-doom/core/core.el" nil t)
  require(core "/home/jeff/.emacs.d-doom/core/core")
  eval-buffer(#<buffer  *load*> nil "/home/jeff/.emacs.d-doom/core/core-cli.el" nil t)  ; Reading at buffer position 2647
  load-with-code-conversion("/home/jeff/.emacs.d-doom/core/core-cli.el" "/home/jeff/.emacs.d-doom/core/core-cli.el" nil t)
  load("/home/jeff/.emacs.d-doom/core/core-cli" nil t)
  command-line-1(("--no-splash" "--load" "/home/jeff/.emacs.d-doom/core/core-cli" "--load" "" "--"))
  command-line()
  normal-top-level()
```

### Describe your attempts to resolve the issue

The root cause is that Chemacs2 needs its ```early-init.el``` executed, but ```doomscript```'s combo of emacs arguments ```-q --no-site-file --batch``` will clearly prevent normal user-land startup.  Without Chemacs2 getting a way to execute, it cannot set ```user-emacs-directory``` properly, and so our ```core-lib``` will never be found.  Hence the error.

The solution is to _also_ load the ```early-init.el``` from Chemacs2 if we're running in that environment.  

Pull request forthcoming.

### Steps to reproduce

```$ doomscript```

### System Information

https://pastebin.com/v7AMjTng

kishvanchee:
[...]

PR #6582, opened 2022-07-21
fix: doomscript fails under Chemacs2 #6581

jeffkowalski:
Detect if running under Chemacs2.
If so, additionally load its early-init.el in order to properly
set the user-emacs-directory to the location of doomemacs.

-----
- [X ] I searched the issue tracker and this hasn't been PRed before.
- [X ] My changes are not on [the do-not-PR list](https://doomemacs.org/d/do-not-pr) for this project.
- [X ] My commits conform to [the git conventions](https://doomemacs.org/d/git-conventions).

### DIFF
--- a/bin/doomscript
+++ b/bin/doomscript
@@ -30,6 +30,13 @@ if [ -z "$TMPDIR" ]; then
 fi
 
 export EMACSDIR="${EMACSDIR:-$(cd $(dirname "$BASH_SOURCE")/.. && pwd)}"
+# If running under Chemacs2, emacs-user-directory will be different than our desired path.
+# We then need to load chemacs manually to allow it to set the directory correctly.
+user_emacs_directory="$($emacs --batch --eval '(princ user-emacs-directory)' 2>/dev/null)"
+if [ "$EMACSDIR" != "$user_emacs_directory" ]; then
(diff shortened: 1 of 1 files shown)


hlissner:
Unfortunately, I must turn down this PR, because:

1. It is not desirable that `doomscript` ever load `early-init.el`, in any config, Chemacs or otherwise, as it loads *all* of your private config. It is up to the consumer of `doomscript` to load their config if it wants to, but it shouldn't be imposed on _all_ its consumers by default.
2. This assumes that the user must be using chemacs if `doomscript`'s parent directory doesn't live in `user-emacs-directory`. This harms `doomscript`'s portability, as the user may have Doom cloned to a non-standard location, and may have a non-chemacs config installed at `~/.config/emacs` or `~/.emacs.d`.
3. Once Doom v3.0 is released, its CLI will replace chemacs. Because `doomscript` is forced to operate on as few assumptions as possible (because polling Emacs/your initfiles is expensive and error prone), I'm not sure it's worth the trouble to support Chemacs. I'll revisit that possibility when I get there.

In any case, thanks for bringing this to my attention. I will follow up in #6581 once I have an acceptable solution to this problem.

PR #6568, opened 2022-07-18
fix(dirvish): include dired-x and newer defaults

Dadams2:
- Bumps the version to latest as doom is very far behind.
- fixes the void-variable `dired-omit-files` experienced by users who have unpinned dirvish 
- adds a sensible quit binding for fullscreen 
- adds a sensible `dirvish-mode-line-format` consistent with the package authors recommendations (see https://github.com/alexluigit/dirvish/blob/main/CUSTOMIZING.org)
- fixes popup window sizing issue experienced by users who have `+all` enabled for `popup` (see https://github.com/alexluigit/dirvish/issues/18)

This also implements the feedback of the package author in #6397 for the original pull request for this module that includes:
- replaces `fd-dired` with `dirvish-fd`
- removes the `+dired/quit-all` binding
- adds a binding for `dired-dispatch` similar to that of magit and treemacs
- replaces `diff-hl-dired-mode` with dirvish's inbuilt `vc-state`

the `+dirvish` flag is currently undocumented in the modules. But I am assuming this will need to be a separate PR.

Fixes #6562
References #6397
Replaces #6566 #6567

Before:
image

After:
image

-----
- [x ] I searched the issue tracker and this hasn't been PRed before.
- [ x] My changes are not on [the do-not-PR list](https://doomemacs.org/d/do-not-pr) for this project.
- [ x] My commits conform to [the git conventions](https://doomemacs.org/d/git-conventions).
- [ x] My changes are visual; I've included before and after screenshots.
- [x ] Any relevant issues or PRs have been linked to.

PR #6585, opened 2022-07-22
Dired fixes and improvements

LemonBreezes:
This PR adds a few small fixes and features to our Dired module:
1. Fixes `dired-do-compress` when done with non-directory files marked.
2. Enables `drag-and-drop` support for users of Emacs29.
3. Adds `dired-subtree` for togging Dired directories with `TAB` as if they were Org headings.
4. Rebinds `dirvish-dispatch` and `dirvish-quicksort` to Dirvish analogues.

### DIFF
--- a/core/core-editor.el
+++ b/core/core-editor.el
@@ -287,7 +287,14 @@ tell you about it. Very annoying. This prevents that."
     "Auto revert stale buffers in visible windows, if necessary."
     (dolist (buf (doom-visible-buffers))
       (with-current-buffer buf
-        (doom-auto-revert-buffer-h)))))
+        (doom-auto-revert-buffer-h))))
+
+  (defadvice! +dired--no-revert-while-compressing-a (oldfun &rest args)
+    "Don't revert while executing `dired-do-compress' on marked lines."
+    :around #'dired-do-compress
+    (advice-add #'doom-auto-revert-buffer-h :override #'ignore)
+    (unwind-protect (apply oldfun args)
+      (advice-remove #'doom-auto-revert-buffer-h #'ignore))))
 
 
 (use-package! recentf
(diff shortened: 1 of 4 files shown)


hlissner:
Does this conflict at all with #6568?

LemonBreezes:
> Does this conflict at all with #6568?

Not anymore. I have just reverted the key binding changes.

alexluigit:
Please guard `dired-subtree` with `(unless (featurep! dirvish))` since dirvish has a native subtree extension and it should not use together with `dired-subtree`.

I hope you don't mind if I am direct here, actually I don't think introducing `dired-subtree` into dired module is a good idea because this package has conflict with other packages within this module

- `diff-hl-dired-mode` is not able to recognize subtrees inserted by `dired-subtree`
- the dired module also has `dired-git-info`, which prevents `dired-subtree` to insert a subtree. See: https://github.com/clemera/dired-git-info/issues/9
... there are probably other conflicts

These issue doesn't exists with `dirvish-subtree` + `vc-state` + `git-msg`.

Plus, dirvish-subtree already provided `dirvish-subtree-toggle/clear/up`, so you don't have to define that `+dired-subtree-remove-all`.

LemonBreezes:
> Please guard `dired-subtree` with `(unless (featurep! dirvish))` since dirvish has a native subtree extension and it should not use together with `dired-subtree`.
> 
> I hope you don't mind if I am direct here, actually I don't think introducing `dired-subtree` into dired module is a good idea because this package has conflict with other packages within this module
> 
> * `diff-hl-dired-mode` is not able to recognize subtrees inserted by `dired-subtree`
> * the dired module also has `dired-git-info`, which prevents `dired-subtree` to insert a subtree. See: [Clash with dired-subtree clemera/dired-git-info#9](https://github.com/clemera/dired-git-info/issues/9)
>   ... there are probably other conflicts
> 
> These issues doesn't exists with `dirvish-subtree` + `vc-state` + `git-msg`.
> 
> Plus, dirvish-subtree already provided `dirvish-subtree-toggle/clear/up`, so you don't have to define that `+dired-subtree-remove-all`.
> 
> See also: https://github.com/alexluigit/dirvish/blob/main/FAQ.org#dired-subtree

Ah, I had no idea of these conflicts. Thanks for bringing that up. I have removed `dired-subtree` from the PR.

PR #6586, opened 2022-07-22
fix(unicode): fix long startup times on Emacs29

LemonBreezes:
This patch fixes Pcache for users of Emacs29 so that users of Unicode
Fonts on Emacs29 do not have to have 30 seconds plus initialization
times on every startup. 

I am currently trying to get this patch into Pcache but it looks like the package is abandoned. I have PM'd Jonas (Tarsius) on Reddit about this and am waiting for him to reply.

### DIFF
--- a/modules/ui/unicode/packages.el
+++ b/modules/ui/unicode/packages.el
@@ -2,3 +2,8 @@
 ;;; ui/unicode/packages.el
 
 (package! unicode-fonts :pin "47f2397ade28eba621baa865fd69e4efb71407a5")
+(package! pcache
+  ;; Use a fork of `pcache', which has been patched to support Emacs 29 (the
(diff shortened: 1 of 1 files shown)


t-e-r-m:
Why was this closed?  I have the same issue.

LemonBreezes:
> Why was this closed?
> I have the same issue BTW.

This patch to Pcache was upstreamed so try `doom upgrade` to see if that fixes your issue. If not, I will figure out why the code is not being pulled in by Doom Emacs.

t-e-r-m:
That fixed it, thank you!

https://github.com/sigma/pcache/commit/cae29ddbc3d12fac18ab5cfc26fa3ef13eb97dad

PR #6597, opened 2022-07-26
Add override to reflect breaking `fd` change.

dominicm00:
`v8.3.0` of `fd` introduced a breaking change to how paths are displayed; see https://github.com/sharkdp/fd/issues/760 and https://github.com/sharkdp/fd/pull/861. #6535 introduced a new `fd` flag that reverts the new behavior, but it is only compatible with `>=v8.3.0`. 

This pull request adds an optional override to use a binary other than `fd` for users with older versions, and reflects the new minimum version in the README.

References #6535

-----

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ Check out [the FAQ][FAQ] for answers to common questions about the project.
 + Emacs 27.1+ (*28.1 is recommended*, or [native-comp](https://www.emacswiki.org/emacs/GccEmacs). **29+ is not supported**).
 + [ripgrep] 11.0+
 + GNU `find`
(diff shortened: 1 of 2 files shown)


dominicm00:
I'm not sure how the git conventions work in this case, since #6535 was accidentally a breaking change but not marked as such. Should one of these commits be marked as breaking with an `Amend` reference?

hlissner:
Thanks for the PR! I've changed it to a draft because I'm not comfortable with depending on fd 8.3.0 due of its [low availability](https://repology.org/project/fd-find/versions) at this time. I'll explore alternatives over the weekend before I decide what to do here.

PR #6629, opened 2022-08-03
`:tools tree-sitter` syntax highlighting 

jeetelongname:
as well as documentation updates to reflect this update
this is still a work in progress as there is also a lazy load bug in this change as well as update for the latest docs

-----
- [X] This a draft PR; I need more time to finish it.

### DIFF
--- a/modules/tools/tree-sitter/autoload.el
+++ b/modules/tools/tree-sitter/autoload.el
@@ -2,8 +2,21 @@
 
 ;;;###autodef (fset 'tree-sitter! #'ignore)
 (defun tree-sitter! ()
-  (interactive)
-  (turn-on-tree-sitter-mode))
+  "Dispatch to turn on tree sitter.
+
+Used as a hook function which turns on `tree-sitter-mode'
+and selectively turn on `tree-sitter-hl-mode'.
+according to `+tree-sitter-hl-enabled-modes'"
+  (turn-on-tree-sitter-mode)
+  ;; conditionally enable `tree-sitter-hl-mode'
+  (let ((mode (bound-and-true-p tree-sitter-hl-mode)))
+    (when-let (mode (if (pcase +tree-sitter-hl-enabled-modes
+                          (`(not . ,modes) (not (memq major-mode modes)))
+                          ((and `(,_ . ,_) modes) (memq major-mode modes))
+                          (bool bool))
+                        (unless mode +1)
+                      (if mode -1)))
+      (tree-sitter-hl-mode mode))))
 
 ;; HACK: Remove and refactor when `use-package' eager macro expansion is solved or `use-package!' is removed
 ;;;###autoload
--- a/modules/tools/tree-sitter/config.el
+++ b/modules/tools/tree-sitter/config.el
@@ -1,7 +1,17 @@
 ;;; tools/tree-sitter/config.el -*- lexical-binding: t; -*-
 
+(defvar +tree-sitter-hl-enabled-modes '(not web-mode typescript-tsx-mode)
+  "A list of major modes which should be highlighted by tree-sitter.
+
+If this list begins with `not', then it negates the list.
+If it is t, it is enabled in all modes.
+If nil, it is disabled in all modes")
+
+;;
+;;; Packages
+
 (use-package! tree-sitter
-  :hook (tree-sitter-after-on . tree-sitter-hl-mode)
+  :defer t
   :config
   (require 'tree-sitter-langs)
   ;; This makes every node a link to a section of code
--- a/modules/tools/tree-sitter/README.org
+++ b/modules/tools/tree-sitter/README.org
@@ -124,16 +124,28 @@ outer}-text-objects-map~:
               (rust-mode . [(use_declaration) @import]))))
 #+end_src
 
-** Disable syntax highlighting for certain modes
-To disable syntax highlighting across the board you can do:
+** Configuring Tree sitter highlighting
+Highlighting is controlled by the variable =+tree-sitter-hl-enabled-modes=
+This list allows you to whitelist, blacklist, fully enable or fully disable
+tree-sitter highlighting for all major modes
+
+To use highlighting in select modes add major-modes to
+~+tree-sitter-hl-enabled-modes~. This disables highlighting in all other modes.
+The symbol that should be used is the major mode symbol, *not the package
+symbol*.
 #+begin_src emacs-lisp
-(remove-hook 'tree-sitter-after-on-hook #'tree-sitter-hl-mode)
+(setq +tree-sitter-hl-enabled-modes '(python-mode go-mode))
 #+end_src
 
-Then it can be enabled for specific major modes:
+If you want to disallow highlighting in certain modes then the car of
+~+tree-sitter-hl-enabled-modes~ should be =not=. This enables highlighting in
+all modes except the ones disallowed.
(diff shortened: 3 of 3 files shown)


jeetelongname:
Ok found the lazy loading bug, forgot that `:hook` added an implict `:defer` which I needed to add back

modules/tools/tree-sitter/config.el
```diff
@@ -1,15 +1,21 @@
...
-  :hook (tree-sitter-after-on . tree-sitter-hl-mode)
+  :defer t
   :config
   (require 'tree-sitter-langs)
   ;; This makes every node a link to a section of code
   (setq tree-sitter-debug-jump-buttons t
         ;; and this highlights the entire sub tree in your code
         tree-sitter-debug-highlight-jump-region t))
 
-
```
hlissner, line 12: Could you restore this blank line? It's a Doom convention that top-level `use-package!` blocks should be separated by two blank lines.

modules/tools/tree-sitter/config.el
```diff
@@ -1,15 +1,21 @@
 ;;; tools/tree-sitter/config.el -*- lexical-binding: t; -*-
 
+(defvar +tree-sitter-hl-enabled-modes '(not web-mode typescript-tsx-mode)
+  "A list of major modes which should be highlighted by tree-sitter.
+
+If this list begins with `not', then it negates the list.
+If it is t, it is enabled in all modes.
+If nil, it is disabled in all modes")
+
```
hlissner, line 9: For consistency with other modules, please insert:

```diff
 If nil, it is disabled in all modes")

+
+;;
+;;; Packages
+
 (use-package! tree-sitter
```

modules/tools/tree-sitter/autoload.el
```diff
@@ -2,8 +2,24 @@
...
+
+;;;###autoload
+(defun +tree-sitter--hl-enable-maybe ()
+  "Enable `tree-sitter-hl-mode' according to `+tree-sitter-hl-enabled-modes'."
+  (tree-sitter-hl-mode (if (pcase +tree-sitter-hl-enabled-modes
+                             (`(not . ,modes) (not (memq major-mode modes)))
+                             ((and `(,_ . ,_) modes) (memq major-mode modes))
+                             (bool bool))
+                           +1
+                         -1)))
```
hlissner, line 22: Is there any reason we can't merge `+tree-sitter--hl-enable-maybe` and `tree-sitter!`? I'd like to avoid autoloading private/internal functions, and it doesn't seem to be used anywhere else.

hlissner: Also, this `tree-sitter-hl-mode` call should be guarded. Even if called only to disable it, we'd be autoloading it and performing some shut-down logic (which will likely be inexpensive, but it's a potential point of failure).

Something like:

```lisp
(when-let* ((mode (bound-and-true-p tree-sitter-hl-mode))
            (mode (if (pcase +tree-sitter-hl-enabled-modes
                        (`(not . ,modes) (not (memq major-mode modes)))
                        ((and `(,_ . ,_) modes) (memq major-mode modes))
                        (bool bool))
                      (unless mode +1)
                    (if mode -1))))
  (tree-sitter-hl-mode mode))
```

hlissner (changes requested):

modules/tools/tree-sitter/autoload.el
jeetelongname: If this is meant to be a toggle then we also need to change the line `turn-on-tree-sitter-mode` to a toggle

jeetelongname:
```lisp
(if (bound-and-true-p tree-sitter-mode)
    (ignore-errors
      (tree-sitter-mode +1))
  (tree-sitter-mode -1))
```

something like this

hlissner: Ah, `tree-sitter!` is not supposed to be a toggle (or an interactive command, for that matter). It's only intended to be a hook. Can you remove the `(interactive)` in it?

jeetelongname: I think i was mirroring `lsp!` in so far as it is also an "interactive" hook function,
In any case that makes sense, Ill remove it, I am guessing we also will not need the toggle logic you have come up with

hlissner: > I am guessing we also will not need the toggle logic you have come up with

We still do, actually. The problem is that `tree-sitter-hl.el` is lazy loaded. By calling `tree-sitter-hl-mode` (even to turn it off when it isn't enabled in the first place), we're loading it, so I'd like it guarded.

jeetelongname: Ah I understand now, makes sense, in any case that has been pushed, thanks for the help!

modules/tools/tree-sitter/autoload.el
```diff
@@ -2,8 +2,21 @@
...
+  (turn-on-tree-sitter-mode)
+  ;; conditionally enable `tree-sitter-hl-mode'
+  (when-let* ((mode (bound-and-true-p tree-sitter-hl-mode))
+              (mode (if (pcase +tree-sitter-hl-enabled-modes
+                          (`(not . ,modes) (not (memq major-mode modes)))
+                          ((and `(,_ . ,_) modes) (memq major-mode modes))
+                          (bool bool))
+                        (unless mode +1)
+                      (if mode -1))))
+    (tree-sitter-hl-mode mode)))
```
hlissner, line 19: Ack, sorry to keep doing this to you, but I've noticed a bug in this guard (if `tree-sitter-hl-mode` isn't enabled already, it won't be enabled). Can you replace it with:

```diff
(defun tree-sitter! ()
  (interactive)
  (turn-on-tree-sitter-mode)
  ;; conditionally enable `tree-sitter-hl-mode'
- (when-let* ((mode (bound-and-true-p tree-sitter-hl-mode))
-             (mode (if (pcase +tree-sitter-hl-enabled-modes
-                         (`(not . ,modes) (not (memq major-mode modes)))
-                         ((and `(,_ . ,_) modes) (memq major-mode modes))
-                         (bool bool))
-                       (unless mode +1)
-                     (if mode -1))))
-   (tree-sitter-hl-mode mode)))
+(let ((mode (bound-and-true-p tree-sitter-hl-mode)))
+  (when-let (mode (if (pcase +tree-sitter-hl-enabled-modes
+                        (`(not . ,modes) (not (memq major-mode modes)))
+                        ((and `(,_ . ,_) modes) (memq major-mode modes))
+                        (bool bool))
+                      (unless mode +1)
+                    (if mode -1)))
+    (tree-sitter-hl-mode mode))))
```

hlissner (changes requested):

modules/tools/tree-sitter/autoload.el
jeetelongname: No worries, I should be checking these things more thoroughly myself as well. Ill give this a test and after that we should be good to go

jeetelongname: With a little testing all seems to be good now, seems to work (when entering a non supported mode nothing is loaded, but it does enable for modes that are supported) 

thanks for workshopping this (and bareing with me)

hlissner (approved):

hlissner:
Great work (as usual)! Thanks again!

PR #6632, opened 2022-08-04
fix(mu4e): force the account stripe face to be non-italic

seanfarley:
With mu4e came a new face for the header view. While this is nice for
the text, it had the consequence of making the account stripe (the
vertical bar) a bit jarring. This patch fixes that behavior to make the
stripe non-italic by append `'default` to the face.

### DIFF
--- a/modules/email/mu4e/autoload/email.el
+++ b/modules/email/mu4e/autoload/email.el
@@ -134,7 +134,8 @@ a quoted symbol for a alist of current strings and faces provided."
   (put-text-property
    0 (length str)
    'face
-   (if (not unique)
+   (list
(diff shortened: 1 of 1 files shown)


seanfarley:
Here are some screenshots of before and after:

Screen Shot 2022-08-04 at 1 48 36 PM

Screen Shot 2022-08-04 at 4 07 26 PM

seanfarley:
Also, @tecosaur should probably take a look since I have no idea what I'm doing with lisp.

hlissner (approved):

hlissner:
LGTM. Thanks for your help!

PR #6641, opened 2022-08-06
feat(eshell): looking up documenation and history search

aikrahguzar:

 Some (hopefully) improvements to eshell module,
1) Add a lookup handler documentation. The documentation can be a buffer with tldr page, a man page or the helpful for elisp. This adds [tldr](https://github.com/kuanyui/tldr.el) to the packages needed for the module.
2) For shell-command tldr pages are preferred but if both a man page and tldr page exist, it is possible to switch between the two by pressing `x` is the buffer with the page.
3) Adds a `eshell/help` command which is an alternative to pressing `K`.
4) Changes the history navigation to use the `completion-styles`. This means for example that `orderless` can be used to navigate history using `<up>` and `<down>`. The current match offered is highlighted to show the matching substring but highlighting is removed before sending the input. Kind of inspired by the history navigation in fish shell but more powerful. This has required probably too much change to play well with syntax-highlighting but seems to be working well in my limited testing.
5) Fixes a problem with `+eshell-search-history` where it would fail if the point was at the end of buffer.
6) Set various `ansi-color-*` faces to the colors from the current doom-theme using `custom-set-faces!`. I am not sure what should be done with `ansi-color-bright-*` so I just have gone for colors defined in the doom themes. Someone with more aesthetic sense should take a look if possible.
7) ~~Add advice wrapping syntax highlighting function with `with-silent-modifications`. This ensures that syntax highlighting doesn't pollute undo data.~~ Merged upstream https://github.com/akreisher/eshell-syntax-highlighting/pull/10
8) Add annotations to fish-completions candidates when using vertico. This done by defining a capf building on the code in https://github.com/minad/marginalia/issues/87.  Since this is a capf the same annotations are also present in company but I have gated this under the vertico. Maybe that shouldn't be the case but it seems like helm already has the annotations. This capf also takes care of some of the problems with `fish-completion` like subcommand completion. The fix itself can be done in fish-completion but it is unmaintained https://gitlab.com/ambrevar/emacs-fish-completion. Right now the only code fomr the package that the capf uses is essentially a call to `fish -c complete -C command` so that dependency can be removed relatively easily.

### DIFF
--- a/lisp/cli/doctor.el
+++ b/lisp/cli/doctor.el
@@ -141,7 +141,8 @@ in."
 
   (print! (start "Checking for stale elc files..."))
   (elc-check-dir doom-core-dir)
-  (elc-check-dir doom-local-dir)
+  (elc-check-dir doom-modules-dir)
+  (elc-check-dir (doom-path doom-local-dir "straight" straight-build-dir))
 
   (print! (start "Checking for problematic git global settings..."))
   (if (executable-find "git")


aikrahguzar:
@hlissner, I had some clumsiness taking stuff out of my personal config that I thought could be generally useful and into the eshell module. That led to a few commits and extending them after opened  this pr. Is that the reason you converted this into a draft? I think I am done making changes for now on my end. Of course, I am quite happy to make changes in response to feedback, as I am not sure how much of this carries its weight outside my config and not quite sure if I have put the stuff where it should go. I have some time to tinker right now so I am asking but no rush.

hlissner:
Sorry for the late response! I converted this to a draft because I approve of the functionality, but there are a couple issues I'd like to iron out before merging it. Trouble is, I've been too swamped to do a proper review. It shouldn't be much more than a week's wait, but I'll get it as soon as I can.

aikrahguzar:
> I converted this to a draft because I approve of the functionality, but there are a couple issues I'd like to iron out before merging it. Trouble is, I've been too swamped to do a proper review. It shouldn't be much more than a week's wait, but I'll get it as soon as I can.

Please take your time. I just wanted to now if there was something to be done on my end.

aikrahguzar:
One thing to look at: `ds` doesn't work in eshell. This is due to `evil-collection-eshell-evil-delete` being bound to `d` but `M-x evil-delete RET s` works. What needs to be changed in `evil-collection-eshell-evil-delete` to make `ds` work?

PR #6659, opened 2022-08-10
fix(org): respect org-table-automatic-realign

efim:
Fixes unsubmitted bug with org tables autoaligning:

In function for automatically align table on exit of insert state
added check for standard variable ~org-table-automatic-realign~

Documentation
Non-nil means automatically re-align table when pressing TAB or RETURN.

When nil, aligning is only done with M-x org-table-align, or after column
removal/insertion.

Troubleshooted that when attempted to figure out how to stop table that realigned for 11 seconds after exit of insert mode.
Found that there's a variable for that and that doom custom function doesn't respect it

-----


hlissner (approved):

hlissner:

PR #6676, opened 2022-08-14
module: add :completion new-helm(helm): bug fixes, more flags, less opiniated

johanwiden:
Bug fixes:
- Now uses emacs completion. This also enables fuzzy completion.
- Use helm-posframe instead of directly posframe for +childframe.
  The old implementation no longer works.
- By default use helm for find-file-at-point. Turning this off seems to
  have been for a problem that no longer exists.
- New key binding C-c L in helm-multi-files to toggle helm source
  locate. The previous key binding C-c p is and was used by
  helm-projectile.
- Increased helm-candidate-number-limit to make it less likely to miss
  interesting completion candidates.
- Set helm-mode explicitly, based on module flag, instead of via a hook.
  This is more reliable.
- helm-descbinds is now enabled explicitly, instead of via a hook.
  This is more reliable.
- Using all-the-icons had stopped working. Now fixed.

Module flags:
- By introducing more module flags, one can choose how
  much of helm to enable, and how helm should behave.

Added features:
- Remap ibuffer-find-file to helm-find-files.
- Disabling helm-ff-lynx-style-map is now a module flag.
- Hiding helm session window mode line is now a module flag.
- Documented that helm session header line may be good for helm novices.
- helm-display-buffer-default-height is now set only if popup window is
  used. Use of popup window is a module flag.
- Added a module flag for: Open helm session window in current window,
  and autoresize the helm session window.
- A module flag to set helm-grep-default-command to ack, instead of grep.
- Remapping of swiper to swiper-helm is now controlled by a module flag.
- Choice between icon set treemacs or all-the-icons is now by a module
  flag.
- packages.el now documents version date for each package.
- Uses helm version 3.8.7 2022-08-03
- Uses helm-projectile version 2022-08-07
- README.org:
  - Added some links to external documentation.
  - Documented module flags.
  - Documented implementation change for +childframe.
  - Documented prerequisites.
  - For key bindings documented to what helm function they are bound.
    This is especially important if one does not use evil.
  - Documented autoloaded functions.
  - Added Configuration documentation for helm-mode, helm,
    helm-posframe, helm-projectile.

Why new module name:
- Have both old and new module available, until there is confidence that
  users can switch to new module without problems.
  Then deprecate old helm module.

Fixes #0000
References #0000
Replaces #0000

-----
- [v] I searched the issue tracker and this hasn't been PRed before.
- [v] My changes are not on [the do-not-PR list](https://doomemacs.org/d/do-not-pr) for this project.
- [v] My commits conform to [the git conventions](https://doomemacs.org/d/git-conventions).


hlissner:
Thanks for the PR! I've converted this to a draft for now because it needs a lot of work to conform to our conventions and to merge this into the existing helm module. Unfortunately, I'm swamped at the moment, so I am unable to do this now or write a formal review of this yet. I will revisit this as soon as I am able.

hlissner:
Sorry for the late response. I'm preparing to move our modules into a separate repo, so I am closing/merging module PRs. For the time being I've incorporated some of the changes suggested here into `master`, but I will revisit this again after the v3 release and new documentation are up. In any case, thanks for bringing these improvements to my attention!

PR #6683, opened 2022-08-16
Feat/add ltex grammar

abougouffa:
Enable fast grammar checking using `lsp-ltex`, which provides integration with the LTeX LS, this last uses LanguageTool to correct grammar. This is way faster and responsive than the currently used `langtool` package.

The configuration can be used enabling `+lsp` for the `grammar` module.

### Example Org document after enabling `:checkers (grammar +lsp)`
Screenshot_20220816_153750

-----

### DIFF
--- a/modules/checkers/grammar/config.el
+++ b/modules/checkers/grammar/config.el
@@ -1,6 +1,7 @@
 ;;; checkers/grammar/config.el -*- lexical-binding: t; -*-
 
 (use-package! langtool
+  :unless (modulep! +lsp)
   :commands (langtool-check
              langtool-check-done
              langtool-show-message-at-point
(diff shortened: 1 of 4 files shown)


numkem:
Is there a way to make this work when composing an email with mu4e?

abougouffa:
> Is there a way to make this work when composing an email with mu4e?

@numkem In general, LSP servers try to load an entire workspace, which doesn't works with mails. I didn't dig deeper to figure out a workaround.

You can use [`flycheck-languagetool`](https://github.com/emacs-languagetool/flycheck-languagetool) as an alternative, you can take a look at my config, [you will find a configuration example there](https://github.com/abougouffa/dotfiles#flycheck).

modules/checkers/grammar/config.el
```diff
@@ -26,6 +27,68 @@
...
+
+  (defun +lsp-ltex-toggle ()
+    "Toggle LTeX LSP for the current buffer."
+    (interactive)
+    (if (+lsp-ltex--enabled-p)
+        (+lsp-ltex-disable)
+      (+lsp-ltex-enable)))
+
+  (map! :localleader
+        :map (text-mode-map latex-mode-map org-mode-map markdown-mode-map)
```
regadas, line 88: Hi @abougouffa great stuff! I just tried this, and I believe we could also add `LaTeX-mode-map` here. wdyt?

abougouffa: Thanks for the feedback @regadas 

Yep, it totally makes sens, I thought `LaTeX-mode` is an alias to `latex-mode`!

I will add it to the map list.

maikol-solis:
Hi!

I just discovered the eglot counterpart for `lsp-ltex` called `eglot-lsp` (https://github.com/emacs-languagetool/eglot-ltex). 

What do you think? 

Is it worth include it in this PR or it could be in a future one?

Thanks.

peterhoeg:
When using ltex-ls with eglot, you don't actually need the mentioned 3rd party package. You can just call:

```elisp
(set-eglot-client! 'text-mode '("ltex-ls"))
```

npajkovsky:
Anybody know how to make it work in programming mode?

peterhoeg:
Disclaimer: I may be sharing outdated information here!



While there is nothing wrong with using it in prog-mode, I don’t think neither lsp-mode nor eglot support using multiple servers per file. And you would most likely want a server for the language you are editing rather than one for just the text bits.

PR #6684, opened 2022-08-16
Update download command in getting_started.org

4goodapp:
Added --depth 1 in the command to download git.

### DIFF
--- a/docs/getting_started.org
+++ b/docs/getting_started.org
@@ -574,7 +574,7 @@ the default config:
    #+BEGIN_SRC bash :eval no
    [ -f ~/.emacs ] && mv ~/.emacs ~/.emacs.bak
    [ -d ~/.emacs.d ] && mv ~/.emacs.d ~/.emacs.legacy
-   git clone https://github.com/plexus/chemacs2.git ~/.emacs.d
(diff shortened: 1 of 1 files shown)


iyefrat:
Hey, Doom currently isn't accepting changes to docs files under `docs`, see the [do-not-pr list](https://discourse.doomemacs.org/t/do-not-pr/2406). Also, it has switched to `master` as the main branch, in case you aren't aware, so pull requests should go to there.

4goodapp:
> Hey, Doom currently isn't accepting changes to docs files under `docs`, see the [do-not-pr list](https://discourse.doomemacs.org/t/do-not-pr/2406). Also, it has switched to `master` as the main branch, in case you aren't aware, so pull requests should go to there.

it's a small change but an important one I think, so I think it's worth adding it cause without the `--depth 1` option, cloning the repose download 116136 files - 40MB+ on your system versus 1148 files - 7MB with that option as of this writing.

hlissner:
Everything in docs/*.org will be rewritten. This includes the chemacs instructions. Not to mention, Doom can replace Chemacs once v3 is out, so this change would not survive long enough to warrant the addition. Thanks for bringing it to our attention, in any case.

4goodapp:
> Not to mention, Doom can replace Chemacs once v3 is out

That's a good news.

PR #6685, opened 2022-08-16
Update README.org

4goodapp:


brongulus:
Didn't get the memo that hacktoberfest was starting in august this time

iyefrat:
This PR constitutes a small wording change, and per our [do-not-pr list](https://discourse.doomemacs.org/t/do-not-pr/2406), we don't accept those to reduce maintainer burden. Module toggling will also have more robust documentation across the board in the near future, once the new docs are fully out.

PR #6686, opened 2022-08-16
Added optional steps to completely remove evil key bindings.

4goodapp:
Added more information / option to completely remove evil key bindings.
see https://github.com/doomemacs/doomemacs/issues/2240#issuecomment-568812647


hlissner:
Thanks for the PR, but I must turn it down, because:

1. The added information is incorrect. `:config (default +bindings)` does not [bind evil keybinds unconditionally](https://github.com/doomemacs/doomemacs/blob/master/modules/config/default/config.el#L495-L497). When evil is disabled, it will bind vanilla-friendly keybinds instead.
2. This was PRed to the wrong branch.
3. The git commit message doesn't conform to our [git conventions](https://discourse.doomemacs.org/t/git-conventions/2407).

PR #6688, opened 2022-08-17
fix: defer-feature! fallback FNS

ivanbrennan:
Fix the behavior of `defer-feature!` when called with one argument.

If `defer-feature!` was called with one argument (as is the case in the `:lang common-lisp` module), `FNS` defaulted to an empty list. As a result, `FEATURE` would be deferred but never added back to the features list, and `after!` blocks would never trigger for `FEATURE`.

Instead of defaulting to an empty list, fallback to a singleton list `(FEATURE)`. This aligns with the behavior the macro had prior to 5b8b04f0c8b6458a3ac4b41417c69413a3bae92e, which generalized `FNS` to support a list of functions rather than just one.

-----

### DIFF
--- a/lisp/doom-lib.el
+++ b/lisp/doom-lib.el
@@ -595,14 +595,15 @@ serve as a predicated alternative to `after!'."
            (add-hook 'after-load-functions #',fn)))))
 
 (defmacro defer-feature! (feature &rest fns)
-  "Pretend FEATURE hasn't been loaded yet, until FEATURE-hook or FN runs.
+  "Pretend FEATURE hasn't been loaded yet, until FEATURE-hook or FNS run.
 
 Some packages (like `elisp-mode' and `lisp-mode') are loaded immediately at
 startup, which will prematurely trigger `after!' (and `with-eval-after-load')
 blocks. To get around this we make Emacs believe FEATURE hasn't been loaded yet,
-then wait until FEATURE-hook (or MODE-hook, if FN is provided) is triggered to
-reverse this and trigger `after!' blocks at a more reasonable time."
-  (let ((advice-fn (intern (format "doom--defer-feature-%s-a" feature))))
+then wait until FEATURE-hook (or any of FNS, if FNS are provided) is triggered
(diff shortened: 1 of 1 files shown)


lisp/doom-lib.el
```diff
@@ -595,14 +595,15 @@ serve as a predicated alternative to `after!'."
...
 (defmacro defer-feature! (feature &rest fns)
-  "Pretend FEATURE hasn't been loaded yet, until FEATURE-hook or FN runs.
+  "Pretend FEATURE hasn't been loaded yet, until FEATURE-hook or FNS run.
 
 Some packages (like `elisp-mode' and `lisp-mode') are loaded immediately at
 startup, which will prematurely trigger `after!' (and `with-eval-after-load')
 blocks. To get around this we make Emacs believe FEATURE hasn't been loaded yet,
-then wait until FEATURE-hook (or MODE-hook, if FN is provided) is triggered to
+then wait until FEATURE-hook (or any of FNS, if FNS are provided) is triggered to
 reverse this and trigger `after!' blocks at a more reasonable time."
```
hlissner, line 604: Please refill this paragraph. Should end up looking like:

```diff
-then wait until FEATURE-hook (or any of FNS, if FNS are provided) is triggered to
+then wait until FEATURE-hook (or any of FNS, if FNS are provided) is triggered
+to reverse this and trigger `after!' blocks at a more reasonable time."
```

hlissner (changes requested):
Thanks for the PR! This looks good to me; just two small nits, one mentioned in the review and:

Please reword the commit message to add the correct [scope](https://discourse.doomemacs.org/t/git-conventions/2407#scopes-3) to the summary: 

```diff
-fix: defer-feature! fallback FNS
+fix(lib): defer-feature! fallback FNS

 If defer-feature! was called with one argument (as is the case in the
```

The rest of the commit message is fine.

hlissner (approved):

hlissner:
LGTM. Thanks again!

PR #6696, opened 2022-08-19
fix(csharp): add dep pkg for sln-mode

dradetsky:
I didn't see any relevant issues/PRs when searching a few things like `sln`, `csharp`, etc, but maybe I missed something

-----
- [x] My commits conform to [the git conventions](https://doomemacs.org/d/git-conventions). I think. I really didn't read them that long.
- ~My changes are visual; I've included before and after screenshots.~
-  ~This a draft PR; I need more time to finish it.~

-----------------------

While trying to use sln-mode via the csharp module, I noticed that it would not load because it was missing a dependency `font-lock-ext`. Examining the source of sln-mode, we see

```
(require 'font-lock-ext) ; https://github.com/sensorflo/font-lock-ext/
```
Now this is a trivial & tiny package (see for yourself), but as I don't want to rewrite any of the packages involved, the simplest fix for this is just to add this package to the module.

On a slightly separate note: I actually originally was trying to enable support for fsharp, not csharp. However, the fsharp module does not add support for .sln or .fsproj. However, the csharp module does add support for csproj-mode, which itself supports .fsproj (since both are just xml anyhow). I'm not sure how this sort of issue should be handled in general. Should we just add all the ancillary modes to fsharp as well? Should we create a `csharp-aux-tools` (or something) module which users of either the csharp or fsharp module would use as well? (Can modules depend on other modules?) Some 3rd option? I dunno, this should probably be a separate issue, but I thought I'd mention it.

### DIFF
--- a/modules/lang/csharp/packages.el
+++ b/modules/lang/csharp/packages.el
@@ -4,6 +4,8 @@
 (package! csharp-mode :pin "856ecbc0a78ae3bdc2db2ae4d16be43e2d9d9c5e")
 (package! csproj-mode :pin "a7f0f4610c976a28c41b9b8299892f88b5d0336c")
 (package! sln-mode :pin "0f91d1b957c7d2a7bab9278ec57b54d57f1dbd9c")
+;; sln-mode depends on font-lock-ext
+(package! font-lock-ext :pin "b6c82e8ac7996d96494a54454015a98ceb883feb")
 (when (modulep! +unity)
   (package! shader-mode :pin "d7dc8d0d6fe8914e8b6d5cf2081ad61e6952359c"))
 (when (modulep! +dotnet)


dradetsky:
I didn't see the thing about conventional-commit when perusing the contrib guidelines; oops

dradetsky:
Should I 

- uncommit
- recommit as (i think) `fix(csharp): add dep for sln` 
- force-push to overwrite this pr branch

or is another resolution preferred?

jeetelongname:
You should be able to edit the last commit in place by using `git commit --ammend` or `c a` in magit

elken:
This seems like it's an issue that should be handled upstream rather than here

dradetsky:
@elken how should it be handled upstream?

elken:
Because it's up to the package to dictate it's dependencies. This only fixes it for doom.

dradetsky:
@elken I said "how" not "why". It's not that I don't understand the idea of handling things upstream. It's that I'm not sure what all of the possible options for actually handling this upstream are, and which of them are reasonable.

At a minimum, handling this upstream would involve making a change to `sln-mode`, or forking it, making the change in the fork, and causing `package!` to pick up the fork (I don't actually know how to do this). I'm not even sure what the correct change would be, and I would have to know to do this. What the correct change would be is an example of the "how" I asked about (do you have an opinion?) The first thing that occurred to me was to use elpa dependencies, but neither sln-mode nor font-lock-ext are in elpa (i checked melpa & gnu elpa), so I would have to package both first, and I don't even know how to package something for elpa in the first place. I'm not sure how using elpa deps would interact with straight. So this seems like an unreasonable option, but maybe there's a more reasonable option I'm not thinking of (I don't know much about elisp development).

The immediate problem is that sln-mode has not been touched since 2015. Its creator has not made a public contribution on Github since 2019. Maybe he's dead. In any case, getting a change into upstream seems like it might not be very easy. Maybe I'm wrong & no change to upstream is actually required (e.g. this could be done at the level of a change in the elpa recipe), but I don't how this would actually work and whether the amount of effort involved would be reasonable.

On the other hand, we could just make the 1-line change I'm proposing. This actually fixes the problem now. And it doesn't preclude anyone who knows what they're doing (unlike me) from fixing this upstream and then removing this 1-line change. I figure we should make the simple change that fixes the problem for users now.

On the third hand, if we don't want to make the 1-line change, we should remove sln-mode, since it's broken. If you try to edit a .sln file it errors & doesnt load. Personally, I prefer to keep sln-mode, so I prefer making the 1-line change necessary to fix it. But IMO doing nothing pending a theoretical upstream fix is the worst option.

hlissner (approved):

hlissner:
Sorry for the confusion! Our maintainers are correct that issues like these ought to be tackled upstream (because they affect everyone using these packages), but that's supposing a typical case where a reasonable turnaround time can be expected from upstream.

Unfortunately, that'd require both packages be published to an ELPA and a `Package-Requires` header be added to `sln-mode`, but the package is likely unmaintained and [a request to publish sln-mode to MELPA was already made](https://github.com/sensorflo/sln-mode/issues/3). All the while, we face irrevocable breakage because of it.

So you're right that Doom should handle this, even if only temporarily. Much later, once https://github.com/doomelpa gets going, I'll fork the two packages there and submit them to [M]ELPA. Or perhaps factor out `font-lock-ext` entirely -- what it offers doesn't quite warrant the extra dependency.

hlissner (approved):

hlissner:
> Should we just add all the ancillary modes to fsharp as well? Should we create a csharp-aux-tools (or something) module which users of either the csharp or fsharp module would use as well? (Can modules depend on other modules?) Some 3rd option? I dunno, this should probably be a separate issue, but I thought I'd mention it.

I'm brainstorming some solutions to this. For example, combining the dotnet modules into a `:lang dotnet`, with languages behind flags (`:lang dotnet +fsharp`), or waiting until v3.1 where categories will become implicit modules: if any `:lang *` module is enabled, `:lang` itself will behave like an activated module, where shared packages can be declared, or their default recipes overrided (without installing them). For example:

```elisp
;; in modules/lang/packages.el
(override! sln-mode :recipe (...) :pin "...")

;; in modules/lang/csharp/packages.el
...
(package! sln-mode)

;; in modules/lang/fsharp/packages.el
...
(package! sln-mode)
```

Or perhaps implement "pseudo" packages for modules to tap into:

```elisp
;; in modules/lang/packages.el
(defpackage! :dotnet
  (package! sln-mode ...)
  ...)

;; in modules/lang/csharp/packages.el
(package! :dotnet)

;; in modules/lang/fsharp/packages.el
(package! :dotnet)
```

elken:
> with languages behind flags

The issue with that is it then becomes harder to do per-language flags like `+lsp` or `+tree-sitter`.

The introduction of metapackages I like best, it would also allow `:lang js` to be split up in js and ts

dradetsky:
I can't run it via gha, but:
```
0.151s (add-dep-to-csharp) ~/c/e/f/d/doom-emacs 00:05 $ doom ci lint-commits HEAD
> Linting 1 commits
  > edfc3f5 fix(csharp): add dep pkg for sln-mode
✓ There were no issues!
```

hlissner:
Thanks for your help! (And sorry for the late response!)

PR #6697, opened 2022-08-19
fix(lib): doom-exec-process use doom-print--format

ivanbrennan:
Update `doom-exec-process` to use `doom-print--format` instead of `doom--format`, which was removed in fe16de690418393c4b1dd79277f01a1c72b3be10add5a48022befe35b5f2e89c

-----


hlissner:

PR #6698, opened 2022-08-19
fix(lib): add cat to doom-system-distro-version

ivanbrennan:
One of the cond clauses in `doom-system-distro-version` calls a non-existent `cat` function. Referencing a similar pattern in the body of `doom-info`, I added a `(defun cat ...` to the `letf!` bindings in `doom-system-distro-version`.

-----

### DIFF
--- a/lisp/lib/system.el
+++ b/lisp/lib/system.el
@@ -27,7 +27,11 @@
 ;;;###autoload
 (defun doom-system-distro-version ()
   "Return a distro name and version string."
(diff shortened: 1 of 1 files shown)


hlissner:
Rather than restore the old binding, I thought it'd be better to simply expand them. I then later replaced them with calls to our new files API (in 9b4973198b8eb54394fd19e9fc0e533ff1b61f8b), which I think are better suited to the task. In any case, this has been resolved. Thanks for your help and for bringing it to my attention!

PR #6713, opened 2022-08-22
fix(dart): enable rainbow-delimiters in dart-mode

elken:


modules/lang/dart/config.el
```diff
@@ -3,6 +3,7 @@
 (use-package! dart-mode
   :defer t
   :config
+  :hook (dart-mode . rainbow-delimiters-mode)
```
hlissner, line 6: The order of keys actually breaks `:config`. Plus, `:hook` makes `:defer t` redundant. Please change this to:

```diff
-  :defer t
+  :hook (dart-mode . rainbow-delimiters-mode)
   :config
-  :hook (dart-mode . rainbow-delimiters-mode)

hlissner (changes requested):

modules/lang/dart/config.el
elken: You're right, I can't read...

I started adding it then panic pushed it because it was 9am. Noted to myself to _not_ do that again

hlissner:
LGTM. Thanks again!

PR #6715, opened 2022-08-23
Fix doom/version (canonicalize filenames in git calls)

ivanbrennan:
If `doom-emacs-dir` contains a "~", calling `doom/version` will generate an error like:
```
fatal: cannot change to '~/.config/emacs/': No such file or directory
```
This is because `git` is invoked via `call-process`, so "~" is not expanded as it would be in a shell command.

Fix this by canonicalizing the filename.

-----


hlissner (approved):

hlissner:

PR #6732, opened 2022-08-29
feat(emacs-lisp): recognize ";;* ..." headlines

LemonBreezes:
Here we patch `:lang emacs-lisp` so that `;;* ...` headings are recognized by Imenu and Outline's functions. We also patch `:editor lispy` so that Lispy does not override these settings and does not introduce its own font-lock keywords for Outline headings.

### DIFF
--- a/modules/editor/lispy/config.el
+++ b/modules/editor/lispy/config.el
@@ -22,7 +22,10 @@
       (local-set-key (vector 'remap (lookup-key lispy-mode-map (kbd "TAB"))) #'completion-at-point)))
   :config
   (setq lispy-close-quotes-at-end-p t)
-  (add-hook 'lispy-mode-hook #'turn-off-smartparens-mode))
+  (add-hook 'lispy-mode-hook #'turn-off-smartparens-mode)
+
+  (when (modulep! :lang emacs-lisp)
+    (add-hook 'lispy-mode-hook #'+lispy-restore-elisp-outline-settings)))
 
 
 (use-package! lispyville
--- a/modules/editor/lispy/config.el
+++ b/modules/editor/lispy/config.el
@@ -24,11 +24,6 @@
   (setq lispy-close-quotes-at-end-p t)
   (add-hook 'lispy-mode-hook #'turn-off-smartparens-mode)
 
-  ;; We recommend `outline-minor-faces-mode' from
-  ;; https://github.com/tarsius/outline-minor-faces as an alternative which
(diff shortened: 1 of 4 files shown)


modules/editor/lispy/config.el
```diff
@@ -22,7 +22,15 @@
       (local-set-key (vector 'remap (lookup-key lispy-mode-map (kbd "TAB"))) #'completion-at-point)))
   :config
   (setq lispy-close-quotes-at-end-p t)
-  (add-hook 'lispy-mode-hook #'turn-off-smartparens-mode))
+  (add-hook 'lispy-mode-hook #'turn-off-smartparens-mode)
+
+  ;; We recommend `outline-minor-faces-mode' from
+  ;; https://github.com/tarsius/outline-minor-faces as an alternative which
+  ;; respects `outline-regexp' and `outline-level'.
+  (setq lispy-font-lock-keywords nil)
```
seanfarley, line 30: This is a minor nit but I sorta think this should go into another PR that adds `outline-minor-faces-mode` (and then sets this variable to `nil`). Currently, before this PR a user has a nice fontified section header but after this it disappears.

That being said, I've already added `outline-minor-faces-mode` to my own config and if this gets merged will submit a PR here, so it's not a blocker for me.

LemonBreezes: I'll do that and I'm also looking into another tweak to get `outline-demote` working too.

LemonBreezes: Okay. I have removed those lines but have decided not to fix `outline-demote` and other outline features. My explanation is as follows:

Outline expects `outline-regexp` to exactly match the heading, so ";;;" or ";;*" rather than ";;; V" (if the heading is ";;; Variables"). However, we want to exclude autoloads from being sections and Emacs does not have a way to negate in regular expressions. The only way to fix this is to advise all the functions in `outline` which use `(match-string 0)` after searching for `outline-regexp`. I don't think it's worth it.

hlissner:
I'm not sure this is desirable. This would make *every* line of every comment block a veritable imenu section and/or outline-mode heading. Three semi-colons should be the minimum for this.

hlissner:
There were a bunch of minor changes needed (mainly to conform git messages to conventions, squash commits, and polish our regexp for `outline-regexp` (to ensure the order of semi-colons and asterisks)), which I've done in a0ccb6b, which should resolve this issue. Let me know if it causes you any issues. Thanks for your help and for bringing it to my attention!

PR #6738, opened 2022-08-31
DWIM for most app-switching commands (=mu4e, =elfeed, =calendar)

LemonBreezes:
In the #emacs-lisp channel of the Doom Emacs Discord, we were discussing how the current behavior of `=mu4e` can be annoying because if a user is already reading an email, leaves the `*mu4e*` workspace, and switches back with `=mu4e`, the Mu4e main buffer is opened again in the active window, forcing the user to manually `switch-to-previous-buffer`. I have implemented an improvement upon this behavior for various modules so that if we call `=mu4e`, `=rss`, etc., and that workspace is already in Mu4e/Elfeed, etc., we do not call `(mu4e)`, `(elfeed)`, etc. again (respectively).

While reviewing these various app modules, I have found some behaviors that seemed odd to me such as `=irc` emitting an error if the workspace is already in an IRC node or `=notmuch` using the workspace name `*notmuch*` and have patched these too.

### DIFF
--- a/modules/app/calendar/autoload.el
+++ b/modules/app/calendar/autoload.el
@@ -1,6 +1,8 @@
 ;;; app/calendar/autoload.el -*- lexical-binding: t; -*-
 
 (defvar +calendar--wconf nil)
+(defvar +calendar-workspace-name "*calendar*"
+  "Name of the workspace created by `=calendar', dedicated to the calendar.")
 
 (defun +calendar--init ()
   (if-let (win (cl-find-if (lambda (b) (string-match-p "^\\*cfw:" (buffer-name b)))
@@ -15,7 +17,7 @@
   (interactive)
   (if (modulep! :ui workspaces)
       (progn
-        (+workspace-switch "Calendar" t)
+        (+workspace-switch +calendar-workspace-name t)
         (doom/switch-to-scratch-buffer)
         (+calendar--init)
         (+workspace/display))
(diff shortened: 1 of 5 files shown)


seanfarley:
Do Want Give It To Me - nh9k1qzeLf99S

seanfarley:
I've been using this PR for 10 days and it's been really great, fwiw

danilevy1212:
I have taken this PR as inspiration to put my global `org-todo-list` in it's own workspace:

```
(defvar +org-todo-list-workspace-name "Org TODOs")

;;;###autoload
(defun dan/=org-todo-list (&optional arg)
  "Like `org-todo-list', but the buffer is within it's own workspace"
  (interactive "P")
  (+workspace-switch +org-todo-list-workspace-name t)
  (unless (eq (buffer-local-value 'major-mode (window-buffer (selected-window)))
              'org-agenda-mode)
    (doom/switch-to-scratch-buffer)
    (org-todo-list arg)
    (with-current-buffer org-agenda-buffer-name
      (add-hook! 'kill-buffer-hook :local t
                 (+workspace-delete +org-todo-list-workspace-name)))
    (+workspace/display)))
```

Wonder if anybody would find this useful, I certainly like it more this way.

hlissner (approved):

hlissner:
Thanks for your help!

seanfarley:
image

PR #6746, opened 2022-09-02
fix(layout): Bépo rotation happening before magit

gagbo:
Fixes an issue where binding rotations in Bépo were happening too soon,
resulting in some remaps not being visible to the user when interacting
with magit, :editor evil +everywhere, and :input layout +bepo

Ref: https://discourse.doomemacs.org/t/layout-bepo-does-not-properly-remap-k-in-magit/3063


hlissner (approved):

hlissner:
Good work, as always! Thanks again!

PR #6749, opened 2022-09-05
fix(vertico): erroring when grep or rg not found

bb010g:
When installing Doom Emacs on Windows, I ran into `doom doctor` erroring out and exiting when `grep` wasn't found. The presence of `grep` and `rg` should be verified before they're presumed to exist in the call to `consult--grep-lookahead-p`.

-----

### DIFF
--- a/modules/completion/vertico/doctor.el
+++ b/modules/completion/vertico/doctor.el
@@ -2,10 +2,14 @@
 
 (require 'consult)
 
-(unless (consult--grep-lookahead-p "grep" "-P")
-  (warn! "The installed grep binary was not built with support for PCRE lookaheads.
-  Some advanced consult filtering features will not work as a result, see the module readme."))
+(if (executable-find "grep")
+  (unless (consult--grep-lookahead-p "grep" "-P")
+    (warn! "The installed grep binary was not built with support for PCRE lookaheads.
+    Some advanced consult filtering features will not work as a result, see the module readme."))
+  (warn! "Couldn't find grep."))
 
-(unless (consult--grep-lookahead-p "rg" "-P")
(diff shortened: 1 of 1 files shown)


elken:
There's already a global check for ripgrep [here](https://github.com/doomemacs/doomemacs/blob/master/lisp/cli/doctor.el#L94) so I wonder if there should probably just be an additional check there for grep

abougouffa:
I think that `ripgrep` is a **hard** dependency of Doom Emacs, as specified in the *[prerequisites](https://github.com/doomemacs/doomemacs#prerequisites)* section. So, it would be safe to assume its presence.

elken:
> I think that `ripgrep` is a **hard** dependency of Doom Emacs, as specified in the _[prerequisites](https://github.com/doomemacs/doomemacs#prerequisites)_ section. So, it would be safe to assume its presence.

Never make assumptions :) Hence why the ripgrep check already exists

abougouffa:
> Never make assumptions :) Hence why the ripgrep check already exists

Sure, and it is coherent for the check to happen in Doom's core files rather than modules.

bb010g:
Assuming the presence in other modules is fine, but those assumptions shouldn't crash `doom doctor`, the thing that tells you when those assumptions aren't met.

hlissner:
I have a rewrite of the doctor system in the works for v3, which will include core-level checks for Doom's dependencies (like ripgrep), so individual modules don't have to do their own checks like this. Plus, they will do a much better job of distinguishing between runtime errors and doctor failures, which will indirectly fix this issue. That, and because v3 is just around the corner, I have to turn down this PR, but thanks for bringing it to my attention!

PR #6755, opened 2022-09-07
fix(debugger): fix dap-ui conditional loading

gagbo:
Probably missed when adding eglot support code in the module


hlissner (approved):

hlissner:
Thanks again!

PR #6758, opened 2022-09-07
feat(vertico): make component separators escapable

LemonBreezes:
Currently, we use the characters \s and & as component separators for autocompletion. This is good, however there are scenarios where a user would like to match these separators literally as part of a search query. This PR enables that.

### DIFF
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -86,13 +86,27 @@ orderless."
      +vertico-basic-remote-try-completion
      +vertico-basic-remote-all-completions
      "Use basic completion on remote files only"))
+
+  (defun +vertico-escapable-split-on-space-or-ampersand (string)
+    "This function splits `string' into"
+    (mapcar
+     (lambda (piece)
+       (replace-regexp-in-string (concat "\0" "\\|" "\1")
+                                 (lambda (x) (pcase x ("\0" " ") ("\1" "&") (_ x)))
+                                 piece))
+     (split-string (replace-regexp-in-string
+                    "\\\\\\\\\\|\\\\ \\|\\\\&"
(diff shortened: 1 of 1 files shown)


LemonBreezes:
> Hey, like I said on the discord I'll look into this issue over the weekend. Ideally I'd like to just use `#'orderless-escapable-split-on-space`, I'm not sure if the reasons for not having gone with that are still relevant (using `&` is rather awkward if we don't need to, and I think this bit of code predates the existence of easily using space as a separator).

The reason we use `&` as a separator is so that when we use autocompletion with Corfu (I'm not sure if Company also uses Orderless), we have an easy way to insert a wildcard into the query.

minad:
@LemonBreezes With Corfu you could still use orderless-escapable-split-on-space and instead bind & to corfu-insert-separator in corfu-map :)

minad:
But this has the bad side effect that you cannot enter literal ampersand :-P You could then bind M-& to insert &. Whatever way you turn it there will be some downside...

minad:
Hmm, I think the best solution is the following:

1. Use orderless-escapable-split-on-space
2. Bind SPC to corfu-insert-separator in corfu-map
3. Bind M-SPC to a custom command which inserts backslash and space

Any downsides of that? At least you leave the poor ampersand alone.

LemonBreezes:
> ~Hmm, I think the best solution is the following:~
> 
> 1. ~Use orderless-escapable-split-on-space~
> 2. ~Bind SPC to corfu-insert-separator in corfu-map~
> 3. ~Bind M-SPC to a custom command which inserts backslash and space~
> 
> ~Any downsides of that? At least you leave the poor ampersand alone.~
> 
> Nevermind, this won't work well. We want to keep space to end unintentional completion after an identifier.

You're right. The best course of action is to not make the ampersand special by default but it is also not too bad of a trick.

PR #6764, opened 2022-09-10
fix(lib): unquote dir! call in load! macro

theschmocker:
This restores previous behavior where `dir!` was called at the time of `load!`'s expansion, which ensured that it resolved to the directory of the source file where load! was called

In my case, this manifested in not being able to use evil-ex because of a "Cannot get this file-path" error _unless_ evil-ex was called with modules/editor/evil/config.el in the active buffer first to load it

AFAICT, this is equivalent to how load worked before https://github.com/doomemacs/doomemacs/commit/a179b8d262a0630e543af24fa5ffd8d33da0ac89, but I'm not sure whether quoting `(dir!)` was significant there and if this will have a negative impact on anything else and constitutes a breaking change

-----

### DIFF
--- a/lisp/doom-lib.el
+++ b/lisp/doom-lib.el
@@ -610,7 +610,7 @@ directory path). If omitted, the lookup is relative to either `load-file-name',
 
 If NOERROR is non-nil, don't throw an error if the file doesn't exist."
   `(doom-load
-    (file-name-concat ,(or path `(dir!)) ,filename)
+    (file-name-concat ,(or path (dir!)) ,filename)
     ,noerror))
 
 (defmacro defer-until! (condition &rest body)


hlissner:
Sorry for the late response. Quoting the `dir!` was intentional, but prematurely committed (it was supposed to be pushed after 7e0c2ed). 7e0c2ed should resolve the issue you were seeing. It is intentional that I've quoted it, so that it is expanded where `load!` is called, rather than expanded in `load!`'s definition (where it would always, unhelpfully return the path to `doom-lib.el` instead). In any case, thanks for bringing it to my attention!

theschmocker:
@hlissner no problem! Thanks for the quick fix and the explanation; all is well with evil-ex again

PR #6778, opened 2022-09-13
fix(config): tangle with the same Org as Doom uses

tecosaur:
I feel like this is pretty simple and self-explanatory.

Without this change the sync and async tangling functions can (will) use different Org versions.

-----
- [x] I searched the issue tracker and this hasn't been PRed before.
- [x] My changes are not on [the do-not-PR list](https://doomemacs.org/d/do-not-pr) for this project.
- [x] My commits conform to [the git conventions](https://doomemacs.org/d/git-conventions).
- [ ] My changes are visual; I've included before and after screenshots.
- [ ] I am blindly checking these off.
- [x] Any relevant issues or PRs have been linked to (none).
- [ ] This a draft PR; I need more time to finish it.

PR #6799, opened 2022-09-17
fix(literate): use same org install when using CLI

tecosaur:
It's good for config tangling behaviour to be consistent across:
1. interactive sessions
2. asynchronous calls
3. the doom CLI

(1) of course uses the Doom-installed org, (2) was taken care of in #6778, and this should make (3) behave the same.

Without this, one can run into bugs and behavioural differences between the emacs-bundled Org and the (more recent) doom-installed Org, for example, see https://git.savannah.gnu.org/cgit/emacs/org-mode.git/commit/?id=cb8bf4a0d which fixes an issue with `:noweb-ref` references not being expanded in some situations.

-----

### DIFF
--- a/modules/config/literate/cli.el
+++ b/modules/config/literate/cli.el
@@ -2,5 +2,18 @@
 
 (load! "autoload")
 
+(defun +literate-add-installed-org-to-load-path-h ()
+  "Use the straight-installed, not bundled Org."
+  (let ((straight-org-build-dir
+         (doom-path straight-base-dir "straight" straight-build-dir "org"))
+        (straight-org-repo-dir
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
This introduces a race condition where, if a user's literate config depends on newer versions of Org, it will likely throw an error on `doom install`/`doom sync` in a fresh Doom install. They'd need to run it once with `:config literate` disabled, then again with it on.

Still, I think this is acceptable until 3.1 where I'll introduce first class support for literate configs in Doom core, and where spin-off profiles will resolve this issue much more elegantly.

In any case, thanks again for your great work and help!

PR #6805, opened 2022-09-18
fix: only use trash on local hosts

### LINKED ISSUE #6804: Doom + emacs-plus 28 cannot delete file using dired in tramp mode

I use doom and emacs-plus 28. When I was in tramp mode, if I open a dried buffer and delete a file(e.g. D) the file should be successfully deleted.

### What actually happened?

I get an error message: 

Finder got an error: Handler can't handle objects of this class. (-10010)
image

In [doomemacs/config.el at b395bd5392dd6ca88b230102363797b47b8f6991 · doomemacs/doomemacs · GitHub 2](https://github.com/doomemacs/doomemacs/blob/b395bd5392dd6ca88b230102363797b47b8f6991/modules/os/macos/config.el#L50), system-move-file-to-trash was set to osx-trash-move-file-to-trash, which leads to this issue in tramp mode. 

A workaround:
```
;; see document of `move-file-to-trash'
  (defun system-move-file-to-trash (filename)
    (process-file-shell-command
     (format "trash %S" (file-local-name filename))))
```

[Reference](https://emacs-china.org/t/bug-doom-emacs-plus28-ssh-dried/22265/2)

### Steps to reproduce

The issue can be reproduced with emacs-plus28 and doom:

1. ssh to remote using tramp mode
2. open dried buffer(space o -)
3. delete a file(d x or D) 
 

### System Information

https://pastebin.com/TA69jGQ6

elken:
Might be worth introducing a "remote trash command", but for now this seems to work.

Closes #6804

### DIFF
--- a/modules/os/macos/config.el
+++ b/modules/os/macos/config.el
@@ -45,6 +45,9 @@
   (setq delete-by-moving-to-trash t)
 
   ;; Lazy load `osx-trash'
-  (and IS-MAC
-       (not (fboundp 'system-move-file-to-trash))
-       (defalias #'system-move-file-to-trash #'osx-trash-move-file-to-trash)))
+  (when (not (fboundp 'system-move-file-to-trash))
+    (defun system-move-file-to-trash (file)
+      "Move FILE to trash."
+      (when (and (not IS-LINUX)
+                 (not (file-remote-p default-directory)))
+        (osx-trash-move-file-to-trash file)))))


modules/os/macos/config.el
```diff
@@ -45,6 +45,10 @@
...
   ;; Lazy load `osx-trash'
-  (and IS-MAC
-       (not (fboundp 'system-move-file-to-trash))
-       (defalias #'system-move-file-to-trash #'osx-trash-move-file-to-trash)))
+  (when (not (fboundp 'system-move-file-to-trash))
+    (defun system-move-file-to-trash (file)
+      "Move FILE to trash."
+      (when (and IS-MAC
+                 (not (file-remote-p default-directory))
+                 (not (string= (string-trim (shell-command-to-string "uname")) "Linux")))
```
hlissner, line 53: Any reason you're not using `(not IS-LINUX)` here? In fact, why is this check necessary at all with `IS-MAC` already in the condition?

hlissner (changes requested):

hlissner (changes requested):
Could you also add the `macos` module scope and Fix: trailer to the commit message? I.e.

```
fix(macos): only use trash on local hosts

Fix: #6804
```

modules/os/macos/config.el
elken: It didn't work for me when remote

hlissner: But why would the condition get past `(not (file-remote-p default-directory))` when remote?

elken: Hm I see your point, I think I overthought it

hlissner (approved):

hlissner:

PR #6826, opened 2022-09-20
feat(workspaces): add mru cache for alt-tab like functionality

seanfarley:
This enables keeping a list of workspaces in most recently used order so that we can enable "alt-tab"-like functionality for switching between workspaces that are not apps (or vice versa, only switching between apps).

This is currently behind an +mru flag to workspaces but it doesn't necessarily need to be since it doesn't change any behavior. Users would need to opt-in to using the new functions.

Need to decide other keybindings for evil, if any. Also, need to decide if this should be a flag or not.

### DIFF
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -408,6 +408,9 @@
         :desc "Delete workspace"             "k" #'+workspace/delete
         :desc "Save workspace"               "S" #'+workspace/save
         :desc "Switch to other workspace"    "o" #'+workspace/other
+        (:when (modulep! :ui workspaces +mru)
+         :desc "Switch to other non-app workspace"
+         "O" #'+workspace/other-non-app)
         :desc "Switch to left workspace"     "p" #'+workspace/switch-left
         :desc "Switch to right workspace"    "n" #'+workspace/switch-right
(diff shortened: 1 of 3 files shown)


seanfarley:
Hypothetically speaking, this could replace the `+workspace--last` variable but I leave that for a future PR since it would be quite the significant change.

hlissner:
Thanks for the PR, but I'm afraid I must turn this down. Though this functionality is something I want for the `:ui workspaces` module, eventually, this implementation introduces some undesired side-effects by reordering `persp-names-cache` (affecting the positional access commands, like `SPC TAB [N]`, and causing the ivy/vertico/helm lookup commands to have non-deterministic listings -- and more for any interfaces that read/write directly from/to `persp-names-cache`).

For a while I've wanted to build a layer over persp-mode to abstract workspace names from their hash table keys. Mainly to resolve the possibility of name collisions (which isn't uncommon among project names, for instance), but this would also enable us to arbitrarily sort workspaces in any way we like, and without side-effects to interfaces that directly read/write to `persp-names-cache`. Once that is done, I'll look into adding these MRU commands.

That said, it's not high on my TODO list at this time, so there's no ETA on it yet, but I'll close this for now, since I'm prepping to move modules into a separate repo (and PRs can't be moved). Thanks for bringing it to my attention, in any case!

PR #6827, opened 2022-09-20
fix(mu4e): cooperative lock file watching

tecosaur:
Some time ago I noticed the cooperative lock file management wasn't working as I remember. I forget what exactly I was thinking, but basically I've poked at the code until it seems to work better.

If some other people could test this and report back, that would be much appreciated.

-----

### DIFF
--- a/modules/email/mu4e/autoload/mu-lock.el
+++ b/modules/email/mu4e/autoload/mu-lock.el
@@ -55,10 +55,11 @@ Else, write to this process' PID to the lock file"
     (delete-file +mu4e-lock-request-file)
     (call-process "touch" nil nil nil +mu4e-lock-request-file)
     (funcall orig-fun callback)
-    (setq +mu4e-lock--request-watcher
-          (file-notify-add-watch +mu4e-lock-request-file
-                                 '(change)
-                                 #'+mu4e-lock-request))))
+    (when +mu4e-lock--request-watcher
+      (file-notify-rm-watch +mu4e-lock--request-watcher))
+    (file-notify-add-watch +mu4e-lock-request-file
+                           '(change)
+                           #'+mu4e-lock--request)))
 
 (defvar +mu4e-lock--file-watcher nil)
 (defvar +mu4e-lock--file-just-deleted nil)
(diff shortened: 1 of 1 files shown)


modules/email/mu4e/autoload/mu-lock.el
```diff
@@ -53,12 +53,12 @@ Else, write to this process' PID to the lock file"
       (user-error "Unfortunately another Emacs is already doing stuff with Mu4e, and you can only have one at a time")
     (write-region (number-to-string (emacs-pid)) nil +mu4e-lock-file)
     (delete-file +mu4e-lock-request-file)
-    (call-process "touch" nil nil nil +mu4e-lock-request-file)
```
seanfarley, line 56: Won't this have the same problem as before on BSD kqueue systems? `file-notify-add-watch` is implementation dependent and doesn't always create the file.

seanfarley: Just confirmed:

```
error in process filter: file-notify-add-watch: File does not exist: No such file or directory, /var/folders/bj/cvgkd64s42v_dt0z1bpy397c0000gn/T/mu4e_lock_request
error in process filter: File does not exist: No such file or directory, /var/folders/bj/cvgkd64s42v_dt0z1bpy397c0000gn/T/mu4e_lock_request
```

seanfarley: Maybe we can add a comment about `kqueue` differences so as to not remove the call to `touch`?

tecosaur: Hmm, how do you find the behaviour if that line is kept? (i.e. does it work)

seanfarley: It works as expected if I add back the `touch` call.

seanfarley: Alternatively, we could patch the emacs functions to create the file to make all the implementations line up. Maybe even submit that upstream.

seanfarley:
image

tecosaur:
Sean ... where did you get all of these images from?

hlissner (approved):

hlissner:
Thanks again!

PR #6829, opened 2022-09-21
fix(csharp): fix +tree-sitter

dradetsky:
The syntax highlighting for basic csharp-mode is busted, and I don't know how to fix it upstream. I'm not even sure if it can be fixed without abandoning the current cc-langs based approach (explaining this is probably more complicated than maintainers want to hear about).

Fortunately, csharp-mode also supports an "experimental" csharp-tree-sitter-mode which actually works correctly. However, doom's +tree-sitter flag does not correctly enable this mode. I think the existing use of the flag may have been based on instructions from before it was a separate major mode. Anyway, this change seems to fix it.

-----

### DIFF
--- a/modules/lang/csharp/config.el
+++ b/modules/lang/csharp/config.el
@@ -1,6 +1,9 @@
 ;;; lang/csharp/config.el -*- lexical-binding: t; -*-
 
 (use-package! csharp-mode
+  :init
+  (when (modulep! +tree-sitter)
+    (add-to-list 'auto-mode-alist '("\\.cs\\'" . csharp-tree-sitter-mode)))
   :hook (csharp-mode . rainbow-delimiters-mode)
   :config
   (set-electric! 'csharp-mode :chars '(?\n ?\}))
--- a/modules/lang/csharp/packages.el
+++ b/modules/lang/csharp/packages.el
@@ -10,5 +10,7 @@
   (package! shader-mode :pin "d7dc8d0d6fe8914e8b6d5cf2081ad61e6952359c"))
 (when (modulep! +dotnet)
   (package! sharper :pin "96edd4a1dbc267afdff0cb97298d1b05b7c2080c"))
+;; In addition to the packages provided by the tools/tree-sitter module,
+;; csharp-tree-sitter-mode also requires tree-sitter-indent.
 (when (modulep! +tree-sitter)
   (package! tree-sitter-indent :pin "4ef246db3e4ff99f672fe5e4b416c890f885c09e"))
--- a/modules/lang/csharp/packages.el
+++ b/modules/lang/csharp/packages.el
@@ -10,3 +10,5 @@
   (package! shader-mode :pin "d7dc8d0d6fe8914e8b6d5cf2081ad61e6952359c"))
 (when (modulep! +dotnet)
   (package! sharper :pin "96edd4a1dbc267afdff0cb97298d1b05b7c2080c"))
+(when (modulep! +tree-sitter)
+  (package! tree-sitter-indent :pin "4ef246db3e4ff99f672fe5e4b416c890f885c09e"))


jeetelongname:
> explaining this is probably more complicated than maintainers want to hear about
If you can it would be a help :slightly_smiling_face: 

Otherwise I can open a PR that adds `csharp-tree-sitter-mode` to the block that enables tree-sitter

jeetelongname:
Ill review this in a bit

dradetsky:
@jeetelongname since you asked for the explanation: here are two files that aren't highlighted properly with regular syntax highlighting:

```
class C {
    const string s = @"ok ""esc""} str";
}
```

and

```
class Klass {
    public static bool operator ==(Klass var0, Klass var1) {
        return true;
    }
}
```

I didn't look into the second case very carefully. I was looking into how I could fix the first case, then noticed the second case, then decided to try tree-sitter, discovered that it actually worked, and decided that I didn't know if fixing regular syntax was worth doing.

Anyway, I'm not great at elisp, but as far as I can tell csharp-mode is using the cc-langs system for "multiline strings" to handle verbatim strings. In verbatim strings, `""` is a single escaped `"`, so when you encounter one you're supposed to keep the string going as far as highlighting is concerned. But the multiline string routine _requires_ that `"` be treated as a closer for the string; it's designed to handle cases like Python multiline strings where the closer is `"""`, and then you match the final quote or whatever. It's possible that I've misunderstood how this system works, but the people who wrote csharp-mode clearly didn't make it work and they're probably much better at both elisp and csharp than me.

The only possible escape hatch seems to be `c-make-ml-string-closer-re-function`. I _think_ this overrides the normal closer-finding behavior, but I'm not totally sure. In any case, I couldn't get a solution based on this to work. This is probably just me being bad at elisp, but my point is that I can't determine if there is or is not a solution within the `c-lang-defconst` way of doing things. I could learn a bunch more about elisp, but I could also just use tree-sitter, which seems to handle every case perfectly so that seemed better.

modules/lang/csharp/config.el
```diff
@@ -1,6 +1,9 @@
 ;;; lang/csharp/config.el -*- lexical-binding: t; -*-
 
 (use-package! csharp-mode
+  :init
+  (when (modulep! +tree-sitter)
```
jeetelongname, line 5: Why can this not be added to the local hook [here?](https://github.com/doomemacs/doomemacs/blob/master/modules/lang/csharp/config.el#L38-L39)

This will run after tree sitter is enabled and will stay local to the buffer.

modules/lang/csharp/packages.el
```diff
@@ -10,3 +10,5 @@
   (package! shader-mode :pin "d7dc8d0d6fe8914e8b6d5cf2081ad61e6952359c"))
 (when (modulep! +dotnet)
   (package! sharper :pin "96edd4a1dbc267afdff0cb97298d1b05b7c2080c"))
+(when (modulep! +tree-sitter)
```
jeetelongname, line 13: Why is tree-sitter-indent needed here? Is it a dependency of of `csharp-tree-sitter-mode`? If not it may be better off in the `:tools tree-sitter` module or in one of the `:editor` modules

jeetelongname (changes requested):

modules/lang/csharp/config.el
dradetsky: I tried that & it didn't work. I think it's because we're trying to change something that's global & which we specifically want enabled before we load tree-sitter or csharp. Anyway, I'm basically doomifying this: https://github.com/emacs-csharp/csharp-mode/#tree-sitter-support

dradetsky: Also, I'm pretty sure the hook you're referring to is obsolete (i.e. we don't to set that thing anymore), but I didn't want to remove it in case somebody could explain what was going on there better.

modules/lang/csharp/packages.el
dradetsky: Yes, it is.

jeetelongname: Can you add a comment than mentions this.

modules/lang/csharp/config.el
jeetelongname: I will discuss this with @hlissner but we use the local vars hook as its only called in the places we wanted (file associated buffers being the main one)

modules/lang/csharp/packages.el
dradetsky: done

modules/lang/csharp/config.el
dradetsky: I checked the source. There are comparatively few instances of `auto-mode-alist` in modules code. Of those, I checked about 8. 2 of them (`lang/cc/autoload.el`, `lang/latex/config.el`) are just top-level usages. All the others I checked did the same sort of thing I do: modify `auto-mode-alist` in the `:init` form. Of those, one (`lang/nix/config.el`) appears to have been written by hlissner (or at least he definitely touched that code according to git-blame). I didn't check who wrote the other examples that do the same thing I do.

So I mean, feel free to discuss it with him if you want, but I dunno that you necessarily have to.

hlissner (approved):

hlissner:
Sorry for the tremendously late response. I've merged this into 883a3fa, where I've grouped the tree-sitter config under csharp-tree-sitter. I've also added tree-sitter-indent to the tree-sitter module, so it can be opted into wherever it is available (and superior to built-in indentation). Let me know how that fairs!

In any case, thanks for your help and the PR, and for the helpful examples!

PR #6830, opened 2022-09-21
fix(lib): doom/delete-this-file nil path handling

ivanbrennan:
Fix the handling of a nil `path` within `doom/delete-this-file`.

If `path` is nil (e.g. called interactively when buffer is not visiting a file), avoid calling `abbreviate-file-name` on nil, otherwise an error will be signaled:
```
(wrong-type-argument stringp nil)
```
Additionally, fix the subsequent path checks. These were treating two distinct scenarios as a "Buffer is not visiting any file" user-error:
- nil path
- non-existent path
Only the first should result in that error. The second should proceed to the next path check (which was previously unreachable), to signal the appropriate error, "File doesn't exist: %s".

_Note_: I noticed that this function calls `user-error` in some cases and `error` in others. I'm not sure whether that's intentional, so I didn't change it.

-----

### DIFF
--- a/lisp/lib/files.el
+++ b/lisp/lib/files.el
@@ -406,8 +406,8 @@ If FORCE-P, delete without confirmation."
    (list (buffer-file-name (buffer-base-buffer))
          current-prefix-arg))
   (let* ((path (or path (buffer-file-name (buffer-base-buffer))))
-         (short-path (abbreviate-file-name path)))
-    (unless (and path (file-exists-p path))
+         (short-path (and path (abbreviate-file-name path))))
+    (unless path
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
LGTM. Thanks for the help! 

> Note: I noticed that this function calls user-error in some cases and error in others. I'm not sure whether that's intentional, so I didn't change it.

That is intentional. I use `user-errors` for errors caused by the caller (the user), e.g. calling the command in a file-less buffer or aborting the confirmation prompt, and `error` for problems caused by the underlying system (e.g. file IO errors).

PR #6843, opened 2022-09-25
fix(lib): doom-info symlink-path

ivanbrennan:
Swap the **THEN** and **ELSE** in `symlink-path` so that symlinks are displayed correctly.

Currently, `bin/doom info` displays a non-symlinked path like,
```
~/.config/emacs/ -> ~/.config/emacs/
```
and a symlinked path like,
```
~/.config/emacs/
```

-----


hlissner (approved):

hlissner:

PR #6866, opened 2022-10-02
module: add `:emacs eww`

agzam:
Adds eww module.

*A few useful things I borrowed from Protesilaos Stavrou. I don't know what the proper attribution protocols we follow in Doom, I simply added a comment, but let me know if that has to be done differently.*


protesilaos:
Hello there!  I am distributing my dotfiles under the GPLv3.  If I can grant you MIT-License-compatible rights, please let me know what is needed and I will do it.  In other words, use whatever you want.

hlissner:
Sorry for the late response! I've put together an `:emacs eww` module inspired by this PR in f27a85e. Let me know if it gives you any trouble. Thanks for bringing it to my attention!

PR #6867, opened 2022-10-03
feat(eshell): looking up documenation and history search

aikrahguzar:

#6641 accidentally got closed because of my inadequate git skills. So I am reopening it here after fixing things on my end.

### DIFF
--- a/modules/term/eshell/autoload/commands.el
+++ b/modules/term/eshell/autoload/commands.el
@@ -28,3 +28,8 @@ For folks with a habit of using \"emacs\" to open files, even in eshell."
   "Create a directory then cd into it."
   (make-directory dir t)
   (eshell/cd dir))
+
+;;;###autoload
+(defun eshell/help (cmd)
+  "Get help for command from tldr, man or helpful."
+  (+eshell-help-run-help cmd))
--- a/modules/term/eshell/autoload/eshell.el
+++ b/modules/term/eshell/autoload/eshell.el
@@ -166,7 +166,7 @@ Once the eshell process is killed, the previous frame layout is restored."
         ((modulep! :completion helm)
          (helm-eshell-history))
         ((modulep! :completion vertico)
-         (forward-char 1) ;; Move outside of read only prompt text.
+         (eshell-bol) ;; Move outside of read only prompt text.
          (consult-history))
         ((eshell-list-history))))
 
--- a/modules/term/eshell/autoload/help.el
+++ b/modules/term/eshell/autoload/help.el
@@ -17,9 +17,9 @@
    ((or (and (string-match-p "^\\*." cmd) (cl-callf substring cmd 1))
         (eshell-search-path cmd))
     (or (and (tldr-get-file-path-from-command-name cmd)
-             (prog1 t (tldr cmd))
+             (prog1 t (tldr cmd)))
         (and (Man-completion-table cmd nil nil)
-             (prog1 t (man cmd))))))
+             (prog1 t (display-buffer (man cmd))))))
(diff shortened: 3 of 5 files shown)


aikrahguzar:
Something pretty exciting on the eshell front is `eshell-eat-mode` from the package [eat](https://codeberg.org/akib/emacs-eat). This mode can run visual commands from within `eshell`. The package is very new so some stability and bugs are to expected. But in my limited experience it works really well. For example I can run `btm` in emacs without the flicker which happens in `term`.

I am wondering if it should be integrated with the eshell module possibly under a `+eat` flag. I don't think much configuration is needed but integrations with evil need to be ironed out, especially how to handle `ESC`. Other than that all that think of is a hook that sets the `eshell-visual-*commands` to `nil` when `eshell-eat-mode` is turned on while saving their value and restores it when turning off the mode.

AkibAzmain:
@aikrahguzar Eat is much more generic, it can be used like Vterm without Eshell, for example with `M-x eat`.  And the Eshell mode is `eat-eshell-mode`.

aikrahguzar:
> And the Eshell mode is `eat-eshell-mode`.

Thanks! I corrected that mistake. 

Eat is of course more general, but my interest is mostly due to eshell since it fixes what was for me the last annoyance of eshell since I don't have to care about visual commands any more. It also improves the rendering of output for some commands in eshell which is a nice bonus.

AkibAzmain:
@aikrahguzar If you add Eat to Doom, I suggest to also add Eat module, in addition to Eshell integration.

AkibAzmain:
And for anybody interested in Eat, you can use your mouse in Eat!

aikrahguzar:
After running into an issue with `eat` which was ultimately caused by the function `+eshell-remove-fringes-h` in `eshell-mode-hook`, I have changed it to a hopefully more robust version suggested by @AkibAzmain

aikrahguzar:
> @aikrahguzar thanks for sharing that; why not use [pcmpl-args-pcomplete-on-man](https://github.com/JonWaltman/pcmpl-args.el/blob/43229e1096f89c277190f09a3d794781f8fa0015/pcmpl-args.el#L1738C1-L1756C58) as a higher priority in the fallback order? That would lessen the need to trust commands in your path.

I want to figure out how to use `pcomplete-from-help` to parse manual pages, but it is something for distant future. It should be a matter of figuring out the correct regexps to pass to `pcomplete-from-help` so not too hard. I can't imagine a scenario where I need to defend against the commands I have explicitly installed so I don't actually see a problem with calling a command with `-h` or `--help` and the results for completions are good enough that I don't want to install an extra package when `pcomplete-from-help` is in emacs core.

Jdogzz:
@hlissner I was just talking with you on Discord about eshell autosuggest and it so happens this PR has a trivial bug fix for the function +eshell/search-history:

https://github.com/doomemacs/doomemacs/pull/6867/files#diff-7e15263b04ed0ca96c91de9ff37708579049b2d56c80b8465fdc41f39d59a9bcR166

Since this overall PR has been sitting stale for some time, would you like me to submit this one bug fix as a separate PR to whittle it down?

hlissner:
@Jdogzz I actually discovered that after speaking to you and already have the fix ready to push. It'll be up shortly.

hlissner:
I've incorporated some of the changes from the PR into the eshell module, but have opted to leave `tldr.el` to the user to install themselves, because of the external dependency.  I also couldn't get the capf backend working; something I'll revisit another time.

Thanks for bringing this to my attention, in any case!

PR #6876, opened 2022-10-05
fix(ligatures): fixed set-fontset-font not found

jessekelly881:
I was having an issue using emacs in the terminal while having ligatures enabled in the doom config. I could not any command that opened a menu(e.g. space + space, space + /, etc) I traced the issue to a missing check for set-fontset-font in the ligature init. Adding a check for whether the function was defined before calling it fixed the issue. 


hlissner:
Since there are quite a few `set-fontset-font` calls in and outside of Doom that don't check before hand, I've used a shim in a2b0294 to no-op its calls instead, which should resolve this issue. Let me know if that isn't the case. And thanks for your help!

PR #6877, opened 2022-10-06
fix(lsp): don't prompt twice when selecting root

juanedi:
When the lsp root can't be computed automatically (via `lsp--calculate-root`), lsp-mode will prompt the user to select it.

Doom's logging hook was calling `lsp--calculate-root` again to know what to log (assuming it would just compute it again based on directory structure), which means we would sometimes as the user to select the root twice.

Here we reimplement the logging using advice instead so we can piggyback on the first call and log whatever it returns.

I didn't find this reported as an issue and the fix was trivial once I found the problem so I decided to just go ahead and make a PR. Hope that works!

-----

### DIFF
--- a/modules/tools/lsp/+lsp.el
+++ b/modules/tools/lsp/+lsp.el
@@ -76,16 +76,16 @@ Can be a list of backends; accepts any value `company-backends' accepts.")
           (setq-local flycheck-checker old-checker))
       (apply fn args)))
 
-  (defadvice! +lsp-display-guessed-project-root (root)
-    "Log what LSP things is the root of the current project."
-    ;; Makes it easier to detect root resolution issues.
-    :filter-return #'lsp--calculate-root
-    (if root
-        (lsp--info "Guessed project root is %s" (abbreviate-file-name root))
-      (lsp--info "Could not guess project root."))
-    root)
-
-  (add-hook! 'lsp-mode-hook #'+lsp-optimization-mode)
+  (add-hook! 'lsp-mode-hook
+    (defun +lsp-display-guessed-project-root-h ()
+      "Log what LSP thinks is the root of the current project (if it was guessed automatically)."
+      ;; Makes it easier to detect root resolution issues.
+      (if lsp-auto-guess-root
+          (let (path (buffer-file-name (buffer-base-buffer)))
+            (if-let (root (lsp--calculate-root (lsp-session) path))
+                (lsp--info "Guessed project root is %s" (abbreviate-file-name root))
+              (lsp--info "Could not guess project root.")))))
+    #'+lsp-optimization-mode)
 
   (when (modulep! :completion company)
     (add-hook! 'lsp-completion-mode-hook
--- a/modules/tools/lsp/+lsp.el
+++ b/modules/tools/lsp/+lsp.el
@@ -76,15 +76,16 @@ Can be a list of backends; accepts any value `company-backends' accepts.")
           (setq-local flycheck-checker old-checker))
       (apply fn args)))
 
-  (add-hook! 'lsp-mode-hook
-    (defun +lsp-display-guessed-project-root-h ()
-      "Log what LSP things is the root of the current project."
-      ;; Makes it easier to detect root resolution issues.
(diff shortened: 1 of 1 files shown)


hlissner (changes requested):
I'd prefer that we use a hook, rather than advice, as the latter affects all call sites for `lsp--calculate-root`. There could be any number of these calls upstream (in lsp-mode) or downstream (in Doom or user config) where I don't want this logging to occur. Once per buffer, when lsp-mode activates, is the ideal target.

Could you see if this change (applied to the original hook) works for you?

```diff
   (add-hook! 'lsp-mode-hook
     (defun +lsp-display-guessed-project-root-h ()
       "Log what LSP things is the root of the current project."
       ;; Makes it easier to detect root resolution issues.
       (when-let (path (buffer-file-name (buffer-base-buffer)))
-        (if-let (root (lsp--calculate-root (lsp-session) path))
-            (lsp--info "Guessed project root is %s" (abbreviate-file-name root))
-          (lsp--info "Could not guess project root."))))
+        (let ((lsp-auto-guess-root t))
+          (if-let (root (lsp--calculate-root (lsp-session) path))
+              (lsp--info "Guessed project root is %s" (abbreviate-file-name root))
+            (lsp--info "Could not guess project root.")))))
     #'+lsp-optimization-mode)
```

juanedi:
Good point about using hooks!

The only problem I see with your proposed change is that we will be logging the auto-guessed root even for users that don't rely on auto-guessing. In my concrete example, I get prompted to select the root interactively but then see a different one logged. Going by the function's docs ("Log what LSP thinks is the root of the current project") that would be wrong and I suppose might even be misleading for troubleshooting purposes (although you would know better!).

Another thing we could do is to log only when the user actually has `lsp-auto-guess-root` enabled:

```elisp
(defun +lsp-display-guessed-project-root-h ()
      "Log what LSP thinks is the root of the current project (if it was guessed automatically)."
      ;; Makes it easier to detect root resolution issues.
      (if lsp-auto-guess-root
          (let (path (buffer-file-name (buffer-base-buffer)))
            (if-let (root (lsp--calculate-root (lsp-session) path))
                (lsp--info "Guessed project root is %s" (abbreviate-file-name root))
              (lsp--info "Could not guess project root.")))))
```

That means you wouldn't always see the root logged, but the cases you are missing are probably the less likely to need troubleshooting.

IMO it would be great to log always but getting the value by inspecting the state lsp-mode instead of calling `lsp--calculate-root` again. After digging through the code of `lsp-mode` I haven't been able to figure out exactly where the to get the root of the current buffer from. The session gives me the list of all roots but I don't see any mapping of buffers to roots anywhere.

What do you think?

ericdallo:
I missed this PR and ended up creating[ this one](https://github.com/doomemacs/doomemacs/pull/6877) with a different approach which IMO would be better, the TLDR is that we only log and check lsp--calculate-root if the lsp-auto-guess-root is enabled, which would make not prompting anymore WDYT @juanedi @hlissner ?
I'm ok doing on this PR as well

ericdallo:
One more update: lsp-mode already logs the project root, there is no need for this hook anymore :) 
I made the changes in my PR removing it

juanedi:
Thank you for resurfacing this, @ericdallo! (and sorry I dropped the ball and never came back here 😅)

Removing the hook seems reasonable if it's already logged by `lsp-mode`. It's not logging it for me but I might be using an older version of the library (I'm currently on [this commit](https://github.com/emacs-lsp/lsp-mode/commit/9faa49269)).

I did go ahead and make the changes I mentioned above anyways in case we do need to manually log for any reason. Otherwise I'd be happy to just close this PR. Up to you, @hlissner.

ericdallo:
> Removing the hook seems reasonable if it's already logged by lsp-mode. It's not logging it for me but I might be using an older version of the library (I'm currently on [this commit](https://github.com/emacs-lsp/lsp-mode/commit/9faa49269)).

I suggest you try latest, we had some fixes recently related to that

hlissner:
Thanks for both of your help. 8b414e6 should resolve this PR as well. Let me know if that isn't the case and I'll look into it again.

PR #6900, opened 2022-10-18
fix(highlight-indent-guides): for terminal users

Sleepful:
`(when (display-graphic-p))` is always `nil` for terminal emacs

I tested it on my machine, I run emacs as daemon in kitty terminal.

Before this change:
- start emacs
- `highlight-indent-guides-mode` is on (through `:ui indent-guides`)
- however there are *no* indent lines visible
- changing doom themes manually with `load-theme` allows indent guides to be shown. Manually calling `highlight-indent-guides-auto-set-faces` has the same effect.

After change:
- start emacs
- indent-guides are displayed on files without any issue

-----

- [ ] I am blindly checking these off. 🤣

### DIFF
--- a/modules/ui/indent-guides/config.el
+++ b/modules/ui/indent-guides/config.el
@@ -6,17 +6,13 @@
   (setq highlight-indent-guides-method 'character
         highlight-indent-guides-suppress-auto-error t)
   :config
-  (defun +indent-guides-init-faces-h (&rest _)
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
As of 9d4689d this should be resolved. Thanks for your help! Let me know if it doesn't fix the issue.

Issue #6714, opened 2022-08-23
next location (C-i) not working in C-o and C-i pair


kujohn:
### What did you expect to happen?

`C-i` would goto next location

### What actually happened?

`C-i` adds indents to the current line

### Describe your attempts to resolve the issue

n/a

### Steps to reproduce

1. Open document
2. Try to jump around few lines
3. `C-o` will jump back to previous lines
4. `C-i` will insert indents

### System Information

https://pastebin.com/dznRwGsF

hlissner:
Unfortunately, this is not an issue Doom can address, as it's an age-old characteristic of terminals, to interpret `C-i` as a `TAB`, which is what Emacs receive. An alternative keybind must be used, in this case, e.g. `C-M-,` -- or bind a new one, for example `g/`:

```elisp
(map! :n "g." #'better-jumper-jump-backward)
```

In GUI Emacs, Doom can and has [taught Emacs to distinguish the two input events](https://github.com/doomemacs/doomemacs/blob/master/lisp/doom-keybinds.el#L44-L57), but there's no way to do this for Emacs. Your best hope is to adopt a more modern terminal capable of remapping "TAB" to the correct event, but I'm not aware of one that exists.

Since there's nothing Doom can do here, I'll close this issue, but feel free to follow up here.

miles170:
I wonder if we can bind `evil-jump-forward` to `TAB` when using emacs under a terminal.
The following code works for either a GUI or a terminal.

```elisp
(map! :m (if (display-graphic-p) [C-i] "TAB") #'evil-jump-forward)
```

kujohn:
@miles170 thank you thank you. This is exactly what I was looking for. There's quiet a lot of muscle memory that's hard to unlearn.

hlissner:
The keybind above will break `C-i` for daemon+gui users, and has the disasterous side-effect of overriding most other `TAB` keybinds in various modes. Use it with caution.

The keybind you likely want to unbind is [this one](https://github.com/doomemacs/doomemacs/blob/master/modules/config/default/+evil-bindings.el#L47-L69):

```elisp
(map! :m [tab] nil)
```

PR #6915, opened 2022-10-26
fix(evil): evil-jump-forward not working under tty

miles170:
Fix `evil-jump-forward` does not bind to `TAB` when using emacs under a terminal.

See #6714.

-----

### DIFF
--- a/modules/editor/evil/config.el
+++ b/modules/editor/evil/config.el
@@ -424,7 +424,7 @@ directives. By default, this only recognizes C directives.")
 ;;   zu{q,w} - undo last marking
 
 (map! :v  "@"     #'+evil:apply-macro
-      :m  [C-i]   #'evil-jump-forward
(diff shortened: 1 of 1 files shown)


hlissner:
Changes to our modules are currently in [our do-not-PR list](https://discourse.doomemacs.org/t/do-not-pr/2406#doomemacsdoomemacs-2).

Besides that, this fix is likely based on an incorrect assumption, Doom uses `[C-i]` because such input events don't fire in terminal Emacs [by design](https://github.com/doomemacs/doomemacs/blob/master/lisp/doom-keybinds.el#L47-L67). Whatever this PR is solves, it is very likely to be accidental.

Plus, guarding this keybind would break the `[C-i]` keybinding for GUI Emacs users who use the daemon (`display-graphic-p` is always `nil` when a daemon starts up).

PR #6918, opened 2022-10-28
tweak: kill-current-buffer: suppress prompt for dired-mode buffers

daanturo:
In dired buffers, `(buffer-modified-p (current-buffer))` frequently returns t although no modifications have been made. Which makes `doom--switch-to-fallback-buffer-maybe-a's `y-or-n-p' is always asked.

Fix by checking if current `major-mode` isn't `dired-mode`. We may need a more generic checker like `derived-mode-p` if needed.

-----

### DIFF
--- a/lisp/doom-ui.el
+++ b/lisp/doom-ui.el
@@ -206,6 +206,7 @@ windows, switch to `doom-fallback-buffer'. Otherwise, delegate to original
            (let ((visible-p (delq (selected-window) (get-buffer-window-list buf nil t))))
              (unless visible-p
                (when (and (buffer-modified-p buf)
+                          (not (memq major-mode '(dired-mode)))
                           (not (y-or-n-p
                                 (format "Buffer %s is modified; kill anyway?"
                                         buf))))


lisp/doom-ui.el
```diff
@@ -206,6 +206,7 @@ windows, switch to `doom-fallback-buffer'. Otherwise, delegate to original
            (let ((visible-p (delq (selected-window) (get-buffer-window-list buf nil t))))
              (unless visible-p
                (when (and (buffer-modified-p buf)
+                          (not (member major-mode '(dired-mode)))
```
hlissner, line 209: Please replace `member` with `memq`, as it is optimal for symbol lists.

hlissner (changes requested):
Can you provide steps to reproduce this? In what case(s) does dired change its buffer's modified flag?

Also, could you reword the commit message to:

```
tweak: kill-current-buffer: suppress prompt for dired-mode buffers

In dired buffers, `(buffer-modified-p (current-buffer))` frequently
returns t although no modifications have been made. Which makes
`doom--switch-to-fallback-buffer-maybe-a's `y-or-n-p' is always asked.
```

(The scope should specify where the change occurs, not what package(s) it affects. Since the change is in core, there should be no scope)

daanturo:
> Can you provide steps to reproduce this? In what case(s) does dired change its buffer's modified flag?

I just try with
```bash
emacs -q ~ --eval "(print (buffer-modified-p (current-buffer)))"
# Emacs prints `t`
```
It's reproducible whether in vanilla Emacs or Doom. I happens for me nearly in every directory under `$HOME`.

daanturo:
Excuse me, I'm on Emacs 29, I'll investigate more and maybe report upstream.

aikrahguzar:
> Can you provide steps to reproduce this? In what case(s) does dired change its buffer's modified flag?

Generally this is due to some package modifying text properties without using `with-silent-modifications`. In dired's case marking files also does that. That can actually be an argument against this change since those marks might have been due to some reason but unmarking everything doesn't reset the flag and for me at least marks don't need to exist long term. If I am killing a dired buffer with marked files, it is because I am done with what I wanted to do.

daanturo:
This is indeed an Emacs 29.0 regression, now fixed https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=e9bdf3d4bd5aa2be7448b0f7a968af5abb058db0.

Unless there's a need to avoid the prompt for certain modes, I think we can close this PR.

hlissner:
> Unless there's a need to avoid the prompt for certain modes, I think we can close this PR.

In that case I think it can be left as is. Thanks for looking into it and for your help!

PR #6927, opened 2022-11-01
Add projectile support for Fossil and Pijul VCSes

zilti:


zilti (commented):
This pull request adds the Projectile support for Fossil and Pijul to Doom.

hlissner:
Thanks for the PR, but I must turn it down for a couple reasons:

- The project root markers for these VCs were intentionally omitted because I don't think they're common enough to be included by default (leaving it to users to add, since it is trivial to do so).
- I plan to drop Projectile (in favor of project.el) in the future, and I'd like to limit extension of its defaults for the time being.
- I plan to add flags for various VCSs to the `:emacs vc` module, where I'll centralize a more general effort to support these VCSs (e.g. pijul and fossil), the ones we already support (e.g. git, mercurial, svn), and others (e.g. p4 and darcs).

PR #6935, opened 2022-11-04
feat(javascript): add coffee-mode

offbyone:
I noticed when opening coffeescript files that coffee-mode is missing; I added the latest commit (pinned) for it here.

-----

### DIFF
--- a/modules/lang/javascript/README.org
+++ b/modules/lang/javascript/README.org
@@ -11,7 +11,8 @@ This module adds [[https://www.javascript.com/][JavaScript]] and [[https://www.t
 - Refactoring commands ([[doom-package:js2-refactor]])
 - Syntax checking ([[doom-package:flycheck]])
 - Browser code injection with [[doom-package:skewer-mode]]
-- Coffeescript & JSX support
+- Coffeescript ([[doom-package:coffee-mode]])
+- JSX support
 - Jump-to-definitions and references support ([[doom-package:xref]])
 
 ** Maintainers
@@ -29,8 +30,10 @@ This module adds [[https://www.javascript.com/][JavaScript]] and [[https://www.t
 - +tree-sitter ::
   Leverages tree-sitter for better syntax highlighting and structural text
   editing. Requires [[doom-module::tools tree-sitter]].
+- +coffeescript :: Enables support for [[https://coffeescript.org/][Coffeescript]].
 
 ** Packages
+- [[doom-package:coffee-mode]]* if [[doom-module::+coffeescript]]
 - [[doom-package:js2-refactor]]
 - [[doom-package:nodejs-repl]]
 - [[doom-package:npm-mode]]
--- a/modules/lang/javascript/config.el
+++ b/modules/lang/javascript/config.el
@@ -162,6 +162,9 @@
     ;; `emmet-mode' is used.
     emmet-expand-jsx-className? t))
 
+(use-package! coffee-mode
+  :when (modulep! +coffeescript)
+  :hook (coffee-mode . rainbow-delimiters-mode))
 
 ;;
 ;;; Tools
--- a/modules/lang/javascript/packages.el
+++ b/modules/lang/javascript/packages.el
@@ -4,6 +4,8 @@
 ;; Major modes
 (package! rjsx-mode :pin "b697fe4d92cc84fa99a7bcb476f815935ea0d919")
 (package! typescript-mode :pin "88f317f0b6aef8f8d232e912fdbc679799580c56")
+(when (modulep! +coffeescript)
+  (package! coffee-mode :pin "35a41c7d8233eac0b267d9593e67fb8b6235e134"))
 
 ;; Tools
 (package! js2-refactor :pin "a0977c4ce1918cc266db9d6cd7a2ab63f3a76b9a")


iyefrat (changes requested):
Can you put this behind a `+coffeescript` flag? I don't think coffeescript is used widely enough to be on by default.

(heads up, this PR might be closed before it gets merged due to a pending repo split. If this is the case you should re open it on the `modules` repo once that's a thing)

offbyone:
> Can you put this behind a `+coffeescript` flag? I don't think coffeescript is used widely enough to be on by default.

Done.

iyefrat (changes requested):
few points:
1. I forgot to say last time, but please add the package to the list of packages in the readme
2. you added the configuration of the package under `+tree-sitter` by mistake
3. you also need to flag-gate the package install in `packages.el`

modules/lang/javascript/README.org
```diff
@@ -29,8 +29,12 @@ This module adds [[https://www.javascript.com/][JavaScript]] and [[https://www.t
 - +tree-sitter ::
   Leverages tree-sitter for better syntax highlighting and structural text
   editing. Requires [[doom-module::tools tree-sitter]].
+- +coffeescript ::
+  [[https://coffeescript.org/][Coffeescript]] is a language that compiles into javascript. This flag will installs and
```
elken, line 33: *install and enable

modules/lang/javascript/README.org
```diff
@@ -29,8 +29,12 @@ This module adds [[https://www.javascript.com/][JavaScript]] and [[https://www.t
...
+  enables `coffee-mode`.
 
 ** Packages
+- doom-package:coffee-mode if [[doom-module::+coffeescript]]
```
elken, line 37: [[doom-package:coffee-mode]]

offbyone: Man... it is _hard_ to tell which things you need to add in org-mode sometimes :) 

The text for that in my editor looked _exactly_ the same as the others. I had to actually `find-file-literally` to see what the difference was.

modules/lang/javascript/README.org
```diff
@@ -29,8 +29,12 @@ This module adds [[https://www.javascript.com/][JavaScript]] and [[https://www.t
 - +tree-sitter ::
   Leverages tree-sitter for better syntax highlighting and structural text
   editing. Requires [[doom-module::tools tree-sitter]].
+- +coffeescript ::
+  [[https://coffeescript.org/][Coffeescript]] is a language that compiles into javascript. This flag will install and
+  enables `coffee-mode`.
```
iyefrat, line 34:
```suggestion
- +coffeescript ::
 Enables support for [[https://coffeescript.org/][Coffeescript]].
```

Sorry to nitpick, but doom packages have a special syntax (also, \`foo\` is markdown verbatim, regular or verbatim is =foo=, and doom's docs have a separate thing), but either way I think that explaining this here is redundant with the `Packages` section

modules/lang/javascript/README.org
```diff
@@ -11,7 +11,7 @@ This module adds [[https://www.javascript.com/][JavaScript]] and [[https://www.t
 - Refactoring commands ([[doom-package:js2-refactor]])
 - Syntax checking ([[doom-package:flycheck]])
 - Browser code injection with [[doom-package:skewer-mode]]
-- Coffeescript & JSX support
+- Coffeescript & JSX support ([[doom-package:coffee-mode]])
```
iyefrat, line 14: These should be split to separate lines because `coffee-mode` only effects the coffeescript part (as far as i'm aware).

modules/lang/javascript/README.org
```diff
@@ -29,8 +29,12 @@ This module adds [[https://www.javascript.com/][JavaScript]] and [[https://www.t
 - +tree-sitter ::
   Leverages tree-sitter for better syntax highlighting and structural text
   editing. Requires [[doom-module::tools tree-sitter]].
+- +coffeescript ::
+  [[https://coffeescript.org/][Coffeescript]] is a language that compiles into javascript. This flag will install and
+  enables `coffee-mode`.
 
 ** Packages
+- [[doom-package:coffee-mode]] if [[doom-module::+coffeescript]]
```
iyefrat, line 37:
```suggestion
- [[doom-package:coffee-mode]]* if [[doom-module::+coffeescript]]
```
see the `ivy` module for examples of this syntax. i don't think it's documented in the repo yet, but i'm not sure maybe i missed it.

iyefrat (changes requested):
Hey, a few more points, sorry for the repeated revisions on such a small PR but I think they're important to have. Also, please squash everything to one commit once you're done.

modules/lang/javascript/README.org
iyefrat: hot tip: doing `SPC m l t` will **t**oggle the **l**inks to appear in plaintext mode so they are easier to edit

modules/lang/javascript/README.org
elken: It's only recently gone live, I think documenting is a good shout

modules/lang/javascript/README.org
elken: Yeah coffee-mode doesn't do anything with jsx so they should be different lines

modules/lang/javascript/config.el
```diff
@@ -162,6 +162,9 @@
     ;; `emmet-mode' is used.
     emmet-expand-jsx-className? t))
 
+(when (modulep! +coffeescript)
```
elken, line 165: I think stylistically we prefer `:when` over wrapping in a `when` form

offbyone:
Is this suitable to merge?

elken:
After speaking to Henrik about this, it seems this has [already been removed](https://github.com/doomemacs/doomemacs/commit/ec95669710dead9e7e21dc429553a7026dbdefb5) one for being too trivial, so this is likely to be rejected for the same reason.

I'll leave it to Henrik for the final say though :)

hlissner:
Sorry for the late response. Unfortunately, I must turn this PR down. coffee-mode was intentionally removed from this module in 657b326e92a3 because I deemed the language too niche to be included by default, and its configuration is trivial enough for users to install themselves. Thanks for the PR, in any case!

PR #6943, opened 2022-11-14
bump: :lang scheme geiser

fosskers:
emacsmirror/geiser@550d57d347b6 -> emacsmirror/geiser@04dbdacfeca0 
emacsmirror/geiser-guile@b2d6f398e33c -> emacsmirror/geiser-guile@f5e82dc0f5a0

Currently, Scheme code can't be executed in `org-mode` due to a missing `run-geiser` function. This was fixed in a more recent commit, which this PR includes.

Fixes #6916 

-----
- [x] I searched the issue tracker and this hasn't been PRed before.
- [x] My changes are not on [the do-not-PR list](https://doomemacs.org/d/do-not-pr) for this project.
- [x] My commits conform to [the git conventions](https://doomemacs.org/d/git-conventions).
- [ ] I am blindly checking these off.
- [x] Any relevant issues or PRs have been linked to.

PR #6944, opened 2022-11-15
Fixes for Scheme/Geiser REPL integration

flatwhatson:
Fix the popup rule not matching Geiser buffers (buffer naming changed upstream).

Fix repl/eval behavior to use `geiser-eval-region` instead of the default `+eval/send-region-to-repl` which doesn't work properly with Geiser.

Replaced calls to deprecated `switch-to-geiser` to avoid warning messages.

Introduced a scheme-mode package which backports `scheme-mode` fixes from master for Emacs <29.

-----

### DIFF
--- a/modules/lang/scheme/autoload.el
+++ b/modules/lang/scheme/autoload.el
@@ -69,5 +69,5 @@ property lists and names starting with 'default'."
 (defun +scheme/open-repl ()
   "Open the Scheme REPL."
   (interactive)
-  (call-interactively #'switch-to-geiser)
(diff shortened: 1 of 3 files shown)


hlissner (approved):

fosskers:
Otherwise works as desired. Thank you!

flatwhatson:
#6943 isn't a prerequisite, confirmed this works with both the current & bumped geiser.

hlissner (dismissed):

hlissner:
LGTM. Thanks again for the help!

PR #6945, opened 2022-11-15
fix(org): Switch to bzg's up-to-date org-contrib repo

ncihnegn:


hlissner:
Thanks for the PR, but I must turn it down. I have avoided the source intentionally, due to occasional issues with sr.ht uptime, and prefer to have (the officially installed) packages be served from one infrastructure. 

That said, I realize emacs-straight/org-mode doesn't update very frequently. I plan to create a more up-to-date mirror for it and a number of other packages at https://github.com/doomelpa after 3.1 or 3.2 is finished. Thanks for bringing it to my attention, in any case.

PR #6955, opened 2022-11-19
fix(mu4e): Do not shadow `mu4e` arguments

bergheim:
`(mu4e)` accepts an optional argument, `background`, which is useful if you want to jump straight to `mu4e-headers-mode`. `=mu4e` does useful things in addition, but ignores this parameter. It should pass it on.

(In general, `=FUNC` commands should not mask optional arguments, if any.)

-----

### DIFF
--- a/modules/email/mu4e/autoload/email.el
+++ b/modules/email/mu4e/autoload/email.el
@@ -52,7 +52,7 @@ default/fallback account."
 (add-hook 'mu4e-main-mode-hook #'+mu4e-init-h)
 
 ;;;###autoload
-(defun =mu4e ()
+(defun =mu4e (&optional background)
   "Start email client."
   (interactive)
   (require 'mu4e)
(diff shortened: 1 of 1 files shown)


tecosaur:
To help me understand this, do you think you could give an example of when one might want to call `(=mu4e t)` over `(mu4e t)`?

bergheim:
Hi Teco, saur:

I have a bunch of keybindings set up which jump straight to the work inbox, personal inbox, or specialized views which has for instance threading on (which I do not have by default). The way I make this work is by starting `(mu4e t)` (in the background), and then immediately follow that up with a `(mu4e-headers-search-bookmark)` command. This takes me straight to my desired search. I never use the initial mu4e dashboard.

You can look at [a function that I have](https://github.com/bergheim/dotfiles/blob/025b8e6c9730f09fe1f9df6d3f2ceb9d69d413e1/.config/doom/config.el#L183) for an actual example of this (the dates in the docstring are not current). Without my PR, this will not work, since `(=mu4e)` does not accept `background`.

However, I do not think my personal anecdote should be relevant - unless there is a reason for it, `(=cmd)` commands in doom that wrap `(cmd)` should not change the original signature.

I could just switch to `(mu4e t)` but because of the nice buffer additions you added to `(=mu4e)`, I would like to keep using that instead.

hlissner (approved):

hlissner:
I may rethink our `=*` commands after v3.1 (as they're kind of clumsy as-is), but I think this is a sensible change for the time being. Thanks for your help!

PR #6956, opened 2022-11-20
fix(mu4e): issues preventing README fontification/loading

elken:
Fixes a couple of issues RE: fontification on the README in Emacs, as well as (un)relying on the nix module


elken:
Ah fine, set that to close this one then nbd

tecosaur:
Nah, thanks for raising this, I wouldn't have noticed this without you bringing it up.

PR #6959, opened 2022-11-24
docs(grammar): add nixos languagetool config

venikx:
The setup required for NixOS with LanguageTool is not as simple as installing LanguageTool and enabling the module. NixOS builds both the UI, CLI, and server jars. So, Emacs has to focus the correct script to start the CLI version.

The contribution docs indicate changes to docs itself should be done on the `rewrite-docs` branch, but this one hasn't been updated with the latest changes, and it seemed other PR have been merged onto master which were docs related.

-----
=> Green light was given on Discord by @hlissner, even though it's part of the do-not-pr list.

### DIFF
--- a/modules/checkers/grammar/README.org
+++ b/modules/checkers/grammar/README.org
@@ -33,6 +33,14 @@ It is available from either https://languagetool.org/ or your OS's package
 manager. E.g.
 - macOS: ~$ brew install languagetool~
 - Arch Linux: ~$ pacman -S languagetool~
+- NixOS:
+#+begin_src nix
+{
+  environment.systemPackages = with pkgs; [
+    languagetool
+  ];
+}
+#+end_src
 
 This module tries to guess the location of =languagetool-commandline.jar=. If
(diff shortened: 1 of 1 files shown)


Infinidoge:
I have a similar snippet in my own configuration, though in my case I set it up as 
```elisp
(use-package! langtool
  :config
  (setq langtool-bin "languagetool-commandline")
  )
```
I'm not sure which would be better, but thought to bring it up.

Emiller88 (approved):
Looks great to me! I've always done
```elisp
(after! langtool
  (setq langtool-bin "languagetool-commandline"))
```

venikx:
What would be the more idiomatic way?

hlissner:
Sorry for the late response. I've linked to the package on search.nixos.org (and will do the same wherever I refer to nix packages from now on), and added an extra `executable-find` lookup for `languagetool-commandline`, which should resolve the issues mentioned here (and so it works out of the box for users). Let me know if it gives you any trouble. Thanks for bringing it to my attention!

PR #6964, opened 2022-11-26
feat: add doom/copy-this-file-and-open

liuzhishan:
I found that `doom/copy-this-file` is not very convenient. When I copy a file, normally, the next step would be to edit it. So I have to execute a find-file command to open that new file. And many times I forgot to open the new file, just edit the origin buffer. Sometimes it take a while to notice it.

So I added a new function `doom/copy-this-file-and-open`. It copy the file of current buffer and then open the new file, then switch immediately to the new buffer.

It can also be bind to some convenient key:

```(map! :leader :desc "Copy this file and open" :n "f c" #'doom/copy-this-file-and-open)```

-----

### DIFF
--- a/lisp/lib/files.el
+++ b/lisp/lib/files.el
@@ -442,7 +442,9 @@ If FORCE-P, overwrite the destination file if it exists, without confirmation."
 
 ;;;###autoload
 (defun doom/copy-this-file-and-open (new-path &optional force-p)
-  "Copy current buffer's file to NEW-PATH, and then open the file."
+  "Copy current buffer's file to NEW-PATH, and then open the file.
+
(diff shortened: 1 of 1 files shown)


hlissner:
Sorry for the late response (I've been preoccupied with moving and work the past two months). I've changed `doom/copy-this-file` to open the copied file afterwards, in 2c65225. I didn't think it was worth having two separate commands for it, when `M-x copy-file` already exists for the old purpose. In any case, thanks for bringing it to my attention!

undergroundquizscene:
Thanks for fixing this! It has caught me out a few times, I’m glad it’ll no longer be an issue.

PR #6965, opened 2022-11-26
nit(doc): clone to .config/emacs by default

gagbo:
Since support moved to Emacs 27+, all versions should support cloning to .config/emacs.

Maybe it would be worth adding a command that moves or deletes legacy configuration locations.

The PR is mostly created as a glorified TODO item

References: doomemacs/.github#2

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,14 +111,14 @@ doctor` to check for any that you may have missed.
 
 # Install
 ``` sh
-git clone --depth 1 https://github.com/doomemacs/doomemacs ~/.emacs.d
-~/.emacs.d/bin/doom install
+git clone --depth 1 https://github.com/doomemacs/doomemacs ~/.config/emacs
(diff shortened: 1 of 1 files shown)


hlissner:
Agreed. From now on, $EMACSDIR will be ~/.config/emacs by default. There were a few other references to update so I integrated this PR into c1c966c. I'll update the user manual separately. Thanks for bringing it to my attention!

brorbw:
While I do think that its generally good to follow convention using `SPC SPC` in `.config` is now filled with files from the emacs dir. Besides moving the directory out of `~/.config` are there other possible solutions to ignore the emacs folder from `.config`

verhovsky:
This is a mistake, I moved my doom install to ~/.config/emacs but something keeps re-creating ~/.emacs.d and I end up with a fresh Emacs config without any doom stuff.

PR #5730, opened 2021-11-04
feat(clojure): Improve REPL icon modeline for cider

ericdallo:
This PR use the new cider hooks to improve the REPL icon on modeline. At the moment we just show the REPL icon as yellow if cider not connected and green if connected, with this PR we add a new intermediate state of REPL in sync:

Now, If the REPL is not connected we show the icon:
image

If the REPL is connected but nothing was evaluated or cider is still evaluating any code, we show:
image

If cider is in sync/not evaluating anything, we show:
image

This is a follow up of https://github.com/clojure-emacs/cider/pull/3084 and https://github.com/clojure-emacs/cider/pull/3089

Another gif showing the new feature:
cider-modeline-eval

PR #6979, opened 2022-12-01
fix(clojure): undefined face on +light modeline

ymarco:
Introduced in #5730. `breakpoint-disabled` is defined in `gdb-mi.el`, not sure why that specifically was used here, but `shadow` is better than spamming `*Messages*`.


hlissner (approved):

hlissner:
Thanks again for your help!

PR #6982, opened 2022-12-06
feat(lookup): add yandex search providers

Layerex:


sirenko:
@Layerex why would you want to use Yandex's services?

wrt: the PR
IMO, it's a bad idea to give users of an open-source project a possibility to (shoot themselves in the foot to) send their requests to a search engine, who's [parent company is about to be spear-headed by a former Putin's finance minister](https://www.yahoo.com/now/putin-ally-kudrin-yandex-role-134534872.html) while [they're looking for ways to escape imposed sanctions](https://fortune.com/2022/11/25/yandex-leaves-russia-ukraine-invasion-sanctions-search-aleksei-kudrin/).

Layerex:
> Why would you want to use Yandex's services?

- Superior image search
- Better indexing of Russian segment of the Internet
- Fine maps

Yandex is widely used. Free software is about choice. Users themselves should choose, if they want to use Yandex or not.

hlissner:
This has given me something to think about, regarding privacy and our search providers. While I value user choice, I value privacy just as much, and the potential to send data (especially by accident) to providers we don't like, know, or agree with (including Google) is something I'd like to take more seriously. How, exactly, needs more thought. Perhaps by gating them behind more flags, like `+yandex`/`+ru`, `+baidu`/`+ch`, `+google`, and so on?

Anyhow, this needs more thought, so I will close this PR for now and will revisit it once I've moved our modules out to https://github.com/doomemacs/modules in preparation for v3. Thanks for bringing it to my attention, in any case.

PR #6986, opened 2022-12-08
fix(docker): load docker-tramp conditionnally

gagbo:
Trying to use `docker-tramp` in Emacs in version 29 or more triggers a warning[^1] about `docker-tramp` being deprecated in favour of `tramp-container`, apparently stopping the loading of modules altogether. This lead at least notably to `:config default` not being loaded, so no binding is available in my sessions without that fix (EDIT: I'm finally not sure, couldn't repro the "no modules loading" part on other computers)

[^1]: https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/net/tramp-compat.el?h=emacs-29#n53

```lisp
(with-eval-after-load 'docker-tramp
  (warn (concat "Package `docker-tramp' has been obsoleted, "
		"please use integrated package `tramp-container'")))
```

### DIFF
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -185,7 +185,7 @@ orderless."
          ("C-x C-d" . consult-dir)
          ("C-x C-j" . consult-dir-jump-file))
   :config
-  (when (modulep! :tools docker)
+  (when (and (not EMACS29+) (modulep! :tools docker))
     (defun +vertico--consult-dir-docker-hosts ()
       "Get a list of hosts from docker."
       (when (require 'docker-tramp nil t)
--- a/modules/tools/docker/packages.el
+++ b/modules/tools/docker/packages.el
@@ -2,5 +2,6 @@
 ;;; tools/docker/packages.el
 
 (package! docker :pin "44f0bbec9b3deb027d17f4c10d8ec4106ed89dfb")
-(package! docker-tramp :pin "930d7b46c180d8a13240a028c1b40af84f2a3219")
+(unless EMACS29+
+  (package! docker-tramp :pin "930d7b46c180d8a13240a028c1b40af84f2a3219"))
 (package! dockerfile-mode :pin "b63a3d12b7dea0cb9efc7f78d7ad5672ceab2a3f")


avelino:
```
⛔ Warning (emacs): Package ‘docker-tramp’ has been obsoleted, please use integrated package ‘tramp-container’
```

We must not use the package `tramp-container`?

gagbo:
Maybe, I don't know what it does really (I can take a guess sure, but I've never used tramp so 🤷 ). This really only gets rid of the warning, I don't know if other things are necessary to integrated tramp-container

ReneFroger:
I have the same issue as well since upgrading to Emacs 30. What did you specifically to get rid of the warning, if I might ask, @gagbo ?

snizovtsev:
`docker.el` still [require](https://github.com/Silex/docker.el/blob/44f0bbec9b3deb027d17f4c10d8ec4106ed89dfb/docker.el#L7) `docker-trump` unless bumped to cc0046e. Here is my version:
```diff
-(package! docker :pin "44f0bbec9b3deb027d17f4c10d8ec4106ed89dfb")
-(package! docker-tramp :pin "930d7b46c180d8a13240a028c1b40af84f2a3219")
+(package! docker :pin "cc0046e6a557dce0ccc4108dd22e04f21ba8b0dc")
+(unless (>= emacs-major-version 29)
+  (package! docker-tramp :pin "930d7b46c180d8a13240a028c1b40af84f2a3219"))
```

gagbo:
> I have the same issue as well since upgrading to Emacs 30. What did you specifically to get rid of the warning, if I might ask, @gagbo ?

The page here is a PR page, so you can go to the ["Files changed" tab](https://github.com/doomemacs/doomemacs/pull/6986/files) to see the change

offbyone:
While I'm okay to work off a branch occasionally here, this would be really nice to get merged. Is there anything one could do to assist with that?

elken:
https://github.com/doomemacs/doomemacs/blob/818efb263fdb5f564e9e37f79bce17d50b559888/modules/completion/vertico/config.el#L190 also needs adjusting

hlissner:
The module needed a larger bump, so I did have of this PR in d41cf4e and will bump `:lang haskell` separated in the next few hours. Thanks for bringing it to my attention, in any case!

hlissner:
I merged this as part of a larger bump in d41cf4e. Thanks again for your help! 

> also needs adjusting

Forgot the last bit; pushed it in 1b9f93859.

ReneFroger:
When starting with an empty configuration and adding the following line: 

    (use-package! helm-projectile
        :custom
        (helm-projectile-on))

Save it and restart Emacs, then I get this warning too. It's caused by that only one line, the rest of my configuration is empty. This puzzles me because I'm using helm-projectile only in my configuration. When I remove the use-package declaration of helm-projectile, this warning doesn't come up anymore.

I think this is caused by the same issue, but I can't figure out how I could adjust the helm-projectile source code for Doom Emacs. I tried to look into variables of helm-projectile which would allow me to disable docker-tramp, but alas, there are no references of this.

What would be the best way to approach this, to get rid of the warning?

PR #6992, opened 2022-12-15
module: add :lang cc +nortags to allow a cleaner cc-mode without rtags/irony

xuning97:
Users can now choose among 
1. rtags/irony
2. lsp
3. no preconfigured tags or lsp

With their previous :lang cc configuration compatible
-----


hlissner:
Sorry for the late response: unfortunately, this PR rests on a few decisions I haven't made yet for this module, and am holding off on until v3 is done. Among them, is a plan to remove irony altogether, replace company with corfu, and add support for cscope/gnu globals (possibly with a flag for each of them and rtags). 

I also don't want to set a precedent for negated flags with a `+no...` prefix at this time, because v3 will introduce a native syntax for negated flags (e.g. `-lsp`), and don't want to introduce flags that will be removed so soon after.

In any case, I am closing this PR for now to prepare to move all our modules into a separate repo, but I will revisit this again after v3, when I can. In the meantime, users can disable these packages with:

```lisp
(package! irony :disable t)
(package! rtags :disable t)
```

Thanks for bringing this to my attention, however!

PR #7001, opened 2022-12-20
feat(default): add C-h/C-l evil keybinds for vertico

BenediktBroich:
Add C-h/C-l keybinding for directory up/directory enter or embark preview to vertico-map. The keybindings are currently not used. I try to replicate the C-h/C-l behavior from helm that i added in these pull requests: [pr](https://github.com/doomemacs/doomemacs/pull/6853) and was merged [here](https://github.com/doomemacs/doomemacs/commit/6887998c23d31430bde619e52d86de7b0ddf5c96)
 
Fixes 0
References 0
Replaces 0

-----

### DIFF
--- a/modules/completion/vertico/autoload/vertico.el
+++ b/modules/completion/vertico/autoload/vertico.el
@@ -116,6 +116,23 @@ Supports exporting consult-grep to wgrep, file to wdeired, and consult-location 
       (let ((embark-quit-after-action nil))
         (embark-dwim)))))
 
+;;;###autoload
+(defun +vertico/directory-enter ()
+  "Enter directory or embark preview on current candidate."
+  (interactive)
+  (if (>= vertico--index 0)
+      (if (and (let ((cand (vertico--candidate)))
+                 (or (string-suffix-p "/" cand)
+                     (and (vertico--remote-p cand)
+                          (string-suffix-p ":" cand))))
+               (not (equal vertico--base ""))
+               (eq 'file (vertico--metadata-get 'category)))
+          (vertico-insert)
+        (unless (bound-and-true-p consult--preview-function)
+          (save-selected-window
+            (let ((embark-quit-after-action nil))
+              (embark-dwim)))))))
+
 (defvar +vertico/find-file-in--history nil)
 ;;;###autoload
 (defun +vertico/find-file-in (&optional dir initial)
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -216,7 +216,10 @@
          "C-j"   #'vertico-next
          "C-M-j" #'vertico-next-group
          "C-k"   #'vertico-previous
-         "C-M-k" #'vertico-previous-group)))
+         "C-M-k" #'vertico-previous-group
+         (:when (modulep! :editor evil +everywhere)
+          "C-h"  #'vertico-directory-up
+          "C-l"  #'+vertico/directory-enter))))
 
 
 ;;; :ui


modules/completion/vertico/autoload/vertico.el
```diff
@@ -116,6 +116,23 @@ Supports exporting consult-grep to wgrep, file to wdeired, and consult-location
       (let ((embark-quit-after-action nil))
         (embark-dwim)))))
 
+;;;###autoload
+(defun +vertico/directory-enter ()
+  "Enter directory or embark preview on current candidate."
```
iyefrat, line 121: What is the reason for using this custom function over the built in `vertico-directory-enter`?

modules/completion/vertico/autoload/vertico.el
```diff
@@ -116,6 +116,23 @@ Supports exporting consult-grep to wgrep, file to wdeired, and consult-location
...
+                 (or (string-suffix-p "/" cand)
+                     (and (vertico--remote-p cand)
+                          (string-suffix-p ":" cand))))
+               (not (equal vertico--base ""))
+               (eq 'file (vertico--metadata-get 'category)))
+          (vertico-insert)
+        (unless (bound-and-true-p consult--preview-function)
+          (save-selected-window
+            (let ((embark-quit-after-action nil))
+              (embark-dwim)))))))
```
iyefrat, line 134: don't really love the use of `embark-dwim` here, makes it harder to reason what this function is meant to be doing.

modules/config/default/+evil-bindings.el
```diff
@@ -216,7 +216,10 @@
          "C-j"   #'vertico-next
          "C-M-j" #'vertico-next-group
          "C-k"   #'vertico-previous
-         "C-M-k" #'vertico-previous-group)))
+         "C-M-k" #'vertico-previous-group
+         (:when (modulep! :editor evil +everywhere)
+          "C-h"  #'vertico-directory-up
+          "C-l"  #'+vertico/directory-enter))))
```
iyefrat, line 222: I know this PR got accepted for helm, but binding something on `C-h` in a vertico session does have a side effect, which is losing the ability to `C-h k` in a vertico completion session, which is pretty useful. Not really sure which way that should go, Henrik will weight in when he gets to this PR.

iyefrat (changes requested):

modules/completion/vertico/autoload/vertico.el
BenediktBroich: So `vertico-directory-enter` exits the search buffer. This function allows you to complete the directory and preview files.

modules/config/default/+evil-bindings.el
BenediktBroich: True. :)

modules/completion/vertico/autoload/vertico.el
```diff
@@ -116,6 +116,23 @@ Supports exporting consult-grep to wgrep, file to wdeired, and consult-location
       (let ((embark-quit-after-action nil))
         (embark-dwim)))))
 
+;;;###autoload
+(defun +vertico/directory-enter ()
+  "Enter directory or embark preview on current candidate."
+  (interactive)
+  (if (>= vertico--index 0)
```
agzam, line 123: FYI, lispers prefer to use `<` because prefix notation allows multiple arguments and can be confusing.

```suggestion
  (if (<= 0 vertico--index)
```

hlissner:
After a couple tweaks I've merged this into 890f200. The conditional keybinds are a bit verbose, but I'll revisit its elegance after v3.

> I know this PR got accepted for helm, but binding something on C-h in a vertico session does have a side effect, which is losing the ability to C-h k in a vertico completion session, which is pretty useful. Not really sure which way that should go, Henrik will weight in when he gets to this PR.

To help soften the blow, I've made C-h (and C-l) conditional on whether the current session is a file command. That said, in situations where C-h is remapped, you can always use F1 instead.

> don't really love the use of embark-dwim here

I agree with that. Since I plan to dig into embark's guts post-v3, I'll see if I can make the embark call there more explicit then, unless another PR can beat me to it.

-----

Anyhow, thanks for the PR and input!

PR #7027, opened 2023-01-04
tweak(default): Allow disabling OS-specific key bindings

jaybosamiya:
Adds a module flag `+no-OS-specific-bindings` to `default` that allows disabling the automatically assigned keybindings on MacOS. I did not notice any Windows-specific keybindings, so this currently only disables the MacOS specific bindings, but I have named the flag to try to maintain forward compatibility _if_ Windows stuff is needed too.

The intention with having it be a _negative_ flag (i.e., `+no-....`) is so that it stays backwards compatible with people's setups, and folks should remain unimpacted by this change, _unless_ they explicitly want to disable the automatically assigned OS-specific keybindings.

-----

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -238,7 +238,7 @@
        :g "M-8"   #'+workspace/switch-to-7
        :g "M-9"   #'+workspace/switch-to-8
        :g "M-0"   #'+workspace/switch-to-final
-       (:when IS-MAC
+       (:when (and IS-MAC (not (modulep! +no-OS-specific-bindings)))
         :g "s-t"   #'+workspace/new
         :g "s-T"   #'+workspace/display
         :n "s-1"   #'+workspace/switch-to-0
--- a/modules/config/default/README.org
+++ b/modules/config/default/README.org
@@ -21,6 +21,7 @@ This module provides a set of reasonable defaults, including:
 ** Module flags
 - +bindings :: ...
 - +smartparens :: ...
+- +no-OS-specific-bindings :: Disable introduction of any OS-specific keybindings that this module would otherwise introduce.
 
 ** Packages
 - [[doom-package:avy]]
--- a/modules/config/default/config.el
+++ b/modules/config/default/config.el
@@ -295,7 +295,7 @@ Continues comments if executed from a commented line. Consults
   (define-key tabulated-list-mode-map "q" #'quit-window))
 
 ;; OS specific fixes
-(when IS-MAC
+(when (and IS-MAC (not (modulep! +no-OS-specific-bindings)))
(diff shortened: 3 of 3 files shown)


hlissner:
Thanks for the PR. I must turn it down for these two reasons:

1. I have plans for introducing negated flags in v3.0 (prefixed with a `-` rather than `+`) and I want to avoid introducing a precedent for a `+no-` prefix.
2. Doom only binds OS keys for macOS, and will never bind them for Windows or Linux. That's because MacOS has clear cut conventions for that modifier (e.g. `cmd+t` = new tab, `cmd+s` = save, `cmd+c` = copy, etc), and this extends that into Emacs only so far as those conventions reach. As such I consider it a "fix" rather than reasonable defaults. As this does not conflict with Emacs defaults, user config, or packages; it imposes no hit on performance; and is executed early enough that they are trivially re(un)bind, I don't think this warrants a flag.

I may revisit this once v3.0 hits, but I'm preparing to move modules into a separate repo, so I will close this for now. Thanks for bringing it to my attention, in any case!

jaybosamiya:
Thanks @hlissner for taking a look at it.

Fair reasons for turning it down. However, for someone who uses Cmd,Opt as Meta,Super respectively (since that leads to key positions on the keyboard more closely matching other OSs, maintaining muscle memory, rather than the reverse), the regular MacOS conventions don't apply within Emacs anyways (since they would clash too heavily: eg, I expect Cmd-t = M-t = transpose-words, rather than new-tab).

I agree that for most people, this setup is a "fix" rather than just reasonable defaults; but some someone with (say) my particular setup, it is something that they'll have to literally disable every single changed key, and also keep track of any new keys that may potentially be added to that list in the future, and switch those out too.

As a few alternatives:
1. Negative flags (with the `-`), as you suggested
2. Wrap up mac-specific things into a `(defun ...)`, in which case one can disable it from executing via custom user config (eg: via an advice), without needing to expose any specific flag, negative or positive.

I'm happy to make a PR for (2) if you think that'd be acceptable.

hlissner:
It sounds like the `{mac,ns}-command-modifier` and `{mac,ns}-option-modifier` variables are what you're asking for. Doom sets them [here](https://github.com/doomemacs/doomemacs/blob/master/lisp/doom-keybinds.el#L36-L42) to ensure command = super and option = meta.

By binding OS keys to super (rather than meta), we avoid overwriting Emacs' default keys, and it makes it easy to swap them out and restore the behavior you're asking for:

```elisp
(setq ns-command-modifier 'meta
      ns-option-modifier 'alt)

;; If you're using the emacs-mac build off homebrew, set `mac-command-modifier'
;; and `mac-option-modifier' instead.
```

`C-h v ns-command-modifier` will tell you what values these accept.

PR #7030, opened 2023-01-05
fix(lib): don't overwrite kill-ring on doom/backward-kill-word

alek-sys:
That's a bit tricky to explain, but here it is. I'm on macOs Ventura 13.1 (not that it matters), Emacs is
```
GNU Emacs     v27.2            nil
Doom core     v3.0.0-pre       HEAD -> master e96624926 2023-01-01 21:55:13 -0500
Doom modules  v22.10.0-pre     HEAD -> master e96624926 2023-01-01 21:55:13 -0500
```

When I copy something outside of Emacs (terminal / Safari / whatever) and then then call `doom/backward-delete-word` (usually via `c-w` hotkey or `m-backspace` in my case) before I paste (yank) right afterwards, instead of what I copied before I get some rubbish from the end of the current `kill-ring`. Not an expert in Emacs, but it does look to me that even though `kill-ring` is overridden in `let` block and not actually modified, the pointer is still modified which is causing this weirdness.

So my attempt to fix this is to not only override `kill-ring` variable, but also `kill-ring-yank-pointer`.

-----

### DIFF
--- a/lisp/lib/text.el
+++ b/lisp/lib/text.el
@@ -221,7 +221,7 @@ line to beginning of line. Same as `evil-delete-back-to-indentation'."
 (defun doom/delete-backward-word (arg)
   "Like `backward-kill-word', but doesn't affect the kill-ring."
   (interactive "p")
-  (let (kill-ring)
+  (let ((kill-ring nil) (kill-ring-yank-pointer nil))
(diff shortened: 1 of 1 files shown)


hlissner:

PR #7038, opened 2023-01-10
fix: correct name of doom-after-modules-init-hook

amosbird:
The old name is doom-after-init-modules-hook, should be changed to `doom-after-modules-init-hook`.

-----


hlissner (approved):

hlissner:

PR #7079, opened 2023-02-13
:lang clojure module refactor

elken:
- Update the README
- Better gate clj-refactor to not get in the way of LSP
- Add support for jet & neil

### DIFF
--- a/modules/lang/clojure/config.el
+++ b/modules/lang/clojure/config.el
@@ -168,6 +168,8 @@
             "C"  #'cider-connect-cljs
             "m"  #'cider-macroexpand-1
             "M"  #'cider-macroexpand-all
+            "j"  #'jet
+            "f"  #'neil-find-clojure-package
             (:prefix ("d" . "debug")
              "d" #'cider-debug-defun-at-point)
             (:prefix ("e" . "eval")
@@ -259,3 +261,14 @@
   :when (and (modulep! :checkers syntax)
              (not (modulep! +lsp)))
   :after flycheck)
+
+
+(use-package! neil
+  :commands (neil-find-clojure-package)
+  :config
+  (setq neil-prompt-for-version-p nil
+        neil-inject-dep-to-project-p t))
+
+
+(use-package! jet
+  :commands (jet))
--- a/modules/lang/clojure/config.el
+++ b/modules/lang/clojure/config.el
@@ -7,6 +7,8 @@
 ;; it should have a lower threshold too.
 (add-to-list 'doom-large-file-size-alist '("\\.\\(?:clj[sc]?\\|dtm\\|edn\\)\\'" . 0.5))
 
+(defvar +clojure-load-clj-refactor-with-lsp nil
+  "Whether or not to include clj-refactor along with clojure-lsp.")
 
 ;;
 ;;; Packages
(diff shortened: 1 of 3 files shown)


ericdallo (commented):
Awesome!

jcf:
It's exciting to see work done to improve the built-in Clojure support in Doom. I have a lot of customisation in my private configuration that I've shared with colleagues in the past, and seeing more open collaboration is commendable. Thank you for sharing!

Having reviewed the changes here, I can see some additional documentation, two new bindings, and some tweaks to when LSP or clj-refactor is used.

Regarding the addition of the packages…

I understand Eric's excitement at seeing his Jet package being included, and using Michiel Borkent's Jet to transform data could be helpful outside the Clojure ecosystem. That said, I've not found a use for it in all my years of Clojure development, so I wonder if this package should be installed for every Doom user wanting to write Clojure.

Neil is another fantastic creation from Michiel. It overlaps with clj-refactor, a package I've seen installed by almost every Clojure developer using Emacs. Again, I'm not sure this is something every user of Doom needs to install when working with Clojure.

Perhaps I'm an outlier, and having these two packages upstreamed into Doom would centralise many people's work. I'd be interested to know how other people are using these packages and if there's something I'm missing out on.

Improvements to lookup handlers and navigation to implementations would be fantastic. I've encountered issues here numerous times, both with my Emacs configuration and I've seen others similarly affected (both with and without Doom). Do the changes here solve some or all of the issues around code navigation you're experiencing? That would be a massive quality of life improvement!

A better understanding of the motivation behind these changes and some description of the problems being solved would enable me to provide higher-quality feedback. I'd also like to politely request the changes to indentation be reconsidered to improve the signal-to-noise ratio (not everyone knows how to ignore whitespace) as things progress.

Thanks again for sharing!

elken:
> so I wonder if this package should be installed for every Doom user wanting to write Clojure
If the inclusion is undesired, it's possible to do `(disable-packages! jet)` in your private config, and it won't be pulled in. If reports stack in about the inclusion, I'm happy to revisit but I think it constitutes useful enough functionality.

> Neil is another fantastic creation from Michiel. It overlaps with clj-refactor, a package I've seen installed by almost every Clojure developer using Emacs. Again, I'm not sure this is something every user of Doom needs to install when working with Clojure.
Same as above, and the same can be said for clj-refactor itself. clojure-lsp supercedes it in a lot of ways, that's the unfortunate nature of modules. The inclusion of neil specifically adds a fractional amount of elisp currently, but that's almost certainly going to improve over time.

> Improvements to lookup handlers and navigation to implementations would be fantastic
What are these issues? With or without lsp?

> A better understanding of the motivation behind these changes and some description of the problems being solved
Nothing more than a touch up for now really, mostly documentation. I had ideas about pulling in clojure-ts-mode but it currently only handles treesit not emacs-tree-sitter so that's on hold.

> I'd also like to politely request the changes to indentation be reconsidered to improve the signal-to-noise ratio
Hence the PR being draft, it's nowhere near polished

ericdallo:
> What are these issues? With or without lsp?

I mention some of then [here](https://emacs-lsp.github.io/lsp-mode/tutorials/clojure-guide/), some minor conflicts between cider/cljr/lsp that would be nice if doom fixes it out of the box

elken:
> > What are these issues? With or without lsp?
> 
> I mention some of them with LSP [here](https://emacs-lsp.github.io/lsp-mode/tutorials/clojure-guide/), some minor conflicts between cider/cljr/lsp that would be nice if doom fixes it out of the box

The lookup handlers? Both should be fixed

ericdallo (commented):
Looks pretty good to me, thanks for the effort of improving the clojure lang module

modules/lang/clojure/config.el
```diff
@@ -259,3 +261,14 @@
...
+
+(use-package! neil
+  :commands (neil-find-clojure-package)
+  :config
+  (setq neil-prompt-for-version-p nil
+        neil-inject-dep-to-project-p t))
+
+
+(use-package jet
+  :commands (jet))
```
hlissner, line 274: s/use-package/use-package!/

elken: Wow, thanks xD

hlissner (approved):

hlissner:

PR #7082, opened 2023-02-17
fix(mu4e): from address selection in mu 1.8+

tecosaur:
In brief, this should make `+mu4e-set-from-address-h` work again.


hlissner (approved):

hlissner:
This needed rebasing so I've done and merged it in c7c686c. Thanks for the help!

PR #7083, opened 2023-02-17
tweak(everywhere): add icon to right of modeline

tecosaur:
This adds an icon to indicate that Emacs Everywhere is being used, to ease at-a-glance identification. The "exit_to_app" material icon was chosen as it seems like a good fit for a window that will return you to the application it was invoked from. The padding on the right stops the icon from being right up against the very edge of the window.

It's a minor thing, but I think it's visually nice.

#### Sample image

image


iyefrat:
I like it!

hlissner (approved):

PR #7087, opened 2023-02-18
fix(popup): get all popup windows in all frames

taquangtrung:
Hi,

I'm running Doom Emacs in daemon mode with multiple frames opened.
If a popup window is opened in one frame, pressing `C-g` in another frame will not close that popup window.

I fixed this issue by enabling the function `+popup-windows` to get all popup windows in all frames.

Can you review and merge this PR if it is useful?

Thanks!

-----

### DIFF
--- a/modules/ui/popup/autoload/popup.el
+++ b/modules/ui/popup/autoload/popup.el
@@ -224,8 +224,12 @@ with ARGS to get its return value."
 
 ;;;###autoload
 (defun +popup-windows ()
-  "Returns a list of all popup windows."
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but I must turn it down: popup windows are intentionally scoped to frames to limit their effect and prevent operations occuring in frames that might be buried or outside the user's focus. I'll look into making this more configurable, but for now I'm prepping to move our modules to a separate repo, so I'll close this for now. Thanks for bringing it to my attention in any case!

PR #7096, opened 2023-02-21
Glamorous doc links

tecosaur:
This is not the final form of the Docs, but it allows for this work to be reorganised and then further improved down the line.

For fun, here are a few screenshots of the Org docs with these changes.

image

image

image

### DIFF
--- a/lisp/lib/docs.el
+++ b/lisp/lib/docs.el
@@ -16,36 +16,145 @@
 (defvar doom-docs-dir (file-name-concat doom-emacs-dir "docs/")
   "Where Doom's documentation files are stored. Must end with a slash.")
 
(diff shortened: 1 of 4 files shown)


elken:
Think this has been yak shaved to death now xD Keen to get this in

tecosaur:
I think there's still more yack shaving to go, but I think it's at least reached the point we can all agree it's an *improvement* :slightly_smiling_face:

elken:
> I think there's still more yak shaving to go, but I think it's at least reached the point we can all agree it's an _improvement_ slightly_smiling_face

It was always an improvement :P

hlissner (approved):

hlissner:
Great work, as usual! Thanks again for your help!

PR #7098, opened 2023-02-21
fix(vertico): docker-tramp => tramp-container

elken:
Works on my 29, I don't have easy access to an earlier version so would need to be verified there also


hlissner (approved):

hlissner:
Thanks again!

PR #7107, opened 2023-02-24
:tools rework REPL selection UI

fosskers:
This PR fixes a previously unreported bug, where not all available REPL-opening functions would be displayed via `SPC o r` in major modes that don't have an associated REPL (like the `*doom*` buffer). Further, this PR improves the UI a bit, displaying more human-readable names for the available modes.

Before:

repl-modes-before

After:

repl-modes

**NOTE:** There is still a strange bug here involving loading. In particular, from a __fresh__ restart of Doom, I'm able to open a Lua, Elisp, and Scheme repls, but not:

- Sh: `Symbol's value as variable is void: sh-shell`
- Lisp (`sly-mrepl`): `Couldn't find a valid REPL for +doom-dashboard-mode`
  - Indeed `sly-mrepl` (the symbol found in `+eval-repls`) is not a `commandp`, while `sly` is! __Is the `commandp` requirement too strict?__
- Haskell: `Symbol’s function definition is void: haskell-session-interactive-buffer`

The Lisp failure might be trivially fixable by a tweak to the `common-lisp` module, but I'm not sure about the other two. Thoughts?

Unrelated: Should we also attempt to put major mode icons next to the `completing-read` selections?

-----

### DIFF
--- a/modules/tools/eval/autoload/repl.el
+++ b/modules/tools/eval/autoload/repl.el
@@ -50,7 +50,16 @@
 
 (defun +eval-repl-known-repls ()
   "Yield the available repl functions as a list of symbols."
-  (seq-uniq (mapcar #'cl-second +eval-repls)))
+  (seq-uniq (mapcar (pcase-lambda (`(,mode ,fn . _)) (list mode fn)) +eval-repls)))
+
+(defun +doom-pretty-mode-name (mode)
+  "Convert a mode name into a variant nicer for human eyes."
+  (let ((mode (if (symbolp mode) (symbol-name mode) mode)))
+    (if (not (string-match "^\\([a-z-]+\\)-mode$" mode))
+        (error "Given string/symbol is not a major mode: %s" mode)
+      (string-join (split-string (capitalize (match-string-no-properties 1 mode))
+                                 "-")
+                   " "))))
 
 (defun +eval-repl-found-repls ()
   "Search the interned symbol list for functions that looks like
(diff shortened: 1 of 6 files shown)


modules/tools/eval/autoload/repl.el
```diff
@@ -48,42 +48,74 @@
...
       buffer)))
 
-(defun +eval-open-repl (prompt-p &optional displayfn)
-  (cl-destructuring-bind (_mode fn . plist)
-      (or (assq major-mode +eval-repls)
-          (list nil nil))
-    (when (or (not fn) prompt-p)
-      (let* ((choices (or (cl-loop for sym being the symbols
-                                   for sym-name = (symbol-name sym)
-                                   if (string-match "^\\(?:\\+\\)?\\([^/]+\\)/open-\\(?:\\(.+\\)-\\)?repl$" sym-name)
```
fosskers, line 58: This regex alone does not match all known repl symbols. The `loop` here comes up with a different list than what is known in `+eval-repls`. This PR unifies these two sources.

modules/tools/eval/autoload/repl.el
```diff
@@ -48,42 +48,74 @@
...
-        (when (bound-and-true-p evil-mode)
-          (call-interactively #'evil-append-line))
-        (when region
-          (insert region))
-        t))))
+(defun +eval-repl-known-repls ()
+  "Yield the available repl functions as a list of symbols."
+  (seq-uniq (mapcar (pcase-lambda (`(,mode ,fn . _)) (list mode fn)) +eval-repls)))
+
+(defun +doom-pretty-mode-name (mode)
```
fosskers, line 55: This could potentially live somewhere else, as it seems generally useful. Major modes do already seem to know their own "pretty name" (for display in the mode line), but this pretty version is only bound when the major mode is active, and is set to `mode-name`. So it's not generally searchable. This function instead determines the pretty name on-the-fly.

modules/tools/eval/autoload/repl.el
```diff
@@ -48,42 +48,74 @@
...
+    (if (not (string-match "^\\([a-z-]+\\)-mode$" mode))
+        (error "Given string/symbol is not a major mode: %s" mode)
+      (string-join (split-string (capitalize (match-string-no-properties 1 mode))
+                                 "-")
+                   " "))))
+
+(defun +eval-repl-found-repls ()
+  "Search the interned symbol list for functions that looks like
+repl openers."
+  (cl-loop for sym being the symbols
```
fosskers, line 67: I was really scratching my head last night over where the `symbols` symbol came from. I suppose it's a `loop` macro built-in that knows how to enumerate all the currently interned symbols?

modules/tools/eval/autoload/repl.el
```diff
@@ -48,42 +48,74 @@
...
+                     (user-error "Aborting"))))
+    (cl-second (assoc choice repls))))
+
+(defun +eval-repl-from-major-mode ()
+  "Fetch the repl associated with the current major mode, if there
+is one."
+  (pcase-let ((`(_ ,fn . ,plist) (assq major-mode +eval-repls)))
+    (list fn plist)))
+
+(defun +eval-open-repl (prompt-p &optional displayfn)
```
fosskers, line 101: Modularized to more easily test the separate logical components.

modules/tools/eval/autoload/repl.el
hlissner: Yes, that's correct.

fosskers:
https://github.com/joaotavora/sly/pull/572 will auto-fix the Lisp issue.

hlissner (approved):

hlissner:

PR #7112, opened 2023-02-24
fix(csharp): csharp-tree-sitter-mode: void-function error in Emacs 29+

passeride:
in Emacs 29.0.60 the function (csharp-tree-sitter-mode) is void and deprecated, and seems that tree-sitter works just fine without it. So i have Removed it

Fixes #0000
References #0000
Replaces #0000

-----

### DIFF
--- a/modules/lang/csharp/config.el
+++ b/modules/lang/csharp/config.el
@@ -49,7 +49,7 @@ or terminating simple string."
   :defer t
   :init
   (add-hook 'csharp-mode-local-vars-hook #'tree-sitter! 'append)
-  (if (version<= emacs-version "28")
+  (if (fboundp #'csharp-tree-sitter-mode)
       (add-to-list 'auto-mode-alist '("\\.cs\\'" . csharp-tree-sitter-mode))))
 
 
--- a/modules/lang/csharp/config.el
+++ b/modules/lang/csharp/config.el
@@ -48,8 +48,7 @@ or terminating simple string."
   :when (modulep! +tree-sitter)
   :defer t
   :init
-  (add-hook 'csharp-mode-local-vars-hook #'tree-sitter! 'append)
-  (add-to-list 'auto-mode-alist '("\\.cs\\'" . csharp-tree-sitter-mode)))
+  (add-hook 'csharp-mode-local-vars-hook #'tree-sitter! 'append))
 
 
 ;; Unity shaders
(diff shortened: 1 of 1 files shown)


passeride:
Emacs 29 might be out of scope, and i see that tree-sitter was included as of 29, if this is out of scope of doom, feel free to decline this PR.

passeride:
Added version check so this does not break anything on 28.X

hlissner:
I'm certain this is not a bug with Doom or the module. `csharp-tree-sitter-mode` is defined by [a third party package](https://github.com/emacs-csharp/csharp-mode/blob/master/csharp-tree-sitter.el#L397). The installed version of Emacs doesn't change whether it exists or not.

Rather, I suspect this is a bug with the new `loaddefs-gen.el` library in Emacs 29, because `csharp-tree-sitter-mode` is an autoloaded function, and I've been hearing bug reports here and there about various autoloads issues on Discord and [Github](https://github.com/doomemacs/doomemacs/pull/6482#issuecomment-1441509169).

As removing one autoloaded function is not a proper solution to this, I cannot accept this PR. Thanks for bringing to my attention, in any case.

modules/lang/csharp/config.el
```diff
@@ -49,7 +49,8 @@ or terminating simple string."
   :defer t
   :init
   (add-hook 'csharp-mode-local-vars-hook #'tree-sitter! 'append)
-  (add-to-list 'auto-mode-alist '("\\.cs\\'" . csharp-tree-sitter-mode)))
+  (if (version<= emacs-version "28")
```
hlissner, line 52: Use `fboundp` instead of a version check, because it is still possible to be on a build of Emacs 29 that predates this change. I.e.

```diff
-(if (version<= emacs-version "28")
+(if (fboundp #'csharp-tree-sitter-mode)
```

Since the function is autoloaded, it can be detected before the package is loaded.

hlissner (changes requested):
My bad! I forgot that csharp-mode had been absorbed into (much later commits of) Emacs 29. So this PR is correct. Just one small correction and it'll be ready for merging:

modules/lang/csharp/config.el
passeride: :+1: I updated it to recomendastion, thanks 4 feedback

hlissner:
Thanks for your help!

PR #7113, opened 2023-02-24
fix(lib): letf! in doom--help-search

iyefrat:
Slipped by manual testing due to consult not being loaded at the time
and doom--help-search going to default grep-find usage instead of
hitting the failiure.

Amend: 54c4340740ec


hlissner (approved):

hlissner:
Good catch. Thanks again!

PR #7127, opened 2023-03-02
feat(rss): allow disable auto elfeed-db-compact

merrickluo:
running elfeed-db-compact automatically when killing the *elfeed-search* buffer is a convenient way to keep the elfeed db files clean, but it can also take a lot of time to do so, which blocks emacs from doing anything, adding a variable to disable it.

on my setup, it would take almost 10s to kill the buffer, which is annoying, the default value of `+rss-auto-db-compact` is true, so the default behavior is unchanged.

-----


hlissner:
In the interest of avoiding introducing more variables than is necessary, I've moved the `elfeed-db-compact` call to `kill-emacs-hook` instead in f1f010f, where a brief, blocking delay should be more acceptable. Thanks for bringing it to my attention, in any case!

PR #7129, opened 2023-03-02
fix(company): hook company-abort when +childframe enabled

teeann:
Add hook for `company-abort` when `+childframe` option is enabled.

-----


hlissner (approved):

hlissner:
I don't think this `evil-normal-state-entry-hook` hack is necessary anymore (with or without +childframe), but I'll look into this further after v3.0. Thanks for your help, in any case!

PR #7136, opened 2023-03-06
fix(clojure): revert cljr-refactor change

elken:
Creates an odd regression where the localleader map isn't loaded. In
lieu of an actual fix, revert the change.

Amend: ca90c5e83caa
Fix: #7130

### DIFF
--- a/modules/lang/clojure/config.el
+++ b/modules/lang/clojure/config.el
@@ -7,8 +7,6 @@
 ;; it should have a lower threshold too.
 (add-to-list 'doom-large-file-size-alist '("\\.\\(?:clj[sc]?\\|dtm\\|edn\\)\\'" . 0.5))
 
-(defvar +clojure-load-clj-refactor-with-lsp nil
-  "Whether or not to include clj-refactor along with clojure-lsp.")
 
 ;;
 ;;; Packages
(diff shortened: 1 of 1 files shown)


elken:
Converted to draft as it might not be needed

elken:
Seems this is indeed no longer needed, it's triggered by setting `doom-localleader-key` in the wrong place and though it does highlight an issue with the package loading it it's been agreed that moving the `doom-localleader-key` to the right place is a good enough fix

hlissner:
doom-localleader-key _is_ meant to be set in config.el. This _is_ a problem with the clojure module eagerly loading `clj-refactor`; it just needs to be deferred, rather than eagerly loaded, so I'll look into that soon.

elken:
Oh, I found an older comment from you that said it had to be set in init.el. My bad!

hlissner:
> Oh, I found an older comment from you that said it had to be set in init.el. My bad!

No worries! That used to be the case before [this commit](https://github.com/doomemacs/doomemacs/commit/c9e7d2731792a8412a8f8a91723859fed747ca7f). The only caveat is that modules have to be careful not to bind (local)leader keys too early.

PR #7140, opened 2023-03-11
fix(clojure): load against the correct package

elken:
The cider library is loaded too late for clojure-mode to rely on, so we
load after the cider-mode package instead.

Fix: #7130

### DIFF
--- a/modules/lang/clojure/config.el
+++ b/modules/lang/clojure/config.el
@@ -39,7 +39,10 @@
                #'tree-sitter!)))
 
 
-(use-package! cider
+;; `cider-mode' is used instead of the typical `cider' package due to the main
+;; library being loaded only when is absolutely needed, which is too late for
+;; our purposes
+(use-package! cider-mode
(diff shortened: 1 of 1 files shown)


elken:
From Henrik: 

>  Modules shouldn't be eagerly binding localleader keys

So this PR will need adjustment still

modules/lang/clojure/config.el
```diff
@@ -274,8 +275,15 @@
   :commands (neil-find-clojure-package)
   :config
   (setq neil-prompt-for-version-p nil
-        neil-inject-dep-to-project-p t))
+        neil-inject-dep-to-project-p t)
+ (map! :map (clojure-mode-map clojurescript-mode-map clojurec-mode-map)
+       :localleader
+       "f"  #'neil-find-clojure-package))
```
hlissner, line 281: The indentation for this `map!` call is off by one.

hlissner (changes requested):

hlissner (approved):

hlissner:
Thanks again for your help!

PR #7146, opened 2023-03-13
feat(common-lisp): add project loading and testing commands

fosskers:
Doing so manually in a Sly REPL is otherwise time consuming and repetitive.

This is a breaking change due to the restoration of `SPC m t` as a testing prefix, as it is commonly used in other language modules. The previous binding to `SPC m t` has been moved to `SPC m T`.

-----

### DIFF
--- a/modules/lang/common-lisp/autoload/common-lisp.el
+++ b/modules/lang/common-lisp/autoload/common-lisp.el
@@ -17,3 +17,65 @@
                         ((> attempt 5) (error "Failed to start Slynk process."))
                         (t (recurse (1+ attempt))))))
       (recurse 1))))
+
+;;;###autoload
+(defun +lisp/load-project-systems ()
+  "Load all systems of the current Lisp project into Sly."
+  (interactive)
+  (thread-last (+lisp--project-asd-file)
+               (+lisp--systems-from-asd)
+               (mapcar (lambda (s) (format ":%s" s)))
+               (funcall (lambda (ss) (string-join ss " ")))
+               (format "(ql:quickload '(%s))")
+               (sly-interactive-eval)))
+
+(defun +lisp--project-asd-file ()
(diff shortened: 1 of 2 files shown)


modules/lang/common-lisp/config.el
```diff
@@ -146,7 +149,9 @@
           :desc "Fetch stickers"           "f" #'sly-stickers-fetch
           :desc "Replay stickers"          "r" #'sly-stickers-replay
           :desc "Add/remove sticker"       "s" #'sly-stickers-dwim)
-         (:prefix ("t" . "trace")
+         (:prefix ("t" . "test")
+          :desc "Test System" "s" #'+lisp/test-system)
+         (:prefix ("T" . "trace")
```
fosskers, line 154: This is the source of the breaking change.

modules/lang/common-lisp/autoload/common-lisp.el
```diff
@@ -17,3 +17,65 @@
...
+;;;###autoload
+(defun +lisp/reload-project ()
+  "Restart the Sly session and reload the current project."
+  (interactive)
+  (sly-restart-inferior-lisp)
+  (cl-labels ((recurse (attempt)
+                (sleep-for 1)
+                (condition-case nil
+                    (sly-eval "PONG")
+                  (error (if (= 5 attempt)
```
fosskers, line 71: There is a brief moment when Sly is asynchronously restarting that attempting to connect to / send Lisp will cause the whole thing to get quite angry. Waiting for even a single second beforehand seems to be enough.

hlissner:
I removed the `!` in the commit message since keybindings don't quite constitue "breaking changes", at least as I had intended, but I understand it's not well documented. That's something I'll fix once the new docs are up (and the modules are in their own repo). Thanks again for your help, in any case!

PR #7163, opened 2023-03-22
feat(vertico): search pattern after consult-line

agzam:
Reuse search pattern for Evil's `<n>` or `<N>` after consult-line

Very often, after searching for something with `consult-line`, I want to jump to the next/previous occurrences with `<n>|<N>`

### DIFF
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -175,7 +175,13 @@ orderless."
                          (lambda (x)
                            (eq (buffer-local-value 'major-mode x) 'org-mode))
                          (buffer-list)))))))
-    (add-to-list 'consult-buffer-sources '+vertico--consult-org-source 'append)))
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but sorry to say it comes at a bad time: I'm preparing to move our modules into [a separate repo](https://github.com/doomemacs/modules), so I'm [not currently accepting new PRs to our modules](https://discourse.doomemacs.org/do-not-pr) for the next week or so.

That said, I like the idea, so I'll definitely revisit this PR shortly afterwards. I'm uncertain where these advice belong though (maybe in a [conditional autoloads file](https://github.com/doomemacs/doomemacs/blob/master/modules/lang/python/autoload/conda.el#L2) or PRed to evil-collection-consult), and I'd like to extend it to consult-grep (plus its ripgrep and git-grep variants) and consult-imenu, but I'll think about it in the meantime.

PR #7169, opened 2023-03-24
feat(lua): recognize `.fenneldoc` files

fosskers:
This turns on Fennel syntax highlighting for `.fenneldoc` files.

### DIFF
--- a/modules/lang/lua/config.el
+++ b/modules/lang/lua/config.el
@@ -63,6 +63,7 @@ lua-language-server.")
 (use-package! fennel-mode
   :when (modulep! +fennel)
   :defer t
+  :mode ("\\.fenneldoc" . fennel-mode)
   :config
   (set-lookup-handlers! 'fennel-mode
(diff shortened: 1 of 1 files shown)


modules/lang/lua/config.el
```diff
@@ -63,6 +63,7 @@ lua-language-server.")
 (use-package! fennel-mode
   :when (modulep! +fennel)
   :defer t
+  :mode ("\\.fenneldoc" . fennel-mode)
```
fosskers, line 66: I believe this technically  matches on any file that ends with `.fenneldoc`, whereas the config file in question is always just named `.fenneldoc`. I tried a few different regex variations but none seemed to work, so I fell back to this.

hlissner:
I've merged this in 705a439 with some slight changes. Let me know if that causes you trouble. Thanks for bringing it to my attention!

fosskers:
Should be fine, thanks!

PR #7176, opened 2023-03-29
tweak(config): feature-gate company keybindings

LemonBreezes:
In anticipation of the Corfu module, this PR feature-gates the company-mode keybindings for non-Evil users. This is already done for Evil users. 

While feature-gating keybindings is on the do-not-PR list, I am hoping that an exception will be made on the basis that the company keybindings are currently feature-gated for Evil users but not for non-Evil users.

### DIFF
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -506,8 +506,9 @@
       "C-x K"       #'doom/kill-this-buffer-in-all-windows
 
       ;;; company-mode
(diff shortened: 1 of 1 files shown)


UndeadKernel (approved):
From my side, I'd say the change makes sense for the default emacs bindings.

hlissner (approved):

hlissner:
Good catch! I'll happily make an exception for this (and any case where we're breaking top-level keybinds). Thanks for your help!

PR #7181, opened 2023-04-03
tweak(lib): preserve cursor/window position on doom/sudo-this-file

YourFin:
When running `doom/sudo-this-file`, save the cursor and window start position prior to transitioning to the sudo'd buffer, and reapply them in the sudo'd version.

This change resolves my personal annoyance arising from situations like the following:
1. Open a config file owned by root (typically somewhere in /etc/)
2. Find a line I would like to edit buried deep in the file
3. Attempt to make a change, and have emacs warn me that the file is in readonly mode
4. Facepalm, run `doom/sudo-this-file`
5. Find my cursor at the top of the file
6. Annoyed, re-navigate my cursor to where it just was in the readonly buffer

I have attempted to preserve the original functionality when `doom/sudo-this-file` is called as a function - however, I see a strong argument that the the additional complexity in keeping the corner-case backwards compatibility is not worthwhile, and I am happy to cut the relevant machinery.

-----
- [x] I am _not_ blindly checking these off. :P

### DIFF
--- a/lisp/lib/files.el
+++ b/lisp/lib/files.el
@@ -480,15 +480,36 @@ If FORCE-P, overwrite the destination file if it exists, without confirmation."
   (find-file (doom--sudo-file-path file)))
 
 ;;;###autoload
-(defun doom/sudo-this-file ()
-  "Open the current file as root."
-  (interactive)
-  (find-file
-   (doom--sudo-file-path
-    (or buffer-file-name
-        (when (or (derived-mode-p 'dired-mode)
-                  (derived-mode-p 'wdired-mode))
-          default-directory)))))
+(defun doom/sudo-this-file (&optional copy-cursor-position)
+  "Open the current file as root.
+
+If COPY-CURSOR-POSITION is non-nil (as when called interactively),
+matches the cursor and window position in the sudo'd buffer with that of the
+current buffer. Additionally, push a mark ring entry for the old position of
+the sudo buffer should it be updating from a position other than the start of
+the buffer."
+  (interactive "p")
+  (let ((nonsudo-window-start (window-start))
+        (nonsudo-point (point))
+        (nonsudo-buffer (current-buffer))
+        (result (find-file
+                 (doom--sudo-file-path
+                  (or buffer-file-name
+                      (when (or (derived-mode-p 'dired-mode)
+                                (derived-mode-p 'wdired-mode))
+                        default-directory))))))
+    (when (and
+           copy-cursor-position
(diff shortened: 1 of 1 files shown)


lisp/lib/files.el
```diff
@@ -480,15 +480,33 @@ If FORCE-P, overwrite the destination file if it exists, without confirmation."
...
+  (interactive "p")
+  (let ((nonsudo-window-start (window-start))
+        (nonsudo-point (point))
+        (result (find-file
+                 (doom--sudo-file-path
+                  (or buffer-file-name
+                      (when (or (derived-mode-p 'dired-mode)
+                                (derived-mode-p 'wdired-mode))
+                        default-directory))))))
+    (when copy-cursor-position
```
YourFin, line 500: I am tempted to add `(not (buffer-modified-p))` as a conditional here, but it wouldn't be perfect. There's also a decent argument for checking if the buffers are identical up to `nonsudo-point`, but I figured I'd keep things simple by default.

lisp/lib/files.el
```diff
@@ -480,15 +480,33 @@ If FORCE-P, overwrite the destination file if it exists, without confirmation."
...
-                  (derived-mode-p 'wdired-mode))
-          default-directory)))))
+(defun doom/sudo-this-file (&optional copy-cursor-position)
+  "Open the current file as root.
+
+If COPY-CURSOR-POSITION is non-nil (as when called interactively),
+matches the cursor and window position in the sudo'd buffer with that of the
+current buffer. Additionally, push a mark ring entry for the old position of
+the sudo buffer should it be updating from a position other than the start of
+the buffer."
```
YourFin, line 490: Optionally:

`s/updating from a position other than the start of
the buffer./updating./`

hlissner:
Sorry for taking so long to get to this, I've been on the fence about the implementation. Yours is very clean, which I like, but I felt it would be better UX to lean on `save-place` here (the benefits of which would extend beyond these two commands), and to not rely on buffer comparisons (since it's reasonable to assume they're the same file, at least in `doom/sudo-this-file`'s case). I've implemented this in f18603e. Let me know if it gives you any trouble. Thanks for bringing it to my attention, in any case!

YourFin:
@hlissner No worries, I'm thrilled with this outcome! Completely agreed that your approach is superior, and successful nerd sniping is always a delight. 

With respect to the buffer comparison - I believe its inclusion was on grounds of conservatism. I wanted to keep proposed behavior to improvement only in the strictest sense. My hope was that it would be cut, though :)

PR #7199, opened 2023-04-14
fix(lib): doom-region-*: only use evil visual range in visual mode

taquangtrung:
Hi,

I found out that Doom Emacs get regions incorrectly for `evil-emacs-state-mode`, because `evil-visual-beginning` and `evil-visual-end` are used to get regions in emacs state mode.

I added additional conditions check to fix this issue.

Can you review and merge this PR?

Thanks!

-----


hlissner (approved):

hlissner:
Sorry for the late reply! This is a good catch. Thanks for fixing it!

PR #7215, opened 2023-05-02
feat(cli): added mid-upgrade hook

offbyone:
Add a hook to run after upgrading doom, but before sync.

This is to enable me to have a hook to upgrade my private doom modules,
which I only want to do when upgrading doom itself, to keep the two in
sync.

### DIFF
--- a/lisp/cli/upgrade.el
+++ b/lisp/cli/upgrade.el
@@ -15,6 +15,8 @@
 (defvar doom-upgrade-remote "_upgrade"
   "The name to use as our staging remote.")
 
+(defvar doom-upgrade-before-sync-hook ()
+  "Hooks run between upgrading Doom itself and performing an upgrading sync.")
 
 ;;
 ;;; Commands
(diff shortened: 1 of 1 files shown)


offbyone:
this is surprisingly hard to test, btw! :D

hlissner:
Thanks for the PR, but I must turn it down. The placement of this `run-hooks` call within the containing `cond` is strange and is treated as a clause of the containing `cond`. That said, since I am rewriting much of `doom upgrade` and `doom sync` under the hood, I'd rather not add (and therefore promise) new hooks at the moment.

For the time being, I suggest you use `doom-before-sync-hook` and use one of the following to check if the sync is happening from a `doom upgrade`:

- `(doom-cli-context-get doom-cli--context 'upgrading)`
- Or if you've updated to the latest Doom (containing 3bced4dbbe783767e10a47f9a4b57f32616f94bc), then `(get! 'upgrading)`.

Thanks for bringing this to my attention, in any case!

PR #7218, opened 2023-05-03
tweak(evil): make window-move-* cyclic

agzam:
Attempting to move far-right window further to the right, makes it appear on the left side of the frame.
Works similarly for other directions.

*I understand that this might be the subject of preference - someone likes it this way, others may not. Feel free to reject it.*

-----


hlissner (approved):

hlissner:
I've gone with a slightly different implementation, along with introducing a new `+evil-want-move-window-to-wrap-around` option, and the ability to pass the prefix arg to any of `+evil/window-move-*` commands to once-off invert `+evil-want-move-window-to-wrap-around`, which should fulfill the intention of this PR. Let me know if this implementation gives you any trouble. Thanks for bringing it to my attention, in any case!

PR #7219, opened 2023-05-04
Minor Eglot improvements

elken:
A couple of small changes that improve the Eglot experience.

### DIFF
--- a/modules/editor/evil/config.el
+++ b/modules/editor/evil/config.el
@@ -533,7 +533,11 @@ directives. By default, this only recognizes C directives.")
        (:after elfeed
         :map elfeed-search-mode-map
         :n "gr" #'elfeed-search-update--force
-        :n "gR" #'elfeed-search-fetch))
+        :n "gR" #'elfeed-search-fetch)
+       (:after eglot
+        :map eglot-mode-map
+        :nv "gd" #'+lookup/definition
+        :nv "gD" #'+lookup/references))
 
       ;; custom evil keybinds
       :nv "zn"    #'+evil:narrow-buffer
--- a/modules/tools/lsp/+eglot.el
+++ b/modules/tools/lsp/+eglot.el
@@ -4,8 +4,16 @@
   :commands eglot eglot-ensure
   :hook (eglot-managed-mode . +lsp-optimization-mode)
   :init
+  (defadvice! +eglot--ensure-available-mode (fn)
+    "Run `eglot-ensure' if the current mode has support."
+    :around #'eglot-ensure
+    (when (alist-get major-mode eglot-server-programs nil nil
+                     (lambda (modes key)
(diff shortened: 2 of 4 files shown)


aikrahguzar:
> A couple of small changes that improve the Eglot experience.

I think it would be a good idea to also add `markdwon-mode` to `packages.el` since otherwise eldoc doesn't render docs very well.

elken:
No need, it'll get pulled in if you have the markdown module enabled. It's a weak dependency and docs look fine without it.

image

aikrahguzar:
> No need, it'll get pulled in if you have the markdown module enabled. It's a weak dependency and docs look fine without it.
> 
> image

It is a weak dependency but I think quite important for a good experience. At least for haskell, I will get just
```haskell`
in the first line of docs without markdown-mode.

I use `(setq eldoc-echo-area-use-multiline-p nil)` so that the mini-window is not constantly resized which I find distracting. But without markdown-mode  the only thing I see is the opening delimiter of the haskell block.

By the way, I am seeing some fontification in the docs in your picture which makes me think that the markdown mode is involved in rendering those.

**Comment by elken** [comment]
It's a bare `emacs -Q` so there's nothing involved except eglot & clojure-mode

File: modules/editor/evil/config.el [hunk 1]
```diff
@@ -533,7 +533,11 @@ directives. By default, this only recognizes C directives.")
        (:after elfeed
         :map elfeed-search-mode-map
         :n "gr" #'elfeed-search-update--force
-        :n "gR" #'elfeed-search-fetch))
+        :n "gR" #'elfeed-search-fetch)
+       (:after eglot
+        :map eglot-mode-map
+        :nv "gd" #'+lookup/definition
+        :nv "gD" #'+lookup/references))
```
regadas, line 540: 🙌

hlissner:
Excellent work, as always! Thanks again!

yuanl:
advice around #'eglot-ensure may have issue here. All [javascript major mode define](https://github.com/joaotavora/eglot/blob/master/eglot.el#L230-L234) has another form like `(MAJOR-MODE-SYMBOL :LANGUAGE-ID ID)` in `eglot-server-programs`.

```lisp
(((js-mode :language-id "javascript")
  (js-ts-mode :language-id "javascript")
  (tsx-ts-mode :language-id "typescriptreact")
  (typescript-ts-mode :language-id "typescript")
  (typescript-mode :language-id "typescript"))
 . ("typescript-language-server" "--stdio"))
```

The adviced function will always return `nil` in js-mode or rjsx-mode which cause eglot-ensure can not be use in javascript. A tide server will started instead.

Deleting this advice is a workaround in my test.

hlissner:
@yuanl Should be fixed in 173842c.

PR #7231, opened 2023-05-11
module(chatgpt): add poc

brorbw:
This module adds integration with chatgpt via [chat.el](https://github.com/iwahbe/chat.el/blob/main/chat.el). 
I had several packages up for review and `chat.el` was the most simple one. I would like to try out `chatgpt-shell` which is why I left a future task to rewrite the module to be able to change the "backend". 

- https://github.com/iwahbe/chat.el
- https://github.com/emacs-openai/chatgpt
- https://github.com/xenodium/chatgpt-shell

Feel free to scrutinize as this is my first module. If this is of interest there should of course be made bindings for it elsewhere and not _in_ the module afaict from the convention in other modules.

Also feel free to ignore and close if this falls into the do-not-PR category. It wasn't clear to me since the list was a bit old and the information on the site seemed outdated.

-----
 this PR -->


rassie:
Not a review, just a couple of notes:

* ChatGPT is a trademark AFAIK, so it might be a good idea not to call the module `chatgpt`. Which is probably a good idea anyway, since replacing it with whatever is top of the breed in six months wouldn't require a rename then.
* There are certainly a myriad of different, but at the same time similar, options for ChatGPT integration, I've used https://github.com/karthink/gptel quite successfully and was somewhat surprised to to see it in the list of possible backends you provided. Would be nice if you at least considered it for this module.

brorbw:
I opted for `chat.el` because it offered a native implementation not using Python or other external packages. The PR is only a draft since I wanted to open a discussion about if this is wanted at all. Truth be told I pushed this mostly for a discussion and feedback and I mostly forgot it since. 
As to why I didn't choose https://github.com/karthink/gptel I simply wasn't aware of it. I used a compiled list of chatGPT modules since there was too many to audit them all and it wasn't on there. It does indeed look like a good option and I would like to investigate switching.

quinn-dougherty:
(I still want this :/ -- Having trouble wiring up `gpt.el` even tho it's on melpa)

brorbw:
@quinn-dougherty Althoug I'm not actively keeping this branch up to date you could add this branch on your own. Currently I have made this into my own module.

quinn-dougherty:
Cool! I wonder if ideally / in a few years we prefer one chatbot / LLM api doom module or one for each vendor. From a fingerwork perspective I would want the same UX to either backend in most cases. Claude also seems like a great fit https://docs.anthropic.com/claude/reference/getting-started-with-the-api

brorbw:
I think down the road a package with integration to more than one LLM is ideal. Something like `forge` but for LLM's. Considering how LLM's are being deployed maybe down the road even a local LLM.

templates/init.example.el
```diff
@@ -108,6 +108,7 @@
        ;;tmux              ; an API for interacting with tmux
        ;;tree-sitter       ; syntax and parsing, sitting in a tree...
        ;;upload            ; map local to remote projects via ssh/ftp
+       ;;chatgpt           ; the last friend you will ever need
```
peaceful-james, line 111: This comment is hilarious!

quinn-dougherty:
> Something like forge but for LLM's. 

There's no standard like `git` such that there APIs would be so compatible.

brorbw:
> There's no standard like git such that there APIs would be so compatible.

There are no API standard between the different forges

oatmealm:
I'm not sure where this is heading but surely we can do better than chat.el today? At least not a solution hardwired to OpenAI. I also think gptel is leading the field, but ellama/elisa (RAG) and couple of others are for interesting too.

hlissner:
I intend to revisit this post-v3 and implement a gptel version of the module. I've left this open as a reminder and a reference.

hlissner:
As of 6898a9c there is now a rudimentary `:tools llm` module, based on gptel. Thanks for bringing this to my attention!

PR #7236, opened 2023-05-12
fix: enable rainbow delimiters by default

casswedson:
I had this snippet in my config, doom installs rainbow delimites but it doesn't enable them by default - just enable it on `prog-mode`

Fixes #0000
References #0000
Replaces #0000

-----
other prs enabled rainbow-delimiters for specific modes, so uhhh

### DIFF
--- a/lisp/doom-ui.el
+++ b/lisp/doom-ui.el
@@ -472,8 +472,9 @@ windows, switch to `doom-fallback-buffer'. Otherwise, delegate to original
 ;; languages like Lisp. I reduce it from it's default of 9 to reduce the
 ;; complexity of the font-lock keyword and hopefully buy us a few ms of
 ;; performance.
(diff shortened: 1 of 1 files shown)


elken:
A non-zero amount of users have complained about wanting to disable the modules that already have them enabled, so I don't think enabling this globally will be well received but I think when global flags are added something like this will be much simpler to add in.

hlissner:
Thanks for the PR, but I must turn it down. I plan to move rainbow-delimiters out of core soon, and I haven't yet decided what to do with it (or if I'll be removing it altogether; it's config is so trivial I could leave it to users to set up themselves) and, as mentioned, not everyone likes the package. I'm more inclined to define a `global-rainbow-delimiters-mode` to make it easier to toggle. 

In any case, that's something to consider once I decide its fate in or after v3.0. Thanks for bringing it to my attention!

Issue #4553, opened 2021-01-21
Change default email address from "jon@doe.com" to "jon@example.com"


bee-san:
In the default `config.el` we get:

```
user-mail-address "jon@doe.com"
```

`doe.com` is a private domain name.

We should use [example.com](https://en.wikipedia.org/wiki/Example.com) for the email, such as `jon@example.com`. This means that if someone uses a service and doesn't change the domain, someone might buy `doe.com` and bad things could happen.

We can guarantee that example.com will always be the de-facto example email.

hlissner:
Thanks for the suggestion, but this is not a real problem. The closest `user-mail-address` comes to being used for sensitive information is to look up your _public_ ssh/pgp key by email address (instead of by fingerprint). If this were truly exploitable, there would be nothing stopping malicious third parties from using `john@example.com` instead. There is no legitimate security concern here.

PR #7258, opened 2023-05-31
docs: don't set user-* variables by default

zaz:
Do not default to setting the user's name to John Doe or their email to john@doe.com. It is better to leave these variables unset than to set them incorrectly.

Currently, if the user edits the Doom config and ignores those two lines, their name and email get set to "John Doe" <john@doe.com>, which is probably not what they want.

I could not find any explanation for why the lines are uncommented when you added them in 18d8ea22f6c81e6b46c87448476d62c9be6685f2

Thank you!

-----

Related #4553

### DIFF
--- a/templates/config.example.el
+++ b/templates/config.example.el
@@ -6,8 +6,8 @@
 
 ;; Some functionality uses this to identify you, e.g. GPG configuration, email
 ;; clients, file templates and snippets. It is optional.
-(setq user-full-name "John Doe"
-      user-mail-address "john@doe.com")
+;; (setq user-full-name "John Doe"
+;;       user-mail-address "john@doe.com")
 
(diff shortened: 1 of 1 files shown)


hlissner:
Hmm, I'm not convinced this is an issue, since these variables aren't used for anything sensitive, and because the intention of these templates is to be changed, but I'm coming around to the fact people just don't like them. 

I plan to add pre-configurators in the future. One, where you can preconfigure Doom on our website (which produces a one-line `doom install` command that'll generate your local config accordingly), and a second CLI wizard to do so from your terminal. Either of which can be used to generate a dummy config you can play with before committing to running `doom install`, so issues like these will eventually go away, but I'll accept this PR in the meantime.

PR #7261, opened 2023-06-02
tweak(ledger): map `ledger-report` in `ledger-report-mode-map`

krydos:
This PR maps `<localleader> r` in `ledger-report-mode-map` to `ledger-report` function. 

`ledger-report` function is very often used and the keybinding `<localleader> r` is already mapped to `ledger-report` in `ledger-mode-map` (notice not a "report" mode map).

The issue is after a report is generated you can't type the same key sequence to choose another report to generate.
It's because the active buffer is changed to the report one and `ledger-report-mode-map` doesn't have the very same keybinding defined. 
It means you have to switch the buffer back to your ledger file and type the sequence again. 

In the ledger-mode.el the keybinding to generate a report works in both buffers. 

Hope it makes sense.
 
That's my first time here, please let m know if I missed anything important. 

-----

### DIFF
--- a/modules/lang/ledger/config.el
+++ b/modules/lang/ledger/config.el
@@ -61,6 +61,10 @@
          :m "[[" #'ledger-navigate-prev-xact-or-directive)
 
         (:localleader
+         :map ledger-report-mode-map
+         "r" #'ledger-report)
+
+        (:localleader
          :map ledger-mode-map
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Makes good sense to me. Thanks for your help!

PR #7286, opened 2023-07-03
bump: tree-sitter-langs.

acristoffers:
This bumps the commit of `tree-sitter-langs` to the latest to include the `MATLAB` parser that just got accepted.
To make the highlight work I had to `(setq! matlab-file-font-lock-keywords '())` in my configuration (inside `matlab-mode-hook`).

### DIFF
--- a/modules/tools/tree-sitter/packages.el
+++ b/modules/tools/tree-sitter/packages.el
@@ -7,4 +7,4 @@
 
 (when (modulep! :editor evil +everywhere)
   (package! evil-textobj-tree-sitter
-    :pin "02f8253034042d8f171bc0ef93e3538b71a29153"))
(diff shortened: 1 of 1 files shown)


acristoffers:
Text-objects are now also available in latest `evil-textobj-tree-sitter`.

hlissner:
The module needed a larger (and more recent) bump, so I've done so in 2b79787. In any case, thanks for bringing it to my attention!

> To make the highlight work I had to (setq! matlab-file-font-lock-keywords '()) in my configuration (inside matlab-mode-hook).

Is this still necessary with the bump?

acristoffers:
Yes, otherwise `tree-sitter-mode` complains about it being void. I believe that the `tree-sitter-langs` expect this change to be made by `matlab-mode`.

PR #7304, opened 2023-07-22
fix(javascript): use top-level node_modules in monorepos

tien:
JavaScript monorepo packages can contain a child `node_modules`, hence we need to grab the one at the workspace root if possible.


hlissner (approved):

hlissner:
I suspect this may be an issue for folks who want multiple node_modules/.bin in their monorepo, and want all of them in their `$PATH` (or have a setup where their `$HOME` is a project, as far as projectile is concerned). Perhaps the ideal would be to walk up the file tree and collect _all_ node_modules/.bin into `$PATH`...  I'll revisit this another time, after some testing.

In any case, thanks for the help!

PR #7313, opened 2023-07-27
tweak(workspaces): prefill current name on rename

undergroundquizscene:
This change uses completing-read to get the new name from the user when renaming workspaces, to allow them to more easily make a small change to the existing name of a workspace.

Note: I haven’t included screenshots because I don’t think this is a big enough visual change to need them. Let me know if this is the wrong conclusion!

-----

### DIFF
--- a/modules/ui/workspaces/autoload/workspaces.el
+++ b/modules/ui/workspaces/autoload/workspaces.el
@@ -221,7 +221,7 @@ workspace."
 ;;;###autoload
 (defun +workspace/rename (new-name)
   "Rename the current workspace."
-  (interactive (list (read-from-minibuffer "New workspace name: ")))
+  (interactive (list (completing-read "New workspace name: " (list (+workspace-current-name)))))
   (condition-case-unless-debug ex
       (let* ((current-name (+workspace-current-name))
              (old-name (+workspace-rename current-name new-name)))


elken:
`read-from-minibuffer` also takes initial contents as an argument, so why the change to `completing-read`?

undergroundquizscene:
> `read-from-minibuffer` also takes initial contents as an argument, so why the change to `completing-read`?

The docs for `'read-from-minibuffer` scared me off a bit: 

> The optional second arg INITIAL-CONTENTS is an obsolete alternative to
> DEFAULT-VALUE.  It normally should be nil in new code, except when
> HIST is a cons.  It is discussed in more detail below.

> Fifth arg HIST, if non-nil, specifies a history list and optionally
> the initial position in the list.  It can be a symbol, which is the
> history list variable to use, or a cons cell (HISTVAR . HISTPOS).
> In that case, HISTVAR is the history list variable to use, and
> HISTPOS is the initial position for use by the minibuffer history
> commands.  For consistency, you should also specify that element of
> the history as the value of INITIAL-CONTENTS.  Positions are counted
> starting from 1 at the beginning of the list.  If HIST is t, history
> is not recorded.

> The remainder of this documentation string describes the
> INITIAL-CONTENTS argument in more detail.  It is only relevant when
> studying existing code, or when HIST is a cons.  If non-nil,
> INITIAL-CONTENTS is a string to be inserted into the minibuffer before
> reading input.  Normally, point is put at the end of that string.
> However, if INITIAL-CONTENTS is (STRING . POSITION), the initial
> input is STRING, but point is placed at _one-indexed_ position
> POSITION in the minibuffer.  Any integer value less than or equal to
> one puts point at the beginning of the string.  *Note* that this
> behavior differs from the way such arguments are used in completing-read
> and some related functions, which use zero-indexing for POSITION.

These make it sound like I should only use the initial contents argument when I’ve specified a history list variable and a position, which seemed more complicated. What history list variable should I use? Do I need to declare that in advance somewhere?

I guess something like the following might work?

``` emacs-lisp
(interactive (list (read-from-minibuffer "New workspace name: " (+workspace-current-name) nil nil (cons '+workspace--rename-history 1))))
```

elken:
> I guess something like the following might work?

>` (interactive (list (read-from-minibuffer "New workspace name: " (+workspace-current-name) nil nil (cons '+workspace--rename-history 1))))`
Yeah that works still, Henrik will have to answer decisively here but I think this change is fine

hlissner (approved):

hlissner:
From my testing it seems various completion frameworks (helm, ivy, vertico, etc) handle read-from-minibuffer's `INITIAL-CONTENTS` argument *slightly* different from one another, but are more consistent with `completing-read`'s. This need more testing, but I'll settle for the latter for now. Thanks for your help, in any case!

undergroundquizscene:
> From my testing it seems various completion frameworks (helm, ivy, vertico, etc) handle read-from-minibuffer's `INITIAL-CONTENTS` argument _slightly_ different from one another, but are more consistent with `completing-read`'s. This need more testing, but I'll settle for the latter for now. Thanks for your help, in any case!

Interesting, I hadn’t expected that. Glad to help :)

PR #7333, opened 2023-08-02
Lua Love project root function fix

LemonBreezes:
This PR addresses an error I get every time I open a Love2D project with Doom Emacs. It occurs whenever `projectile-update-mode-line` is called.

Currently, the project root function for Love2D projects is broken
because it does not take an optional `dir` argument. This commit fixes
that.

### DIFF
--- a/modules/lang/lua/autoload/lua.el
+++ b/modules/lang/lua/autoload/lua.el
@@ -9,26 +9,28 @@
             (shell-quote-argument root))))
 
 ;;;###autoload
-(defun +lua-love-project-root ()
-  "Returns the directory where a main.lua or main.moon exists.
+(defun +lua-love-project-root (&optional dir)
+  "Search up from DIR for the directory where a main.lua or main.moon exists.
 
 Returns nil if 'love' executable can't be found."
   (when (executable-find "love")
-    (if (doom-project-p)
+    (if (doom-project-p dir)
         (file-name-directory
-         (or (project-file-exists-p! (or "main.lua" "src/main.lua"))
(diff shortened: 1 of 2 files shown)


modules/lang/lua/autoload/lua.el
```diff
@@ -9,26 +9,28 @@
             (shell-quote-argument root))))
 
 ;;;###autoload
-(defun +lua-love-project-root ()
+(defun +lua-love-project-root (&optional dir)
   "Returns the directory where a main.lua or main.moon exists.
 
-Returns nil if 'love' executable can't be found."
+Returns nil if `love' executable can't be found."
```
fosskers, line 15: Emacs will link this to a non-existent symbol.

LemonBreezes: Huh. Changing the quoting style from `'...'` to `\`...'` resolves a byte compiler warning because the former quoting style is now deprecated.

fosskers: Rock and a hard place eh. Emacs 29 has a new quoting style that won't create a link, but I don't know that the linters support it yet.

LemonBreezes: I removed the quotes change. That can be changed at a later point.

hlissner:
Your original suggestion would make `project-file-exists-p!` a bit redundant with the `file-exists-p!` macro, so I've slightly tried to find a middle point with b17257c that would still fix your issue. Let me know if that isn't the case!

Issue #1413, opened 2019-05-14
Remapping prefixes loses descriptions


nate:
```elisp
  (map! :leader
        "TAB" nil
        "l" doom-leader-workspaces-map)
```

This very excellent commit https://github.com/hlissner/doom-emacs/commit/b7d1702484129063df7354430aafe08a5e9ca3a6 allows us to now remap leader prefixes (yay!) but when they're remapped they lose the descriptions of their sub-commands (oh no!).

If you eval the form above you'll see that `TAB` does indeed disappear but `l` says "prefix" and the subcommand descriptions are all either the function name or they say "lambda".

And if you try to do something foolish like this:

```elisp
(map! :leader "TAB" 'evil-switch-to-windows-last-buffer "l" doom-leader-workspace-map)
```

Then `TAB` will retain its old description, "workspace", but will have the new functionality.

hlissner:
I'm not aware of a good solution to this problem. `which-key` associates its labels with key sequence regexps. This means they won't change if the keys have. We'd need non-trivial effort to remap these regexps after-the-fact or find a way to associate which-key labels with keymaps/commands from the get go (which would follow the commands, no matter where they're bound) instead of the current, global map which-key currently uses.

nate:
Well, if we can't change it after the fact, could we configure the keys for various sections before they're initially set?

danielstaleiny:
I had the same issue with old description and I find out that you can overwrite description with `:desc`. 

so write this instead to create new description.

`(map! :leader 
            :desc "your new description" "TAB" #'evil-switch-to-windows-last-buffer 
            :desc "your other new description" "l" #'doom-leader-workspace-map)`

agzam:
So I did something like this:

```
(map! :leader
  (:prefix ("t" . "my toggles")
    :desc "wrap-it" :w #'my-wrap-fn)
```

This would work as expected. Even though there's already a function bound to "SPC tw", it will correctly replace it. 
Yet `map!` completely ignores the descriptions. "SPC t", still would show "toggle" (instead of "my toggles") in the which-key HUD, and "SPC tw", still says "Soft Line wrapping", even though the key points to a different function.

Moreover, `helpful-key` for `SPC tw` can't properly pinpoint the location of the `#'my-wrap-fn`, it says, it's closure converted to defun. It properly shows the body of the function, and incorrectly says that it's defined in `~/.emacs.d/.local/etc/@/init.29.el` (I'm using Emacs 29)

[...]

PR #7189, opened 2023-04-09
feat: allow setting evil states for leader keys

TheJJ:
This allows adjusting the evil-states a leader key is active for.

-----
- [x] I searched the issue tracker and this hasn't been PRed before.
- [x] My changes are not on [the do-not-PR list](https://doomemacs.org/d/do-not-pr) for this project.
- [x] My commits conform to [the git conventions](https://doomemacs.org/d/git-conventions).
- [x] Any relevant issues or PRs have been linked to.

PR #7339, opened 2023-08-06
fix: make leader key descriptions in the keymap itself

LemonBreezes:
This commit moves leader key descriptions from `which-key-replacement-alist` to the keymap itself. This helps with performance because that way, which-key does not have to calculate every single leader key description for each keypress. Furthermore, this fixes issues like #1413 where relocating leader keymaps resulted in which-key not showing the correct leader key descriptions.
 
I also made the leader prefix maps have their own prefix commands by populating the function slot of the keymap variables. This is an Emacs convention. I made `doom-leader-map` follow this convention as well, though I kept the `doom/leader` so as to not break existing configs.

### DIFF
--- a/lisp/doom-keybinds.el
+++ b/lisp/doom-keybinds.el
@@ -145,11 +145,8 @@ all hooks after it are ignored.")
                     forms))
             (when-let (desc (cadr (memq :which-key udef)))
               (prependq!
-               wkforms `((which-key-add-key-based-replacements
-                           (general--concat t doom-leader-alt-key ,key)
-                           ,desc)
-                         (which-key-add-key-based-replacements
-                           (general--concat t doom-leader-key ,key)
+               wkforms `((which-key-add-keymap-based-replacements doom-leader-map
+                           ,key
                            ,desc))))))))
     (macroexp-progn
      (append (and wkforms `((after! which-key ,@(nreverse wkforms))))
@@ -195,7 +192,7 @@ localleader prefix."
 ;;   :prefix/:non-normal-prefix properties because general is incredibly slow
 ;;   binding keys en mass with them in conjunction with :states -- an effective
 ;;   doubling of Doom's startup time!
-(define-prefix-command 'doom/leader 'doom-leader-map)
+(define-prefix-command 'doom-leader-map)
 (define-key doom-leader-map [override-state] 'all)
 
 ;; Bind `doom-leader-key' and `doom-leader-alt-key' as late as possible to give
@@ -210,9 +207,9 @@ localleader prefix."
                    (set-keymap-parent doom-leader-map mode-specific-map))
                   ((equal doom-leader-alt-key "C-x")
                    (set-keymap-parent doom-leader-map ctl-x-map)))
-            (define-key map (kbd doom-leader-alt-key) 'doom/leader))
-        (evil-define-key* '(normal visual motion) map (kbd doom-leader-key) 'doom/leader)
-        (evil-define-key* '(emacs insert) map (kbd doom-leader-alt-key) 'doom/leader))
+            (define-key map (kbd doom-leader-alt-key) #'doom-leader-map))
+        (evil-define-key* '(normal visual motion) map (kbd doom-leader-key) #'doom-leader-map)
+        (evil-define-key* '(emacs insert) map (kbd doom-leader-alt-key) #'doom-leader-map))
       (general-override-mode +1))))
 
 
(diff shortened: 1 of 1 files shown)


LemonBreezes:
I removed `doom/leader` because that line was breaking my config for some reason.

LemonBreezes:
So I removed the `doom/leader` line because ot broke my config somehow. Dunno why. Also, this doesn't make the `localleader` map relocatable but that can be a separate commit or PR when I get to it.

LemonBreezes:
With this commit, now you can define any key to `doom-locallleader-map`
symbol and it will work as expected. The `which-key` descriptions will
work as expected too. The only caveat is that the localleader map is now
updated using hooks so there the potential for bugs where the incorrect
map is selected.

LemonBreezes:
I just pushed an update to exclude some minibuffer modes from `doom-init-localleader-key-h` and to also keep track of which major mode is being used by `doom-localleader-key` with `doom-localleader-current-major-mode`.

LemonBreezes:
Ok, so I need a bit of help with this block of code:
```
(add-hook! '(after-change-major-mode-hook
             doom-switch-buffer-hook
             doom-switch-window-hook
             doom-switch-frame-hook)
  (defun doom-init-localleader-key-h ()
    "Set the localleader keys for the current major-mode."
    (unless (or (derived-mode-p 'special-mode)
                (derived-mode-p 'minibuffer-inactive-mode)
                (derived-mode-p 'minibuffer-mode)
                (derived-mode-p 'compilation-mode)
                (eq major-mode 'fundamental-mode))
      (setq doom-localleader-current-major-mode major-mode)
      (set-keymap-parent doom-localleader-map
                         (cdr (assq major-mode doom-localleader-map-alist))))))
```
What I don't know is how to set those hooks and the condition so that the correct localleader map is always enabled. I need to exclude temporary buffers and such. I would like for more people to test this branch.

LemonBreezes:
Okay. I am currently testing using `doom-unreal-buffer-p` instead of pattern-matching on modes for `doom-init-localleader-key-h`.

LemonBreezes:
I've been testing this PR for a few months now and IMO, I think it needs a bit of help in terms of choosing the correct hook(s) and condition to update the local-leader map's major mode. So if anyone can, please contribute solutions. I am thinking that might have to be in a `post-command-hook`.

LemonBreezes:
Hi. I have updated this to use `post-command-hook` and made a backwards compatibility fix. This PR is ready for review now.

lisp/doom-keybinds.el
```diff
@@ -195,8 +234,17 @@ localleader prefix."
...
+(define-prefix-command 'doom-leader-map)
+(define-prefix-command 'doom-localleader-map)
+(fset 'doom/leader doom-leader-map)     ; For backwards compatibility.
+
+(defun doom-update-localleader-key-h ()
+  "Set the localleader keys for the current major-mode."
+  (set-keymap-parent doom-localleader-map
+                     (cdr (assq major-mode doom-localleader-map-alist))))
+
+(add-hook 'post-command-hook #'doom-update-localleader-key-h)
```
TheJJ, line 247: why does this need to run after each and every command?

LemonBreezes: Originally I had it hooked to `doom-switch-buffer-hook`, `doom-switch-frame-hook`, `doom-switch-window-hook`, and `after-change-major-mode-hook` but then it kept running in unexpected scenarios that happened inter-command like when the Corfu popup would show up so I had to set a bunch of predicates for that hook to run and I was not able to find a working set of predicates. After switching it to a `post-command-hook`, I was able to remove all those predicates and just run the `set-keymap-parent` action itself. It runs fast enough to not cause any noticeable latency:
image

emenel:
Just want to say that I've been testing this pr with my meow config, and it fixes the issues with using meow and doom's leader/localleader instead of meow's keypad mode. localleader now appears where it should and works perfectly along with meow's modes and doom's standard leader key (spc). Before, with meow, the localleader would not work via the doom leader (it had to be bound to a different top level key combination).

LemonBreezes:
I just discovered issues with edge cases where the localleader map is used for a minor mode or for a derived mode, so for `org-journal-map`, the `org-mode` localleader bindings will not show up. I'm gonna be working on this. This is a pretty tough code to write.

TheJJ:
could we merge or consider #7189 as a previous step please?

for me the ability to activate the leader key or alt-key combos in more evil states already solves my use-case already. and it extends the more general approach from this patch to specify which evil states the leader and localleader map is for.

LemonBreezes:
Ok. I am going to wait for some time since Hlissner said he was going to make major changes to the core keybindings code.

emenel:
I realize this is on hold pending bigger changes ... but at the moment it's holding me back from updating with more recent doom commits. When I update to the current main branch with this PR I get errors from modules about bindings:

`Error in a Doom module: modules/emacs/dired/config.el, (error Eager macro-expansion failure: (wrong-type-argument symbolp (dired-mode-map)))`

I've looked around to see if I can figure out how to adapt the changes, but it's beyond my knowledge of Doom's inner workings and the time i have at the moment.

For now I'm just not updating, but I'd love to be able to keep my bindings working and update to get the latest changes occasionally.

PR #7341, opened 2023-08-07
 feat(emacs-lisp): add reduced-flymake-byte-compile

jeetelongname:
This PR reduces the amount of false positives and erronious warnings the elisp flymake backends throw off. this PR brings flymake up to UX parity with flycheck. 

-----

### DIFF
--- a/modules/lang/emacs-lisp/autoload.el
+++ b/modules/lang/emacs-lisp/autoload.el
@@ -338,6 +338,47 @@ as `+emacs-lisp-non-package-mode' will enable it and disable the other checkers.
                  (ignore-errors (delete-file tmp-file))
                  (kill-buffer out-buf))))))))
 
+(define-minor-mode +emacs-lisp--flymake-non-package-mode
+  ""
+  :since "3.0.0"
+  (if +emacs-lisp--flymake-non-package-mode
+      (progn
+        (remove-hook! 'flymake-diagnostic-functions :local #'elisp-flymake-checkdoc #'elisp-flymake-byte-compile)
+        (add-hook 'flymake-diagnostic-functions #'+emacs-lisp-reduced-flymake-byte-compile nil t))
+    (add-hook! 'flymake-diagnostic-functions :local #'elisp-flymake-checkdoc #'elisp-flymake-byte-compile)
+    (remove-hook 'flymake-diagnostic-functions #'+emacs-lisp-reduced-flymake-byte-compile t)))
+
+(define-minor-mode +emacs-lisp--flycheck-non-package-mode
+  ""
+  :since "3.0.0"
+  (if (not +emacs-lisp--flycheck-non-package-mode)
+      (when (get 'flycheck-disabled-checkers 'initial-value)
+        (setq-local flycheck-disabled-checkers (get 'flycheck-disabled-checkers 'initial-value))
+        (kill-local-variable 'flycheck-emacs-lisp-check-form))
+    (with-memoization (get 'flycheck-disabled-checkers 'initial-value)
+      flycheck-disabled-checkers)
+    (setq-local flycheck-emacs-lisp-check-form
+                (prin1-to-string
+                 `(progn
+                    (setq doom-modules ',doom-modules
+                          doom-disabled-packages ',doom-disabled-packages
+                          byte-compile-warnings ',+emacs-lisp-linter-warnings)
+                    (condition-case e
+                        (progn
+                          (require 'doom)
+                          (require 'doom-cli)
+                          (require 'doom-start))
+                      (error
+                       (princ
+                        (format "%s:%d:%d:Error:Failed to load Doom: %s\n"
+                                (or ,(ignore-errors
+                                       (file-name-nondirectory
+                                        (buffer-file-name (buffer-base-buffer))))
+                                    (car command-line-args-left))
+                                0 0 (error-message-string e)))))
+                    ,(read (default-toplevel-value 'flycheck-emacs-lisp-check-form))))
+                flycheck-disabled-checkers (cons 'emacs-lisp-checkdoc
+                                                 flycheck-disabled-checkers))))
 ;;;###autoload
 (define-minor-mode +emacs-lisp-non-package-mode
   "Reduce flycheck/flymake verbosity where it is appropriate.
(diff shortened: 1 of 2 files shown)


aikrahguzar:
For a while I had a version of `+emacs-lisp-reduced-flymake-byte-compile` that loaded some of the core doom elisp files using `-l` switch. This slowed the linting but the results were much better since byte compiler could understand the doom apis against which config is written. IIRC in that approach the only errors that needed to be silenced were `noruntime` ones that depended on load order which the byte compiler can't reason about. However the problem was that I couldn't figure out a set of files such that the  doom library would be loaded but not my config and that bit me when one time my config had an unbalanced parenthesis. So I went back to `emacs -Q` approach. However since doom has a (mostly unannounced) profile system now, it would be good to have a checker that uses a `vanilla doom` profile instead of `emacs -Q` for checking user configs.

jeetelongname:
that is an option i can explores. ATM this backend follows the same linter warnings and the same api that flycheck uses (`+emacs-lisp-linter-warnings`) so I don't really want to mess with that all too much.  
Add to that a bunch of profile and package management changes are forecasted to come sooner rather than later I don't want to rely on an old API to then have to rework it again.

Its not a bad idea by any means mind you. When / if flycheck is dropped we can defo explore a proper doom linter. not only including the doom runtime but also including specific checks and so on.

aikrahguzar:
> that is an option i can explores. ATM this backend follows the same linter warnings and the same api that flycheck uses (`+emacs-lisp-linter-warnings`) so I don't really want to mess with that all too much. Add to that a bunch of profile and package management changes are forecasted to come sooner rather than later I don't want to rely on an old API to then have to rework it again.
> 
> Its not a bad idea by any means mind you. When / if flycheck is dropped we can defo explore a proper doom linter. not only including the doom runtime but also including specific checks and so on.

This is definitely an idea for the future after doom 3 is released. On the same note, as a user of `(modeline +light)` I lost the indicator provided by `+modeline-checker`.  I was able to hack together a solution looking at how it worked with flycheck,
```emacs-lisp
(after! flymake
  (advice-add 'flymake--update-diagnostics-listings :after
              (defun +modeline-flymake-info (buf)
                (when flymake--state
                  (let ((all-done t) (nerrors 0) (nwarnings 0) (nnotes 0) (total 0))
                    (maphash (lambda (_ bstate) (if (flymake--state-reported-p bstate)
                                               (dolist (diag (flymake--state-diags bstate))
                                                 (pcase (flymake-diagnostic-type diag)
                                                   ((or :error 'eglot-error) (cl-incf nerrors))
                                                   ((or :warning 'eglot-warning) (cl-incf nwarnings))
                                                   ((or :note 'eglot-note) (cl-incf nnotes))))
                                             (unless (flymake--state-disabled bstate) (setq all-done nil))))
                             flymake--state)
                    (setq total (number-to-string (+ nerrors nwarnings nnotes)))
                    (with-current-buffer buf
                      (let ((echo-text (format "Errors: %d, Warnings: %d, Notes: %d" nerrors nwarnings nnotes)))
                        (setq +modeline-checker
                              (if all-done
                                  (cond ((> nerrors 0) (+modeline-format-icon "warning" total 'error echo-text))
                                        ((> nwarnings 0) (+modeline-format-icon "do_not_disturb_alt" total 'warning echo-text))
                                        ((> nnotes 0) (+modeline-format-icon "note_add" total 'success echo-text))
                                        (t (+modeline-format-icon "check" "" 'success nil -0.1)))
                                (+modeline-format-icon "av_timer" "" 'mode-line-highlight "Running...")))))))))

  (advice-add 'flymake-start :before
              (defun +modeline-flymake-init (&rest _)
                (setq +modeline-checker (+modeline-format-icon "av_timer" "" 'mode-line-highlight "Running...")))))
```
It works but uses a lot of private flymake api so not really suitable for general use but that is another thing needed to bring UX on par with flycheck.

jeetelongname:
I can have a look at the modeline but I do not use `+light` so i was not aware of it. 

This reminds me I do need to play with the modeline segments. Ill see if I can get some time and do so.

modules/lang/emacs-lisp/autoload.el
```diff
@@ -274,9 +274,71 @@ https://emacs.stackexchange.com/questions/10230/how-to-indent-keywords-aligned"
...
+    ;; write the content to a temp file
+    (save-restriction
+      (widen)
+      (write-region nil nil tmp-file nil 'nomessage))
+    ;; make the process
+    (setq +emacs-lisp-reduced-flymake-byte-compile--process
+          (make-process
+           :name "+emacs-reduced-flymake"
+           :noquery t
+           :connetion-type 'pipe
```
hpfr, line 304:
```suggestion
           :connection-type 'pipe
```

modules/lang/emacs-lisp/autoload.el
```diff
@@ -274,9 +274,71 @@ https://emacs.stackexchange.com/questions/10230/how-to-indent-keywords-aligned"
                                  file-base)))))))
          (not (locate-dominating-file default-directory ".doommodule")))))
 
+(defvar-local +emacs-lisp-reduced-flymake-byte-compile--process nil)
+
+(defun +emacs-lisp-reduced-flymake-byte-compile (report-fn &rest _args)
+  "A Flymake backend for warnings from the elisp byte compiler
+
+This checker reduces the amount of false positives the byte compiler throws off
+compared too `elisp-flymake-byte-compile'. The linter warnings that are enabled
```
hpfr, line 283:
```suggestion
compared to `elisp-flymake-byte-compile'. The linter warnings that are enabled
```

modules/lang/emacs-lisp/autoload.el
```diff
@@ -274,9 +274,71 @@ https://emacs.stackexchange.com/questions/10230/how-to-indent-keywords-aligned"
                                  file-base)))))))
          (not (locate-dominating-file default-directory ".doommodule")))))
 
+(defvar-local +emacs-lisp-reduced-flymake-byte-compile--process nil)
+
+(defun +emacs-lisp-reduced-flymake-byte-compile (report-fn &rest _args)
+  "A Flymake backend for warnings from the elisp byte compiler
```
hpfr, line 280:
```suggestion
  "A Flymake backend for byte compilation in non-package elisp.
```

modules/lang/emacs-lisp/autoload.el
```diff
@@ -287,39 +349,52 @@ Essentially, this means in any elisp file that either:
...
-                        (format "%s:%d:%d:Error:Failed to load Doom: %s\n"
-                                (or ,(ignore-errors
-                                       (file-name-nondirectory
-                                        (buffer-file-name (buffer-base-buffer))))
-                                    (car command-line-args-left))
-                                0 0 (error-message-string e)))))
-                    ,(read (default-toplevel-value 'flycheck-emacs-lisp-check-form))))
-                flycheck-disabled-checkers (cons 'emacs-lisp-checkdoc
-                                                 flycheck-disabled-checkers))))
+      (if (modulep! :checkers syntax +flymake)
```
hpfr, line 359: Could we just have two definitions of this minor mode, one for flycheck and one for flymake, rather than nesting all these conditionals inside one minor mode body?

modules/lang/emacs-lisp/autoload.el
```diff
@@ -274,9 +274,71 @@ https://emacs.stackexchange.com/questions/10230/how-to-indent-keywords-aligned"
...
+           :connetion-type 'pipe
+           :buffer out-buf
+           :command `(,(expand-file-name invocation-name invocation-directory)
+                      "-Q"
+                      "--batch"
+                      ,@(mapcan (fn! (list "-L" %)) elisp-flymake-byte-compile-load-path)
+                      ;; this is what silences the byte compiler
+                      "--eval" ,(prin1-to-string `(setq doom-modules ',doom-modules
+                                                        doom-disabled-packages ',doom-disabled-packages
+                                                        byte-compile-warnings ',+emacs-lisp-linter-warnings)
```
hpfr, line 313: `+emacs-lisp-reduced-flymake-byte-compile` is [`elisp-flymake-byte-compile`](https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/progmodes/elisp-mode.el?h=emacs-29.1#n2141), modulo some reshuffling, Doom macros, and comments. As you note, this is the important change. Is there some way to advise the original diagnostic function rather than vendoring all the boilerplate? I don’t see an easy way to do this either, but I wanted to at least note this somewhere

jeetelongname: The ideal solution is to abstract the command into a variable that a person could change if needed. 
This is something that would need to be upstreamed. 
If/when that patch is sent this all becomes easier. 

But advice here would just lead to overriding the current implementation with a new one. It would add another layer of indirection which I would avoid if I can just use hooks. There is just no good entry point here other than the back end.

I know you mention that this would not be easy but I wanted to highlight why I went for a clean room implementation instead of trying to advise it. Its annoying but IMO the cleaner solution.

jeetelongname:
What a catch! I was struggling to debug this and would love too know how you tracked this down. for the time being ill remove the `fn!` call and replace it with the lambda

hlissner (approved):

PR #7342, opened 2023-08-08
tweak(default): better vanilla multiple-cursor bindings

UndeadKernel:
Better default bindings for multiple-cursors in vanilla emacs.

A transient map takes care of repeating useful commands without having to input entire shortcut sequences.
Added the binding "C-+" to create a new cursor like what is selected (following the recommendation of the package's author).

-----
- [ x] My changes are not on [the do-not-PR list](https://doomemacs.org/d/do-not-pr) for this project.
- [ x] My commits conform to [the git conventions](https://doomemacs.org/d/git-conventions).
- [ x] Any relevant issues or PRs have been linked to.


LemonBreezes:
Hi. I'll look over the MC bindings later today but I disagree with binding `C--` to contracting the expanded region because that overrides `negative-argument`. Maybe we can make a repeat map instead so that `C--` is only bound transiently.

UndeadKernel:
The reason for changing "C--" to "contract region" is that there are other ways to set negative arguments and setting arguments is not as common as working with the expansion and contraction of regions. I feel that I only sporadically use negative arguments and, when I need to, there is always "C-u -<number>".

LemonBreezes:
> The reason for changing "C--" to "contract region" is that there are other ways to set negative arguments and setting negative arguments is not as common as working with the expansion and contraction of regions. I feel that I only sporadically use negative arguments and, when I need to, there is always "C-u -".

With respect to `expand-region`, doesn't this accomplish the same effect without binding that key globally?
```
(setq expand-region-fast-keys-enabled nil)
(add-hook 'repeat-mode-hook
          (lambda nil
            (defvar expand-region-repeat-map (make-sparse-keymap)
              "Defined by `define-repeat-map'.")
            (progn
              (define-key expand-region-repeat-map (kbd "=") #'er/expand-region)
              (define-key expand-region-repeat-map (kbd "-") #'er/contract-region)
              (define-key expand-region-repeat-map (kbd "C-=") #'er/expand-region)
              (define-key expand-region-repeat-map (kbd "C--") #'er/contract-region)
              (put 'er/expand-region 'repeat-map 'expand-region-repeat-map)
              (put 'er/contract-region 'repeat-map 'expand-region-repeat-map))))
```

Or if you use the `define-repeat-map` package, the above block is just

```
(setq expand-region-fast-keys-enabled nil)
(define-repeat-map expand-region
  ("=" er/expand-region
   "-" er/contract-region
   "C-=" er/expand-region
   "C--" er/contract-region))
```

Similarly, the transient map you wrote can be condensed to
```
(define-repeat-map multiple-cursors
    ("n" mc/mark-next-like-this
     "p" mc/mark-previous-like-this
     "N" mc/unmark-next-like-this
     "P" mc/unmark-previous-like-this))
```
So in my opinion, we should add 
```
(use-package! repeat
  :init
  (add-hook 'doom-first-input-hook #'repeat-mode)
  :config
  (map! :map help-map "C-r" #'describe-repeat-maps)
  (setq repeat-exit-key "TAB"
        repeat-check-key nil
        repeat-echo-mode-line-string nil))
```
and 
```
(package! define-repeat-map :recipe (:repo "https://tildegit.org/acdw/define-repeat-map.el.git"))
```
and just condense that along with future repeat maps (or let that be delegated to the user).

UndeadKernel:
I agree that indeed "C--" could be a transient binding. I'll look into this.

UndeadKernel:
I propose to not use other minor modes (e.g., repeat-mode) and external packages to create transients for multiple-cursors and expand-region. I don't like the idea of relying on yet more minor modes when it's possible to do it without any for just a little boiler plate.

Note that the "boiler plate" for the transients used in multiple-cursors (MC) is needed. Because of the way MC handles the creation of multiple cursors, all custom bindings for MC need to be named functions.

LemonBreezes:
> I propose to not use other minor modes (e.g., repeat-mode) and external packages to create transients for multiple-cursors and expand-region. I don't like the idea of relying on yet more minor modes when it's possible to do it without any for just a little boiler plate.
> 
> Note that the "boiler plate" for the transients used in multiple-cursors (MC) is needed. Because of the way MC handles the creation of multiple cursors, all custom bindings for MC need to be named functions.

I can understand why you would not want to pull in the `define-repeat-map` macro but I still believe that `repeat-mode` is a superior solution to using raw transient maps because:

1. `repeat-mode` provides a visual hint to the user indicating that a repeat map is currently active and shows which keys are in the transient map. With the lower level `transient-map` solution, a user doesn't have any indication that a transient map is active. The visual hint can also be integrated with `which-key` or `embark` using https://github.com/karthink/repeat-help if desired.
2. `repeat-mode` also provides a `describe-repeat-maps` command and allows the user to toggle the use of repeat maps globally using `(repeat-mode -1)`, etc., so it provides an overall more complete and unified user experience for transient maps.
3. `repeat-mode` is in use by many packages such as `tab-bar` (built into Emacs) and `diff-hl` so simply enabling `repeat-mode` without further configuration will enable a lot of convenient repeat maps that package developers thought of when writing their packages.

UndeadKernel:
@hlissner, I'd like to have your opinion on this matter: As a summary, do you see any issue in using (and enabling) a new minor-mode to create transient bindings? I originally created some transient bindings for the multiple-cursors minor mode using `transient-mode-map`. @LemonBreezes made a good point that we could also create those transient bindings configuring and enabling the minor mode `repeat-mode`. What's your take on this? If you don't have any preference, I suppose I'd go with @LemonBreezes's suggestion of configuring and enabling `repeat-mode`.

@LemonBreezes, is `repeat-mode` a global minor mode or does it have to be enabled when visiting new buffers?

LemonBreezes:
> @hlissner, I'd like to have your opinion on this matter: As a summary, do you see any issue in using (and enabling) a new minor-mode to create transient bindings? I originally created some transient bindings for the multiple-cursors minor mode using `transient-mode-map`. @LemonBreezes made a good point that we could also create those transient bindings configuring and enabling the minor mode `repeat-mode`. What's your take on this? If you don't have any preference, I suppose I'd go with @LemonBreezes's suggestion of configuring and enabling `repeat-mode`.
> 
> @LemonBreezes, is `repeat-mode` a global minor mode or does it have to be enabled when visiting new buffers?
> 

Yup, `repeat-mode` is a global minor mode.

hlissner:
Sorry for the tremendously late response, but I'd prefer we use `repeat-mode`. In fact, it's already on my TODO list to enable it by default for non-evil users, and perhaps even extend it for evil or leader keys. That said, I don't think it should be this module's responsibility to *enable* it, only exploit it if it happens to be enabled.

PR #7349, opened 2023-08-16
fix(multiple-cursors): check evil-local-mode instead of the global one (+multiple-cursors-compat-switch-to-emacs-state-h)

daanturo:
For evil compatibility (switching to emacs state during multiple cursors), when evil is enabled locally by evil-local-mode but not globally by evil-mode, using the former to check is more accurate.

Also make the state variables local since MC maybe used for multiple buffers: activate current buffer and while not deactivating here, go to another buffer and activate there.

-----


hlissner (approved):

hlissner:
Good catch! Thanks again for your help!

PR #7358, opened 2023-08-19
refactor(vertico): consult-dir: drop dependence on docker-tramp

elken:
Removes the dependence on docker-tramp, can later be cleaned up with
`tramp-container--completion-function` when we drop support for <29.

Aware this probably isn't the _cleanest_ way to handle this, happy to drop ootb support for docker & podman and leave it up to the user to add whichever they want.


hlissner (approved):

hlissner:
LGTM. Thanks again!

PR #7366, opened 2023-08-22
feat(default): add use consult-man if available

LemonBreezes:
`consult-man` is a better version of `man` that computes its candidates asynchronously and allows the user to pass arguments directly to the man command. It's basically like `consult-grep` but for `man`. Before merging this, please double-check that `consult-man` is defined in our pinned version of `consult`.

### DIFF
--- a/modules/config/default/autoload/default.el
+++ b/modules/config/default/autoload/default.el
@@ -25,7 +25,8 @@ generate `completing-read' candidates."
   (interactive)
   (call-interactively
    (if (and (not IS-MAC) (executable-find "man"))
-       #'man
+       (or (command-remapping #'man)
+           #'man)
      #'woman)))
 
 ;;;###autoload


UndeadKernel (approved):
All looks good from my side.

hlissner (changes requested):
Instead of explicitly calling `consult-man`, please use `command-remapping` instead:

```elisp
(defun +default/man-or-woman ()
  "Invoke `man' if man is installed and the platform is not MacOS, otherwise use `woman'.

`man -k \"^\"` is very slow on MacOS, which is what `Man-completion-table' uses to
generate `completing-read' candidates."
  (interactive)
  (call-interactively
   (if (and (not IS-MAC) (executable-find "man"))
       (or (command-remapping #'man)
           #'man)
     #'woman)))
```

This is so that the command respects remaps, such as the one that already exists for consult-man [here](https://github.com/doomemacs/doomemacs/blob/master/modules/completion/vertico/config.el#L125). It should achieve the same result, and should defer to ivy/helm's man commands, as well, if they are available.

LemonBreezes:
> Instead of explicitly calling `consult-man`, please use `command-remapping` instead:
> 
> ```emacs-lisp
> (defun +default/man-or-woman ()
>   "Invoke `man' if man is installed and the platform is not MacOS, otherwise use `woman'.
> 
> `man -k \"^\"` is very slow on MacOS, which is what `Man-completion-table' uses to
> generate `completing-read' candidates."
>   (interactive)
>   (call-interactively
>    (if (and (not IS-MAC) (executable-find "man"))
>        (or (command-remapping #'man)
>            #'man)
>      #'woman)))
> ```
> 
> This is so that the command respects remaps, such as the one that already exists for consult-man [here](https://github.com/doomemacs/doomemacs/blob/master/modules/completion/vertico/config.el?rgh-link-date=2023-08-23T00%3A40%3A09Z#L125). It should achieve the same result, and should defer to ivy/helm's man commands, as well, if they are available.

Ok! Just finished making that change.

hlissner (approved):

hlissner:
LGTM! Thanks again for your help.

PR #7373, opened 2023-08-25
fix(lib): doom/help-packages: handle missing homepage

mosquito-magnet:
When a package is added via straight local-repo recipe, no homepage can be determined. Prevent `doom--help-insert-button` throwing an error when `help-packages` is called for such a package. Instead insert "n/a" for this section, similar to other sections without data.

-----
    - while trivial, this fixes an error preventing display of help


hlissner:
Good catch! Thank for your help (and for your first PR to the project)!

PR #7375, opened 2023-08-26
fix(evil): update evil-collection-mode-list

LemonBreezes:
Hi. I realized that `evil-collection` already lazy-loads all of these modules:
image
so I refactored our `evil-collection` configuration, removing most of the code. I do not notice any startup time difference after removing the optimizations. This way we don't have to hard-code `evil-collection-mode-list`, which is currently out-of-date in Doom anyways. 

The only feature really missing is the debug output telling us if an Evil Collection module was loaded too early.

### DIFF
--- a/modules/editor/evil/init.el
+++ b/modules/editor/evil/init.el
@@ -123,6 +123,7 @@ variable for an explanation of the defaults (in comments). See
       dired
       dired-sidebar
       disk-usage
+      distel
       doc-view
       docker
       ebib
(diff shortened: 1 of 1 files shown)


LemonBreezes:
It seems like with this PR, Doom is pulling in Dired at startup because of https://github.com/emacs-evil/evil-collection/issues/746. I'm gonna make a PR for that issue and try to get it merged.

LemonBreezes:
Gonna close this since the `evil-collection-mode-list` was updated and this PR has gotten stale.

PR #7385, opened 2023-08-31
fix(org): check org-fold-outline in invisible property

LemonBreezes:
This PR addresses https://github.com/doomemacs/doomemacs/issues/7206. Currently, `+org/open-fold` is not working because `(outline-visible-p)` does not check for the `org-fold-outline` property. So I replaced all calls of `(outline-visible-p)` with the redefinition by jesseylin, which checks for `org-fold-outline`.


hlissner:
This may warrant a more systematic fix (maybe an advice), but I'll look into that later. Thanks for the help, in any case!

PR #7402, opened 2023-09-11
fix(evil): more reliable window detection

fosskers:
At some point `windmove-find-other-window` stopped loading reliably, such that `SPC w H` would complain of a missing symbol. Indeed navigating to the implementation of `windmove` within Emacs's Lisp we find warnings that `windmove-find-other-window` has mostly been superceded by newer window APIs since 2013, and is implemented in terms of `window-in-direction`.

By using `window-in-direction` directly, the loading problem disappears.

-----


hlissner (approved):

hlissner:
LGTM. Thanks again!

PR #7404, opened 2023-09-11
feat(vertico): add projectile-completion-system fn

elken:
Allows a nicer view to be displayed. This could, and pending discussion
should, end up upstream instead.

### DIFF
--- a/modules/completion/vertico/autoload/vertico.el
+++ b/modules/completion/vertico/autoload/vertico.el
@@ -259,3 +259,17 @@ targets."
 (defun +vertico-basic-remote-all-completions (string table pred point)
   (and (vertico--remote-p string)
        (completion-basic-all-completions string table pred point)))
+
+(autoload #'consult--read "consult")
+;;;###autoload
+(defun +vertico/projectile-completion-fn (prompt choices)
+  "Given a PROMPT and a list of CHOICES, filter a list of files for
+`projectile-find-file'."
+  (interactive)
+  (consult--read
+   choices
+   :prompt prompt
+   :sort nil
+   :add-history (thing-at-point 'filename)
+   :category 'file
+   :history '(:input +vertico/find-file-in--history)))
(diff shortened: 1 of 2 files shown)


hlissner (approved):

minad:
@elken My knowledge of Doom is quite limited, but I think this change is not needed. Projectile itself uses `completing-read` which should work as is with Vertico and other completion UIs. I know this, since I've contributed the completion UI auto detection to Projectile a while back myself. ;)

What is the advantage of `consult--read` here? My usual recommendation is to only use `consult--read` if you use Consult-specific preview or narrowing, otherwise the basic `completing-read` is the better option, simply since it is generic. If there is an actual advantage ("nicer view"), maybe the `completing-read` call in Projectile could be improved directly instead?

elken:
> maybe the `completing-read` call in Projectile could be improved directly instead

Yes, hence my original comment

minad:
> Yes, hence my original comment

Makes sense. I wondered why you went with `consult--read` here. I suspect only because of the simplicity associated with keyword arguments?

elken:
> Makes sense. I wondered why you went with `consult--read` here. I suspect only because of the simplicity associated with keyword arguments?

The regular `completing-read` here doesn't show icons or other marginalia annotations, I figured this was simpler than deep-diving into setting up hooks/advice

minad:
> The regular completing-read here doesn't show icons or other marginalia annotations, I figured this was simpler than deep-diving into setting up hooks/advice

Hmm, the problem is likely only that the `completing-read` call doesn't specify a completion category. It is a bit painful, but you can do that via the `metadata` action, see https://www.gnu.org/software/emacs/manual/html_node/elisp/Programmed-Completion.html.

minad:
If you need some examples you can also take a look at `cape--table-with-properties` or `embark--with-category`. These helpers alleviate the pain of writing a completion table slightly.

hlissner:
With plans to replace projectile with project.el on the horizon, I'm closing any open PRs that tweak projectile's behavior. Thanks for the PR, in any case!

PR #7405, opened 2023-09-11
tweak(vterm): vterm-send-next-key mapped to C-q

stfl:
when using vterm and accidently opening nano (wrong $EDITOR set and using git via cli) closing nano is not possible because sending C-x is not possible.
not even in `<insert-state>`

With `C-q` mapped to `vterm-send-next-key` it's possible to send it via `C-q C-x`


hlissner:
I agree this is out of Doom's purview, so I'll leave it to users to choose which keys to rebind this way (or upstream to find a better workaround). Thanks for bringing it to my attention, in any case.

hlissner (approved):

PR #7406, opened 2023-09-11
docs: use `pacman -Syu` in Arch Linux install instructions

daemonburrito:
Use safest pacman incantation, '-Syu'. The outcome in package list state could be okay for a majority, but it's 100% safe to upgrade when installing.

From the Arch wiki:

* https://wiki.archlinux.org/title/Pacman#Installing_packages
* https://wiki.archlinux.org/title/System_maintenance#Partial_upgrades_are_unsupported (could happen...)

Trivial docs, no issue.

Fix: #0000
Ref: #0000
Close: #0000

-----

### DIFF
--- a/docs/getting_started.org
+++ b/docs/getting_started.org
@@ -158,9 +158,9 @@ dnf install fd-find    # is 'fd' in Fedora <28
 **** Arch Linux
 #+BEGIN_SRC bash
 # required dependencies
-pacman -S git emacs ripgrep
+pacman -Syu git emacs ripgrep
 # optional dependencies
-pacman -S fd
+pacman -Syu fd
 #+END_SRC
 
 The above installs Emacs 27 (at the time of writing).


MithicSpirit:
`pacman -S` is completely safe, the only issue is `pacman -Sy` since that updates the package list without updating the packages themselves. There is no need to update the system in this case (unless the system is so old that it cannot find the packages to install).

daemonburrito:
> `pacman -S` is completely safe, the only issue is `pacman -Sy` since that updates the package list without updating the packages themselves. There is no need to update the system in this case (unless the system is so old that it cannot find the packages to install).

tldr Feel free to reject, read below for rationale to approve.

I did expect some push-back, and I stated my feelings about "100%" reliability when comes to the packages list. The "unless" is important, as is possible accidental "y"s. But in the end, it's legal to do `-Sy`. It just seems like it would be in a project's interest to instruct users in installation instructions with foolproof steps, especially with a "high impact, low probability" event like a broken packages list.

Thanks anyway for your time.

MithicSpirit:
Considering that these instructions are for Arch Linux, which is the very antonym of "foolproof", I think that this is fine. Maybe having an additional caveat to use `pacman -Syu` to update the system if the latest version is desired or if it cannot find the packages (exact message is `warning: failed to retrieve some files` followed by `error: failed to commit transaction (failed to retrieve some files)`), with possibly a warning against `pacman -Sy`, would be a good compromise though.

hlissner:

PR #7411, opened 2023-09-12
refactor!: replace all-the-icons with nerd-icons

elken:
Commits will likely need rebasing in some form, but I've gone through and tested as much as I can.

Per discussion on Discord, we can either disable `all-the-icons` here or in a future PR down the line; makes no difference to me

### DIFF
--- a/docs/getting_started.org
+++ b/docs/getting_started.org
@@ -551,9 +551,7 @@ doom sync
 doom env
 
 # Lastly, install the icon fonts Doom uses:
-emacs --batch -f all-the-icons-install-fonts
-# On Windows, `all-the-icons-install-fonts` will only download the fonts, you'll
-# have to install them by hand afterwards!
+emacs --batch -f nerd-icons-install-fonts
 #+END_SRC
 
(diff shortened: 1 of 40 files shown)


hlissner (changes requested):
Could you remove the `doom-ui` scope from 483836d7aa2e2c13ef12f0a52a442ef80cc30925's commit message? Changes to core should have no scope (except for `lib` for the stuff in `lisp/lib/*.el`).

elken:
> ~Could you remove the `doom-ui` scope from [483836d](https://github.com/doomemacs/doomemacs/commit/483836d7aa2e2c13ef12f0a52a442ef80cc30925)'s commit message? Changes to core should have no scope (except for `lib` for the stuff in `lisp/lib/*.el`).~
> 
> Actually, scratch that. I'll likely squash this PR, and will correct it then.

Yeah this should probably be 1 commit, I just wanted to split it up in case there were breaking changes with ivy/helm/company since they were more involved than simple changes

hlissner (approved):

PR #7420, opened 2023-09-14
fix(tools): replaced obsolete featurep! macro

lagman:
The macro  `featurep!` has been obsolete as of `3.0.0`. 
-----
It was lingering on in the module `collab`. I replaced it with the `modulep!`, macro


hlissner:
Good catch! There was another `featurep!` reference in doctor.el, so I've incorporated this into cd2477e. Thanks for bringing it to my attention, in any case!

PR #7421, opened 2023-09-14
fix(format): load after correct library

elken:
This allows the macro to run late enough that user config overrides the
internal one


elken:
Hmm this doesn't quite seem to work still

elken:
Not needed anymore

PR #7427, opened 2023-09-15
fix(format): update missing values & set buffer modification

elken:
This prevents apheleia complaining about the editing buffer being
killed.

From https://github.com/doomemacs/doomemacs/pull/6369#issuecomment-1720717000


hlissner (dismissed):

hlissner:
Thanks again!

PR #7429, opened 2023-09-16
fix(format): use +vc-gutter-update-h instead

LemonBreezes:
The function `+vc-gutter-init-maybe-h` is not defined for `+diff-hl` so I am getting a void function error without this PR.


hlissner:
Good catch. Thanks again!

PR #7436, opened 2023-09-17
fix(ligatures): proper resetting of font-ligatures

gagbo:
resetting font-ligatures means passing `nil` as the second argument of `(set-font-ligatures!)`

Using `ligatures-ignored-major-modes` to cancel ligatures is too brittle, because "resetting ligatures for `prog-mode`" would have bad semantics with all its derived modes. The user probably just wants to remove the default ligatures and then configure `foo-mode`, but pushing `prog-mode` to `ignored-major-modes` might just disable ligatures across all modes _derived from_ `prog-mode`.

This commit changes things in 2 ways:
- resetting ligatures now directly manipulates ligature.el internal state (the `ligature-composition-table` alist)
- in order to work, Doom must maintain an extra invariant on that alist, multi-modes keys (`'(foo-mode bar-mode)`), cannot be used, only single modes.

That mostly means that users should _not_ use
`ligature-set-ligatures` themselves in private config, but instead always rely on `set-font-ligatures!` which does splicing behind curtains. Failing to do so would be mostly harmless though (it would just make "resetting ligatures" only partially remove set ligatures).

Fix: #7433 
Close: #7433

-----
- [x] ~~My changes are visual; I've included before and after screenshots.~~
- [x] I am _not_ blindly checking these off.
- [x] ~~This a draft PR; I need more time to finish it.~~

### DIFF
--- a/modules/ui/ligatures/autoload/ligatures.el
+++ b/modules/ui/ligatures/autoload/ligatures.el
@@ -67,10 +67,15 @@ Font ligatures can be unset for emacs-lisp-mode with:
 Note that this will keep all ligatures in `+ligatures-prog-mode-list' active, as
 `emacs-lisp-mode' is derived from `prog-mode'."
   (declare (indent defun))
+  ;; NOTE: Doom enforces `ligature-composition-table' to have a single mode per key in the alist.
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
LGTM. Thanks again!

PR #7443, opened 2023-09-20
Fix Org docs links

hpfr:
The fancy Org docs links had some regressions with the switch to Nerd Fonts. In the process of addressing those, I ended up finding and taking care of several other quirks. See the commit messages for more details.

Before:

image

Note the changed codepoint for the variable icon resulting in a folder glyph from Nerd Fonts as well as the scissors in place of the package Octicon. Also, the face icon (U+E90F, ) will appear as tofu for people who don’t happen to have all-the-icons installed, Nerd Fonts don’t have a glyph at that code point.

After:

image

Note the shadowed unbound symbols and the passthrough of the `fixed-pitch-serif` face (it’s a bit subtle in this particular case, but you can at least see the difference in the `f` glyphs). Packages (italic) and modules (bold) that are found on their load paths get blue links, and disabled modules and unloaded packages get `warning` icons to align with their `help-echo` messages. Since executables don’t have a follow function for more information, they don’t look like links.

There were often breadcrumbs in the existing implementation indicating that some of these behaviors were intended from the start; they just needed a little polish.

Here’s the source for the example:
```org
- [[var:doom-font]]
- [[var:foo]]
- [[fn:doom-big-font-mode]]
- [[fn:foo]]
- [[face:fixed-pitch-serif]]
- [[face:foo]]
  
- [[doom-package:org]]
- [[doom-package:eglot]]
- [[doom-package:lsp-mode]]
- [[doom-module::tools lsp]]
- [[doom-module::tools ansible]]
- [[doom-module::tools foo]]
- [[doom-executable:latex]]
- [[doom-executable:gnuplot]]
```

-----
- [x] *n/a* Any relevant issues or PRs have been linked to.

### DIFF
--- a/lisp/lib/docs.el
+++ b/lisp/lib/docs.el
@@ -372,7 +372,7 @@ depending.")
 (defvar doom-docs--cookies nil)
 ;;;###autoload
 (define-minor-mode doom-docs-mode
-  "Hides metadata, tags, & drawers and activates all org-mode pretiffications.
+  "Hides metadata, tags, & drawers and activates all org-mode prettifications.
 This primes `org-mode' for reading."
   :lighter " Doom Docs"
   :after-hook (org-restart-font-lock)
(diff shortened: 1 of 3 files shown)


hlissner:
LGTM. Great work! Thanks again for your help!

PR #7444, opened 2023-09-20
feat(pdf): prefer built-in pdf-tools

hpfr:
Nixpkgs provides a pdf-tools with prebuilt epdfinfo, much like vterm. It works on my machine. I added an entry to the docs similar to the vterm module’s, but I held off on distros I’m not able to test myself. People who use a built-in package on other distros can make PRs as desired.

Based on your unconditional use of the `:built-in` keyword for vterm, I assume this gracefully falls back to installing pdf-tools like normal if no built-in package is available?

-----
- [x] *n/a* Any relevant issues or PRs have been linked to.

### DIFF
--- a/modules/tools/pdf/README.org
+++ b/modules/tools/pdf/README.org
@@ -42,6 +42,16 @@ This module requires the =epdfinfo= program. Unless you're on Windows, the
 pdf-tools-install~ command. See the next section for instructions on how to
 build the =epdfinfo= program on Windows.
 
+Some Linux distributions package [[doom-package:pdf-tools]] with a prebuilt
+=epdfinfo= program.
+- NixOS (home-manager):
+  #+begin_src nix
+  programs.emacs = {
+    enable = true;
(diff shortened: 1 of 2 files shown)


hlissner:
I'm conflicted about this PR. vterm gets a `:built-in 'prefer` because `libvterm` is a required, but unmanaged, external dependency. Packages like mu4e, vterm, and agda2-mode are similar in this way, and so they are some of the most troublesome packages for Doom to support (since they can't reasonably be pinned).

pdf-tools, on the other hand, *does* manage its own dependency (epdfinfo), and with an installer that is reasonably competent on many systems. I'd rather `:built-in 'prefer` be opt-in where possible, rather than opt-out.

That said, pdf-tools doesn't change anywhere near as often as any of the other packages, so I might be willing to make it an exception. I'll sit on this for a while.

> Based on your unconditional use of the :built-in keyword for vterm, I assume this gracefully falls back to installing pdf-tools like normal if no built-in package is available?

Yeah, that's correct.

PR #7451, opened 2023-09-21
fix(ligatures): set-font-ligatures! reset on nil

andresilva:

Fix: #7433
Ref: #7433
Close: #7433

https://github.com/doomemacs/doomemacs/issues/7433#issuecomment-1729267699

cc @gagbo

-----

### DIFF
--- a/modules/ui/ligatures/autoload/ligatures.el
+++ b/modules/ui/ligatures/autoload/ligatures.el
@@ -70,7 +70,7 @@ Note that this will keep all ligatures in `+ligatures-prog-mode-list' active, as
   ;; NOTE: Doom enforces `ligature-composition-table' to have a single mode per key in the alist.
   ;; This is less efficient than what ligature.el can do (i.e. use a list of modes, or `t' as a key),
   ;; but holding this invariant allows resetting with `(set-font-ligatures! 'mode nil)` to work reliably.
-  (if (null ligatures)
+  (if (or (null ligatures) (equal ligatures '(nil)))
       (dolist (mode (ensure-list modes))
         (delq! mode ligature-composition-table 'assq))
     (after! ligature


andresilva:
FWIW I still find the interface of this function to be weird. Since you first need to call this function with `nil` to clean up the existing ligatures for a certain mode, and then do another call with the ligatures you want. IMO calling this function should reset the existing ligatures for that mode and then set the given ligatures, that's what I would expect from something called "set".

gagbo:
> FWIW I still find the interface of this function to be weird. Since you first need to call this function with `nil` to clean up the existing ligatures for a certain mode, and then do another call with the ligatures you want. IMO calling this function should reset the existing ligatures for that mode and then set the given ligatures, that's what I would expect from something called "set".

The reason for making this additive by default is that it allows other modules to add ligatures using the same interface without worrying about other calls or order of the calls. Theoretically the reset with `nil` is only called once in user config if you want to reclaim full control of the ligations happening. I guess we could split the `set-font-ligatures` in 2, 1 `add` and 1 `reset`, but always resetting is going to make debugging user issues a lot harder when file load order starts to become relevant

gagbo (approved):
And yeah, I fully missed that having an explicit `nil` argument would change the `&rest` value. Thanks for that!

andresilva:
> The reason for making this additive by default is that it allows other modules to add ligatures using the same interface without worrying about other calls or order of the calls. Theoretically the reset with `nil` is only called once in user config if you want to reclaim full control of the ligations happening. I guess we could split the `set-font-ligatures` in 2, 1 `add` and 1 `reset`, but always resetting is going to make debugging user issues a lot harder when file load order starts to become relevant

Hadn't considered this, makes sense. Thanks for the explanation.

andresilva:
ping

hlissner:
LGTM! Sorry for taking so long to get around to this. Thanks for your help!

PR #7458, opened 2023-09-24
fix(vertico): use consult-dir/fd for find

LemonBreezes:
Use the built-in functions for `fd` and `find` instead for `vertico`. These are better because ours loads all the files into `vertico` immediately, which is slow for large directories, whereas `consult-fd` and `consult-find` does the searches asynchronously. I also added an optional prefix arg to query for a directory before searching.

Please check that these vertico commands exist in your pinned version of `vertico`, since I have all my packages unpinned.

-----

### DIFF
--- a/lisp/lib/projects.el
+++ b/lisp/lib/projects.el
@@ -149,9 +149,10 @@ If DIR is not a project, it will be indexed (but not cached)."
             (if (doom-module-p :completion 'ivy)
                 #'counsel-projectile-find-file
               #'projectile-find-file)))
-          ((and (bound-and-true-p vertico-mode)
-                (fboundp '+vertico/find-file-in))
-           (+vertico/find-file-in default-directory))
+          ((bound-and-true-p vertico-mode)
+           (if (executable-find "fd")
+               (consult-fd dir)
+             (consult-find dir)))
           ((and (bound-and-true-p ivy-mode)
                 (fboundp 'counsel-file-jump))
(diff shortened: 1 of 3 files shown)


LemonBreezes:
Also, it might be worth replacing `projectile-find-file` with these commands since they don't front-load all the candidates and are asynchronous.

iyefrat:
Hey, unfortunately this PR is pretty incomplete - it doesn't bump the packages needed for these features to work, has the wrong name for the commits (not using `consult-fd` isn't a bug, so it should be `fix:`), it doesn't remove all uses of `+vertico/find-file-in`, and doesn't remove the custom `+vertico/consult-fd` that doom currently has.

Reviewing it properly would take more time than writing a vertico bump PR that would have incorporated this migration anyway, where I've added you as a co-author: #7463. Ar any rate, thanks for bringing this to our attention!

LemonBreezes:
> Hey, unfortunately this PR is pretty incomplete - it doesn't bump the packages needed for these features to work, has the wrong name for the commits (not using `consult-fd` isn't a bug, so it shouldn't be `fix:`), it doesn't remove all uses of `+vertico/find-file-in`, and doesn't remove the custom `+vertico/consult-fd` that doom currently has.
> 
> Reviewing it properly would take more time than writing a vertico bump PR that would have incorporated this migration anyway, where I've added you as a co-author: #7463. Ar any rate, thanks for bringing this to our attention!

That's great. Thanks!

PR #7473, opened 2023-09-29
fix: set doom-unicode-font back to nil by default

seanfarley:
Previous to this commit, common emoji didn't display correctly on macOS.
This variable was changed in 9787022b839d2d22cb772c6d61cc7f35b9dd5e95
but it *seems* to be an oversight because,

1) in `+unicode-setup-fonts-h' there is logic that checks for
`doom-unicode-font' being nil as a default setting for
`doom-unicode-font-family'

and

2) the commit that changed this was about changing all-the-icons to
nerd-icons which doesn't seem to need a change to doom-unicode-font.

### DIFF
--- a/lisp/doom-ui.el
+++ b/lisp/doom-ui.el
@@ -36,7 +36,7 @@ Must be a `font-spec', a font object, an XFT font string, or an XLFD string. See
 
 An omitted font size means to inherit `doom-font''s size.")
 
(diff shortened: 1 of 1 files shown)


seanfarley:
Before, missing font:

Screenshot 2023-09-28 at 8 13 12 PM

After,
Screenshot 2023-09-28 at 8 39 56 PM

seanfarley:
Oh snap! That does indeed fix this issue, thanks!

PR #7474, opened 2023-09-29
mu4e: add icons for mu4e's internal modeline

seanfarley:
`mu4e` now has its own modeline that is quite flexible. These changes add icons inline with doom's style.

Also adds `mu4e`'s buffers to a popup rule since it now uses display-buffer-alist.

### DIFF
--- a/modules/email/mu4e/autoload/email.el
+++ b/modules/email/mu4e/autoload/email.el
@@ -113,7 +113,7 @@ will also be the width of all other printable characters."
       (insert str)
       (car (window-text-pixel-size)))))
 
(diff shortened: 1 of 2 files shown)


LemonBreezes:
Looks okay to me. In fact, I just noticed that since we fontify the search symbols as help buttons, they show up as underlined in my modeline. X_X
image

seanfarley:
Ah, so that's why they're underlined!

LemonBreezes:
> Ah, so that's why they're underlined!

My guess is that before they gave no help echo but Mu4e added them and now that code is obsolete.

hlissner (approved):

hlissner:

PR #7487, opened 2023-10-05
fix(rust): make cargo popups consistent

arnfaldur:
When invoking `rustic-cargo-run` a buffer called `*cargo-run*` is opened. This adds a rule such that this buffer opens in a popup window that is the same as the popup windows for other rustic cargo commands.

This PR is adding to the solution to [this issue](https://github.com/doomemacs/doomemacs/issues/2623).

-----


hlissner (approved):

hlissner:

PR #7497, opened 2023-10-11
fix(nix): make doctor check against nix-nixfmt-bin

goofansu:
Doom doctor should check against `nix-nixfmt-bin` instead of hard-coded.

- [x]  I searched the issue tracker and this hasn't been PRed before.
- [x]  My changes are not on [the do-not-PR list](https://doomemacs.org/d/do-not-pr) for this project.
- [x]  My commits conform to [the git conventions](https://doomemacs.org/d/git-conventions).


hlissner (approved):

hlissner:
I want to avoid setting a precedent for relying on runtime configuration in doctor checks (it's very brittle and depends a lot on the integrity of the user's config), but I'll merge this for the time being, at least until I come up with a better alternative. Thanks for the help, in any case!

goofansu:
@hlissner Thanks, the default formatter in `nix-mode` is `nixfmt`, so it is not necessary to set manually unless you use a different formatter such as `nixpkgs-fmt`.

PR #7498, opened 2023-10-11
fix(projectile): disable colors in fd invocation

noisycomputation:
bbatsov/projectile@971cd5c4f25f -> bbatsov/projectile@d24b8173223f

The pinned `projectile` version erroneously calls `fd` without explicitly disabling colors, which inserts color terminal codes into filenames when used over `tramp`. This completely breaks projectile over `tramp`.

The linked change adds `-c never` (shorthand form of `--color=never`) to `projectile-git-fd-args`. This is the only change in `projectile` between the two versions linked above, and it completely resolves the issue.

Ref: https://github.com/bbatsov/projectile/issues/1860
Ref: https://github.com/bbatsov/projectile/pull/1859

-----

### DIFF
--- a/lisp/packages.el
+++ b/lisp/packages.el
@@ -40,7 +40,7 @@
   :pin "572a10c11b6cb88293de48acbb59a059d36f9ba5")
 
 ;; doom-projects.el
-(package! projectile :pin "971cd5c4f25ff1f84ab7e8337ffc7f89f67a1b52")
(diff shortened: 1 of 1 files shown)


hlissner:
My bad, I only just realized that `projectile` was bumped in c5f561b0c54f9adab1a133ed91d2f9cbd50eeda0, which should address this issue. Thanks for bringing it to my attention, in any case!

PR #7504, opened 2023-10-15
fix(treemacs): do not overwrite git mode when treemacs-python-executable is set

kiana-S:
If the variable `treemacs-python-executable` is set, then the git mode shouldn't be overwritten, even if python cannot be found in the PATH.

-----


hlissner (approved):

hlissner:
LGTM. Thanks for the help!

PR #7509, opened 2023-10-17
fix(org): call org-reveal in correct buffer

Zetagon:
Sometimes, `org-reveal` is called in the wrong buffer which throws an error.  For example, `org-link-open-from-string` creates an temporary org-mode buffer that gets killed very quickly which means that `org-reveal` gets called in a different buffer.  I have also had issues with org-reveal getting called in the org-roam buffer, which is why this commit also saves the buffer it was called in.

-----

### DIFF
--- a/modules/lang/org/autoload/org.el
+++ b/modules/lang/org/autoload/org.el
@@ -510,8 +510,9 @@ All my (performant) foldings needs are met between this and `org-show-subtree'
        ;; when `org-mode' first initializes.
        (let ((buf (current-buffer)))
          (unless (doom-temp-buffer-p buf)
-           (run-at-time 0.1 nil (λ! (with-current-buffer buf
-                                      (org-reveal '(4)))))))))
+           (run-at-time 0.1 nil (lambda ()
+                                  (with-current-buffer buf
+                                    (org-reveal '(4)))))))))
 
 ;;;###autoload
 (defun +org-remove-occur-highlights-h ()
--- a/modules/lang/org/autoload/org.el
+++ b/modules/lang/org/autoload/org.el
@@ -508,7 +508,10 @@ All my (performant) foldings needs are met between this and `org-show-subtree'
        ;; Must be done on a timer because `org-show-set-visibility' (used by
        ;; `org-reveal') relies on overlays that aren't immediately available
        ;; when `org-mode' first initializes.
-       (run-at-time 0.1 nil #'org-reveal '(4))))
+       (let ((buf (current-buffer)))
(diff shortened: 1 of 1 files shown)


Zetagon:
> I think you can drop the commented code though?

You saw nothing :eyes:

modules/lang/org/autoload/org.el
```diff
@@ -508,7 +508,10 @@ All my (performant) foldings needs are met between this and `org-show-subtree'
        ;; Must be done on a timer because `org-show-set-visibility' (used by
        ;; `org-reveal') relies on overlays that aren't immediately available
        ;; when `org-mode' first initializes.
-       (run-at-time 0.1 nil #'org-reveal '(4))))
+       (let ((buf (current-buffer)))
+         (unless (doom-temp-buffer-p buf)
+           (run-at-time 0.1 nil (λ! (with-current-buffer buf
+                                      (org-reveal '(4)))))))))
```
hlissner, line 514: Please use `lambda` instead of `λ!`. 

`λ!` is an alias for `cmd!` which only exists for backwards compatibility in user elisp, and shouldn't be used within Doom's source. It also produces an interactive closure, which `run-at-time` does not benefit from.

hlissner (changes requested):

modules/lang/org/autoload/org.el
Zetagon: Fixed.

hlissner (approved):

hlissner:
LGTM. Thanks for your help!

modules/lang/org/autoload/org.el
```diff
@@ -508,7 +508,11 @@ All my (performant) foldings needs are met between this and `org-show-subtree'
        ;; Must be done on a timer because `org-show-set-visibility' (used by
        ;; `org-reveal') relies on overlays that aren't immediately available
        ;; when `org-mode' first initializes.
-       (run-at-time 0.1 nil #'org-reveal '(4))))
+       (let ((buf (current-buffer)))
+         (unless (doom-temp-buffer-p buf)
+           (run-at-time 0.1 nil (lambda ()
+                                  (with-current-buffer buf
+                                    (org-reveal '(4)))))))))
```
real-or-random, line 515:
```suggestion

                                  (when (buffer-live-p buf)
                                    (with-current-buffer buf
                                      (org-reveal '(4))))))))))
```

real-or-random (commented):
This breaks `org-roam-capture` (and `org-capture`) for me because this command uses multiple org buffers and the initial one will be dead when the timer runs.

PR #7521, opened 2023-10-26
perf(lsp): turn off eglot events buffer

vlcek:
The `eglot-events-buffer-size` setting disables the `eglot-events-buffer` when 0, enabling more consistent performance on long running emacs instance. Default is 2000000 lines. After each new event the events buffer is pretty printed as a whole, which causes steady performance decrease over time. Quite a bit of CPU is spent on pretty printing and Emacs GC is put under high pressure.

Fix: #0000
Ref: #0000
Close: #0000

-----

### DIFF
--- a/modules/tools/lsp/+eglot.el
+++ b/modules/tools/lsp/+eglot.el
@@ -16,6 +16,12 @@
   (setq eglot-sync-connect 1
         eglot-autoshutdown t
         eglot-send-changes-idle-time 0.5
+        ;; NOTE This setting disable the eglot-events-buffer enabling more
+        ;;      consistent performance on long running emacs instance.
+        ;;      Default is 2000000 lines. After each new event the whole buffer
+        ;;      is pretty printed which causes steady performance decrease over time.
+        ;;      CPU is spent on pretty priting and Emacs GC is put under high pressure.
+        eglot-events-buffer-size 0
         ;; NOTE We disable eglot-auto-display-help-buffer because :select t in
         ;;      its popup rule causes eglot to steal focus too often.
         eglot-auto-display-help-buffer nil)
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
lsp-mode has a similar issue with its log buffers and oversized payloads from chatty clients (necessitating the increase of `read-process-output-max`). In any case, I prefer the 'log as little as possible if debug mode is off' approach for a number of reason (not least of which are packages that abuse temp buffers and writes to `*Messages*` triggering premature redraws of the Emacs frame), so I am happy to default `eglot-events-buffer-size` to 0.

Anyhow, thanks for the PR!

vlcek:
> Got any profiling data? I wonder if setting it to a smaller value (like 100000) would bound the performance impact more tightly while preserving the ability to browse recent communication with LSP servers.

I had. I does not remember the details, however, when you got to the point the events buffer was quite big (~1M+) something like ~50-70% CPU was consumed just by pretty printing the whole buffer, rest (30-40%) was garbage collection.

At first, I tried to find ways to turn off the pretty printing but to no avail (except nasty advice functions). I also thought about limiting the number of events in the buffer, however, when I saw the debug mode option then I saw no real point of leaving it enabled.
Using LSPs is already a big hit for the responsiveness so I want to minimize it.

> lsp-mode has a similar issue with its log buffers and oversized payloads from chatty clients (necessitating the increase of `read-process-output-max`). In any case, I prefer the 'log as little as possible if debug mode is off' approach for a number of reason (not least of which are packages that abuse temp buffers and writes to `*Messages*` triggering premature redraws of the Emacs frame), so I am happy to default `eglot-events-buffer-size` to 0.
> 
> Anyhow, thanks for the PR!

You're welcome! It was about time for me to do some contribution since I am using doomemacs daily for the last few years :) Thanks for the awesome work you all are doing here!

PR #7525, opened 2023-10-28
refactor(evil): remove lisp prefix-function embrace pair

LemonBreezes:
Hello. So my PR to Embrace to add prefix-function to the default pairs was accepted (https://github.com/cute-jumper/embrace.el/pull/26). Now we can use those anywhere using `C-f`. This is better than our current DWIM-style `f` because that way we can write Lisp anywhere and not have to remember if we are actually in a Lisp buffer when we call `embrace-commander`.

By the way, Embrace is also not pinned, I see. Not sure if that is important.

-----

### DIFF
--- a/modules/editor/evil/config.el
+++ b/modules/editor/evil/config.el
@@ -245,8 +245,6 @@ directives. By default, this only recognizes C directives.")
   :hook (org-mode . embrace-org-mode-hook)
   :hook (ruby-mode . embrace-ruby-mode-hook)
   :hook (emacs-lisp-mode . embrace-emacs-lisp-mode-hook)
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
LGTM. THanks again!

hey-kader:
LITHP

LemonBreezes:
> After updating doom again, I no longer get the error, but the behavior is still different from before:
> `dsC-f` does what `dsf` used to do before.
> `dsf` seems to just remove the closing parenthesis of a sexp.
> @LemonBreezes is it possible to restore the previous behavior easily? I think it was much better.

I'm not at my computer for a few days but I expect you can re-add the code that I deleted and that will work basically.

PR #7527, opened 2023-10-28
feat: use setq! in helpful

LemonBreezes:
Hi. Recently my `helpful` PR for using a custom set variable function has been accepted (https://github.com/Wilfred/helpful/pull/332). This means that now we can use `setq!` instead of `setq` for the "Set"  button in `helpful`. This is good because that way we trigger the variable's `:set` property.

-----

### DIFF
--- a/lisp/doom-editor.el
+++ b/lisp/doom-editor.el
@@ -584,7 +584,9 @@ current buffer."
                    filename))
           (prog1 (apply fn args)
             (when (buffer-live-p buf)
-              (with-current-buffer buf (goto-char pos)))))))))
+              (with-current-buffer buf (goto-char pos))))))))
+  :config
+  (setq helpful-set-variable-function #'setq!))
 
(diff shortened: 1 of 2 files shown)


hlissner:
I totally jumped the gun on this PR. I thought `setopt` was the better choice for this particular use-case (what sets `setq!` apart is that it avoids some package loading machinery tied to `custom-set`'s type checker, but that's only useful for some performance gains at startup; less so for interactive variable changes in helpful buffers), but what I forgot about was backwards compatibility (i.e. `setopt` was introduced in 29.x and is therefore unavailable to 27/28 users). (Side note: I also thought this should've been a bump commit, rather than a `feat`).

Anyhow, sorry about that! After 1bc5f441f00a06e61087e59e94e8ce1642ffeea6 this PR has effectively been fully merged. Thanks for your help, as always!

PR #7529, opened 2023-10-29
fix(eshell): fix elisp autocompletion in Eshell

LemonBreezes:
For a long time I have wondered why Elisp autocompletion is broken in Eshell. After reading the code for `eshell-syntax-highlighting`, it is clear:
image
The `post-command-hook` is deleting the entire prompt string and re-inserting it with syntax highlighting. This is likely messing with overlays and/or markers that Corfu & Company use to store their completion state. I think the best decision is to disable this feature entirely.

I don't personally know how to highlight a portion of the buffer with Elisp highlighting without modifying the text but if anyone does, please follow up.

EDIT: I made an issue about this in https://github.com/akreisher/eshell-syntax-highlighting/issues/14.

-----

### DIFF
--- a/modules/term/eshell/config.el
+++ b/modules/term/eshell/config.el
@@ -229,8 +229,10 @@ Emacs versions < 29."
   (eshell-did-you-mean-output-filter "catt: command not found"))
 
 
-(use-package eshell-syntax-highlighting
-  :hook (eshell-mode . eshell-syntax-highlighting-mode))
+(use-package! eshell-syntax-highlighting
+  :hook (eshell-mode . eshell-syntax-highlighting-mode)
+  :init
+  (add-hook 'eshell-syntax-highlighting-elisp-buffer-setup-hook #'highlight-quoted-mode))
 
 
 (use-package! fish-completion
--- a/modules/term/eshell/packages.el
+++ b/modules/term/eshell/packages.el
@@ -6,7 +6,7 @@
 (package! shrink-path :pin "c14882c8599aec79a6e8ef2d06454254bb3e1e41")
 (package! esh-help :pin "417673ed18a983930a66a6692dbfb288a995cb80")
 (package! eshell-did-you-mean :pin "80cd8c4b186a2fb29621cf634bcf2bcd914f1e3d")
-(package! eshell-syntax-highlighting :pin "8bf0494ca71944b9d4bfb8ec3c93ea29d46bc2f9")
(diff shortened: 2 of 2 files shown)


aikrahguzar:
> I don't personally know how to highlight a portion of the buffer with Elisp highlighting without modifying the text but if anyone does, please follow up.

I think one approach can be to look at `comint-fontify-input-mode`. From a cursory glance, it uses an indirect buffer which should work pretty well since I think text properties are shared between the indirect buffer and base buffer.

LemonBreezes:
This PR is pending the following PR to be merged for `eshell-syntax-highlighting`: https://github.com/akreisher/eshell-syntax-highlighting/pull/15

After that, I will convert this PR to a simple bump.

modules/term/eshell/config.el
```diff
@@ -230,7 +230,9 @@ Emacs versions < 29."
 
 
 (use-package eshell-syntax-highlighting
```
akreisher, line 232: @LemonBreezes Not sure if you know this, but is there a reason this is the only defun here that does not use the doom `use-package!` ]macro? I don't use doom myself, so not sure how the doom macro defers from the stock use-package.

modules/term/eshell/config.el
```diff
@@ -230,7 +230,9 @@ Emacs versions < 29."
...
-  :hook (eshell-mode . eshell-syntax-highlighting-mode))
+  :hook (eshell-mode . eshell-syntax-highlighting-mode)
+  :init
+  (add-hook 'eshell-syntax-highlighting-indirect-buffer-setup-hook #'highlight-quoted-mode))
```
akreisher, line 235: This hook was renamed:
```suggestion
  (add-hook 'eshell-syntax-highlighting-elisp-buffer-setup-hook #'highlight-quoted-mode))
```

LemonBreezes: Thanks. I did not realize I was not on the latest commit.

modules/term/eshell/config.el
LemonBreezes: Actually, that is a mistake in the code I did not notice. The impact that has is that the user cannot disable that `use-package` block using `(disable-packages! eshell-syntax-highlighting)` and the user cannot use the `:after-call` nor `:defer-incrementally` properties.

hlissner:
Thanks for the PR! Since another package needed a bump in that module, I've merged this into a larger bump commit (plus, this needed to be a `bump` commit rather than a `fix`, since it involves changing a package's pin). Thanks for bringing it to my attention in any case!

PR #7536, opened 2023-11-06
fix(fold): reorder fold type checks

valrus:
`+fold--ts-fold-p` just checks that treesitter mode is active and the `ts-fold` feature is active, which will block any other fold-method checks following it. So having the `cond` conditions for folds out of order can lead to inconsistent behavior if using specific folding mechanisms directly.

Move it to the end so other fold types have a chance, and consistently order fold type checks so unfolding will (more likely) use the same underlying mechanism as folding.

Here's a repro, which (for me) this PR resolves:
Create this Python file and, with point on `class`, call `(hs-hide-level)`.
```
class TestHideShow:
    def __init__(self):
        pass
```
You get:
```
class TestHideShow:
    def __init__(self):  [ ... ]
```
Move point to `def` and call `+fold/open`. It will do nothing — it's a `hideshow` fold but `+fold/open` tries to unfold it with 
`ts-fold-open`. It can be opened with `hs-show-block`.

As an aside, when I searched for related issues, I found https://github.com/doomemacs/doomemacs/issues/6737 and was unable to repro it, with or without this fix.

-----

### DIFF
--- a/modules/editor/fold/autoload/fold.el
+++ b/modules/editor/fold/autoload/fold.el
@@ -75,8 +75,8 @@ Targets `vimmish-fold', `hideshow', `ts-fold' and `outline' folds."
            (cl-letf (((symbol-function #'outline-hide-subtree)
                       (symbol-function #'outline-hide-entry)))
              (outline-toggle-children)))
-          ((+fold--ts-fold-p) (ts-fold-toggle))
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Sorry for taking so long to get to this. Thanks for your help!

PR #7543, opened 2023-11-11
feat(racket): add support for +hash-lang

sorawee:
Racket Mode recently added `racket-hash-lang-mode` as an alternative major mode to the old `racket-mode`. This PR adds `+hash-lang` to support this new major mode. It also bumps the pinned commit to the latest version which includes the `racket-hash-lang-mode` feature.

-----


hlissner:
With 1baebda, I've opened for a slightly different, less macro-heavy approach, but now the racket module as a `+hash-lang` flag. Let me know if it gives you any trouble. Thanks for bringing it to my attention!

PR #7546, opened 2023-11-14
module: add `:lang graphviz`

nbfalcon:
- org-babel
- graphviz-dot-mode -> syntax highlighting
- error-checking using flycheck
- doom doctor

I've used lang/plantuml as a template (cp -r).

- [X] I searched the issue tracker and this hasn't been PRed before.
- [X] My changes are not on [the do-not-PR list](https://doomemacs.org/d/do-not-pr) for this project.
- [X] My commits conform to [the git conventions](https://doomemacs.org/d/git-conventions).
- [ ] I am blindly checking these off.
- [X] Any relevant issues or PRs have been linked to.
- [ ] This a draft PR; I need more time to finish it.

PR #7547, opened 2023-11-14
fix(plantuml): flycheck: executable support

nbfalcon:
The flycheck-plantuml is by default configured to always run plantuml via "java". This only works with a downloaded plantuml. However, I would prefer to have plantuml installed via my package manager (fedora/dnf).

A locally installed PlantUML executable is already detected by default for normal use (export/preview), we can also use it for flycheck.

If plantuml is downloaded using plantuml-download-jar, this jar is still used by default (see setq plantuml-exec-mode), so this should not affect previous setups.

Fix: #0000
Ref: #7546 (implemented together with that feature/diagrams as well)
Close: #0000

-----
- N/A My changes are visual; I've included before and after screenshots.

### DIFF
--- a/modules/lang/plantuml/config.el
+++ b/modules/lang/plantuml/config.el
@@ -17,7 +17,11 @@
 (use-package! flycheck-plantuml
   :when (modulep! :checkers syntax)
   :after plantuml-mode
-  :config (flycheck-plantuml-setup))
+  :config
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
LGTM. Thanks for your help!

PR #7549, opened 2023-11-17
fix(format): correct behaviour when disabling apheleia

John-Goff:
Edit 2023/11/21 - added an extra commit and updated the title to better reflect the PR's contents.

Previously, it was impossible to disable saving in certain modes. This was due to 
1) Improper feature name in `after!` call
2) Docs referencing behaviour which was not implemented in code
3) Docs not updated properly after 4ecd616cd8f60640d6b0be2ffc3d762e88099bb4 updated this module to use apheleia

Original PR text follows:

------
I was looking into forcing apheleia to always use the LSP formatter if available and wasn't sure if that should be done here or directly in the apheleia code. In the process of investigating that, I found several issues with the format module docs. I figured those should be addressed in a separate PR, I'm still looking into the LSP stuff but that may or may not come after this depending on if I can get it working right.

-----

### DIFF
--- a/modules/editor/format/README.org
+++ b/modules/editor/format/README.org
@@ -60,8 +60,9 @@ When this flag is enabled, you shouldn't need to do anything other than write
 code and save it.
 
 ** Without +onsave
-Without the flag, formatting will only occur when either =apheleia-format-buffer=
-or =+format/buffer= is called.
+Without the flag, formatting will only occur when either =+format/buffer=
+or =apheleia-format-buffer= is called. The difference between them is
+=+format/buffer= will use a LSP server if configured and available.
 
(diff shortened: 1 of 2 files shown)


John-Goff:
Done! That was the one commit I was unsure about, the function name felt kinda confusing to me but I'd be happy to revert that change if it's not desired.

hlissner (approved):

hlissner:
LGTM! Thanks for your help!

>  the function name felt kinda confusing to me but I'd be happy to revert that change if it's not desired

That's alright. I plan to do a naming convention 'check-up' sometime next year to revisit these oddities, so I'm not too fussy about them for now.

PR #7559, opened 2023-11-24
fix(latex): avoid stealing focus after compilation

maoli17:
The current LaTeX module uses `TeX-command-run-all` for compilation, which by default opens the compiled document in a viewer. This behavior causes a loss of focus from the Emacs window, interrupting the editing workflow.

I address this by adding a custom compilation function, which saves the documents and compiles the project, but does not open it in a viewer. This allows users to continue editing without the need to manually switch back the focus to Emacs.

`TeX-command-run-all` is still a useful automation, so I keep it there but change the key binding.

-----


hlissner (approved):

hlissner:
Thanks for your help!

PR #7515, opened 2023-10-20
bump(clojure): new CIDER release

zilti:
- [X] I searched the issue tracker and this hasn't been PRed before.
- [X] My changes are not on [the do-not-PR list](https://doomemacs.org/d/do-not-pr) for this project.
- [X] My commits conform to [the git conventions](https://doomemacs.org/d/git-conventions).
- [ ] My changes are visual; I've included before and after screenshots.
- [ ] I am blindly checking these off.
- [X] Any relevant issues or PRs have been linked to.
- [ ] This a draft PR; I need more time to finish it.

This is not a draft, but I wonder if we can remove the hack for Paredit - it seems that CIDER does not (anymore) depend on paredit.

PR #7561, opened 2023-11-24
Bump :lang clojure

elken:
Bump to the latest versions of everything & add a note about using `enrich-classpath`. 

@vemv my understanding from the docs is this is still a wip flag, or is it ready to be a default yet?

Closes #7515

### DIFF
--- a/modules/lang/clojure/README.org
+++ b/modules/lang/clojure/README.org
@@ -43,6 +43,8 @@ This module adds support for the Clojure(Script) language.
 
 This module requires:
 - [[https://clojure.org/][clojure]]
+
+This module optionally requires:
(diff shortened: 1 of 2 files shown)


vemv:
I'd consider it almost ready, but it also would be a fair policy to simply stick to CIDER's default - we'll change it in CIDER 2.0.0 when we consider this a time-proven solution.

Keeping it disabled is also a good way of ensuring that users understand that there's new machinery undernath, that _may_ have to do with any perceived issues.

elken:
Cool, was thinking the same :smile: I'll leave it as a README note then

hlissner (approved):

hlissner:
Thanks again!

PR #7565, opened 2023-11-26
fix(lsp): eglot: prefer built-in eglot

tbszlg:
Eglot installed via straight caused issues with calls to `consult-imenu` for my emacs installation of version 29.1. The issue caused symbols to not be categorized according to class, function, method, etc. for all languages (other than elisp). Using the built-in version of eglot solves this issue.

As far as I could tell (elisp newbie) the issue was a result of `eglot-imenu` performing wrong formatting for the completion candidates of symbols in the current buffer. The formatting of the candidate list in `consult-imenu` did not result in correct symbol type resolution because of this.

Ref: #7340 - Found this related issue that has the same change as a workaround for a different issue for Emacs 29 installations.

-----

### DIFF
--- a/modules/tools/lsp/packages.el
+++ b/modules/tools/lsp/packages.el
@@ -3,7 +3,9 @@
 
 (if (modulep! +eglot)
     (progn
-      (package! eglot :pin "2b145778ba5e57f393e50aea76b28e518c401828")
+      (package! eglot
+        :pin "2b145778ba5e57f393e50aea76b28e518c401828"
+        :built-in 'prefer)  ; Built-in as of Emacs 29
       (when (modulep! :completion vertico)
         (package! consult-eglot :pin "db9d41c9812a5a8a7b9a22fa7f3c314e37584d41"))
       (when (and (modulep! :checkers syntax)


tbszlg:
I honestly have not verified if a newer version of Eglot installed via straight would have fixed my issues as well.

My reasoning to prefer the built-in installation of Eglot was that this would have been my expected behavior for Emacs 29.1 when using doom. And I imagined that going for a different version than the built-in one would require me to modify the installation in my `packages.el` file.

Would the general philosophy for package version selection prefer to always use a pinned version installed via straight instead?

LemonBreezes:
> I honestly have not verified if a newer version of Eglot installed via straight would have fixed my issues as well.
> 
> My reasoning to prefer the built-in installation of Eglot was that this would have been my expected behavior for Emacs 29.1 when using doom. And I imagined that going for a different version than the built-in one would require me to modify the installation in my `packages.el` file.
> 
> Would the general philosophy for package version selection prefer to always use a pinned version installed via straight instead?

Yes because this makes it easier to troubleshoot bugs and report them upstream. If the package doesn't work, the general process is to bump it / unpin it to see if the bug is fixed upstream and if not, to report it upstream, possibly opening a linked issue here as well.

tbszlg:
> Yes because people are using different versions of Emacs and pinning `eglot` gives Doom Emacs more control since everyone will be using the same version regardless of their Emacs version.

I see, that makes sense.

I have tested the latest upstream version of `eglot` and the issue still exists for me.
With that I will close this PR again and move the discussion to `consult` to see if the root cause of the issue lies with it or with `eglot`.

PR #7570, opened 2023-11-29
docs: update install instructions for macos

rossampro:
Updating macOS docs because compiling railwaycat/emacs-mac --with-modules --with-native-comp doubles performance on macOs using M2 mac as a backdrop


hlissner:

PR #7584, opened 2023-12-07
fix(org): correct :face parameter for doom-user links

OceanS2000:
Currently the documentation for `:lang org`, among some others, does not load correctly. It would not enter `doom-docs-org-mode` with error
```
File mode specification error: (user-error Invalid action)
```

It turns out that `org-link-set-parameters` accepts `:face` for a function returning a face or a symbol naming a face. Passing `org-priority` directly would cause it being evaluated as a function, preventing documents containing `doom-user:` links from loading.

-----


hlissner (approved):

hlissner:

PR #7590, opened 2023-12-14
bump: :term eshell

vitallium:
Hi. Attempting to upgrade `doomemacs` with the `upgrade` or the `sync` command fails if you have enabled the `eshell` module:

```
✓ (212/338) emacs-fish-completion: d34d0b9 -> df42e15 [1 commit(s)]
        ERROR: Couldn't collect commit list because: fatal: bad object df42e153082927536763bdf408184152a7c938c3
```

It seems that `emacs-fish-completion` package has a new owner and it was moved to GitHub now - https://github.com/Ambrevar/emacs-fish-completion

This merge request updates the `emacs-fish-completion` pkg recipe to the new home. I think that the fix here should work. Let me know if this should be done in another way. Thanks!

ambrevar/emacs-fish-completion@df42e1530829 -> LemonBreezes/emacs-fish-completion@d34d0b96fde6

-----

### DIFF
--- a/modules/term/eshell/packages.el
+++ b/modules/term/eshell/packages.el
@@ -9,5 +9,5 @@
 (package! eshell-syntax-highlighting :pin "4ac27eec6595ba116a6151dfaf0b0e0440101e10")
 
 (unless IS-WINDOWS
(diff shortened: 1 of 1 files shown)


hlissner:
LGTM. Thanks for your help!

PR #7604, opened 2023-12-31
fix(lib): doom/bumpify-diff: respect structure of given list

fosskers:
The old check was a bug fix to work around noisy values that somehow made it into diff checks. The `package!` symbol is never actually present in the list of values yielded by the search in `read-package`, so this commit alters the lookup to respect what is actually present, thus guaranteeing that `destructuring-bind` succeeds and bump diffs are actually detected.

-----


hlissner (approved):

hlissner:

PR #7611, opened 2024-01-04
feat: reintroduce the `format--org-region` back

lroolle:
The `format--org-region` function, highly useful and convenient for frequent Org mode users, was removed for some reason from the `format` module in the last refactor (#6369). I suggest re-adding it, as it could be useful.

Fix: https://github.com/doomemacs/doomemacs/pull/6369#discussion_r1441274403
Ref: #6369
Close: https://github.com/doomemacs/doomemacs/pull/6369#discussion_r1441274403

### DIFF
--- a/modules/editor/format/autoload/format.el
+++ b/modules/editor/format/autoload/format.el
@@ -53,7 +53,33 @@
            (when callback (funcall callback))
            (kill-buffer formatted-buffer)))))))
 
-
+(defun +format--org-region (beg end)
+  "Reformat the region within BEG and END.
+If nil, BEG and/or END will default to the boundaries of the src block at point."
+  (when (eq major-mode 'org-mode)
+    (let ((element (org-element-at-point)))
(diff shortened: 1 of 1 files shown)


hlissner:
Ah, my bad, I forgot to close this PR when I committed 20de3d0f29c1f4924611bfd72d39aae4fa51e3eb, which brings back the region formatter for org buffers. Thanks for bringing it to my attention, in any case!

lroolle:
> Ah, my bad, I forgot to close this PR when I committed [20de3d0](https://github.com/doomemacs/doomemacs/commit/20de3d0f29c1f4924611bfd72d39aae4fa51e3eb), which brings back the region formatter for org buffers. Thanks for bringing it to my attention, in any case!

The new format src blocks fn goes wrong when dealing with the block region, which causing it to format code outside the intended src block, I've created a new PR #7685 to fix this issue.

PR #7621, opened 2024-01-14
tweak: add alert and use native alert on macOS

catap:
This small changes allows to make quite nice setup.

For example after this I may add to my WL configuration:
```
(setq wl-biff-notify-hook '((lambda () (alert "You have new mail!" :title "Wanderlust"))))
```
which leads to alerts like this:
Screenshot 2024-01-14 at 14 05 29

### DIFF
--- a/modules/os/macos/config.el
+++ b/modules/os/macos/config.el
@@ -33,6 +33,9 @@
 (after! auth-source
   (pushnew! auth-sources 'macos-keychain-internet 'macos-keychain-generic))
 
+;; Integrate with notification
+(after! alert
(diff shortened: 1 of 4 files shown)


catap:
rebased to the last master

hlissner:
Sorry for the late response. I have to turn down this PR because it's much too trivial to warrant a module, and niche (as in there multiple options and no compelling reason to treat `alert.el` as *the one* we support). Plus, module development have moved to doomemacs/modules so I will be closing all module-related PRs over time. Thanks for the PR and for bringing it to my attention, in any case!

PR #7624, opened 2024-01-19
fix(vertico): don't shell-quote consult-ripgrep-args

45mm:
Fixes the `:args` parameter of `+vertico-file-search` being unusable: the command-line args given through this parameter were being incorrectly escaped before appending them to `consult-ripgrep-args`. The fix is to not escape them at all.

We were using `shell-quote-argument`, which is meant for passing file
names, strings and so on, not command-line arguments. For example,
`(shell-quote-argument "--foo=bar")` yields "--foo\\=bar", which is
obviously invalid unless we're trying to pass an option named '--foo\'.

At any rate, there is no quoting/escaping for shells in the default
value of `consult-ripgrep-args`, so it doesn't look like this is
something we need to do.

-----


hlissner:
Good catch. Thanks for looking into this and fixing it!

PR #7625, opened 2024-01-20
fix(julia): improve julia-snail settings

ngharrison:
Make multimedia-enable actually work.
Remove boxes around every line of inline output and set text color to plain white instead of matching the line behind them.

Before:
image

After:
image

(shades may differ slightly because of chosen theme)

-----
- [x] I am blind.

### DIFF
--- a/modules/lang/julia/config.el
+++ b/modules/lang/julia/config.el
@@ -110,7 +110,8 @@
   (setq-default julia-snail-multimedia-enable t)
   (setq julia-snail-popup-display-face
         `(:background ,(doom-color 'base3)
-          :foreground ,(doom-color 'base8)))
+          :foreground ,(doom-color 'base8)
+          :weight regular))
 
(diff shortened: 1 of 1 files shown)


RomeoV:
FYI I like the white boxes. I don't know what you fixed about the multimedia, usually the things I need work fine (?)

ngharrison:
Hmm, if I remember correctly, `julia-snail-multimedia-enable` was only buffer-local for me and its default value was nil so it was having no effect.

hlissner:
73f19ac should resolve the issues presented here (changing the default value of `julia-snail-multimedia-enable` and `julia-snail-popup-display-face`). I've opted to not set `julia-snail-popup-display-face` at all, since julia-snail will [generate one automatically](https://github.com/gcv/julia-snail/blob/master/julia-snail.el#L1509-L1513) if it is unset. Plus, the `doom-color` function (and `baseX` variables) belong to doomemacs/themes. If the user did not have it installed, they would see errors, and I'd rather avoid a hard dependency on any single theme.

PR #7639, opened 2024-01-30
fix(evil): obsolete `evil-command-window` advice

45mm:
As noted in #7556, the advice `+popup--evil-command-window-a` fails, preventing the command window from being opened. This is because `evil-command-window` has been rewritten extensively.

In particular, https://github.com/emacs-evil/evil/commit/a09fdca0b35ef4289cf55d7b0ffadf4cf3a5c9fc made it use `display-buffer` instead of `switch-to-buffer`, so that users could customize how the window opens. Since this was the function of this advice, it is obsolete and can be removed.

Fix: #7556 

-----

### DIFF
--- a/modules/ui/popup/+hacks.el
+++ b/modules/ui/popup/+hacks.el
@@ -102,26 +102,6 @@ were followed."
 
 ;;;###package evil
 (progn
-  ;; Make evil-mode cooperate with popups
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:

PR #7641, opened 2024-01-30
docs(ivy): mention lowered `ivy-sort-max-size`

45mm:
Depending on the system, this can cause sorting to be disabled long before the performance hit would be felt, and it's not obvious that this is the reason.

-----

### DIFF
--- a/modules/completion/ivy/README.org
+++ b/modules/completion/ivy/README.org
@@ -176,7 +176,16 @@ A wgrep buffer can be opened from swiper with [[kbd:][C-c C-e]].
 ** TODO Change the position of the ivy childframe
 
 * TODO Troubleshooting
-/There are no known problems with this module./ [[doom-report:][Report one?]]
+** Sorting is not applied at all sometimes
+If the number of candidates is greater than ~ivy-sort-max-size~, sorting will be
+disabled completely. Doom lowers the default value to prevent performance
+issues, so increasing the value may fix your issue:
+#+begin_src elisp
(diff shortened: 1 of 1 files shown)


modules/completion/ivy/README.org
```diff
@@ -176,7 +176,8 @@ A wgrep buffer can be opened from swiper with [[kbd:][C-c C-e]].
 ** TODO Change the position of the ivy childframe
 
 * TODO Troubleshooting
-/There are no known problems with this module./ [[doom-report:][Report one?]]
+** Sorting is not applied at all sometimes
+If the number of candidates is greater than ~ivy-sort-max-size~, sorting will be disabled completely. Doom lowers the default value to prevent performance issues, so increasing the value may fix your issue.
```
hlissner, line 180: Could you run `M-x org-fill-paragraph` on this paragraph?

Also, please add an elisp snippet after it:

```org
#+begin_src elisp
;;; add to $DOOMDIR/config.el
(after! ivy
  (setq ivy-sort-max-size 30000))  ; Doom sets this to 7500, but Ivy's default is 30k
#+end_src
```

hlissner (changes requested):

modules/completion/ivy/README.org
45mm: done

hlissner (approved):

hlissner:
LGTM. Thanks for your help!

PR #7643, opened 2024-02-01
feat: set jump on more kill buffer functions

LemonBreezes:
Hi. There are some interactive commands for killing the current buffer that do not allow for undoing the buffer kill by using the jumps. I made those commands set a jump.

-----

### DIFF
--- a/lisp/doom-editor.el
+++ b/lisp/doom-editor.el
@@ -486,6 +486,9 @@ files, so this replace calls to `pp' with the much faster `prin1'."
   ;; I'm not advising `kill-buffer' because I only want this to affect
   ;; interactively killed buffers.
   (advice-add #'kill-current-buffer :around #'doom-set-jump-a)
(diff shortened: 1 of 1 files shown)


hlissner:
This one's a little tricky. better-jumper stores jump history in window parameters (which is preferable/more intuitive to [the alternative](https://github.com/gilbertw1/better-jumper/blob/master/better-jumper.el#L56)), which gets deleted along with windows, so trying to set jump points around potential window deletions will always be a little weird. What's more, I'd like to avoid playing advice whack-a-mole with all the various kill buffer commands.

I'd originally avoided writing a more generic `kill-buffer-hook` hook, because Emacs kills many temporary/buried buffers at any given point, and we don't want runaway jump point allocations in invisible buffers, but I'll revisit the idea. Maybe a simple `get-buffer-window` check is all we need to filter out non-interactively killed buffers.

In any case, I'll need more time to look into this further, refresh what I remember about `better-jumper`, and test out this `kill-buffer-hook` idea.

PR #7646, opened 2024-02-03
add better icons for lsp-headerline

teeann:
Add `all-the-icons` support for `lsp-headerline` icons display.

-----

### DIFF
--- a/modules/tools/lsp/+lsp.el
+++ b/modules/tools/lsp/+lsp.el
@@ -38,6 +38,9 @@ Can be a list of backends; accepts any value `company-backends' accepts.")
   ;; even if its a dependency
   (when (modulep! :checkers syntax +flymake)
     (setq lsp-diagnostics-provider :flymake))
+  ;; Better breadcrumbs default, if enabled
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but I must turn it down. The main reason: I'm in the process of swapping out all-the-icons with nerd-icons across Doom (has wider support (including the terminal), a simpler install process (only one font), and a wider selection of icons).

Other reasons:
- This PR causes `lsp-icons` to be eagerly loaded (at startup).
- It disables `lsp-headerline-breadcrumb-enable-diagnostics`, which is orthogonal to the apparent purpose of the PR (i.e. to change icons),
- The rewritten functions should not be redefined, but instead be advised,
- The git commit doesn't conform to [our git conventions](https://discourse.doomemacs.org/t/git-conventions/2407).

Also, careful what you're checking off:
image

If you can repurpose this to use nerd-fonts, I will gladly reopen this, however. But please include screenshots, if you do. It's difficult to know what you are changing, and why you consider the change to be better than the original.

PR #7649, opened 2024-02-06
fix(dired): void-variable dired-omit-files error

regadas:
Hello,

With recent update https://github.com/doomemacs/doomemacs/commit/e242ac9548487ae50857ec1b169a20d946019816 `dired +dirvish` fails with `void-variable dired-omit-files error`.

Fix: #6562


hlissner (approved):

hlissner:

PR #7661, opened 2024-02-10
tweak(org): honor default command when archiving

real-or-random:
Change the binding for archiving to `org-archive-subtree-default`, which is the recommend "catch-all" command in the org manual. The user can specify the actual command in `org-archive-default-command`. The default for this variable is `org-archive-subtree`, which we previously used for the binding, so this commit changes behavior only for users who have set `org-archive-default-command` explicitly.

-----


hlissner:
LGTM. Thanks for your help!

PR #7662, opened 2024-02-10
fix(lib): sudo file path: expand-file-name

45mm:
It's possible for the user to type shell variables (something like `$MYVAR/dir/filename`) as part of the filepath, so we need to call `expand-file-name`.


hlissner (approved):

hlissner:
LGTM. Thanks for your help!

PR #7674, opened 2024-02-14
fix(vertico) set root user for consult-dir containers

bergmannf:
The tramp string used to connect to the container uses the first 'container-runtime ps' argument as a username. However, the first argument that is returned is the container id, not a username.

Instead of inspecting the container to find the user, just default to 'root' which should work for most cases.

There doesn't seem to be an open issue for this - so if that's needed I can create one first.

-----

### DIFF
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -209,10 +209,8 @@ orderless."
                               (apply #'process-lines +vertico-consult-dir-container-executable
                                      (append +vertico-consult-dir-container-args (list "ps")))))
                for cand = (split-string line "[[:space:]]+" t)
-               collect (let ((user (unless (string-empty-p (car cand))
-                                     (concat (car cand) "@")))
-                             (hostname (car (last cand))))
-                         (format "/%s:%s%s:/" host user hostname))))
+               collect (let ((hostname (car (last cand))))
+                         (format "/%s:root@%s:/" host hostname))))
 
     (defun +vertico--consult-dir-podman-hosts ()
       (let ((+vertico-consult-dir-container-executable "podman"))


elken:
You're right that it seems to have no effect, not sure what I was thinking when I added it but it makes more sense IMO to just default to nothing.

```diff
modified   modules/completion/vertico/config.el
@@ -210,7 +210,7 @@ defvar +vertico-consult-dir-container-args
                                      (append +vertico-consult-dir-container-args (list "ps")))))
                for cand = (split-string line "[[:space:]]+" t)
                collect (let ((hostname (car (last cand))))
-                         (format "/%s:root@%s:/" host hostname))))
+                         (format "/%s:%s:/" host hostname))))
 
     (defun +vertico--consult-dir-podman-hosts ()
       (let ((+vertico-consult-dir-container-executable "podman"))
```

modules/completion/vertico/config.el
```diff
@@ -209,10 +209,8 @@ orderless."
                               (apply #'process-lines +vertico-consult-dir-container-executable
                                      (append +vertico-consult-dir-container-args (list "ps")))))
                for cand = (split-string line "[[:space:]]+" t)
-               collect (let ((user (unless (string-empty-p (car cand))
-                                     (concat (car cand) "@")))
-                             (hostname (car (last cand))))
-                         (format "/%s:%s%s:/" host user hostname))))
+               collect (let ((hostname (car (last cand))))
+                         (format "/%s:root@%s:/" host hostname))))
```
elken, line 213: You're right that it seems to have no effect, not sure what I was thinking when I added it but it makes more sense IMO to just default to nothing.

```suggestion
                         (format "/%s:%s:/" host hostname))))
```

hlissner:
This was merged (with @elken's suggestion) into d22fa5a. Thanks for bringing it to my attention!

PR #7683, opened 2024-02-19
refactor: factorize external executables

pharcosyle:
I'd like to see external executable dependencies (ripgrep, etc.)  better consolidated in the code.

Currently they're scattered all over the place. Ideally there would be a single defvar/defcustom for each that would allow for easy overriding, in particular for Guix/Nix users like me who would like to patch the code in a packaging step so all the various external deps don't have to be installed globally (but it would be useful to everyone else too and it's good practice anyway).

THIS IS A DRAFT PR. Just a start on how this might look and to get feedback.

Some considerations:
- `rg`/`fd` use is inconsistent with respect to remote files
- perform `executable-find` at the point of definition or at each point of use (where necessary)?
- doctor checks for executables depended-on by packages that themselves define a var for the executable should do `executable-find` on that var, not the raw command as a string

WDYT?

### DIFF
--- a/lisp/cli/doctor.el
+++ b/lisp/cli/doctor.el
@@ -92,7 +92,7 @@ in."
                        version))
            (warn! "Cannot determine Git version. Doom requires git 2.23 or newer!")))))
 
-   (unless (executable-find "rg")
+   (unless (executable-find doom-rg-binary)
      (error! "Couldn't find the `rg' binary; this a hard dependecy for Doom, file searches may not work at all")))
 
   (print! (start "Checking for Emacs config conflicts..."))
--- a/lisp/doom-projects.el
+++ b/lisp/doom-projects.el
@@ -19,6 +19,8 @@ debian, and derivatives). On most it's 'fd'.")
 
 (defvar doom-projects--fd-version nil)
 
+(defvar doom-rg-binary "rg")
+
 
 ;;
 ;;; Packages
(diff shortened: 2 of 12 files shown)


hlissner:
Sorry for the late response. I believe the original issue is resolved for Doom's core dependencies (`doom-fd-executable` and `doom-ripgrep-executable`), but I think going beyond for *every* executable (e.g. formatters) is overkill. In most cases upstream offers a variable or a way to trivially redefine such things (like formatters).

Plus the practice of referring to naked executables is an issue that permeates the entire Emacs ecosystem, so solving it at this superficial level in Doom may give the user an impression that Doom *ought* to be handling every one of these cases, which is a maintenance nightmare I don't want to assume responsibility for. Not to mention the extra complication of supporting TRAMP environments. 

If folks are in environments that necessitates this degree of flexibility, I trust that they will also be familiar with alternative means of fixing the issue at a more appropriate layer of the stack (e.g. with shims, altering their `$PATH`, or through their nix/guix configs). Plus, being a nixos user myself, I find nixpkgs (particularly nixpkgs.emacs's site-lisp files) handles this fine enough that little to no reconfiguration is even needed in most cases.

In any case, thanks for the PR and for bringing this to my attention!

(Side-note: module development has been moved to doomemacs/modules, so I am slowly closing all module-related PRs)

pharcosyle:
Excellent thank you!

PR #7685, opened 2024-02-20
fix(format): incorrect region calculation for org src block

lroolle:
Resolves an issue where formatting could unintentionally affect content outside src blocks.

### DIFF
--- a/modules/editor/format/autoload/format.el
+++ b/modules/editor/format/autoload/format.el
@@ -128,21 +128,19 @@ the requested feature."
 (defun +format-in-org-src-blocks-fn (beg end _op)
   "Reformat org src blocks with apheleia as if they were independent buffers."
   (when (derived-mode-p 'org-mode)
-    (goto-char beg)
-    (while (re-search-forward org-babel-src-block-regexp end t)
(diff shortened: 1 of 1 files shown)


hlissner:
The function needs to iterate over all blocks within BEG and END. This change makes it only operate on the one at point, so cannot approve this PR as it is.

lroolle:
We usually use this feature within a single src block most of the time? Otherwise, we would have to manually select the region, or the blocks might get messed up. Regarding the search block issue, how about this approach:

1. if point within a src block -> reformat current block;
2. if point outside any src blocks -> iterate over all blocks within BEG and END -> reformat the last src block;

hlissner:
Has something changed? I'm working on a larger refactor of the format module which will indirectly restore the behavior you're describing.

hlissner:
Sorry for taking so long to get to this. I've added a `+format/org-block` command in 8072762, which I've bound to `gQ` (for Evil users) in org-mode (in f6e65c4), which should fix the original issue. Let me know if that gives you any trouble. Thanks for your patience and for bringing it to my attention!

PR #7695, opened 2024-02-26
dev: shell.nix: use niv and simplify

Lord-Valen:
+ I added `niv` for easier dependency management.
  - This allows us to easily move to new nixpkgs point releases.
  - This allows us to easily add nixpkgs revisions, pinning old emacs versions.
  - This allows future nix code to share dependencies (this is helpful for #2552).
  - This vastly improves the readability of shell.nix.
+ I used `mkShellNoCC`, which is preferred to `mkDerivation` for shell use.
+ Since emacs 26 is not in the supported version range (per the README), I removed it from the shell.
+ I replaced the conditional tree which checked the emacs version with an expression which modifies the string to match the attributes in `pkgs`. This should have feature parity with the previous implementation, but is far more flexible.
  - If the `emacs` argument is "head", we use`emacsGit`.
  - If the modified string does not match any `pkgs` attributes, we fallback to the `emacs` of the current nixpkgs point release.
  - If the modified string matches a `pkgs` attribute, we use that package.
+ I added aliases for the major emacs versions to their latest releases. (This is nice for people who mean "28.2" by "28," for example) 
-----


hlissner:
I wanted to minimize Nix's footprint in the project, considering it may be accompanied by Guix/Docker backends later. And our shell.nix is so short, I don't think its readability is enough of a concern to bring in `niv`. I've also opted for the simplest implementation: modules will have their own shell.nix fragments, able to define their own `packages` and `shellHook`, which will be imported from shell.nix via an `--args modules '[ ... ]'` argument. Future versions of the `doom emacs` and `doom sandbox` commands will provide a nicer interface to bring it all together.

In any case, I've incorporated a couple other things from this PR, like using `mkShellNoCC` and major version aliases (28->28.2, 29->29.4, etc). Thank you for your help!

PR #7700, opened 2024-02-27
fix(workspaces): add check for ivy before setting counsel-projectile-[XXXX]

yaaama:
This change now wraps configuration code for
`counsel-projectile-switch-project-actions` within a conditional check for the Ivy module.

It ensures configuration related to
`counsel-projectile` are only set up if the `:completion ivy` module is enabled by the user.

This is my first PR so forgive me for any mistakes, feedback is highly appreciated.
Thank you

-----
- [x ] My changes are not on [the do-not-PR list](https://doomemacs.org/d/do-not-pr) for this project.
- [x ] My commits conform to [the git conventions](https://doomemacs.org/d/git-conventions).
- [x ] Any relevant issues or PRs have been linked to.

### DIFF
--- a/modules/ui/workspaces/config.el
+++ b/modules/ui/workspaces/config.el
@@ -179,32 +179,34 @@ stored in `persp-save-dir'.")
   (add-hook 'delete-frame-functions #'+workspaces-delete-associated-workspace-h)
   (add-hook 'server-done-hook #'+workspaces-delete-associated-workspace-h)
 
-  ;; per-project workspaces, but reuse current workspace if empty
-  ;; HACK?? needs review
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but I have to refuse it. To explain why:

1. The new value for `projectile-switch-project-action` is necessary for all users, with or without ivy.
2. Setting `counsel-projectile-switch-project-action` is harmless if counsel-projectile isn't available, so I'd prefer to keep it as is than complicate the load-order for these settings (which affects what goes in `HERE` in `(after! HERE ...)` if the user wants to change `counsel-projectile-switch-project-action` or `projectile-switch-project-action`.
3. I plan to phase out the ivy module (and replace `persp-mode` with an alternative) in the not-too-near future, and I'd like to minimize development in and around the `:ui workspace` and `:completion ivy` modules altogether until that happens.

PR #7714, opened 2024-03-08
feat(magit): add +childframe module flag

Dev380:
Add `+childframe`, which uses transient-posframe
to make the magit popups show in a childframe
instead of the minibuffer.

### Screenshots:
Before:
20240307_19h17m36s_grim
After:
20240307_19h18m13s_grim

-----


hlissner:
Thanks for the PR, but I can't accept this one (at least, not yet). One, the PR is incomplete (it doesn't enable `transient-posframe-mode`, for instance), but also, the package has global side-effects that are not trivially isolated to Magit buffers only, i.e. any package using transient will suddenly be childframe-enabled after enabling `+childframe` on `:tools magit`, which I want to avoid. 

After the v3 release, I intend to make transient a core package (and part of a new set of `:core` modules). At that point I'll reconsider introducing a `+childframe` flag and writing window-local a `transient-posframe-mode`, but for the time being, I'll have to turn this down. Thanks for bringing it to my attention, in any case!

PR #7720, opened 2024-03-10
fix(vertico): disable minor mode highlight duly

Dev380:
Minor mode highlights did not disable as long as the mode was enabled (e.g. I had visual-line-mode highlighted even in buffers when it wasn't on). This should fix that.

-----

### DIFF
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -371,7 +371,8 @@ orderless."
       (if (or (eq sym major-mode)
               (and
                (memq sym minor-mode-list)
-               (boundp sym)))
+               (boundp sym)
+               (eval sym)))
           (add-face-text-property 0 (length cmd) 'font-lock-constant-face 'append cmd)))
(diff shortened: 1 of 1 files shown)


modules/completion/vertico/config.el
```diff
@@ -371,7 +371,8 @@ orderless."
       (if (or (eq sym major-mode)
               (and
                (memq sym minor-mode-list)
-               (boundp sym)))
+               (boundp sym)
+               (eval sym)))
```
hlissner, line 375: Please use `(symbol-value sym)` instead of `eval`.

hlissner (changes requested):

modules/completion/vertico/config.el
Dev380: Fixed in 7d7cbec

hlissner (approved):

hlissner:
Thanks again for your help!

PR #7723, opened 2024-03-12
feat(macos): osx-trash is only needed by emacs < 29

seanfarley:
As the title says, osx-trash is no longer needed nowadays for emacs so one less package to install :-)

### DIFF
--- a/modules/os/macos/config.el
+++ b/modules/os/macos/config.el
@@ -33,17 +33,20 @@
 (after! auth-source
   (pushnew! auth-sources 'macos-keychain-internet 'macos-keychain-generic))
 
+;; Delete files to trash on macOS, as an extra layer of precaution against
+;; accidentally deleting wanted files.
+(setq delete-by-moving-to-trash t)
+
 
 ;;
 ;;; Packages
 
 (use-package! osx-trash
+  ;; DEPRECATED: Not needed on Emacs 29+. Remove when dropping 28 support.
+  ;;   Fixed by https://debbugs.gnu.org/cgi/bugreport.cgi?bug=21340.
+  :when (< emacs-major-version 29)
   :commands osx-trash-move-file-to-trash
   :init
-  ;; Delete files to trash on macOS, as an extra layer of precaution against
-  ;; accidentally deleting wanted files.
-  (setq delete-by-moving-to-trash t)
-
   ;; Lazy load `osx-trash'
   (when (not (fboundp 'system-move-file-to-trash))
     (defun system-move-file-to-trash (file)
--- a/modules/os/macos/packages.el
+++ b/modules/os/macos/packages.el
@@ -1,5 +1,6 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; os/macos/packages.el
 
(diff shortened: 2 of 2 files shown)


hlissner (approved):

modules/os/macos/config.el
```diff
@@ -38,6 +38,9 @@
 ;;; Packages
 
 (use-package! osx-trash
+  ;; DEPRECATED Not needed on Emacs 29+. Remove when dropping 28 support.
+  ;; fixed by https://debbugs.gnu.org/cgi/bugreport.cgi?bug=21340
+  :when (< emacs-major-version 29)
   :commands osx-trash-move-file-to-trash
```
hlissner, line 44: Even if the package isn't installed, we still need to set `delete-by-moving-to-trash`, so please make the following changes:

```diff
   :commands osx-trash-move-file-to-trash
   :init
-  ;; Delete files to trash on macOS, as an extra layer of precaution against
-  ;; accidentally deleting wanted files.
-  (setq delete-by-moving-to-trash t)
- 
   ;; Lazy load `osx-trash'
```

```diff
 ;; Integrate with Keychain
 (after! auth-source
   (pushnew! auth-sources 'macos-keychain-internet 'macos-keychain-generic))
+
+;; Delete files to trash on macOS, as an extra layer of precaution against
+;; accidentally deleting wanted files.
+(setq delete-by-moving-to-trash t)
+

 ;;
 ;;; Packages

```

modules/os/macos/config.el
```diff
@@ -38,6 +38,9 @@
 ;;; Packages
 
 (use-package! osx-trash
+  ;; DEPRECATED Not needed on Emacs 29+. Remove when dropping 28 support.
+  ;; fixed by https://debbugs.gnu.org/cgi/bugreport.cgi?bug=21340
```
hlissner, line 42: Please make formatting consistent with other notices:

```diff
 (use-package! osx-trash
-  ;; DEPRECATED Not needed on Emacs 29+. Remove when dropping 28 support.
-  ;; fixed by https://debbugs.gnu.org/cgi/bugreport.cgi?bug=21340
+  ;; DEPRECATED: Not needed on Emacs 29+. Remove when dropping 28 support.
+  ;;   Fixed by https://debbugs.gnu.org/cgi/bugreport.cgi?bug=21340.
   :when (< emacs-major-version 29)
```

hlissner (changes requested):

modules/os/macos/config.el
seanfarley: Oh jeez, that completely slipped by me >_<

modules/os/macos/config.el
seanfarley: I don't suppose there is a doom function to format these type of comments, is there? 😬

hlissner (approved):

modules/os/macos/config.el
hlissner: Afraid not. The closest you'll get is `gq` (`evil-fill-and-move`) or `M-q` (`fill-paragraph`), but only handles line-width, not indentation or colon.

hlissner:

PR #7724, opened 2024-03-12
fix(mu4e): extend popup rule to all mu4e- prefixed buffers

seanfarley:
It was too restrictive before so let's just match anything that is prefixed with `mu4e-`.

### DIFF
--- a/modules/email/mu4e/config.el
+++ b/modules/email/mu4e/config.el
@@ -223,7 +223,7 @@ is non-nil."
   (setq-hook! 'mu4e-view-mode-hook truncate-lines nil)
 
   ;; mu4e now uses `display-buffer-alist' so we need to add some rules of our own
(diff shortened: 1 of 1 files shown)


hlissner:
Can you list a few (or more) example buffers that'd benefit from this? It may still be desirable to capture some and not others, so I can't approve of this PR if there isn't a more specific benefit.

seanfarley:
Ah, yeah, that's a good point. The previous logic didn't consider composing both a reply and a new message, so sometimes the reply would pop up in a really tiny window. I'm actually really curious if this is the correct way or not because I don't really understand the display-alist stuff.

hlissner:
Since there are legitimate mu4e buffers we want to treat as popups (like `*mu4e-log*` and possibly some of the attachment preview buffers), I must turn down this PR. More nuanced popup rules are needed for this module, in any case. Thanks for bringing it to my attention though!

PR #7725, opened 2024-03-12
feat(mu4e): use built-in notifications instead of mu4e-alert

seanfarley:
Now that mu4e has its own notification feature since 1.10 (and as of
this commit version 1.12 is already out), we can just use that. Much
less code to maintain.

Closes: https://github.com/doomemacs/doomemacs/issues/6896

### DIFF
--- a/modules/email/mu4e/config.el
+++ b/modules/email/mu4e/config.el
@@ -75,6 +75,7 @@ is non-nil."
      (setq mu4e-get-mail-command "offlineimap -o -q")))
 
   (setq mu4e-update-interval nil
+        mu4e-notification-support t
         mu4e-sent-messages-behavior 'sent
         mu4e-hide-index-messages t
         ;; configuration for sending mail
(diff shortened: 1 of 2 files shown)


tecosaur:
To me this seems good :+1:. It is less fancy than the prior (`mu4e-alert`) based system, we can set a fancier `mu4e-notification-function` ourselves in the future if wanted.

tecosaur (approved):

seanfarley:
> To me this seems good 👍. It is less fancy than the prior (`mu4e-alert`) based system, we can set a fancier `mu4e-notification-function` ourselves in the future if wanted.

Yeah, that's what I was thinking, too 😄 First we simplify (this PR), then we can discuss more fancy (and hopefully generic) notifications!

hlissner:
LGTM! Thanks for your help! (And thank you for reviewing this, @tecosaur)

PR #7726, opened 2024-03-12
feat(common-lisp): add more sly packages

seanfarley:
Might as well?

### DIFF
--- a/modules/lang/common-lisp/config.el
+++ b/modules/lang/common-lisp/config.el
@@ -174,3 +174,8 @@
   :defer t
   :init
   (add-to-list 'sly-contribs 'sly-asdf 'append))
+
+(use-package! sly-stepper
+  :defer t
+  :init
+  (add-to-list 'sly-contribs 'sly-stepper))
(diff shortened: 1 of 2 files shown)


hlissner:
Can you explain the purpose and benefit of these packages? And why they should be included by default? Are these packages that most/all Common Lisp developers will want?

seanfarley:
I'm still learning common lisp but it seems quicklisp is pretty defacto-ish for modern development? I'm using it with a handful of projects so I admit it's a small sample.

The other two were just nice-ish ones that I thought would be nice to have but mostly just shots in the dark.

ispringle:
Just adding my two cents here, quicklisp is used by pretty much all CL devs, it's the package repo for Lisp. I think sly-quicklisp makes sense as an addition to the CL module. sly-stepper is a good debugging tool and it extends Sly's native stickers feature nicely. I agree it also makes sense in the module. I have the readtable package installed and never use it. I'm sure more advanced CL devs might have a need for multiple readtables but it is not, in my opinion, a regular/normal part of CL dev work.

seanfarley:
That is exactly what I'm currently setup for in my CL environments, too, hehe. I almost never have used readtables but thought, "why not?" so I'll remove that as well. Thanks for the input!

seanfarley:
Ok, I updated the PR to remove `readtables` and think it's pretty good to know now (IMHO)

hlissner (approved):

hlissner:

PR #7727, opened 2024-03-12
[RFC]🔥 feat: add buffer-name-relative

seanfarley:
This will help unify the people and quell, once and for all, the
test.foo<20> buffer names. Power to the people.

In all seriousness, this has greatly helped improve the buffer list with similar-named buffers across many progects (e.g. `README`) but since this changes some core machinery, I thought I'd gather some feedback. (If it's actually not too controversial, then it's ready to be merged)

### DIFF
--- a/lisp/doom-ui.el
+++ b/lisp/doom-ui.el
@@ -201,6 +201,9 @@ or if the current buffer is read-only or not file-visiting."
 ;;
 ;;; Buffers
 
+(use-package! buffer-name-relative
+  :hook (doom-first-buffer . buffer-name-relative-mode))
+
 (defadvice! doom--switch-to-fallback-buffer-maybe-a (&rest _)
   "Switch to `doom-fallback-buffer' if on last real buffer.
 
--- a/lisp/packages.el
+++ b/lisp/packages.el
@@ -21,6 +21,7 @@
 (package! highlight-numbers :pin "8b4744c7f46c72b1d3d599d4fb75ef8183dee307")
 (package! rainbow-delimiters :pin "f40ece58df8b2f0fb6c8576b527755a552a5e763")
 (package! restart-emacs :pin "1607da2bc657fe05ae01f7fdf26f716eafead02c")
+(package! buffer-name-relative :pin "6c1e98f761344b2d2d51f38d587161f71ca0e750")
 
 ;; doom-editor.el
 (package! better-jumper :pin "47622213783ece37d5337dc28d33b530540fc319")


brongulus:
Hi Sean, I'm not familiar with the package so wanted to ask, does this
```lisp
(setq uniquify-buffer-name-style 'forward)
```
not achieve the desired goal of making buffers with same name unique?

brongulus:
Hi Sean, not familiar with the package but
```lisp
(setq uniquify-buffer-name-style 'forward)
```
Wanted to ask, does this not achieve the same functionality?

seanfarley:
Let's find out!

seanfarley:
Here are the results I found. Each test was opening two projects, each with a few files named the same (`README.org`)

* no modifications (aka default doom without customizing `uniquify-buffer-name-style`)

Screenshot 2024-03-12 at 10 49 31 AM
Screenshot 2024-03-12 at 10 49 49 AM

* explicitly setting `(setq uniquify-buffer-name-style 'forward)` in user config

Screenshot 2024-03-12 at 10 58 24 AM
Screenshot 2024-03-12 at 10 58 15 AM

* this PR

Screenshot 2024-03-12 at 11 00 17 AM
Screenshot 2024-03-12 at 11 00 01 AM

That's pretty close so it doesn't seem `buffer-name-relative` would add much here but that got me wondering ... why wasn't  `uniquify-buffer-name-style` already set to `'forward` since it's explicitly set in [doom-ui.el](https://github.com/doomemacs/doomemacs/blob/a0344ffc3ad7844075f6b0bc53c5130545d5a1d5/lisp/doom-ui.el#L148)?

seanfarley:
> That's pretty close so it doesn't seem `buffer-name-relative` would add much here but that got me wondering ... why wasn't `uniquify-buffer-name-style` already set to `'forward` since it's explicitly set in [doom-ui.el](https://github.com/doomemacs/doomemacs/blob/a0344ffc3ad7844075f6b0bc53c5130545d5a1d5/lisp/doom-ui.el#L148)?

Ah, it seems that this is [explicitly disabled for persp-mode](https://github.com/doomemacs/doomemacs/blob/a0344ffc3ad7844075f6b0bc53c5130545d5a1d5/modules/ui/workspaces/config.el#L92-L98). Ok, I guess the question now is if this package fixes that behavior, or not? If so, do we even need it and could we just use the good ol' `uniquify-buffer-name-style`?

hlissner:
Thanks for the PR, but I cannot accept this one. `uniquify-buffer-name-style` does, indeed, address this issue. That said, I plan to replace persp-mode soon (with what, exactly, I haven't decided, but I have [a number of candidates](https://github.com/orgs/doomemacs/projects/5/views/1) that I'm investigating), but as a side-effect we'll be able to use `uniquify-buffer-name-style` again. I'd rather fix the issue that way, that introduce a new package (especially in Doom's core).

seanfarley:
Yeah, totally makes sense! Thought I'd at least give a shot in dark :-) Thanks for the review (and I learned something new as well).

PR #7728, opened 2024-03-12
refactor(mu4e): drop support for 1.8

seanfarley:
mu4e support code and third-party packages are getting difficult to maintain due to pre-1.10 having a wild time with lisp code. Starting with 1.10, mu has a better policy for lisp practices and has even implemented most of the third-party functionality.

By dropping support for 1.8, we can simplify the mu4e module and maintenance burden for doom. For reference, mu 1.12 was just released.

Again, this might be too big a change for now so I labeled this as RFC. If it turns out that it's ok, then it's ready to be merged.

### DIFF
--- a/modules/email/mu4e/autoload/email.el
+++ b/modules/email/mu4e/autoload/email.el
@@ -20,9 +20,6 @@ OPTIONAL:
 DEFAULT-P is a boolean. If non-nil, it marks that email account as the
 default/fallback account."
   (after! mu4e
-    (when (version< mu4e-mu-version "1.4")
-      (when-let (address (cdr (assq 'user-mail-address letvars)))
-        (add-to-list 'mu4e-user-mail-address-list address)))
     ;; remove existing context with same label
     (setq mu4e-contexts
           (cl-loop for context in mu4e-contexts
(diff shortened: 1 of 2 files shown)


aikrahguzar:
I think another thing to reduce the maintenance burden on doom would be to use `gnus-dired-mode` instead of `dired` support in `+mu4e/attach-files` and `dired-mu4e-attach-ctrl-c-ctrl-c`.

seanfarley:
> I think another thing to reduce the maintenance burden on doom would be to use `gnus-dired-mode` instead of `dired` support in `+mu4e/attach-files` and `dired-mu4e-attach-ctrl-c-ctrl-c`.

Yeah, there's a whole other round of changes after this. Namely, like you mention, updating logic to be more inline with upstream changes to use more of the gnus machinery. Also, I'd like to take a stab at upstreaming the locking mechanism because I think it's a bit too much (and perhaps too brittle) to be in core. Anyways, that's all for the next phase of things!

tecosaur:
While I'm not happy with all the compat code myself, https://repology.org/project/maildir-utils/versions makes me a bit wary of this.

seanfarley:
Sure, I could do that. Initially, I was going off a two-point release policy, then I doubted that, then I swayed back, then decided to just submit it and see what the consensus was. If we want to keep 1.8 for a while, then that's fine too!

hlissner:
Since there's still quite a chasm in mu version support across various package managers, I'd like to continue supporting 1.8. I might consider dropping support for <1.4, perhaps, but I'm only now picking mu4e up, so I'll look into it once I've resolved all the other issues with the mu4e module. Thanks for bringing this to my attention, in any case!

PR #7740, opened 2024-03-21
refactor(corfu): consolidate +orderless logic without vertico

seanfarley:
This tidy ups the corfu module to use the logic when a user enables
corfu +orderless without enabling vertico. Since that's probably a niche
situation, we only specify commit to pin to orderless in the vertico
module so that contributors aren't tripped up with trying to remember to
update orderless in two different code locations. Of course, if this
turns out to be a more popular use-case we can re-evaluate this logic.

### DIFF
--- a/modules/completion/corfu/config.el
+++ b/modules/completion/corfu/config.el
@@ -55,8 +55,10 @@ use the minibuffer such as `query-replace'.")
         corfu-count 16
         corfu-max-width 120
         corfu-on-exact-match nil
-        corfu-quit-at-boundary (if (modulep! +orderless) 'separator t)
-        corfu-quit-no-match (if (modulep! +orderless) 'separator t)
+        corfu-quit-at-boundary (if (or (modulep! :completion vertico)
+                                       (modulep! +orderless))
+                                   'separator t)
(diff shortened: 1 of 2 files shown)


hlissner (approved):

hlissner:
Thanks again for your help!

minad:
@seanfarley Instead of the handwritten dispatcher you could also use the `orderless-affix-dispatch` function, which I've added a while ago.

minad:
Sean Farley ***@***.***> writes:

>> @seanfarley Instead of the handwritten dispatcher you could also use the `orderless-affix-dispatch` function, which I've added a while ago.
>
> I was definitely looking at doing that but didn't understand enough of the
> backstory. It seems that the current dispatcher goes all the way back to the
> days of selectrum? @hlissner is the current doom custom dispatch even needed
> now?

The backstory is that Omar didn't originally include a dispatcher as
part of Orderless. Instead we had documented various dispatchers in the
Orderless README and the Consult wiki. The Doom dispatcher originates
from there - it looks similar to the one I've used and documented in the
Consult wiki. At some point I've added a "standardized" affix dispatcher
to Orderless directly such that Orderless becomes easier to configure.

PR #7743, opened 2024-03-21
feat(fold): unify doom's usage of eliding text

seanfarley:
~~Currently, shortening text is hardcoded as variations of "..." throughout the code. This PR adds two variables to control the short elision of one-line / inline type of text (e.g. 222dc4706…) along with an inherited variable to customize the string used in more traditional folding scenarios of coding and whatnot.~~

Previously, eliding strings was hardcoded to " [...] " which isn't customizable 😬 so `ts-fold-replacement` could become mismatched with `hideshow`.

This variable defaults to " […]" if unicode is displayable, otherwise it falls back to " [...]".

### DIFF
--- a/modules/editor/fold/autoload/hideshow.el
+++ b/modules/editor/fold/autoload/hideshow.el
@@ -30,7 +30,7 @@
                          'empty-line
                          'vimish-fold-fringe))))
     (overlay-put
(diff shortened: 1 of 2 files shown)


seanfarley:
Since submitting this, Henrik [removed eliding commit hashes](639fcc6a2e4dd3df9effef2da7b9d605aaf72214) which creates a conflict for this PR so I simplified this PR to just be for folding code.

PR #7748, opened 2024-03-23
feat(corfu): vim-like C-n / C-p completion

45mg:
Adds insert-mode bindings to make C-n and C-p trigger the Corfu popup, with the first or last candidate selected, respectively.

Care has been taken that no candidate is actually inserted, even if there is only one candidate - only the popup will be shown with the candidate selected. Whether the candidate is previewed or not depends on `corfu-preview-current`.

-----

### DIFF
--- a/modules/completion/corfu/autoload.el
+++ b/modules/completion/corfu/autoload.el
@@ -37,3 +37,21 @@
          (save-excursion (backward-char 1)
                          (insert-char ?\\)))
         (t (call-interactively #'corfu-insert-separator))))
+
+;;;###autoload
+(defun +corfu-popup-and-first ()
+  "Trigger corfu popup and select the first candidate."
+  (interactive)
+  (let ((corfu-auto-prefix 1))
+    (when (completion-at-point)
+      (corfu--exhibit)
+      (corfu--goto 0))))
+
+;;;###autoload
+(defun +corfu-popup-and-last ()
+  "Trigger corfu popup and select the last candidate."
+  (interactive)
+  (let ((corfu-auto-prefix 1))
+    (when (completion-at-point)
+      (corfu--exhibit)
+      (corfu-last))))
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -163,6 +163,8 @@
        (:after corfu
         (:map corfu-mode-map
          :i "C-SPC" #'completion-at-point
+         :i "C-n" #'+corfu-popup-and-first
+         :i "C-p" #'+corfu-popup-and-last
          :n "C-SPC" (cmd! (call-interactively #'evil-insert-state)
                           (call-interactively #'completion-at-point))
          :v "C-SPC" (cmd! (call-interactively #'evil-change)


LemonBreezes:
I like the idea but you should wrap the code for the functions to let-bind `corfu-auto-prefix` to 1.

LemonBreezes:
I like the idea but IMO we should do something like 
```elisp
(defun +corfu-popup-and-first ()
  "Trigger corfu popup and select the first candidate."
  ;; HACK using `corfu--auto-complete-deferred' to trigger the completion popup
  ;; without explicitly inserting any candidate
  (interactive)
  (let ((corfu-auto-prefix 1)
        (completion-at-point-functions '(cape-dabbrev))
        (cape-dabbrev-check-other-buffers evil-complete-all-buffers))
    (corfu--auto-complete-deferred)
    (when (> corfu--total 0)
      (corfu--goto 0))))
``` 

to mimic Evil's behavior more closely.

45mg:
Hmm... I get what you're going for, but in my mind these bindings are just another way to trigger Corfu, not Corfu-ized versions of `evil-complete-{next,previous}`. I think it would be confusing if the Corfu popup showed different candidates for exactly the same prompt, based only on the binding used; and if said bindings didn't respect the Capfs set by the user. What's your thought?

LemonBreezes:
> Hmm... I get what you're going for, but I'm not sure that we should turn the Corfu popup into versions of `evil-complete-{next,previous}`. I think it would be confusing if the Corfu popup showed different candidates for exactly the same prompt, based only on the binding used; and if said bindings didn't respect the Capfs set by the user. The way it's implemented now, it's basically a shortcut for TAB followed by C-n or C-p. What's your thought?

Hm. My impression is that if I am autocompleting with all the capfs, I will prefer TAB, because it's very rare that the first candidate is the one I want. In that case, typing `C-n` will usually be worse because if you don't get a candidate you want, you have to type `backspace` and start over. However, if you use `dabbrev` and possibly set `evil-complete-all-buffers` to nil, you are limiting the candidates enough to where you are more likely to hit one you want if you type `C-n`, since there is a logic that you are typing `C-n` for something you typed somewhere else.

45mg:
Thanks for the thought-out response.

> In that case, typing C-n will usually be worse because if you don't get a candidate you want, you have to type backspace and start over.

You can just press `C-p` in such a case, which should be easier than repeated backspace. Plus, in the case where the first candidate isn't the right one, the solution is usually pressing `C-n` a few times, so this binding can at least save you a single keypress.

> However, if you use dabbrev and possibly set evil-complete-all-buffers to nil, you are limiting the candidates enough to where you are more likely to hit one you want if you type C-n, since there is a logic that you are typing C-n for something you typed somewhere else.

I agree with you here, but I still think we shouldn't override the user's choice of Capfs. In addition to the concern of confusing the user (no way to visually distinguish between two different forms 
of completion), there is also the issue of customizability. If we include that let-binding in the functions, the only way for the user to get the regular Corfu popup with `C-p`/`C-n` is to add `:override` advice to the functions in their config - essentially redefining two functions entirely. And then those new function definitions become code they have to maintain themselves. On the other hand, if we leave it out, the user can easily get the behavior you specified by adding the let-binding with `:around` advice, which would not result in code duplication. So, one approach is a lot more flexible towards the user than the other.

In fact - we could suggest the `:around` advice method in the docs as a general way to limit completion sources for a given binding or command, so that users can make this decision for themselves. What do you thinkb

LemonBreezes:
> Thanks for the thought-out response.
> 
> > In that case, typing C-n will usually be worse because if you don't get a candidate you want, you have to type backspace and start over.
> 
> You can just press `C-p` in such a case, which should be easier than repeated backspace. Plus, in the case where the first candidate isn't the right one, the solution is usually pressing `C-n` a few times, so this binding can at least save you a single keypress.
> 
> > However, if you use dabbrev and possibly set evil-complete-all-buffers to nil, you are limiting the candidates enough to where you are more likely to hit one you want if you type C-n, since there is a logic that you are typing C-n for something you typed somewhere else.
> 
> I agree with you here, but I still think we shouldn't override the user's choice of Capfs. In addition to the concern of confusing the user (no way to visually distinguish between two different forms of completion), there is also the issue of customizability:
> 
>     * If we include that let-binding in the functions, the only way for the user to get the regular Corfu popup with `C-p`/`C-n` is to add `:override` advice to the functions in their config - essentially redefining two functions entirely. And then those new function definitions become code they have to maintain themselves.
> 
>     * On the other hand, if we leave it out, the user can easily get the behavior you specified by adding the let-binding with `:around` advice, which would not result in code duplication, and therefore be less likely to break if/when the module's code changes significantly.
> 
> 
> So, one approach is a lot more flexible towards the user than the other.
> 
> In fact - we could suggest the `:around` advice method in the docs as a general way to limit completion sources for a given binding or command, so that users can make this decision for themselves. What do you think?

Hm, so I'm going to try using this a bit more but my opinion right now is that:
1. `C-n`/`C-p` are free to be bound for commands like these.
2. I'm not sure what specific behavior most people would prefer. Should they mimic Evil's commands by invoking `cape-dabbrev`? Should they invoke the general autocompletion popup?

Your style of having `corfu-auto` set to `nil` and using the `C-n`, `C-p`, etc. commands is unique but I don't see why we should not support that style. I invite other people who see this PR to give their input on what they would prefer. For now, I defer to other reviewers.

45mg:
Cool. In case it wasn't clear by the way, this really only becomes useful when auto-completion is off, otherwise you rarely need this binding in the first place.

LemonBreezes:
> Cool. In case it wasn't clear by the way, this really only becomes useful when auto-completion is off, otherwise you rarely need this binding in the first place.

Hm. I think it's fine to merge this and the other PR #7750 to support that style of completion. Before that though, I really think you should let-bind `corfu-auto-prefix` to 1 in those commands so that they work even if you've only typed one character.

45mg:
> Before that though, I really think you should let-bind corfu-auto-prefix to 1 in these commands so that they work even if you've only typed one character.
Makes sense. Done.

modules/completion/corfu/autoload.el
```diff
@@ -37,3 +37,25 @@
...
+      (corfu--goto 0))))
+
+;;;###autoload
+(defun +corfu-popup-and-last ()
+  "Trigger corfu popup and select the last candidate."
+  ;; HACK using `corfu--auto-complete-deferred' to trigger the completion popup
+  ;; without explicitly inserting any candidate
+  (interactive)
+  (let ((corfu-auto-prefix 1))
+    (corfu--auto-complete-deferred)
```
LuigiPiucco, line 59: Is there a particular reason `corfu--auto-complete-deferred` was used here? It's an internal method and meant to be called with a timer TICK, so this seems a bit off. My testing suggests `completion-at-point` suffices; if the behavior you're working around is that of auto-inserting single matches, `completion-at-point` doesn't do that, and `corfu-on-exact-match` is there for configuring this as well. If it's that of inserting if a candidate is already selected, only `corfu-complete` does that.

45mg: This is necessary because Corfu doesn't set up the popup as soon as you call `completion-at-point`. That happens on `post-command-hook`, and `corfu--auto-complete-deferred` is the simplest way I've found to do it manually. If you just do `completion-at-point` followed by `corfu-next`, you're calling `corfu-next` without an active popup, so it does nothing.

I've updated the comments a bit to explain the hack a bit more.

LuigiPiucco: I think the issue is that `corfu--exhibit` is, by default, not called until `post-command-hook`, and the popup can only be interacted with after that. `corfu--auto-complete-deferred` works because it calls that at the end, which is needed for auto-completion, but it also does other things that are not necessary here. The flow to show the popup, from what I could gather, is: 1) get candidates by running `completion-at-point-functions`; 2) call `corfu--setup` to configure the state; 3) call `corfu--exhibit` to configure the popup. Manual completion does 1 and 2 immediately, but 3 happens in `post-command-hook`, and therefore after the `first/last` call, causing no effect. Since auto-completion is called from `post-command-hook` anyway, `corfu--auto-complete-deferred` calls that preemptively, to avoid another `post-command-hook` run when we know we're already in one, and incidentally that works here. But the following also works, and avoids some stuff we don't need here:
```emacs-lisp
(when (completion-at-point)
  (corfu--exhibit)
  (corfu--goto 0)) ; Or `corfu-last'
```
It's wrapped in `when` to avoid acting in case completion doesn't actually happen (such as if no candidates are found). I think it's worth changing to that, it seems more readable and efficient. Then the HACK comment can be removed.

45mg: Nice work. I incorporated your changes.

45mg:
I guess there's no more progress to be made here, so can this PR (and #7750) be approved?

hlissner (approved):

hlissner (changes requested):
To conform to our naming conventions, could you rename those commands to:

- `+corfu-popup-and-first` -> `+corfu/popup-and-first`
- `+corfu-popup-and-last` -> `+corfu/popup-and-last`

hlissner:
I've opted to emulate evil/vim's `C-n` and `C-p` and invoke dabbrev from them, ala `evil-complete-{next,previous}`, but if a Corfu popup is already open, it will behave like `C-j`/`C-k` instead. Thanks for bringing it to my attention, in any case.

PR #7749, opened 2024-03-23
docs(corfu): turning off auto-completion

45mg:
Document how to turn off auto-completion.

-----

### DIFF
--- a/modules/completion/corfu/README.org
+++ b/modules/completion/corfu/README.org
@@ -203,6 +203,15 @@ A few variables may be set to change behavior of this module:
   Whether to enable Corfu in the minibuffer. See its documentation for
   additional tweaks.
 
+** Turning off auto-completion
+To disable auto-completion, add this to your ~config.el~:
+#+begin_src emacs-lisp
+(after! corfu
+  (setq corfu-auto nil)
+  (setq corfu-auto-commands
+        (delete "self-insert-command\\'" corfu-auto-commands)))
+#+end_src
+
 ** Adding CAPFs to a mode
 To add other CAPFs on a mode-per-mode basis, put either of the following in your
 ~config.el~:


hlissner:
Unsetting `corfu-auto` should be enough. Why modify `corfu-auto-commands` as well?

45mg:
> Unsetting `corfu-auto` should be enough. Why modify `corfu-auto-commands` as well?

I'm not sure why, but it doesn't turn off until I do the latter. Could either of you test with just `(setq corfu-auto nil)` and see if it works?

LemonBreezes:
> > Unsetting `corfu-auto` should be enough. Why modify `corfu-auto-commands` as well?
> 
> I'm not sure why, but it doesn't turn off until I do the latter. Could either of you test with just `(setq corfu-auto nil)` and see if it works?

Yeah, for me, just setting `corfu-auto` to nil did not work either. Seems like it might be a bug in Corfu. Maybe.

45mg:
I tested this on a fresh install. Same thing.

hlissner:
Seems to work fine for me, but only at startup. You two sure you're not trying to `M-: (setq corfu-auto nil)` in a running session? 

Corfu consults [`corfu-auto` once, when `corfu-mode` is activated](https://github.com/minad/corfu/blob/c1e7b6190b00158e67347b4db0a8f7964e5d2f8b/corfu.el#L1324). Changing `corfu-auto` won't affect pre-existing `corfu-mode`-enabled buffers.

hlissner:
Unsetting `corfu-auto` disables auto completion in vanilla Emacs as well, as far as I can tell, so for the time being I've reduced this to only unsetting `corfu-auto` in 1598444 until I find an explanation for the behavior you two are seeing. Thanks for bringing this to my attention, in any case!

PR #7750, opened 2024-03-23
feat(corfu): `both` option for RET behavior

45mg:
Adds a new value of `+corfu-want-ret-to-confirm` to cover the case where the user wants to select the candidate as well as perform the normal behavior of RET.

Also, the logic for RET behavior has been refactored for clarity, and the variable's documentation has been clarified a little.

-----

### DIFF
--- a/modules/completion/corfu/README.org
+++ b/modules/completion/corfu/README.org
@@ -193,9 +193,13 @@ A few variables may be set to change behavior of this module:
 - [[var:corfu-preview-current]] ::
   Configures current candidate preview.
 - [[var:+corfu-want-ret-to-confirm]] ::
-  Enables commiting with [[RET]] when the popup is visible. Default is ~t~, may be set to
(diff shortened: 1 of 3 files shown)


LemonBreezes:
I like this refactor and the `both` option. Update the README.org to reflect this change too.

45mg:
The default behavior has not been changed, and the README didn't document the variable's options before, so I didn't touch it. Is it really necessary?

LemonBreezes:
> The default behavior has not been changed, and the README didn't document the variable's options before, so I didn't touch it. Is it really necessary?

Hm. It does currently mention the variable options before, though that is not necessary in general probably.

45mg:
Oh, my bad, it does. I just missed it. I'll add another commit in a bit.

45mg:
Updated the docs as requested.

hlissner (approved):

hlissner:
Sorry for the delay. I've merged this in 10484de. Thanks again for your help!

Issue #6951, opened 2022-11-18
Include @scturtle fixes for lsp-mode and rust-analyzer


gagbo:
## Purpose

Tracking the inclusion of [rust-analyzer specific fixes](https://github.com/emacs-lsp/lsp-mode/pull/1740#issuecomment-1319259794) for the rust module.

### Fixes

- Add a rust-analyzer specific `defmethod` that extracts the signature part of the `hover` response from `rust-analyzer`, so that the eldoc banner is more useful
- Add a forceful cache invalidation for the hover information as `rust-analyzer` car return too big bounds when the source code file is in a transient state (hard to repro though, so I can't say I have met this)

### Notes

Keep in mind lsp now uses either hash-table or plist for the server messages, so while this snippet is easy to add in personal configurations, extra safety steps must be taken to include this in Doom and have it work properly for plist and ht. Trying to remove the dash dependency in the code is an extra goal: it's unlikely that lsp-mode ever removes dash.el dependency but it would still be nice if the code in the repo here didn't have extra dependency to helper packages

### Going upstream

The reasons why those aren't upstreamed is basically the conversation (which I have not fully read yet) in the PR discussion [there](https://github.com/emacs-lsp/lsp-mode/pull/1740)

PR #7754, opened 2024-03-24
fix(rust): make lsp give better hover information

gagbo:
The caching of rust-analyzer (RA) is too aggressive and leads to bad
results.

RA does not give a protocol-compliant way to extract 1 line information
from the hover response, so `lsp-mode` displays the first line of the
result, which currently is the byte alignment of the data

Close: #6951

-----

### DIFF
--- a/modules/lang/rust/config.el
+++ b/modules/lang/rust/config.el
@@ -50,6 +50,34 @@
             'lsp-mode))
     (add-hook 'rustic-mode-local-vars-hook #'rustic-setup-lsp 'append))
 
+  ;; HACK Add @scturtle fix for signatures on hover on LSP mode.
+  ;;   This code has not been upstreamed because it depends on the
+  ;;   exact format of the response of Rust Analyzer, which is not
+  ;;   stable enough for lsp-mode maintainers.
+  ;; Ref: https://github.com/emacs-lsp/lsp-mode/pull/1740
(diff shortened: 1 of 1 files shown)


modules/lang/rust/config.el
```diff
@@ -50,6 +50,34 @@
...
     (add-hook 'rustic-mode-local-vars-hook #'rustic-setup-lsp 'append))
 
+  ;; HACK Add @scturtle fix for signatures on hover on LSP mode.
+  ;;   This code has not been upstreamed because it depends on the
+  ;;   exact format of the response of Rust Analyzer, which is not
+  ;;   stable enough for lsp-mode maintainers.
+  ;; Ref: https://github.com/emacs-lsp/lsp-mode/pull/1740
+  (when (and (modulep! +lsp) (not (modulep! tools lsp +eglot)))
+    ;; do not cache results from rust-analyzer
+    (advice-add #'lsp-eldoc-function :after (lambda (&rest _) (setq lsp--hover-saved-bounds nil)))
```
hlissner, line 60: Won't this affect non-rust buffers as well?

gagbo: Didn't think of it. Yes it would I think. So I need to either make the advice buffer-local somehow (doubt) or add a major-mode check

hlissner:
I've merged this with a few other minor changes into cf6d44d. Let me know if it gives you any trouble. Thanks again for your help!

PR #7758, opened 2024-03-25
fix(fold): properly support outline-minor-mode

45mg:
- Make `+fold/close-all` work for outline-minor-mode.

- Distinguish between `zo` and `zO` for outline-minor-mode alone, by binding `zO` to a new function `+fold/open-rec`. `zO` will now trigger `outline-show-subtree` in outline-minor-mode, while `zo` will only trigger `outline-show-branches`, keeping the subheadings folded. This matches the behavior in org-mode.

- Fix a bug in `+fold/open-all`, introduced in 7b9d00d6616d67cbc92765bb9da0dd0d2d9e9581, that makes it only operate on vimish-fold folds.

Similar fixes were PRed in #5724, but it looks like the author proceeded to use that branch as their main branch, resulting in hundreds of commits from master being added to the PR, before giving up and closing it.

Note that I have not enabled outline-minor-mode anywhere it's not already on. So the concerns raised in #5724 about outline-minor-mode replacing other kinds of folds should not apply here. Users who are aware of outline-minor-mode and enable it explicitly will probably understand if turning on something that's off-by-default doesn't work perfectly. This PR is only intended to ensure that the existing support works as intended.

-----

### DIFF
--- a/modules/editor/fold/autoload/fold.el
+++ b/modules/editor/fold/autoload/fold.el
@@ -79,6 +79,18 @@ Targets `vimmish-fold', `hideshow', `ts-fold' and `outline' folds."
           ((+fold--ts-fold-p) (ts-fold-toggle)))))
 
 ;;;###autoload
+(defun +fold/open-rec ()
+  "Recursively open the folded region at point.
+
+Targets `vimmish-fold', `hideshow', `ts-fold' and `outline' folds."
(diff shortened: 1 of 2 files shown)


hlissner (approved):

hlissner:

PR #7808, opened 2024-04-11
tweak(ligatures): fallback to derived mode

Dev380:
Ligatures now can fall back to derived mode
ligatures if they are not available for the
current mode. This has lower precedence than
ligatures directly defined for the current
mode.

-----


hlissner:
I opted for a slightly simpler implementation in d941078, which should achieve the same result. Thanks for bringing it to my attention, in any case!

PR #7809, opened 2024-04-11
tweak(ocaml): add more extra ligatures

Dev380:
Tuareg mode (for ocaml) current uses
its own ligatures. This extends that
with some extra, non-conflicting ones
using the `:ui ligatures` module's
pre-defined ligatures.

-----

Before:
20240410_21h33m24s_grim

After:
20240410_21h32m05s_grim

I'm using the dot instead of the cartesian product sign for the tuple type because it uses the * character, which is also used for multiplication. One way to resolve this is to add the cross product symbol to the default `+ligatures-extra-symbols` plist as a compromise between X for multiplication and X for cartesian product. I'm not sure if this is optimal, though, should this be done?

I added the tuple symbol just for the screenshot, but I might remove it because of how hacky it is (matches on ", " to avoid it being lopsided, but then tuples not in that format won't work + there are no spaces on each side).

### DIFF
--- a/modules/lang/ocaml/config.el
+++ b/modules/lang/ocaml/config.el
@@ -17,6 +17,24 @@
   (set-ligatures! 'tuareg-mode :alist
     (append tuareg-prettify-symbols-basic-alist
             tuareg-prettify-symbols-extra-alist))
+  ;; some extra ligatures
+  ;; unfortunately, ocaml does not have many keywords
+  ;; that can't be used as identifiers, so in theory
+  ;; some user defined variables could be prettified.
+  ;; However, naming your function "string" is probably
+  ;; as confusing with and without ligatures.
+  (set-ligatures! 'tuareg-mode :alist
+    :true "true"
+    :false "false"
+    :null "[]"
+    :int "int"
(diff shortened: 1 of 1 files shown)


i-am-linja:
Do not like. _Far_ too terse. Doesn't flow at all. And almost all of these symbols are just wrong:
- **S** and **B** are not standard notation for `string` and `bool`.
- The closest concept to `t list` with a "standard" notation would be `t*`, with a raised asterisk. Functor application in general is not necessarily postfix.
- There is no standard notation at all for `true` and `false`. There's notation for "top" and "bottom", but those are _types_, and using them as values is inadvisable.
- A tuple is a cartesian product, not a _tensor_ product as you have on the right.
- ø is the empty _set_, not the empty _list_.

hlissner:
Thanks for the PR, but I must agree with @i-am-linja, plus these seem a little *too* magical, in the sense that they don't map on fixed variables, running a big risk of stepping on people's toes. I'd rather leave this to end users to configure if they want it, than make it the module's default. Thanks for bringing it to my attention, in any case!

PR #7810, opened 2024-04-11
feat(ophints): add goggles for non-evil users

Dev380:
This adds the goggles package as a module flag
to use goggles instead of volatile-highlights-mode as a non-evil user.

-----


hlissner (changes requested):
Is there any compelling reason to keep volatile-highlights? If not, I'd rather do away with it entirely.

Dev380:
I'm not sure how the release schedule works so I kept it in to avoid a breaking change; should vhl be removed and goggles.el be the new default when evil is off?

Dev380:
VHL is removed

hlissner:
I've compressed this and some other modifications to the module in 3431ddd. Thanks for your help and patience, and for bringing it to my attention!

Issue #6416, opened 2022-05-24
flycheck-posframe popup doesn't disappear while the cursor is moving


OnionHat:
### What did you expect to happen?

When I move my cursor from a line with an error the popup should go away almost instantaneously.

### What actually happened?

When I move my cursor from a line with an error the popup doesn't disappear. It only disappears after i stop moving the cursor. This happens both in EVIL mode and Emacs mode. 

https://user-images.githubusercontent.com/46026042/170061954-63cb4736-10a6-4290-ab49-f41ae3b871f1.mp4

I found this issue which might be related to my issue from one year ago https://github.com/alexmurray/flycheck-posframe/issues/30, but it seems like the project isn't maintained or active any more.

### Describe your attempts to resolve the issue

I tried to play set different values for this three variables to see if the popup would disappear faster or slower but non of them had any effect on the issue.
```lisp
(setq flycheck-idle-change-delay 0
      flycheck-idle-buffer-switch-delay 0
      flycheck-display-errors-delay 0.5))
```
I removed my private config and installed a fresh new doom emacs with `doom install` and added `+childframe` flag to `syntax` in `init.el` to still seeing the same issue described over.

### Steps to reproduce

1. Remove private config
2. Reinstall doom emacs with `doom install`
3. Add `+childframe` flag to `:checkers syntax` in `init.el`
4. `M-x doom/reload`
5. Make an intentionally syntax error like removing a parentheses
6. Go to the line with the error and wait til you see a popup with an error message
7. Start moving up/down/left/right without stopping

### System Information

https://pastebin.com/EuDG3nmM

OnionHat:
After some more testing and debugging, doing `doom upgrade` seems to have solved my issue.

PR #7813, opened 2024-04-12
fix(flycheck): make child frame disappear instantly

Dev380:
Currently, if the +childframe flag is on, the frame will linger for a few seconds before disappearing
after leaving the highlighted error. A post command hook needs to be added to fix this, courtesy of
[jadestrong](https://github.com/doomemacs/doomemacs/issues/6416#issuecomment-1156164346).

Fix: #6416
-----


hlissner:
This isn't *truly* a bug, it's just how flycheck works (it doesn't offer a hook-on-a-timer to attach "stop displaying the error" functions). In any case, I've reworked a slightly more isolated implementation in a189c21, which should resolve your issue. Thanks for bringing it to my attention!

PR #7814, opened 2024-04-12
tweak(flycheck): prettify default settings

Dev380:
Makes the default settings of flycheck a
bit prettier in child frame mode, using
faces from the default "pretty config" in
`flycheck-posframe`. Additionally, unicode
characters can be optionally enabled as
prefixes for flycheck messages.

I'm not sure if +icons should be enabled be default, is it a reasonable to assume that users will have fonts that support unicode icons?

-----

Before:
20240411_22h50m08s_grim

After (with +childframe and +icons)
20240411_22h53m13s_grim

### DIFF
--- a/modules/checkers/syntax/README.org
+++ b/modules/checkers/syntax/README.org
@@ -16,12 +16,14 @@ This module provides syntax checking and error highlighting, powered by
   *Requires GUI Emacs.*
 - +flymake :: Leverages the inbuilt [[doom-package:flymake]] for error and
   diagnostics highlighting.
+- +icons :: Uses unicode icons instead of ASCII characters to prefix tooltips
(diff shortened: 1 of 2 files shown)


hlissner:
Went with a slightly tweaked choice of icons in c8b5bf7 because the X was oversized in some common fonts (like JetBrainsMono), and though I use `⚠` twice, the difference in faces distinguish them adequately. Also, I don't think it's the module's (or even the package's) jurisdiction to override faces like that (I'd rather leave it to themes or users).

> I'm not sure if +icons should be enabled be default, is it a reasonable to assume that users will have fonts that support unicode icons?

I don't think it should be. Not only is it not enabled by default anywhere else, but even with a common font like JetBrains Mono, the `X` in the original PR came out an oversized emoji; I'd rather no bet on these little inconsistencies.

PR #7816, opened 2024-04-12
fix: scratch buffer initializing too early

Dev380:
This is a daemon-only issue.
The scratch buffer is initialized very early
in the startup process. However, some minor
modes (such as `flycheck-posframe-mode`) and
other things that should run on every prog-mode
or elisp mode buffer won't get to set a hook and
have it run. This recreates the scratch buffer
the first time it's switched to to fix this.
For some reason, simply calling the major mode
hook for the scratch buffer after doom-after-init-hook
doesn't work.

For some reason, removing the check for the buffer name in the hook function (i.e. running it on the first buffer switch after doom-after-init-hook) also works, and the buffer switch hook runs automatically despite me not switching any buffers, but just doom-after-init-hook by itself doesn't work and I don't know why.

-----

### DIFF
--- a/lisp/doom-start.el
+++ b/lisp/doom-start.el
@@ -320,6 +320,25 @@ If RETURN-P, return the message as a string instead of displaying it."
     (add-hook 'after-change-major-mode-hook #'doom-run-local-var-hooks-h 100)
     (add-hook 'hack-local-variables-hook #'doom-run-local-var-hooks-h)))
 
+;; fix bug in daemon mode where the scratch buffer is initializer too early
+;; HACK: for some reason `run-mode-hooks' does not work
+;; nor does calling `initial-major-mode' (although that
+;; does work interactively for some reason) so the buffer
+;; needs to be regenerated
+(when (daemonp)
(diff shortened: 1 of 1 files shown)


hlissner:
The reason it isn't working is likely because the scratch buffer is created *very* early in Emacs' startup process, before any configuration (much less Doom) is evaluated. This means any configuration we attach to the scratch buffer won't be applied to the initial buffer.

In any case, I don't think I'll do anything about this, because changing `initial-major-mode` to anything besides `fundamental-mode` introduces a lot of race conditions due to the above, not to mention hurts startup time considerably.

What's more, Doom offers its own on-demand scratch buffer behind `doom/open-scratch-buffer` (`SPC x` or `SPC b x`) and `doom/switch-to-scratch-buffer` (`SPC b X`), which is superior in every way. Contents and major mode persist between sessions, and there are also project-local scratch buffers behind `doom/open-project-scratch-buffer` (`SPC p x`) and `doom/switch-to-project-scratch-buffer` (`SPC p X`)).

PR #7818, opened 2024-04-13
Fix rust ts mode priority

Dev380:
If `rust-ts-mode` is used as the base of `rust-mode`
then it will [automatically add itself](https://github.com/emacsmirror/emacs/blob/2fc7e21f5e75ea6b00d6f7344335f44f5663d955/lisp/progmodes/rust-ts-mode.el#L557) to `auto-mode-alist`
before `rustic-mode`, causing every file other than
the first to be opened to default to `rust-ts-mode`
instead. This can be fixed by removing the treesitter
mode from the list after it gets set when the package
loads. This will change behaviour if:

1) both this module and the tree sitter mode are enabled
2) the user loads the `rust-ts-mode` package in some way
3) the user intends for `rust-ts-mode` to be prioritized
over the doom module.

However, this is unlikely to be the case.

-----

### DIFF
--- a/modules/lang/rust/README.org
+++ b/modules/lang/rust/README.org
@@ -32,6 +32,11 @@ e.g. ~cargo~.
 - rustic has been modified /not/ to automatically install lsp-mode or eglot if
   they're missing. Doom expects you to have enabled the [[doom-module::tools lsp]] module
   yourself.
+- If doom-package:rust-ts-mode is loaded, it'll add itself as the default major mode for rust
+  with higher priority than doom-package:rustic-mode. This package disables doom-package:rust-ts-mode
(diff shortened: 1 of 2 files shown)


hlissner:
Thanks for the PR, but I have to turn it down. That `("\\.rs\\'" . rust-mode)` entry won't be moved to the top of `auto-mode-alist` unless you have `(setq rust-mode-treesitter-derive t)` in your config (it's default value is `nil`). 

By merging this (or making any attempt to undo this `auto-mode-alist` entry), I'd be interfering with folks who genuinely want `rust-mode-treesitter-derive` enabled in their configs, for whatever reason. Since it's already opt-in by default, I think prioritizing `rust-mode` ought to be the expected outcome for users, in this case.

PR #7819, opened 2024-04-13
feat(format): format on save only in certain directories

Dev380:
Adds support for formatting on save in certain
directories/files only, with a new list variable
`+format-on-save-enabled-directories`.

-----

### DIFF
--- a/modules/editor/format/README.org
+++ b/modules/editor/format/README.org
@@ -26,7 +26,8 @@ be formatted and returned to the buffer using
 - +onsave ::
   Enable reformatting of a buffer when it is saved. See
   [[var:+format-on-save-disabled-modes]] to disable format on save for certain
-  major modes.
+  major modes and var:+format-on-save-enabled-files-directories to only
+  format on save for specific files/directories.
 
 ** Packages
 - [[doom-package:apheleia]]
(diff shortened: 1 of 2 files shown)


hlissner:
Thanks for the PR, but I cannot accept this. Doing directory-based configuration with variables is not optimal, and apheleia already provides a better solution: its `apheleia-inhibit` variable, which can be set a myriad of ways:

1. For whole projects/directories, it can be set in a `.dir-locals.el` file, containing:

   ```lisp
   ((nil (apheleia-inhibit . t)))  ; To apply to *all* buffers
   ```

   `apheleia-formatter` (or `+format-with`) can also be set from these files to specify a specific formatter. E.g.

   ```lisp
   ((js2-mode (apheleia-formatter . prettier)))
   ```

2. These can also be set from file-local variables at the top of an file (adjust the comment starters according to the major mode), e.g.

   ```elisp
   ;; -*- apheleia-inhibit: t -*-
   ...
   ```
   
   Or at the end of a JS file:
   ```js
   # Local Variables:
   # apheleia-inhibit: t
   # End:
   ```

3. These variables can also be set from your elisp config, using `setq-hook!` or an `apheleia-mode-functions` hook, e.g.

   ```elisp
   (setq-hook! 'js2-mode-hook apheleia-inhibit t)
   ;; or
   (defvar my-projects '("/x/y/z/" "/foo/bar"))
   (add-hook! 'apheleia-mode-functions
     (lambda ()
       (when (cl-find (doom-project-root) my-projects :test #'file-equal-p)
         (setq-local apheleia-inhibit t))))
   ```

Since implementing a more specialized way to do any of these things is rife with edge cases and opinions, I'd rather leave it to the user to select the approach they like best.

PR #7825, opened 2024-04-16
fix(lib): doom-packages-ensure: fix local packages always built

levindu:
If the package is local, the local-repo has absolute local file path, and straight-modified-dir will return as it is, which always exists.

Fix it by passing package only to straight-modified-dir.

-----


hlissner (approved):

hlissner:
Another good catch. Thanks again!

PR #7049, opened 2023-01-20
fix(god): remove hard coding of cursor color

p4v4n:
When switching themes (light to dark or vice versa) with god-mode enabled sometimes the cursor becomes invisible.
This change removes the hard coded value and keeps the cursor color in sync with the current theme.

PR #7841, opened 2024-04-21
fix(god): fix cursor color

darth10:
This was broken in aa079b0e3ccc0548a677907a08cd12c5b878e44c (#7049).

When the cursor changes from color A to color B, it doesn't ever change back to color A, as `(face 'cursor)` will now return color B.
Since the default cursor in the Doom theme uses the `highlight` background color, we can use that when setting the cursor color when this module is enabled.

### Screenshots

For example, when the `:editor god` module is enabled, change the cursor color by moving to the end of a long line, and then move to the beginning of the line so that the default cursor color should be shown.

#### Before
image

#### After
image

-----
- [X] My changes are visual; I've included before and after screenshots.

### DIFF
--- a/modules/editor/god/autoload.el
+++ b/modules/editor/god/autoload.el
@@ -26,7 +26,7 @@
           (cond (buffer-read-only +god-read-only-mode-color)
                 (is-fill-overflow +god-fill-overflow-color)
                 (overwrite-mode +god-overwrite-mode-color)
-                (t previous-cursor-color))))
+                (t (face-background 'highlight)))))
     (setq cursor-type next-cursor-type)
     (unless (eq previous-cursor-color next-cursor-and-modeline-color)
       (set-cursor-color next-cursor-and-modeline-color))


hlissner:
Since only Doom's themes (and only a subset of them) use `highlight`, reading it would be a fragile solution, so I read the frame parameter instead (which is what `set-cursor-color` sets) in c48a5e5, which should resolve the original issue. Let me know if it gives you any trouble. Thanks for bringing this to my attention!

darth10:
Thanks for taking a look @hlissner! 

Unfortunately, c48a5e5f9c2fd171e6a714840992c6370d5f4528 doesn't fix the issue.
I still see the old behaviour (in the PR description) when I do this:
> change the cursor color by moving to the end of a long line, and then move to the beginning of the line so that the default cursor color should be shown.

Also, the cursor color is now gray when I start Emacs, as `(frame-parameter nil 'cursor-color)` is `"Gray"` for some reason.

I might try and fix this without using `(face-background 'highlight)`, but could use some help/ideas.

darth10:
I thought about the original issue (in #7049), and came up with a potentially better solution.
The original issue was:
> When switching themes (light to dark or vice versa) with god-mode enabled sometimes the cursor becomes invisible.

If we go back to [this version of `modules/editor/god/autoload.el`](https://github.com/doomemacs/doomemacs/blob/5b5dbe1c47dea9c29df91c7dee5da331ee513173/modules/editor/god/autoload.el), we could just reload or set `+god-default-color` through `doom-load-theme-hook`.
What do you think @hlissner?

PR #7842, opened 2024-04-23
Due to change in Cargo, update elisp-tree-sitter

meedstrom:
The current pinned commit is around 2 years old.

This PR updates the commit to the one matching https://github.com/emacs-tree-sitter/elisp-tree-sitter/pull/249

That change is backards-compatible with old Cargo behavior, so it should be safe.

### DIFF
--- a/modules/tools/tree-sitter/packages.el
+++ b/modules/tools/tree-sitter/packages.el
@@ -1,7 +1,7 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; tools/tree-sitter/packages.el
 
-(package! tree-sitter :pin "3cfab8a0e945db9b3df84437f27945746a43cc71")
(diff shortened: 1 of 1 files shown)


jeetelongname:
Hiya, thanks for the PR. Please refer to the [git conventions(https://discourse.doomemacs.org/t/git-conventions/2407) and change your commit text to conform with that of a `bump` commit. 

Do note that this will probably be subsumed into a full module commit (and hopefully soon a module rewrite) so don't take it personally if this is closed

meedstrom:
Ok, I think now it's the convention you want?  And no problem.

hlissner:
The tree sitter module needed a larger bump, so I've incorporated this into ee10764. Thanks for bringing this to my attention, in any case!

PR #7848, opened 2024-04-26
fix(window-select): load winum earlier

LemonBreezes:
This is so that a user can use `winum-auto-setup-mode-line` with value `t`. Since that modifies the default mode line, the entire package needs to be loaded very early for that option to be available to a user.

This is cheaper and better than mapping over every buffer and modifying their buffer-local value for mode-line-format.

Also, I disagree with that comment in the sense that modifying `mode-line-format` is not important at all since it is done correctly. The issue to me is that since it modifies the default value of `mode-line-format`, the package should be loaded early for all buffers to have that mode-line segment.

### DIFF
--- a/modules/ui/window-select/config.el
+++ b/modules/ui/window-select/config.el
@@ -23,12 +23,9 @@
 
 (use-package! winum
   :when (modulep! +numbers)
-  :after-call doom-switch-window-hook
+  :hook (doom-after-init . winum-mode)
   :config
-  ;; winum modifies `mode-line-format' in a destructive manner. I'd rather leave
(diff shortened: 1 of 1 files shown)


hlissner:
I think modifying the global value of `mode-line-format` is poor practice. At the very least, it should add its segment to `global-mode-string`, which is the standard "dumping ground" for extra mode-line segments, and otherwise be left to mode-line plugins to have their own segment for.

I might accept a PR that prepends `winum--mode-line-segment` to `global-mode-string` if `:ui modeline` isn't enabled, but as is, I cannot accept this PR. Thanks for bringing it to my attention, in any case.

PR #7849, opened 2024-04-26
refactor(lib): use ripgrep instead of `git grep`

marienz:
Although this is a refactor for normal users of Doom, it is a bugfix for one (I assume) unsupported configuration: when Doom is not running from a git checkout. In that case, `doom--help-package-configs` currently returns Git's error messages. This breaks `doom/help-packages` because it expects each returned line to contain at least one `:` character, so it errors out with `Wrong number of arguments: (file line _match &rest), 2`.

Using ripgrep here should be equivalent for normal users as long as they have not added untracked files not covered by Doom's .gitignore.

Assuming Doom outside of a git checkout is indeed unsupported, please close this PR if you don't want to spend time on it: when I saw the TODO in that function I figured I'd send this in, but it shouldn't be too hard to work around locally.

(I'm running into this running Doom from the Nix store. So far, this is the only issue I've encountered doing so. I've tested this change with that setup and with a more normal Doom install.)

-----

### DIFF
--- a/lisp/lib/help.el
+++ b/lisp/lib/help.el
@@ -468,14 +468,12 @@ will open with point on that line."
 
 (defun doom--help-package-configs (package)
   (let ((default-directory doom-emacs-dir))
-    ;; TODO Use ripgrep instead
     (split-string
      (cdr (doom-call-process
-           "git" "grep" "--no-break" "--no-heading" "--line-number"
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
This is precisely what I'd wanted to do, but never got around to, so it's a welcome change. Thanks for your help!

PR #7852, opened 2024-04-29
Add binding for find-file-at-point

carmenbianca:
This is a very simple PR to bind 'SPC f h' to `find-file-at-point`. This is immensely useful in terminal emulators. 'h' is a mnemonic for 'here'. 'c' and 'C' (for 'cursor') were already taken.

Fix: N/A
Ref: N/A
Close: N/A

-----

### DIFF
--- a/modules/config/default/+emacs-bindings.el
+++ b/modules/config/default/+emacs-bindings.el
@@ -79,6 +79,7 @@
        :desc "Browse emacs.d"              "E"   #'doom/browse-in-emacsd
        :desc "Find file"                   "f"   #'find-file
        :desc "Find file from here"         "F"   #'+default/find-file-under-here
+       :desc "Find file under cursor"      "h"   #'find-file-at-point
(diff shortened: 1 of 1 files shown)


LemonBreezes:
If you are using the `:completion vertico` module, you can use `find-file`, and then `M-n` to select the file name at point. As a general rule, if you are doing minibuffer completion and you type `M-n` when there are no history items ahead of you, Emacs will try to guess what you are trying to complete. This is called "future history".

hlissner:
Thanks for the PR, but I must turn it down. `+lookup/file` already fulfills `find-file-at-point`'s purpose (given you've enabled the `:tools lookup` module), and it's already bound to `C-c s f` along with other lookup (i.e. search) keybinds. I might consider replacing `C-c f f` with `find-file-at-point`, since the command falls back to `find-file` if the cursor isn't on a recognizable file, but I would like to avoid adding redundant keys to the leader keymap, given how precious the real estate on it is.

PR #7868, opened 2024-05-19
Bump Parinfer to v0.9.0, make use of some niceties

welcome-linja:
parinfer-rust-mode came out with version 0.9 last week, the headline feature of which is using their own fork of parinfer-rust with better Emacs integration. Notably, the nominated macOS binary is now built for arm64e rather than x86_64, which will break some setups; this is inevitable without a complex workaround (as vanilla parinfer-rust is no longer supported, and the fork doesn't provide an x86_64 binary), and it makes just as many setups now work when they previously did not.

I also enabled the option parinfer-rust-disable-troublesome-modes, which is behind a prompt by default. I was unsure about this but I believe it is in line with the existing decision to enable auto-download and the DOOM philosophy in general: reduce the amount of hassle the user goes through to get things working together.

(Apologies about the commits; I discovered the guidelines when opening this PR. If I have to redo them I will.)

-----

### DIFF
--- a/modules/editor/parinfer/config.el
+++ b/modules/editor/parinfer/config.el
@@ -17,7 +17,8 @@
                ((featurep :system 'linux)   "parinfer-rust-linux.so")
                ((featurep :system 'windows) "parinfer-rust-windows.dll")
                ((featurep :system 'bsd)     "libparinfer_rust.so")))
-        parinfer-rust-auto-download (not (featurep :system 'bsd)))
+        parinfer-rust-auto-download (not (featurep :system 'bsd))
+        parinfer-rust-disable-troublesome-modes t)
   :config
   (map! :map parinfer-rust-mode-map
         :localleader
--- a/modules/editor/parinfer/packages.el
+++ b/modules/editor/parinfer/packages.el
@@ -1,4 +1,4 @@
 ;; -*- no-byte-compile: t; -*-
 ;;; editor/parinfer/packages.el
 
-(package! parinfer-rust-mode :pin "8df117a3b54d9e01266a3905b132a1d082944702")
+(package! parinfer-rust-mode :pin "d3bfb2745cc0858e2741dc2a2f00a86f456656ec")


welcome-linja:
Shit, you know what I just realised? `troublesome-modes` by default only includes built-in modes, but plenty of Spaceman's options would be considered troublesome. Merging in this state would cause _crazy_ issues, and I really can't be bothered to do it properly. Sorry.

gagbo:
What are the "Spaceman’s options" ?

I’ve worked a bit on that parinfer version, and it’s a pretty important change to be able to handle lisps as good as neovim, and to deal with parens and quotes in comments too (which currently break a lot of setups).

I’m unpinned (to test updates) so it doesn’t really matter to me now, but those are good changes. Anyway, you _definitely_ don’t have to make a PR here, and instead just change the pin in your `packages.el` and do your tweaks, I was just curious about this part of the post

welcome-linja:
“Spaceman’s options” are what happens when autocorrect gets overzealous and I forget which Emacs distro I’m PRing :P

Yeah it looks great and I was excited about it before I remembered I’m actively resisting procrastediting my configs and I’m not otherwise touching any lisps (really wish I had a use for them because they’re cool), but the problem with this PR was it was unfinished: it would switch off _some_ troublesome modes, but not all the ones we were using (I thought of smartparens and maybe Apheleia, but I’m in no condition to trawl all the packages in all the modules and figure out how to make each one append to the variable). This PR in this state would cause smartparens and parinfer to be active at the same time, which would be a disaster; a DOOM-standard one would make sure only one is active at a time, and same for every plugin that auto-inserts whitespace or delimiters.

But actually, now I’m thinking about it, if that were a problem with the _existing_ parinfer module it would have been noticed and fixed. So I guess _I_ might have been overzealous in closing this. Whatever, it was a trivial amount of work that can easily be redone (and annotate the commits right this time), and I really should understand how it already works before jamming in something that looks like a solution. Or maybe I’m definitely overthinking it, I don’t know, give me a break, I’m conked today.

PR #7876, opened 2024-05-30
Parinfer bump, take 2

welcome-linja:
Ok, this time my commits conform to convention, the upstream maintainer has confirmed I was overzealous closing the last one, and there's even a new version out. Same caveats from last time apply: this now breaks Intel macOS, but fixes Apple Silicon, and Arm anything else still doesn't work. Perhaps we could provide some auto-compile for upstream-unpackaged platforms but that's out of my wheelhouse.

Next step is to enable for other sexp-based modes, like `dune-project` and `wat`. Maybe I'll get onto that at some point.

-----

### DIFF
--- a/modules/editor/parinfer/config.el
+++ b/modules/editor/parinfer/config.el
@@ -17,7 +17,8 @@
                ((featurep :system 'linux)   "parinfer-rust-linux.so")
                ((featurep :system 'windows) "parinfer-rust-windows.dll")
                ((featurep :system 'bsd)     "libparinfer_rust.so")))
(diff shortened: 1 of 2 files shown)


lyallcooper:
Was just about to open a PR to bump the version myself. The current version `8df117a` has some issues that prevented it from working for me, but they seem to be fixed in `e9a23e13`.

Thanks for beating me to it!

hlissner:
To conform the bump commit to our conventions, I've adjusted and merged this into 7c20ef2. Thanks for bringing it to my attention!

PR #7890, opened 2024-06-20
add support for relative path for gf

liuzhishan:
In evil mode, `gf` doest work when path under cursor has common ancstor with current buffer file. This seems to be a common situation. For example, suppose I have the following files:

```
project
└── src
    ├── a
    │   └── a.h
    └── b
        └── b.h
```
and in b.h, the file `a/a.h` is included:

```
// b/b.h
#include "a/a.h"
```
The `gf` command was mapped to `+lookup/file`. If I want to open `a/a.h` file using `gf` in `b.h` file, it doest not work. `+lookup/file` seems to lookup files in some predefined directories.

To solve this problem, I add a new function `+lookup/open-file-under-cursor`:
1. Find path under cursor.
2. If the path is absolute path, then try opening the file using `find-file`.
3. If the path is a relative path, find directory of current buffer file, then try every combination of sub directory and path under cursor. If any of them exists, then we found the target, try opening the file.
4. Otherwise, fallback to `+lookup/file` as the origin logic.

And also I changed the `gf` key mapping in `modules/editor/evil/config.el`:
```
:nv "gf"  #'+lookup/open-file-under-cursor
```

-----

### DIFF
--- a/modules/editor/evil/config.el
+++ b/modules/editor/evil/config.el
@@ -469,7 +469,7 @@ directives. By default, this only recognizes C directives.")
        :nv "K"   #'+lookup/documentation
        :nv "gd"  #'+lookup/definition
        :nv "gD"  #'+lookup/references
-       :nv "gf"  #'+lookup/file
+       :nv "gf"  #'+lookup/open-file-under-cursor
        :nv "gI"  #'+lookup/implementations
        :nv "gA"  #'+lookup/assignments)
       (:when (modulep! :tools eval)
--- a/modules/tools/lookup/autoload/lookup.el
+++ b/modules/tools/lookup/autoload/lookup.el
@@ -399,6 +399,30 @@ Otherwise, falls back on `find-file-at-point'."
 
         ((user-error "Couldn't find any files here"))))
 
+;;;###autoload
+(defun +lookup/open-file-under-cursor ()
+  "Open the file under cursor and go to position if present.
+
+Supports both relative path and absolute path. If path is relative,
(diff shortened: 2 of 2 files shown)


hlissner:
I've incorporated the idea into `+lookup/file` itself in 0a2bcf9, though only if the given path has more than one segment (i.e. `a/a.h` will work, `a.h` won't), since the latter would produce too many false positives. Let me know if it gives you any trouble.

In any case, `ffap-alist` already does a project-root lookup, so `src/a/a.h` would've worked too, and if users want this guess work to be less or more DWIM, I'll leave it to them to add their own resolvers to `ffap-alist`. Thanks for bringing it to my attention, anyhow!

liuzhishan:
> I've incorporated the idea into `+lookup/file` itself in [0a2bcf9](https://github.com/doomemacs/doomemacs/commit/0a2bcf928c868db5cf41041bd707d3f395504130), though only if the given path has more than one segment (i.e. `a/a.h` will work, `a.h` won't), since the latter would produce too many false positives. Let me know if it gives you any trouble.
> 
> In any case, `ffap-alist` already does a project-root lookup, so `src/a/a.h` would've worked too, and if users want this guess work to be less or more DWIM, I'll leave it to them to add their own resolvers to `ffap-alist`. Thanks for bringing it to my attention, anyhow!

Great! Seems to be a better solution. Indeed if given path has no segment, there would be confusion. It should be ignored.

Glad I'm helpful!

PR #7894, opened 2024-06-24
fix(lib): doom-project-find-file: use transient project

132ikl:
If DIR is not a project and does not have root markers, create a transient project instead. This ensures that `project-find-file-in`, and by extension Vertico, is still used in non-project directories instead of falling back to `find-file`

`+vertico/consult-fd-or-find` was removed from `doom-project-find-file` in 1b0af3bfc747, with the intention that Vertico users would instead use Vertico's integration with `project-find-file-in`. However, `project-current` returns `nil` for folders that are not projects and do not have any root markers. This causes `doom-project-find-file` to fallback to `find-file` for commands which use `doom-project-find-file` on directories which aren't necessarily projects, such as `doom/find-file-in-private-config`, `+default/find-in-notes`, and especially `+default/find-file-under-here`. This means that each of these commands ends up having the same behavior as their `browse` counterparts rather than recursively searching through the target directory.

I moved the `project-find-file-in` clause below the `helm-find-files` clause, since the transient project would always take priority over `helm-find-files`, rather than only taking priority when a project exists in that directory as before. If this approach isn't ideal, perhaps there could be two blocks for `project-find-file-in`, one before `helm-find-files` which does not set `project-current-directory-override`, and one after `helm-find-files` which does.

Ref: 1b0af3bfc747

-----

### DIFF
--- a/lisp/lib/projects.el
+++ b/lisp/lib/projects.el
@@ -152,11 +152,12 @@ If DIR is not a project, it will be indexed (but not cached)."
           ((and (bound-and-true-p ivy-mode)
                 (fboundp 'counsel-file-jump))
            (call-interactively #'counsel-file-jump))
-          ((when-let ((pr (project-current nil dir)))
-             (project-find-file-in nil nil pr)))
           ((and (bound-and-true-p helm-mode)
                 (fboundp 'helm-find-files))
            (call-interactively #'helm-find-files))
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
LGTM. Thanks for your help!

PR #7901, opened 2024-06-26
fix(fold): don't expect evil-vimish-fold

45mg:
This module is also used by non-Evil users. Also,
`evil-vimish-fold/next-fold` and friends just wrap the corresponding vimish-fold functions in `dotimes`, anyway.


hlissner (approved):

hlissner:
LGTM. Thanks for your help!

PR #7902, opened 2024-06-26
fix(fold): properly handle recursive folds

45mg:
With four folding systems in play, there are numerous ways to create nested folds (eg. hideshow folds inside folded outline headings with `+fold/close-all`). So `+fold/open-rec' needs to be able to open all kinds of folds within the newly unfolded region. It can't stop at unfolding the type of fold it was called on.


hlissner (approved):

hlissner:
Thanks again!

PR #7903, opened 2024-06-26
fix(fold): +fold/next: actually support outlines

45mg:
The docstring says that this function will move to the next outline heading, but this had not been implemented.


hlissner:
This one is tricky. `+fold/next` and `+fold/previous` should jump to the next/previous fold, not foldables (i.e. unfolded regions), which is what `outline-{next,previous}-heading` do in `outline-minor-mode`. Can this be tweaked to skipped unfolded headings?

45mg:
Ok, this version should do what you described.

hlissner (approved):

hlissner:
LGTM. Thanks again!

PR #7907, opened 2024-06-27
fix(fold): fold-levels off-by-one

45mg:
Without this change, we need to pass a prefix arg that is one more than the actual fold-level we're trying to achieve; `C-u 2` is equivalent to not passing a digit arg. While this looks intentional, I don't see what purpose it serves.

This change provides saner behavior that matches other commands that handle fold-levels (eg. `org-fold-hide-sublevels`).


hlissner (approved):

hlissner:

PR #7924, opened 2024-07-09
fix(format): apheleia doesn't invoke lsp formatter

fdeitylink:
apheleia runs `+format-lsp-buffer` in a scratch buffer (where lsp is disabled) so checking for an active lsp mode to invoke the right formatter fails. Change `+format--lsp-fn` to run `+format--lsp-fn` in context of main buffer rather than apheleia's scratch buffer so that the lsp mode can be discovered.

-----


hlissner (approved):

hlissner:

PR #7929, opened 2024-07-09
fix(lib): avoid writing custom faces to custom.el

real-or-random:
custom-theme-set-faces! calls custom-theme-set-faces, which will mark the customizations for saving to custom.el. But custom-theme-set-faces! is not intended for permanent customizations. Fix this by intercepting the relevant put call.

The actual saving is performed by custom-save-all, which will be triggered whenever anything else writes to custom.el, e.g., when lisp code uses customize-save-variable.

----

I've seen this issue in my config, and it has also been reported here:
https://discourse.doomemacs.org/t/how-to-best-change-face-settings-custom-set-faces-or-set-face-attribute/2721/3
And I had fixed a very similar bug when I was still using spacemacs:
https://github.com/syl20bnr/spacemacs/issues/5353#issuecomment-1693211609

-----

### DIFF
--- a/lisp/lib/themes.el
+++ b/lisp/lib/themes.el
@@ -31,9 +31,16 @@ all themes. It will apply to all themes once they are loaded."
            (dolist (theme (ensure-list (or ,theme 'user)))
              (when (or (eq theme 'user)
                        (custom-theme-enabled-p theme))
-               (apply #'custom-theme-set-faces theme
-                      (mapcan #'doom--custom-theme-set-face
-                              (list ,@specs)))))))
+               ;; Skip the 'saved-face(-comment) properties to avoid that the
+               ;; customizations are written to custom.el when saving other
+               ;; customized settings such as variables (e.g., triggered by
+               ;; lisp code calling customize-save-variable).
+               (letf! (defun put (sym prop val)
+                        (unless (memq prop (list 'saved-face 'saved-face-comment))
+                          (funcall put sym prop val)))
+                 (apply #'custom-theme-set-faces theme
+                        (mapcan #'doom--custom-theme-set-face
+                                (list ,@specs))))))))
        ;; Apply the changes immediately if the user is using the default theme
        ;; or the theme has already loaded. This allows you to evaluate these
        ;; macros on the fly and customize your faces iteratively.


hlissner:
I believe [unsetting `custom--inhibit-theme-enable`](https://github.com/doomemacs/doomemacs/blob/master/lisp/lib/themes.el#L30) already does this, so unsetting those properties isn't necessary. I'll need to double check.

hlissner (changes requested):
Oh, it's actually the reverse. It should be non-nil to suppress saving to custom.el. Could you use this implementation instead?

```diff
     `(progn
        (defun ,fn ()
-         (let (custom--inhibit-theme-enable)
+         (let ((custom--inhibit-theme-enable t))
            (dolist (theme (ensure-list (or ,theme 'user)))
              (when (or (eq theme 'user)
                        (custom-theme-enabled-p theme))
                (apply #'custom-theme-set-faces theme
```

hlissner:
I've put together a slightly different implementation in 6877115 which should resolve your issue. Let me know if it gives you any trouble. Thanks for bringing it to my attention, in any case!

real-or-random:
> I've put together a slightly different implementation in [6877115](https://github.com/doomemacs/doomemacs/commit/68771150ba3eb4bdfdebaac9674d31c4526b7495) which should resolve your issue. Let me know if it gives you any trouble. Thanks for bringing it to my attention, in any case!

Thanks for the quick reply! But I've tested https://github.com/doomemacs/doomemacs/commit/68771150ba3eb4bdfdebaac9674d31c4526b7495, and it doesn't do the job for me. 

To reproduce, I have this in my config:
```
(custom-set-faces!
  '((outline-1 outline-2 outline-3 outline-4 outline-5 outline-6 link)
    :weight normal)
  '((org-agenda-done)
    :weight normal))
```

After running `(customize-save-variable 'foo "bar")`, these are written to custom.el:
```
(custom-set-variables
 ;; custom-set-variables was added by Custom.
 ;; If you edit it by hand, you could mess it up, so be careful.
 ;; Your init file should contain only one such instance.
 ;; If there is more than one, they won't work right.
 '(foo "bar" t)
(custom-set-faces
 ;; custom-set-faces was added by Custom.
 ;; If you edit it by hand, you could mess it up, so be careful.
 ;; Your init file should contain only one such instance.
 ;; If there is more than one, they won't work right.
 '(link ((t (:weight normal))))
 '(org-agenda-done ((t (:weight normal))))
 '(outline-1 ((t (:weight normal))))
 '(outline-2 ((t (:weight normal))))
 '(outline-3 ((t (:weight normal))))
 '(outline-4 ((t (:weight normal))))
 '(outline-5 ((t (:weight normal))))
 '(outline-6 ((t (:weight normal)))))
```

And this is consistent with running `(get 'org-agenda-done 'saved-face)`, which still gives me something non-nil.

When I was working on this, I tried to avoid intercepting `put` because it looks a bit harsh, but I think it's okay in the end.

real-or-random:
Regarding `custom--inhibit-theme-enable`:

I think we want this to be `nil` so that the changes are applied immediately. (See our own comment "Apply the changes immediately if the user is using the default theme ...") But this means that `custom-theme-set-faces` will also put the `'saved-face-(comment)` props

hlissner:
0b13525 should put this to rest. I was so focused on `custom-theme-set-faces!`, I forgot about `custom-set-faces!` (or more specifically, when `THEME` == `'user`). Plus, It's 4am. I should *really* try this "sleep" thing.

PR #7933, opened 2024-07-10
fix(lib): avoid writing customized faces to custom.el (take 3)

real-or-random:
Overriding custom--should-apply-setting needs to be done also when we call the constructed hook immediately instead of putting it into doom-customize-theme-hook.

Ref: #7929
Amend: 0b1352525233

---

Sorry ... I hope you could get some sleep :)

### DIFF
--- a/lisp/lib/themes.el
+++ b/lisp/lib/themes.el
@@ -6,11 +6,7 @@
 (add-hook! 'doom-load-theme-hook
   (defun doom-apply-customized-faces-h ()
     "Run `doom-customize-theme-hook'."
-    (letf! ((#'custom--should-apply-setting #'ignore)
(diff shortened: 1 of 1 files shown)


hlissner:
I wanted to keep the hackery in its own function so that hooks added to `doom-customize-theme-hook` don't have to be aware of it (I plan to make more use of it later), so I've done so in afb9773, which should hopefully put this all to rest, for now, at least.

real-or-random:
Ok, I've tested your commit, this one does the job. :)

> I wanted to keep the hackery in its own function so that hooks added to doom-customize-theme-hook don't have to be aware of it (I plan to make more use of it later), 

I see, that makes sense!

PR #7936, opened 2024-07-11
fix(format): allow +format/buffer to format org blocks

lroolle:
Fix: #7685

### DIFF
--- a/modules/editor/format/autoload/format.el
+++ b/modules/editor/format/autoload/format.el
@@ -57,46 +57,18 @@
              (when callback (funcall callback)))
            (kill-buffer formatted-buffer)))))))
 
-(defun +format--org-region (beg end)
-  "Reformat the region within BEG and END.
-If nil, BEG and/or END will default to the boundaries of the src block at point."
-  (when (eq major-mode 'org-mode)
-    (let ((element (org-element-at-point)))
-      (save-excursion
-        (let* ((block-beg (save-excursion
-                            (goto-char (org-babel-where-is-src-block-head element))
-                            (line-beginning-position 2)))
-               (block-end (save-excursion
-                            (goto-char (org-element-property :end element))
-                            (skip-chars-backward " \t\n")
-                            (line-beginning-position)))
-               (beg (if beg (max beg block-beg) block-beg))
-               (end (if end (min end block-end) block-end))
-               (lang (org-element-property :language element))
-               (major-mode (org-src-get-lang-mode lang)))
-          (if (eq major-mode 'org-mode)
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but I cannot accept it. I intentionally kept `+format/org-block` and `+format/buffer` separate because it doesn't make sense to tie it to format-on-save or "formatting the whole buffer" behavior, unless I allow it to reformat *a*ll org blocks in the buffer (which is too opinionated, destructive, and error prone for org files with many blocks in many different languages).

Instead, `+format/org-block` reformats the block at point (which I've bound to `gQ` in `org-mode` through `+org/reformat-at-point`), on demand. Alternatively, you can also exploit format-on-save if you open the src block in a separate buffer with `org-edit-special`. 

Otherwise, if you *really* want `+format/org-block` on save, add it to `before-save-hook` with a condition to prevent it throwing errors when you save a buffer with your cursor not in a src block:

```elisp
;;; in $DOOMDIR/config.el
(add-hook! 'org-mode-hook
  (add-hook! 'before-save-hook :local
    (if (org-in-src-block-p)
        (+format/org-block (point)))))
```

But be warned that `+format/org-block` wasn't designed to be automated like this, and it will only work if your cursor is in a src block when you save the buffer.

lroolle:
Thanks for the detailed explanation!

I understand that handling org-mode specially in `format-buffer` might not make the most sense. The reason I did it that way is because the old format module was set up to use `format-region-or-buffer` bound to (SPC c f). It's super convenient to trigger formatting only when needed.

Also, considering that in `org-mode`, formatting is mostly required only within src blocks, it seemed logical at the time.

If I'd like to continue with the previous logic, is there an easy way to automatically detect and handle org src blocks? Maybe we could handle this within `format-region-or-buffer`?

hlissner:
I'm hesitant to integrate it into `+format/region` or `+format/region-or-buffer`, but for the time being, I've added a `+format/org-blocks-in-region` to do this, as well as modified `+org/reformat-at-point` (bound to `gQ` for evil users) to call `+format/org-block-in-region` if a selection is active.

If you want to override `+format/{buffer,region,region-or-buffer}` in org-mode, you can also do:

```elisp
;;; in $DOOMDIR/config.el
(map! :after org-mode
      :map org-mode-map
      [remap +format/buffer] #'+format/org-block
      [remap +format/region] #'+format/org-blocks-in-region
      [remap +format/region-or-buffer] #'+format/org-block-at-point-or-in-region)
```

I hope it scratches that itch until I revisit this later.

PR #7943, opened 2024-07-12
fix(vertico): orderless filtering

kaiwk:
The original implementation of `+vertico-orderless-dispatch` match pattern by prefix and suffix in pairs. In that case, '=&&' will go for branch `(string-suffix-p "&" pattern)`, not `(string-prefix-p "=" pattern)`, which fail to filter literal '&&'.

We probably should match all prefixes first, then all suffixes. Just like orderless does.

Ref: https://github.com/oantolin/orderless/blob/178b0c55f2cb49f27cd972f731ea45e5d3aea262/orderless.el#L159

-----

### DIFF
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -74,30 +74,23 @@ orderless."
           (completion-styles +vertico-company-completion-styles))
       (apply fn args)))
 
+  (setq orderless-affix-dispatch-alist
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
My bad. Forgot to rebase out 2df6de2 in the last batch of commits. Fixed and merged now. Thanks for your help!

PR #7977, opened 2024-08-01
fix(corfu): do not use corfu's minibuffer setup

LemonBreezes:
This PR prevents Corfu minibuffer completion from being activated always, even in Vertico after a future bump.

Upstream, Corfu has added minibuffer completion by default so this PR disables it since we already have our own hook & code that is working better for our purposes, i.e. our solution is tuned to work in tandem with vertico, helm, etc. 

-----

### DIFF
--- a/modules/completion/corfu/config.el
+++ b/modules/completion/corfu/config.el
@@ -56,6 +56,7 @@ TAB/S-TAB.")
   (setq corfu-auto t
         corfu-auto-delay 0.18
         corfu-auto-prefix 2
+        global-corfu-minibuffer nil
         global-corfu-modes '((not erc-mode
                                   circe-mode
                                   help-mode


krshrimali:
Thanks @LemonBreezes for the PR. Apologies to the reviewers/maintainers for jumping in on this, but just curious - is there an option (/planned/can be considered) to enable the minibuffer setup for corfu's completion instead? I kinda like it and was hoping to keep it instead of the popup.

Can convert this into a discussion instead, but thought this might be a valid point to bring up in this PR as it's slightly related. Please let me know if otherwise.

LemonBreezes:
> Thanks @LemonBreezes for the PR. Apologies to the reviewers/maintainers for jumping in on this, but just curious - is there an option (/planned/can be considered) to enable the minibuffer setup for corfu's completion instead? I kinda like it and was hoping to keep it instead of the popup.
> 
> Can convert this into a discussion instead, but thought this might be a valid point to bring up in this PR as it's slightly related. Please let me know if otherwise.

I don't understand what you are saying. The minibuffer setup is not enabled for Corfu because Corfu does not activate the minibuffer. Corfu is its own UI. It sounds like what you want is `(setq +corfu-want-minibuffer-completion nil)`.

krshrimali:
> @krshrimali I don't understand what you are saying. The minibuffer setup is not enabled for Corfu because Corfu does not activate the minibuffer. Corfu is its own UI. It sounds like what you want is `(setq +corfu-want-minibuffer-completion nil)`.

Sorry if I wasn't clear, @LemonBreezes - I meant just the opposite: `(setq +corfu-want-minibuffer-completion t)`. I'll try summarize my questions again (please let me know if they aren't clear, and we can continue this on a discussion/issue):

- Will there be a way to keep this `corfu-want-mini-buffer-completion` even when `tty` option is enabled?
- Post this PR, will we be able to see mini-buffer setup with doom running in terminal (when `tty` option is disabled)? From what I understand, if `tty` option is disabled, `corfu-terminal` isn't installed and by default a mini-buffer setup is shown for auto-complete.

LemonBreezes:
@krshrimali Regarding this PR, this PR does not change the current behavior whatsoever. This PR is only so that our configuration does not break after bumping Corfu in the future.

Regarding using Corfu in the terminal, I never tested that but I am doing that right now when Corfu tries to popup in the minibuffer while in the terminal, I get this error https://0x0.st/XOpc.txt, and part of the cause is definitely that `corfu--exhibit` in `post-command-hook` is somehow calling `corfu--candidates-popup` with a nil argument. Is this what you are referring to?

krshrimali:
> @krshrimali Regarding this PR, this PR does not change the current behavior whatsoever. This PR is only so that our configuration does not break after bumping Corfu in the future.

Ah, sorry for the noise in that case. I believe we can start the discussion (on what you stated below) in another issue/discussion thread? Please let me know if it makes sense.
 
> Regarding using Corfu in the terminal, I never tested that but I am doing that right now when Corfu tries to popup in the minibuffer while in the terminal, I get this error https://0x0.st/XOpc.txt, and one problem I see there is that `corfu--exhibit` in `post-command-hook` is somehow calling `corfu--candidates-popup` with a nil argument. Is this what you are referring to?

Yes! I was trying to find a way to popup the minibuffer in the terminal for corfu.

Apologies for the delay in my response.

modules/completion/corfu/config.el
```diff
@@ -56,6 +56,7 @@ TAB/S-TAB.")
   (setq corfu-auto t
         corfu-auto-delay 0.18
         corfu-auto-prefix 2
+        global-corfu-minibuffer nil
```
minad, line 59: Btw, I believe you can use `global-corfu-minibuffer` in Doom, just set it to a custom predicate function. But maybe I am missing something.

LemonBreezes: I will check this out in some days. I just had a short surgery yesterday and am recovering.

LemonBreezes: Okay. I have mostly gotten this working but the big issue I'm seeing is that if I add `cape-dabbrev` to CAPFs,
```elisp
(defun +corfu-add-cape-dabbrev-h ()
  (add-hook 'completion-at-point-functions #'cape-dabbrev 20 t))
(add-hook! 'minibuffer-setup-hook :depth 100
           #'+corfu-add-cape-dabbrev-h)
```
it causes completion in the minibuffer to be enabled unconditionally despite having
```elisp
(setq global-corfu-minibuffer
      (defun +corfu-enable-in-minibuffer ()
        "Enable Corfu in the minibuffer."
        (pcase +corfu-want-minibuffer-completion
          ('aggressive
           (not (or (bound-and-true-p mct--active)
                    (where-is-internal #'vertico-exit
                                       (list (current-local-map)))
                    (and (featurep 'auth-source)
                         (eq (current-local-map) read-passwd-map))
                    (and (featurep 'helm-core) (helm--alive-p))
                    (and (featurep 'ido) (ido-active))
                    (where-is-internal #'minibuffer-complete
                                       (list (current-local-map)))
                    (memq #'ivy--queue-exhibit post-command-hook))))
          ('nil nil)
          (_ (where-is-internal #'completion-at-point
                                (list (current-local-map)))))))
```

minad: > I thought that since the predicate is a function, it wouldn't matter whether completion-at-point-functions was a local variable.

Yes. I don't understand why it doesn't work in your case. The condition is this:

~~~ emacs-lisp
(if (functionp global-corfu-minibuffer)
    (funcall global-corfu-minibuffer)
  (local-variable-p 'completion-at-point-functions))
~~~

LemonBreezes: Okay. I will do a deeper investigation using `emacs -Q`.

hlissner:
I've incorporated the `global-corfu-minibuffer` predicate approach in 1ec4bac and can't reproduce any of local `completion-at-point-functions` strangeness you've mentioned. Can you confirm?

LemonBreezes:
> I've incorporated the `global-corfu-minibuffer` predicate approach in [1ec4bac](https://github.com/doomemacs/doomemacs/commit/1ec4bac75e01ddcc896b8c00d8264a3306a9149e) and can't reproduce any of local `completion-at-point-functions` strangeness you've mentioned. Can you confirm?

Yes. In fact, I can reproduce the issue from Vanilla Emacs:
``elisp
;; Bootstrap straight
(defvar bootstrap-version)

(let ((bootstrap-file
       (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
      (bootstrap-version 5))
  (unless (file-exists-p bootstrap-file)
    (with-current-buffer
        (url-retrieve-synchronously "https://raw.githubusercontent.com/raxod502/straight.el/develop/install.el"
                                    'silent 'inhibit-cookies)
      (goto-char (point-max))
      (eval-print-last-sexp)))
  (load bootstrap-file nil 'nomessage))

(straight-use-package 'corfu)
(require 'corfu)

(setq global-corfu-minibuffer
      (defun +corfu-enable-in-minibuffer-p ()
        "Return non-nil if Corfu should be enabled in the minibuffer.
See `+corfu-want-minibuffer-completion'."
        (not (or (bound-and-true-p mct--active)
                 (bound-and-true-p vertico--input)
                 (and (featurep 'auth-source)
                      (eq (current-local-map) read-passwd-map))
                 (and (featurep 'helm-core) (helm--alive-p))
                 (and (featurep 'ido) (ido-active))
                 (where-is-internal 'minibuffer-complete
                                    (list (current-local-map)))
                 (memq #'ivy--queue-exhibit post-command-hook)))))

(straight-use-package 'cape)
(require 'cape)

(add-hook 'minibuffer-setup-hook
          (defun +corfu-add-cape-dabbrev-h ()
            (add-hook 'completion-at-point-functions #'cape-dabbrev 20 t)))

(setq corfu-auto t
      corfu-auto-delay 0.05)
(global-corfu-mode +1)

(straight-use-package 'vertico)
(require 'vertico)
(vertico-mode +1)
```

It seems that the issue happens because of the `+corfu-add-cape-dabbrev-h` @minad  and a workaround is to wrap that hook in a `(when (+corfu-enable-in-minibuffer-p) ...)`.

LemonBreezes:
Just opened a new PR with that workaround.

hlissner:
The issue is that `global-corfu-modes` overrides `global-corfu-minibuffer` in `corfu--on`, which I think should constitute an upstream bug (e.g. I imagine `global-corfu-modes` should dictate which buffers to even try the predicate in). I've pushed a workaround in 84230a4, in the meantime.

minad:
> EDIT: Buuut (not (eq (aref (buffer-name) 0) ?\s)) should prevent this... I'm confused now. https://github.com/doomemacs/doomemacs/commit/84230a437dfca913ba94c39f7a61c41d24a7cfbd still seems to fix it though? I'll take a better look at this when I'm not commuting.

Exactly. I don't understand *why* this happens.

minad:
I cannot reproduce the problem with the following code:

~~~ elisp
(add-hook 'minibuffer-setup-hook
          (defun +corfu-add-cape-dabbrev-h ()
            (add-hook 'completion-at-point-functions #'cape-dabbrev 20 t)))

(setq global-corfu-minibuffer
      (lambda ()
        (not (or (bound-and-true-p mct--active)
                 (bound-and-true-p vertico--input)
                 (eq (current-local-map) read-passwd-map)))))
~~~

Maybe the problem lies in `+corfu-enable-in-minibuffer-p` or the order of execution of the hooks?

PR #7992, opened 2024-08-12
feat(plantuml): support fallback to system jar

leezu:
If plantuml.jar does not exist in doom-data-dir, but system plantuml.jar exists, fallback to system plantuml.jar. Advice plantuml-donwload-jar function to download to doom-data-dir, instead of attempting to overwrite system plantuml.jar.

-----

### DIFF
--- a/modules/lang/plantuml/config.el
+++ b/modules/lang/plantuml/config.el
@@ -5,8 +5,27 @@
   :init
   (setq plantuml-jar-path (concat doom-data-dir "plantuml.jar")
         org-plantuml-jar-path plantuml-jar-path)
+  (let ((doom-jar-path (concat doom-data-dir "plantuml.jar"))
+        (system-jar-path "/usr/share/plantuml/plantuml.jar"))
+    (setq plantuml-jar-path
+          (cond ((file-exists-p doom-jar-path) doom-jar-path)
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but I must turn it down, because of the redundant code, hardcoded path, and an unclear problem. I've made the defaults for these settings a little more robust in d309dca, so there's less likelihood user configuration will be overwritten (and so that the mode works as expected after running `plantuml-download-jar`).

-----

> If plantuml.jar does not exist in doom-data-dir, but system plantuml.jar exists, fallback to system plantuml.jar

Neither `plantuml-mode` nor `ob-plantuml` specify a system path, so Doom isn't overwriting anything.

> Advice plantuml-donwload-jar function to download to doom-data-dir, instead of attempting to overwrite system plantuml.jar.

If users change `plantuml-jar-path` to a non-standard location, then it'd be wise not to *then* run `plantuml-download-jar` (assuming that is what you mean). What's more, `"/usr/share/plantuml/plantuml.jar"` may work for you and your specific distro, but this is not standard, and by policy, Doom should not be writing files anywhere but `$EMACSDIR` or `$DOOMDIR` (and maybe `$TMPDIR` or the `$XDG_*_DIR` directories), by default.

-----

Thank you for bringing it to my attention, in any case!

PR #8005, opened 2024-08-20
fix(format): avoid `+javascript-npm-conf` definition void error

EthanOQuin:
`+javascript-npm-conf` is defined in modules/lang/javascript/autoload.el, so c0a1b9ef broke formatting JSON files when the javascript module isn't enabled in init.el. Gate the call to the function behind a `modulep` check.


hlissner:
I've opted to inline it instead in 67e5dda. Thanks for bringing it to my attention, in any case!

PR #8023, opened 2024-08-23
Add `+hybrid` flag to `:editor evil`? Maybe?

ambirdsall:
This adds a `+hybrid` flag to the `:editor evil` module. The configuration associated with the flag is trivial (just sets `evil-disable-insert-state-bindings`), but ~the real value of the flag is providing a hook for conditionally disabling `evil{, +everywhere}` bindings that clobber important vanilla emacs bindings~ see note below 😅.

Note below: while rebasing my branch to create this PR, I stumbled across the existence of https://github.com/doomemacs/doomemacs/commit/122e3732f70f21ff9bd6f5e263642fd1793e08b1, which dropped one of my (now missing) commits in favor of a shorter, more readable alternative. Now I'm less convinced that this flag is needed—it does significantly improve the discoverability of this configuration, which isn't nothing, but perhaps it's not enough to justify the flag. I have no problem with reducing the PR to a similar patch on `modules/lang/org/config.el`.

Ref: #107 

-----
  Well, it does edit a module readme, but only trivially; that's not the core change.
- [X] This a draft PR; I need more time to ~finish it~ discuss whether it makes sense to merge in its current form.

### DIFF
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -1208,18 +1208,31 @@ between the two."
             :ni [C-return]   #'+org/insert-item-below
             :ni [C-S-return] #'+org/insert-item-above
             ;; navigate table cells (from insert-mode)
-            :i Cright (cmds! (org-at-table-p) #'org-table-next-field
-                             #'org-end-of-line)
-            :i Cleft  (cmds! (org-at-table-p) #'org-table-previous-field
-                             #'org-beginning-of-line)
-            :i Cup    (cmds! (org-at-table-p) #'+org/table-previous-row
-                             #'org-up-element)
-            :i Cdown  (cmds! (org-at-table-p) #'org-table-next-row
-                             #'org-down-element)
-            :ni CSright   #'org-shiftright
-            :ni CSleft    #'org-shiftleft
-            :ni CSup      #'org-shiftup
-            :ni CSdown    #'org-shiftdown
+            (:unless (modulep! :editor evil +hybrid)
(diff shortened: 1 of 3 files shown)


modules/lang/org/config.el
```diff
@@ -1208,18 +1208,31 @@ between the two."
...
-            :ni CSup      #'org-shiftup
-            :ni CSdown    #'org-shiftdown
+            (:unless (modulep! :editor evil +hybrid)
+              :i Cright (cmds! (org-at-table-p) #'org-table-next-field #'org-end-of-line))
+            (:unless (modulep! :editor evil +hybrid)
+              :i Cleft  (cmds! (org-at-table-p) #'org-table-previous-field #'org-beginning-of-line))
+            (:unless (modulep! :editor evil +hybrid)
+              :i Cup    (cmds! (org-at-table-p) #'+org/table-previous-row #'org-up-element))
+            (:unless (modulep! :editor evil +hybrid)
+              :i Cdown  (cmds! (org-at-table-p) #'org-table-next-row #'org-down-element))
```
ambirdsall, line 1218: I think I'd still argue that a `+hybrid` flag provides value by un-hiding a very useful evil feature, but speaking as an `evil-disable-insert-state-bindings` enjoyer, un-clobbering `C-h` and `C-k` when I'm typing stuff in org-mode is by far the most important behavioral change in here.

hlissner:
I've implemented in this in c8d82f3 based on `evil-disable-insert-state-bindings` instead. I'd rather preserve `+hybrid` for when I eventually implment a `evil-hybrid-state`. Thanks for bringing it to my attention, in any case!

PR #8052, opened 2024-09-06
tweak(indent-guides): disable bitmap by default

nightkr:
Bitmap mode is incredibly slow (on my machine, it takes C-d from near-instant to taking about a second), and looks virtually identical.

-----

### DIFF
--- a/modules/ui/indent-guides/README.org
+++ b/modules/ui/indent-guides/README.org
@@ -10,7 +10,8 @@
 /This module has no dedicated maintainers./ [[doom-contrib-maintainer:][Become a maintainer?]]
 
 ** Module flags
-/This module has no flags./
+- +bitmap ::
+  Use bitmaps when using a non-terminal Emacs. Beware that this can cause massive slowdowns for even moderately large files.
 
 ** Packages
 - [[doom-package:highlight-indent-guides]]
--- a/modules/ui/indent-guides/config.el
+++ b/modules/ui/indent-guides/config.el
@@ -16,7 +16,7 @@ non-nil, the mode will not be activated."
 (use-package! highlight-indent-guides
   :hook ((prog-mode text-mode conf-mode) . +indent-guides-init-maybe-h)
   :init
-  (setq highlight-indent-guides-method (if (display-graphic-p) 'bitmap 'character)
+  (setq highlight-indent-guides-method (if (and (display-graphic-p) (modulep! +bitmap)) 'bitmap 'character)
         highlight-indent-guides-bitmap-function #'highlight-indent-guides--bitmap-line)
 
   (defun +indent-guides-init-maybe-h ()


nightkr:
Not sure whether it makes sense add the flag at all, you can always setq it yourself, should you so desire.

hlissner:
In https://github.com/doomemacs/doomemacs/commit/e0a926dc1ec9e1fa66f31751d33b3e4f41b6801a, I opted to simply default to `character`, since the two do indeed look identical and because `highlight-indent-guides` is one of the more expensive packages in Doom. Thanks for bringing it to my attention!

hlissner:
Just a heads up, in 0d9e188b26ca1d01a57598e37171e5333272cad7, I swapped out the `highlight-indent-guides` package with [indent-bars](https://github.com/jdtsmith/indent-bars) which has proven faster in my tests. I've disabled bitmaps on MacOS or on PGTK builds (on Emacs 29 or older), where there are known performance and stability issues, but I've enabled it by default everywhere else (still undecided on Windows). 

I'm not sure what system you're on, but I'd appreciate if you tried it and let me know if it's any better or worse than your experience *before* e0a926d.

nightkr:
I'm on 0461c5de5fa34d6b374f27eff1965f171049f987 and getting similar performance to highlight-indent-guides' `'character` mode, but with far more visible guides (yay!).

For the record, I'm on Arch's `emacs-nativecomp` 29.4-2, which apparently means `indent-bars-prefer-character` is `nil`.

nightkr:
Ookay, looks like `indent-bars` is having an.. unfortunate interaction with lsp-mode's jump previews:

image

hlissner:
How do you repro those jump previews? Should be a simple matter of adding a function to `+indent-guides-inhibit-functions` to detect that context, but I can't tell what exactly I'm looking at from just the screenshot.

nightkr:
M-x `lsp-ui-peek-find-references` (in a project that uses LSP).

hlissner:
Does this fix it for you?

```elisp
(add-hook '+indent-guides-inhibit-functions #'doom-temp-buffer-p)
```

nightkr:
That seems to break lsp-mode entirely:
```
File mode specification error: (wrong-number-of-arguments ((t) (buf) Returns non-nil if BUF is temporary. (char-equal 32 (aref (buffer-name buf) 0))) 0)
Multiple snippets with same identity: "fn"
Note: rustic-indent-offset, tab-width, evil-shift-width adjusted to 4
File mode specification error: (wrong-number-of-arguments ((t) (buf) Returns non-nil if BUF is temporary. (char-equal 32 (aref (buffer-name buf) 0))) 0)
```

hlissner:
Derp, my bad. Here's the revised snippet:

```elisp
(add-hook '+indent-guides-inhibit-functions (lambda () (doom-temp-buffer-p (current-buffer))))
```

nightkr:
Looks identical to me:
image

hlissner:
Odd. That doesn't look like a childframe. Will keep digging.

hlissner:
The issue is messy and really should be handled upstream, but for the time being, I'll just disable indent-guides temporarily when lsp-ui-peek is active. Give this a try:

```elisp
(after! indent-bars
  (defvar-local +indent-guides--in-buffer-p nil)
  (add-hook! 'lsp-ui-peek-mode-hook
    (defun +indent-guides--toggle-when-peeking-h ()
      (if lsp-ui-peek-mode
          (when indent-bars-mode
            (setq-local +indent-guides--in-buffer-p t)
            (indent-bars-mode -1))
        (when +indent-guides--in-buffer-p
          (setq +indent-guides--in-buffer-p nil)
          (indent-bars-mode +1))))))
```

hlissner:
Here's a slightly more sophisticated fix that seems to work better (selectively removing the indent-guides in the peek popup). Give it a try:

```elisp
(after! indent-bars
  (defadvice! +fixed-lsp-ui-peek--peek-new-a (&rest args)
    :after #'lsp-ui-peek--peek-new
    (when indent-bars-mode
      (save-excursion
        (let ((indent-bars--display-function #'ignore)
              (indent-bars--display-blank-lines-function #'ignore))
          (indent-bars--fontify (overlay-start lsp-ui-peek--overlay)
                                (1+ (overlay-end lsp-ui-peek--overlay))
                                nil)))))
  (add-hook! 'lsp-ui-peek-mode-hook
    (defun +indent-guides--toggle-when-peeking-h ()
      (unless lsp-ui-peek-mode
        (indent-bars-setup)))))
```

nightkr:
That seems to work, thanks!

nqminhuit:
Hi @hlissner 
The indent-bars also seems to affect the git blame feature, here is the result when I run "M-x magit-blame-addition"
image

hlissner:
[af814ff](https://github.com/doomemacs/doomemacs/commit/af814ff86a3961d38e1fc741fa840461ded4bd82) and [bcdb945](https://github.com/doomemacs/doomemacs/commit/bcdb945465854c315c2a90e8f2465cf0beddbe53) should fix the issue for both lsp-ui-peek and magit-blame. Thanks for letting me know!

PR #8067, opened 2024-09-13
fix(fold): avoid Hideshow-not-supported error

45mg:
Some modes are not supported by `hs-minor-mode`, but can still support some of the `+fold/` commands. For example, `pdf-outline-minor-mode` recognizes the same commands as `outline-minor-mode`, but `hs-minor-mode` is not applicable. In cases like these, we shouldn't try to enable `hs-minor-mode`, as this will produce an error that will terminate the command.

-----


hlissner (approved):

hlissner:

PR #8070, opened 2024-09-14
fix(emacs-lisp): always try Helpful for doc lookup

45mg:
As per the description in 6671adc68, this module should always use Helpful's functions as long as Helpful is available (ie. not explicitly disabled by the user in packages.el). The remapping of `describe-symbol` is irrelevant here - the user might prefer to rebind `C-h C-o` to `describe-symbol` (as `helpful-symbol` cannot look up types), but that doesn't necessarily mean they want this module not to use it.


hlissner (approved):

hlissner:
Thanks again for your help!

PR #8088, opened 2024-09-21
fix(clojure): type error evilifying cider-debug keys

Prikaz98:
When I call (doom/reload) there is an error like:
```
⛔ Warning (initialization): An error occurred while booting Doom Emacs:

Error in a Doom module: modules/lang/clojure/config.el, (wrong-type-argument consp nil)
```

There is small bug with replacing param which doesn't handle case where param already replaced


hlissner (approved):

hlissner (approved):

hlissner:

PR #8097, opened 2024-09-28
module: add :lang hledger

mrkajetanp:
This PR adds a `hledger` module based on `hledger-mode`, separate from the currently existing `ledger` module.

I first tried to extend the `ledger` module with optional hledger support, but it was effectively becoming two modules in one and they could each exist independently so it seems like separating them is a cleaner solution.

Despite being very similar they're still different tools with different behaviours, even the website for hledger only describes it as "comparable" to ledger rather than being a different implementation of the same tool.

-----

### DIFF
--- a/modules/lang/hledger/README.org
+++ b/modules/lang/hledger/README.org
@@ -1,0 +1,54 @@
+#+title:     :lang hledger
+#+subtitle:  ledger, but functional
+#+created:   September 28, 2024
+#+since:     3.0.0 (#COMMIT-OR-PR-REF)
+
+* Description :unfold:
+This module adds support for [[https://hledger.org/][hledger]] files. hledger is a command line double-entry
+accounting system that works with simple text files similar to [[https://ledger-cli.org/][ledger]].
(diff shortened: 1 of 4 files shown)


hlissner:
As of v2.2, modules have been moved out of core, so module development has moved to [doomemacs/modules](https://github.com/doomemacs/modules) and [doomemacs/modules-contrib](https://github.com/doomemacs/modules-contrib). All modules PRs are being closed as such. If you're still interested, the modules repo is ready to receive this PR, given these changes: 

- The commit message be changed to `module: add :lang hledger` (to match our git conventions),
- The `set-company-backend!` be moved after `:config`.
- Add `flycheck-hledger` and `flymake-hledger`, based on whether `(modulep! :checkers syntax +flymake)` is enabled or not.
- Change the `(modulep! :editor evil)` check in packages.el to `(modulep! :editor evil +everywhere)`.

If you aren't interested, that's fine too. Sorry for the trouble and thank you for your interest!

PR #8114, opened 2024-10-12
fix(doom-dashboard): right text overflow in modeline

unipro:
This commit fixes a bug where the right string 'DOOM v3.0.0-pre  🕝 6:24PM' in *doom* buffer's mode line gets cut off. The previous screenshot is as follows,
before-screenshot
and the screenshot after this commit is below.
after-screenshot

-----


hlissner:
Rather than set the variable globally, I'd rather it only be set buffer-locally. I've done so in 4f8a8e0, which should resolve your issue. Thanks for bringing it to my attention, in any case!

PR #8125, opened 2024-10-26
fix(eval): remove duplicate REPL entries

floyza:
Using `+eval/open-repl-other-window` in a window that isn't associated with a given repl (for example, the starting screen), I would get duplicate entries in the prompt, because cl-delete-duplicates compares with `eql` by default.

Another issue that I see that's similar is that the different python repls (`+python/open-jupyter-repl`, `+python/open-repl`) are listed under the same name in the prompt. This pr doesn't fix that, however.


hlissner (approved):

hlissner:
LGTM. Thanks for your help!

PR #8127, opened 2024-10-28
feat(lib): factor out window handling hook fn into variable

dradetsky:
If I disable the `popup` module and run `doom/reload` the default hook (presumably designed to work with popup) will close the window it ran in. As a result, I have to fix my window layout after running a reload.

To solve this we:

1. Split the code that does this out of the one-big-hook that gets called after `doom/reload`
2. Put that code into a variable

Then I can get emacs to do what I want by adding the following to my user config

```
(defun dmr:handle-compile-buffer (_buf status)
  (when (equal status "finished\n")
    (kill-buffer _buf)))

(setq doom-handle-compile-buffer-fn #'dmr:handle-compile-buffer)
```

-----
- [ ] My changes are not on [the do-not-PR list](https://doomemacs.org/donotpr) for this project.


dradetsky:
@hlissner I just thought of an easier hack for myself to solve the underlying problem, making this less urgent. I don't really like it, it's obviously suboptimal in many ways, but it solves the problem, so there you go. What I've done is add the following to my config:

```
(remove-hook 'doom-init-ui-hook #'+popup-mode)

(defun dmr/doom/reload ()
  (interactive)
  (+popup-mode +1)
  (doom/reload)
  (+popup-mode -1))

(map! :leader
      (:prefix "h"
               (:prefix "r"
                :desc "Reload user config" "r" #'dmr/doom/reload)))
```

This way, on reload we turn on popup mode just long enough to not have the issue.

PR #7806, opened 2024-04-10
Switch to using opam-switch-mode for :lang ocaml

punchagan:
The PR adds configuration to install `opam-switch-mode` to manage OCaml's opam switches, instead of using a deprecated method in the `tuareg` package. 

-----
- [x] I searched the issue tracker and this hasn't been PRed before.
- [x] My changes are not on [the do-not-PR list](https://doomemacs.org/d/do-not-pr) for this project.
- [x] My commits conform to [the git conventions](https://doomemacs.org/d/git-conventions).

I'm not sure if the DNP on modules is still in place, since there's been no activity on the modules repository since the last comment on the DNP list.

PR #8132, opened 2024-10-31
Add opam-switch-mode to  lang: coq and fix configuration of opam-switch-mode in lang: ocaml

jeremyThibault:
This PR adds configuration to install and use `opam-switch-mode` to manage `opam` switches to the `coq` module. The configuration is taken almost straight from PR #7806 which added the same capability to the `ocaml` module. The only difference is that, unlike in the `ocaml` module, we do not initialize the environment right away.
Incidentally, we need to wrap this initialization inside the `ocaml` module, or it will try to execute even if `tuareg` isn't loaded and produce an error.

Note that this is my first PR, and that I am not very familiar with Emacs and `use-package`, so I'm not very confident that commit bdb4d6064650a5435c5f153b9180aac82b7c5db5 is doing the right thing; at least, it seems to work on my machine.
I'm also not sure if it's fine to use the same `package!` in two different, non-mutually-exclusive modules.

-----

### DIFF
--- a/modules/lang/coq/config.el
+++ b/modules/lang/coq/config.el
@@ -93,3 +93,17 @@
           "e" #'company-coq-document-error
           "E" #'company-coq-browse-error-messages
           "h" #'company-coq-doc)))
+
+(use-package! opam-switch-mode
+  :hook (coq-mode . opam-switch-mode)
+  :preface
+  (map! :after coq
+        :localleader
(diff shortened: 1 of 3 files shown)


bamarsha:
@hlissner Why was this marked as draft? What can be done to help get this ready to merge?

PR #8133, opened 2024-10-31
feat(file-templates): add apply method

dradetsky:
Factor template application code into command file-templates-apply. This allows it to be invoked by the user.

Personally, I find file templates very annoying, and 9/10 times I would end up just deleting the template boilerplate, so I disabled it. However, in some cases I might want to use it. Rather than having to reconfigure whether templates are enabled and re-create the file, we simply pull the couple of lines where the template is actually applied out into a function and add `(interactive)`. Now I can just `M-x file-templates-apply` in the buffer where I want to use the templates.

-----

### DIFF
--- a/modules/editor/file-templates/config.el
+++ b/modules/editor/file-templates/config.el
@@ -148,10 +148,10 @@ must be non-read-only, empty, and there must be a rule in
        (null (buffer-base-buffer))  ; ...isn't an indirect clone
        (file-templates-apply)))
 
-(defun file-templates-apply ()
+(defun +file-templates/apply ()
   "Actually expand a file template if one exists"
   (interactive)
-  (when-let (rule (cl-find-if #'+file-template-p +file-templates-alist))
+  (when-let* ((rule (cl-find-if #'+file-template-p +file-templates-alist)))
     (apply #'+file-templates--expand rule)))
 
 ;;
--- a/modules/editor/file-templates/config.el
+++ b/modules/editor/file-templates/config.el
@@ -146,7 +146,7 @@ must be non-read-only, empty, and there must be a rule in
        (not (file-exists-p buffer-file-name))  ; ...is a new file
        (not (buffer-modified-p))    ; ...hasn't been modified
        (null (buffer-base-buffer))  ; ...isn't an indirect clone
-       (file-templates-apply)))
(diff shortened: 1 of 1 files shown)


modules/editor/file-templates/config.el
```diff
@@ -146,9 +146,13 @@ must be non-read-only, empty, and there must be a rule in
        (not (file-exists-p buffer-file-name))  ; ...is a new file
        (not (buffer-modified-p))    ; ...hasn't been modified
        (null (buffer-base-buffer))  ; ...isn't an indirect clone
-       (when-let (rule (cl-find-if #'+file-template-p +file-templates-alist))
-         (apply #'+file-templates--expand rule))))
+       (file-templates-apply)))
 
+(defun file-templates-apply ()
```
hlissner, line 151: To conform with our naming conventions, please rename this to `+file-templates/apply`.

modules/editor/file-templates/config.el
```diff
@@ -146,9 +146,13 @@ must be non-read-only, empty, and there must be a rule in
...
+  "Actually expand a file template if one exists"
+  (interactive)
+  (when-let (rule (cl-find-if #'+file-template-p +file-templates-alist))
```
hlissner, line 154: `when-let` is deprecated in Emacs 31; replace this with `when-let*`. I.e.

```diff
   (interactive)
-  (when-let (rule (cl-find-if #'+file-template-p +file-templates-alist))
+  (when-let* ((rule (cl-find-if #'+file-template-p +file-templates-alist)))
    (apply #'+file-templates--expand rule)))

hlissner (changes requested):
Side-note: file templates are also simply yasnippet snippets with a name that starts with `__`, so they can expanded like any other snippet: e.g. `__` + `TAB`. This was the intended way folks would expand them, and why a command wasn't initially available.

image

modules/editor/file-templates/config.el
dradetsky: Ironically, I noticed this & wanted to change it, but I was worried that you had too much to deal with already & didn't want to add cognitive load to what was a simple PR.

hlissner:
LGTM! Thanks for your help!

PR #8151, opened 2024-11-06
fix: no default to hl-line in describe-face

dradetsky:
One issue that always bugged me as a new user of doom and other distros, especially when I had less elisp skills, was that I could never use `describe-face` to get useful info about faces (in order to figure anything out) since `hl-line-face` was always in the way. I never wanted info about `hl-line-face`. So just disable the mode while running the proc that gets the default prompt value for `describe-face`.

-----

### DIFF
--- a/lisp/doom-ui.el
+++ b/lisp/doom-ui.el
@@ -398,7 +398,16 @@ windows, switch to `doom-fallback-buffer'. Otherwise, delegate to original
   (add-hook! 'evil-visual-state-exit-hook
     (defun doom-enable-hl-line-maybe-h ()
       (when doom--hl-line-mode
-        (hl-line-mode +1)))))
+        (hl-line-mode +1))))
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but I cannot accept it. Advising such a commonly used, low-level function has too broad an effect and too high a chance of introducing edge cases.

I use `M-x describe-char` instead, and recommend you do, as well. That, or [this custom variant of `describe-char` I intend to add to Doom later](https://gist.github.com/hlissner/65de1f8d726f206a9ac0ed7958bd9426) that runs the command wherever you click the mouse. `describe-char` lists all text properties and overlays at point, plus the faces associated with them.

Otherwise, I'll leave it to users to deploy such an advice in their own configs.

dradetsky:
@hlissner fair

PR #8180, opened 2024-11-26
tweak: consistent order for profile init functions

marienz:
Doom embeds the partials generated by functions on `doom-profile-generators` in lexicographical order by filename, but after recent refactoring all these partials generally do is define a function which Doom calls from `doom-startup`. When Doom generates `doom-startup`, it inserts these calls in `doom-profile-generators` order.

This does not affect Doom itself, since it initializes `doom-profile-generators` in a sensible order. But any additional entries pushed onto `doom-profile-generators` run first, which is surprising (and might break them if they depend on loaddefs or modules having initialized).

Fix it by sorting the list. Do so up front just for simplicity: although the order in which we generate the partials is unlikely to matter, if we have to sort the list we might as well use the sorted one throughout.

Amend: b3aa41fd7424

-----

(I'm guessing barely anyone outside Doom uses `doom-profile-generators`, but figured I'd send this your way just in case, as I doubt this behavior was intended and it seemed easy enough to fix...)

### DIFF
--- a/lisp/lib/profiles.el
+++ b/lisp/lib/profiles.el
@@ -242,7 +242,8 @@ caches them in `doom--profiles'. If RELOAD? is non-nil, refresh the cache."
   (doom-initialize-packages)
   (let* ((default-directory doom-profile-dir)
          (init-dir  doom-profile-init-dir-name)
-         (init-file (doom-profile-init-file doom-profile)))
+         (init-file (doom-profile-init-file doom-profile))
+         (generators (seq-sort-by #'car #'string< doom-profile-generators)))
     (print! (start "(Re)building profile in %s/...") (path default-directory))
     (condition-case-unless-debug e
       (with-file-modes #o750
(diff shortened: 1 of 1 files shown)


hlissner:
This was indirectly fixed in https://github.com/doomemacs/core/commit/dd7903fdcb1ed770faa4ca37bf9adfa9b41f7c42.

A heads up, this will break [this `seq-sort` call](https://github.com/marienz/nix-doom-emacs-unstraightened/blob/main/build-helpers/build-profile#L94-L95) in your build-profile script. `doom-profile-generators` is now a flat list of function names (though, for backwards compatibility it will still accept the old `'("FILE" GENERATOR-FUNCTION STARTUP-FUNCTION)`).

Then comes init function call order. You can either adjust the sorter function *or* adapt to the new format, which allows precise control over the order they're a) inserted into the profile init file (using the numeric prefix in the filename) and b) the order that they're called at startup (with the new `doom-startup-functions` hook). E.g.

```elisp
(defun generate-nix-autoloads (_profile)
  "Pull in autoloads for non-ELPA emacsWithPackages dependencies."
  (doom-file-write
   "80-loaddefs-nix.init.el"
   `((defun unstraightened--startup-nix-autoloads (_profile)
       [...])
     (add-hook 'doom-startup-functions #'unstraightened--startup-nix-autoloads 80))))
(add-to-list 'doom-profile-generate-functions #'generate-nix-autoloads)

(defun generate-unstraightened-autoloads (_profile)
  "Like doom-profile--generate-package-autoloads but for package.el."
  (doom-file-write
   "90-loaddefs-unstraightened.init.el"
   `((defun unstraightened--startup-unstraightened-autoloads (_profile)
       [...])
     (add-hook 'doom-startup-functions #'unstraightened--startup-unstraightened-autoloads 90))))
(add-to-list 'doom-profile-generate-functions #'generate-unstraightened-autoloads)
```

(Function bodies truncated for brevity's sake, but they are unchanged)

PR #8197, opened 2024-12-09
fix(fold): make fold functions consistent

SouthFox-D:
The +fold/{open,close}-all functions prioritized the ts-fold package, while +fold/{toggle,open,close} prioritized the hideshow package.

This will break +fold/{toggle,open,close} for using +fold/close-all buffer when tree-sitter is enabled.

-----

### DIFF
--- a/modules/editor/fold/autoload/fold.el
+++ b/modules/editor/fold/autoload/fold.el
@@ -142,13 +142,13 @@ Return non-nil if successful in doing so."
 Targets `vimmish-fold', `hideshow', `ts-fold' and `outline' folds."
   (interactive)
   (save-excursion
-    (cond ((+fold--vimish-fold-p) (vimish-fold-toggle))
+    (cond ((+fold--ts-fold-p) (ts-fold-toggle))
+          ((+fold--vimish-fold-p) (vimish-fold-toggle))
           ((+fold--outline-fold-p)
            (cl-letf (((symbol-function #'outline-hide-subtree)
                       (symbol-function #'outline-hide-entry)))
              (outline-toggle-children)))
-          ((+fold--hideshow-fold-p) (+fold-from-eol (hs-toggle-hiding)))
-          ((+fold--ts-fold-p) (ts-fold-toggle)))))
+          ((+fold--hideshow-fold-p) (+fold-from-eol (hs-toggle-hiding))))))
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


KendrickLamarck:
Nice, I just saw this as I wanted to create my own PR with the same fix. One question:

The order in `+fold/{open,close}-all` is ts-fold -> vimish-fold -> hideshow -> outline. Should the order in `+fold/{toggle,open,close}` be the same? You have outline and hideshow swapped right now. I don't know if it makes a difference.

I found a few other issues with fold as well:

- `+fold/open-rec` acts like `+fold/open-all` because the `(beg end)` arguments are unused in some places. I can make my own PR for this.
- `+fold/next` doesn't work, but `+fold/previous` does. Still need to figure this out.

SouthFox-D:
> The order in `+fold/{open,close}-all` is ts-fold -> vimish-fold -> hideshow -> outline. Should the order in `+fold/{toggle,open,close}` be the same? You have outline and hideshow swapped right now. I don't know if it makes a difference.

Updated. Thanks for the reminder.

>     * `+fold/next` doesn't work, but `+fold/previous` does. Still need to figure this out.

I think the reason for this is that the  `(overlays-in (point) (point-max))` for +fold/next always includes the current point's fold, thus causing it to get stuck in place.

A quick fix might be:
```diff
@@ -275,7 +275,8 @@ region."
                                        (eq (overlay-get ov 'creator) 'ts-fold))
                                      ;; `overlays-in' does not provide a list that is sorted
                                      ;; (in the way we need it atleast) so we need to sort it based on direction
-                                     (cl-sort (apply #'overlays-in arg-list) comp-fun :key #'overlay-start))))
+                                     (cl-sort (apply #'overlays-in arg-list) comp-fun :key #'overlay-start)))
+                               (ovs (if (> count 0) (cdr ovs) ovs)))
                           (if (and ovs (<= (abs count) (length ovs)))
                               (goto-char (overlay-start (nth (- (abs count) 1) ovs))))))))
            if (save-excursion (funcall fn))
```

hlissner:
ec6a839 should fix the folding package priority issue (and for the new treesit-fold, instead of the old ts-fold) and 7e745e624dab698ef70a83f6f7ec821b8bfdc60b fixed the next/previous commands. Thanks for bringing it to my attention!

PR #8189, opened 2024-12-01
fix: project-current-directory-override takes a directory

dwb:
`project-current-directory-override` takes a string, not a `t`. [It overrides `default-directory`](https://github.com/emacs-mirror/emacs/blob/6c1c3120b98652de149ee9d8c241cd3636755171/lisp/progmodes/project.el#L196-L198) which is always a string.

This was causing a crash for me when using `doom/find-file-in-private-config` from a buffer outside a project:

```
  Debugger entered--Lisp error: (wrong-type-argument stringp t)
    file-remote-p(t)
    projectile-project-root(t)
    project-projectile(t)
    run-hook-with-args-until-success(project-projectile t)
    project--find-in-directory(t)
  [...lots, available on request, but it doesn't seem relevant...]
    doom-project-find-file("/Users/user/.doom.d/")
    doom/find-file-in-private-config()
    funcall-interactively(doom/find-file-in-private-config)
    command-execute(doom/find-file-in-private-config)
```

-----
- [x] I searched the issue tracker and this hasn't been PRed before.
- [x] My changes are not on [the do-not-PR list](https://doomemacs.org/donotpr) for this project.
- [x] My commits conform to [Doom's git conventions](https://doomemacs.org/d/git-conventions).

PR #8200, opened 2024-12-10
fix(lib): pass correct argument `project-current-directory-override`

prashantvithani:
`project-current-directory-override` is `nil` by default and expect the value of directory name i.e. string when non-nil. The value set to this variable is to detect directory which fails with:
`(wrong-argument-type stringp t)`

### DIFF
--- a/lisp/lib/projects.el
+++ b/lisp/lib/projects.el
@@ -160,7 +160,7 @@ If DIR is not a project, it will be indexed (but not cached)."
           ((and (bound-and-true-p helm-mode)
                 (fboundp 'helm-find-files))
            (call-interactively #'helm-find-files))
-          ((when-let* ((project-current-directory-override t)
+          ((when-let* ((project-current-directory-override nil)
                        (pr (project-current t dir)))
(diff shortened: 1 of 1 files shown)


lisp/lib/projects.el
```diff
@@ -160,7 +160,7 @@ If DIR is not a project, it will be indexed (but not cached)."
           ((and (bound-and-true-p helm-mode)
                 (fboundp 'helm-find-files))
            (call-interactively #'helm-find-files))
-          ((when-let* ((project-current-directory-override t)
+          ((when-let* ((project-current-directory-override nil)
```
prashantvithani, line 163: I set this to `nil` (default value) for the lack of better value. `project-current t dir` would prompt only if this value is set to `nil` anyway.

hlissner:
Sorry for the late reply. This appears to have been resolved in https://github.com/doomemacs/doomemacs/commit/1fc41d719c19973c8f45fc93c8e5ffcfa4ff2477 (#8189). Thanks for your help, in any case!

PR #8201, opened 2024-12-11
fix(notmuch): afew: process only new mail

45mg:
The `-a` option will cause afew to process all the user's mail. If the user has a lot of mail, this will take an extremely long time, every time. The expected usage is to process only mail that has been tagged 'new' by notmuch (which requires the [correct
setup](https://github.com/afewmail/afew/blob/master/docs/quickstart.rst)).

-----

### DIFF
--- a/modules/email/notmuch/autoload.el
+++ b/modules/email/notmuch/autoload.el
@@ -36,7 +36,7 @@
 
 (defun +notmuch-get-sync-command ()
   "Return a shell command string to synchronize your notmuch mail with."
-  (let* ((afew-cmd "afew -a -t")
+  (let* ((afew-cmd "afew -n -t")
          (sync-cmd
           (pcase +notmuch-sync-backend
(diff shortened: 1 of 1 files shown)


hlissner (approved):

45mg:
> We should probably deprecate this flag while we’re here...

I agree, all the flag does anyway is append `&& afew -n -t` to the return value of `+notmuch-get-sync-command`.

But maybe we'd want to give users some warning about the deprecation? I don't think Doom even shows a warning about unimplemented flags, so for people not following the commit log (is it expected of everyone to do so?) their afew filters would just silently stop applying, and a bunch of mail might end up wrongly tagged before they realize (assuming they never bothered to do the `post-new` hook thing, like me, since it already worked without that).

And a deprecation warning feels like a bit more of a long-term solution, so I guess we should at least have the correct behavior for now. Does that make sense?

PR #8208, opened 2024-12-19
fix(fold): fix recursive unfolding

KendrickLamarck:
`point-min` and `point-max` were used in some places in `+fold--open-rec-between` instead of `beg` and `end`. This caused `+fold/open-rec` to open all folds rather than only the one the curser is on.

-----


hlissner (approved):

hlissner:

PR #8212, opened 2024-12-25
fix(org): allow tab to expand yasnippets in org

LemonBreezes:
I do not see any evidence of a [tab] binding in `outline-mode-cycle-map` in Emacs29, Emacs30, nor Emacs31. Moreover, this keybinding is overriding our TAB keybinding in GUI Emacs, so YaSnippets are not being expanded.

This references "yasnippet in org-mode. Fix TAB for jumping to $1 and $2" from the Discord server.

Also references
conversations spread around other issues & PRs.

### DIFF
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -839,11 +839,6 @@ between the two."
             #'+org-delete-backward-char-and-realign-table-maybe-h)
 
   (map! :map org-mode-map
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
I suspect this keybind bubbled down from evil-collection or evil-org at some point, but I don't see it anymore either, so I'll accept the change. Thanks again for your help!

PR #8234, opened 2025-01-09
fix: `auto-mode-alist`: make `/*rc` rule low-priority

AjaiKN:
When `nconc` was changed to `add-to-list`, that made this fallback rule be added to the beginning of `auto-mode-alist` instead of the end.


hlissner (approved):

hlissner:

PR #8252, opened 2025-01-23
fix: diff-hl revert not jumping back when aborted

JasZhe:
This has been a pet peeve of mine for a while and I finally decided enough was enough.

`diff-hl-revert-hunk-1` [will signal a user-error](https://github.com/dgutov/diff-hl/blame/65a5de16e21c87b7c12a78a63fc3b57e07c03c86/diff-hl.el#L656) if the user decides to abort the revert hunk operation which will prevent the goto-char function from executing and jumping us back to where we were previously. 

Wrapping the `apply` with `ignore-error 'user-error` sort of works, but not completely since hitting C-g or `abort-recursive-edit` wouldn't signal the `user-error` but it would still move the cursor in an odd way.

I thought about replacing it with a `save-excursion` since that seems like the intent of the original method anyways, but I'm not sure if there are other ramifications with that. In the end I think going with an `unwind-protect` felt somewhat safer and more inline with the existing solution.

Fix: #0000
Ref: #0000
Close: #0000

EDIT: I tested this with a before/after by running with an empty doom config and then manually applying the popup rule change. 
```
doom run --doomdir nodoomconfig/
```

https://github.com/user-attachments/assets/7854aeeb-a020-4513-bcf7-4b70e9afae30

-----

### DIFF
--- a/modules/ui/vc-gutter/config.el
+++ b/modules/ui/vc-gutter/config.el
@@ -93,7 +93,7 @@ Respects `diff-hl-disable-on-remote'."
                 (put 'diff-hl-mode 'last graphic?))))))))
 
   :config
-  (set-popup-rule! "^\\*diff-hl" :select nil :size '+popup-shrink-to-fit)
+  (set-popup-rule! "^\\*diff-hl" :select nil)
 
   (setq diff-hl-global-modes '(not image-mode pdf-view-mode))
   ;; PERF: A slightly faster algorithm for diffing.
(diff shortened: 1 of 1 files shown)


JasZhe:
I played around with this a bit more, and at least for my config (and I may need to try and find some time to reproduce this in a "vanilla" doom config state) it actually might be cause of the popup rule we set for diff-hl buffers. 

What worked better for me was actually 
```
(after! diff-hl (add-to-list 'display-buffer-alist '("^\\*diff-hl" (display-buffer-in-side-window) (side . bottom))))
```

hlissner:
After some experimentation, removing the `:size` rule should be sufficient (and resizing the popup was unnecessary -- and jarring -- to begin with), so I've cherry picked that change in bf528be. With that, the issue should be resolved, but let me know if that isn't the case.

Thanks for your help, in any case!

PR #8255, opened 2025-01-25
fix: use dumb-jump xref instead of dumb-jump-go

JasZhe:
Without this dumb-jump uses that older deprecated style of completion either via the popup library or completing read, but it didn't conform as nicely with the all the xref goodies like line numbers, file separators, consult previewing etc..

This (albeit a little hacky) way of temporarily setting the xref-backend-functions to be dumb jump after lsp-mode completion fails, works quite nicely. 

Fix: #0000
Ref: #0000
Close: #0000

# Screenshots
### With 'dumb-jump-selector as 'popup (default)
Screenshot 2025-01-25 at 1 21 35 AM

### as 'completing-read
image

## After:
image

-----
- [x] Any relevant issues or PRs have been linked to. The ones I found relating to dumb jump were closed already.

### DIFF
--- a/modules/tools/lookup/autoload/lookup.el
+++ b/modules/tools/lookup/autoload/lookup.el
@@ -223,14 +223,18 @@ This can be passed nil as its second argument to unset handlers for MODES. e.g.
       (+lookup--xref-show 'xref-backend-references identifier #'xref--show-xrefs)
     (cl-no-applicable-method nil)))
 
-(defun +lookup-dumb-jump-backend-fn (_identifier)
+(defun +lookup-dumb-jump-backend-fn (identifier)
   "Look up the symbol at point (or selection) with `dumb-jump', which conducts a
 project search with ag, rg, pt, or git-grep, combined with extra heuristics to
 reduce false positives.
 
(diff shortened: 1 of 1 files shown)


JasZhe:
I think the initial reason, and someone can correct me if I'm wrong, why we had this `+lookup-dumb-jump-backend-fn` in the first place, is because according to the doc for `xref-backend-functions`

```
Special hook to find the xref backend for the current context.
Each function on this hook is called in turn with no arguments,
and should return either nil to mean that it is not applicable,
or an xref backend, which is a value to be used to dispatch the
generic functions.
```

The problem, is that lsp-mode's `lsp--xref-backend` doesn't [really do the check at all](https://github.com/emacs-lsp/lsp-mode/blob/d28dd6b7e39506c11e0cb86de69b04a0f1963f9a/lsp-mode.el#L6474). (See [this comment](https://github.com/jacktasia/dumb-jump/issues/353#issuecomment-660623242) in the dumb jump repo as well)

hlissner (approved):

hlissner:
LGTM. Thanks for your help!

PR #8266, opened 2025-02-08
fix(emacs-lisp): handle malformed outline-regexp

LemonBreezes:
Sometimes elisp files specify a local outline-regexp which does not match any groups so with this PR, outline-level returns a default value of 0 instead of an error. As an example, take `gptel-rewrite.el`. 

This is part of my goal of reducing the amount of errors I see while using Emacs to as close to 0 as possible. I would rather the highlight be incorrect than see an error, because errors distract us from working on things that are more important to us.


hlissner (approved):

hlissner:

PR #8271, opened 2025-02-10
fix(packages): use git://git.sv.gnu.org

skryvel:
There are a few threads about related issues. For me, consistently the Git endpoint has been working, while the HTTPS one is often down. As far as I know this should be a transparent change.


nightkr:
git:// [is unencrypted](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols#_the_git_protocol). You're basically handing full code execution privileges to any MitM.

skryvel:
Oh, that's a good point :( I'll close this as I think there's another one about using a mirror instead.

PR #8298, opened 2025-03-01
Update +evil-bindings.el void variable fix

wyrtwala:

Fix Line # 356/357

-----
- [ x] My changes are not on [the do-not-PR list](https://doomemacs.org/donotpr) for this project.
- [ x] My commits conform to [Doom's git conventions](https://doomemacs.org/d/git-conventions).

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -353,8 +353,8 @@
       :desc "Org Capture"           "X"    #'org-capture
       ;; C-u is used by evil
       :desc "Universal argument"    "u"    #'universal-argument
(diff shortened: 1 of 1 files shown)


hlissner:
These are not typos. If you are seeing void-variable errors for either of these symbols, something else is wrong with your config (maybe Evil isn't correctly installed or you have an unrelated error in your private config somewhere -- errors can short-circuit other parts of Emacs' startup process). `help-map` is built into Emacs and `evil-window-map` comes with evil, which both should be available at startup. And if you've disabled the evil module, then the `+evil-bindings.el` file shouldn't be loaded at all.

PR #8299, opened 2025-03-03
perf: do not load pp at startup

LemonBreezes:
Hi. I'm thinking we can avoid using `pp` in `setq-hook!` and to avoid loading `pp` and use the faster `prin1-to-string` since it's rare that people look at the form `setq-hook!` outputs anyways.


hlissner (approved):

hlissner:

PR #8301, opened 2025-03-05
Update emacs-plus installation instructions

verhovsky:
Works on #7686


wchmb:
This is indeed the advice you see after installing Emacs-plus, but I don't see any benefit over using the well-known `ln` UNIX tool... moreover, Spotlight does not index symlinks :_( that's why I tend to just copy-paste binary in /Applications

hlissner:

Issue #8269, opened 2025-02-10
`C-i` triggers `indent-for-tab-command` instead of `better-jumper-jump-forward` when running Emacs as daemon


claytharrison:
### I confirm that...

- [x] I have searched the [issue tracker](https://github.com/doomemacs/doomemacs/issues), [documentation](https://docs.doomemacs.org), [FAQ](https://docs.doomemacs.org/-/faq), [Discourse](https://discourse.doomemacs.org), and [Google](https://google.com), in case this issue has already been reported/resolved.

- [x] I have read "[How to Debug Issues](https://doomemacs.org/d/how2debug)", and will use it to provide as much information about this issue as possible.

- [x] The issue can be reproduced on the **latest** available commit of Doom.

- [x] The issue can be reproduced on a stable release of Emacs, such as 27, 28, or 29. *(Unstable versions end in .50, .60, or .9x)*

### Expected behavior

When I press `C-i` (in a programming mode?) when running Emacs as a daemon, it should jump me forward to the next location in the jump list (using `better-jumper-jump-forward`  as it does in non-daemon Emacs on the base Doom config.

### Current behavior

When I use `C-i` to try and jump forward, instead it runs `indent-for-tab-command` - but only in daemon Emacs. Regular Emacs works fine.

`helpful-key` confirms this, but what's interesting is that the keybinding for `better-jumper-jump-forward` is still listed as `<C-i>`, and the (only) listed keybinding for `indent-for-tab-command` is `TAB` - just as it is in regular Emacs.

### Steps to reproduce

1. Open a fresh Doom Emacs install (latest master commit, Emacs 29.4) with default recommended settings (so evil-mode on)
2. Navigate around a bit.
3. Use `C-o` to jump backwards to your last spot and notice that it works.
4. Use `C-i` to jump forwards to the spot you were just at and notice that it works.
5. Kill Emacs and open it as a daemon-client instead (`emacs --daemon; emacsclient -c`)
6. Repeat steps 2-4 but notice that now when you press `C-i`, it instead runs `indent-for-tab-command` rather than `better-jumper-jump-forward`.

This did work for me before but it broke some months ago and I haven't had the time to properly put together a minimal reproducible test case until now.

### System Information

https://pastebin.com/raw/FjPimVkU

exzombie:
Might be related to #7997, although it has been reported in non-daemon mode as well, so it's not clear to me whether it's the same thing.

WhoisDonlee:
[...]

PR #8303, opened 2025-03-05
fix: RET and TAB keybindings for emacs daemon

WhoisDonlee:
Before this commit, the hack that distinguishes between `TAB`/`C-i` and `RET`/`C-m` didn't properly execute when running Emacs as a daemon. This was because `display-graphic-p` returns `nil` during daemon initialization when no frames exist yet, causing `input-decode-map` to lack the necessary entries.

This PR wraps the relevant code in a function and adds conditional logic:
- When `daemonp` is `t`, it adds the function to `server-after-make-frame-hook` so it runs after a frame is created
- Otherwise, it executes the code immediately as before

Fix: #8269
Ref: #7997, #8101
Close: #8269

-----


hlissner:
I've merged this into 0e57680 along with some adjustments to the commit message, a comment to explain the reasoning, and a hook name more in line with our conventions. Thanks for bringing it to my attention, in any case!

PR #8316, opened 2025-03-19
feat(evil): add evil-textobj-line

ryuheechul:
Which enables combinations such as `dil`, `dal`, `vil`, `val`, etc.

Fix: none
Ref: none
Close: none

-----
  - (Although it's not a visual change, there is a visual demonstration of this feature at https://github.com/emacsorphanage/evil-textobj-line.)
- [x] I am _*NOT*_ blindly checking these off.
  - Although it's not a draft, it's my first contribution to Doomemacs (actually Emacs in general), it's possible that I missed any best practices or what not, feel free to point it out or suggest improvements!

### DIFF
--- a/modules/editor/evil/config.el
+++ b/modules/editor/evil/config.el
@@ -338,6 +338,9 @@ directives. By default, this only recognizes C directives.")
           ("<" . ">"))))
 
 
+(use-package! evil-textobj-line)
+
+
 (use-package! evil-traces
   :after evil-ex
   :config
--- a/modules/editor/evil/README.org
+++ b/modules/editor/evil/README.org
@@ -32,6 +32,7 @@ This holy module brings the Vim editing model to Emacs.
 - [[doom-package:evil-snipe]]
 - [[doom-package:evil-surround]]
 - [[doom-package:evil-textobj-anyblock]]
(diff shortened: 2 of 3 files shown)


modules/editor/evil/config.el
```diff
@@ -338,6 +338,9 @@ directives. By default, this only recognizes C directives.")
           ("<" . ">"))))
 
 
+(use-package! evil-textobj-line)
```
ryuheechul, line 341: Initially, I tried the code below thinking that it might be beneficial for a startup time? (blindly following from the usage right above from `evil-textobj-anyblock`)

```elisp
(use-package! evil-textobj-line
  :defer t)
```

But then two reasons why I didn't include in the PR:
- currently I don't really know what `:defer` does
- also with `:defer t`, it doesn't seem to work

hlissner:
Thanks for the PR, but I must turn it down because `evil-textobj-line` doesn't provide enough value to warrant it being included by default. For one, you get line-wise selection for free, e.g. 

- `dal` -> `dd`
- `dil` -> `^D`
- `val` -> `V`
- `vil` -> `^v$`
- `yil` -> `^Y`
- `yal` -> `yy`

etc. Thanks for bringing it to my attention, in any case!

PR #8262, opened 2025-02-02
fix(corfu): fix default minibuffer completion

LemonBreezes:
After setting +corfu-want-minibuffer-completion to t, I noticed that it Corfu was still being enabled in Consult when finding files. This change fixes that by ensuring that Corfu gets disabled in the minibuffer whenever another completion framework is active.

-----
- [x] I searched the issue tracker and this hasn't been PRed before.
- [x] My changes are not on [the do-not-PR list](https://doomemacs.org/donotpr) for this project.
- [x] My commits conform to [Doom's git conventions](https://doomemacs.org/d/git-conventions).
- [ ] My changes are visual; I've included before and after screenshots.
- [ ] I am blindly checking these off.
- [x] Any relevant issues or PRs have been linked to.
- [ ] This a draft PR; I need more time to finish it.

PR #8334, opened 2025-04-01
fix(corfu): void variable local-map

LemonBreezes:
Fixes the change from #8262 which created a void variable error `local-map`.


hlissner:
Thanks again!

PR #8342, opened 2025-04-05
fix(magit): +evil-bindings: replace deprecated fn

panchoh:
`magit-stage-buffer-file` is an obsolete command (as of Magit 4.3.2); use `magit-file-stage` instead.

I got this error while adding a file to the stage. Fixed this one, but I reckon that there may be others.

-----

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -528,7 +528,7 @@
         :desc "Magit clone"               "C"   #'magit-clone
         :desc "Magit fetch"               "F"   #'magit-fetch
         :desc "Magit buffer log"          "L"   #'magit-log-buffer-file
-        :desc "Git stage this file"       "S"   #'magit-stage-buffer-file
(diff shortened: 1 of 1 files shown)


panchoh:
(sorry, after two blunders, I now understand that PRs must go first to draft state; won't happen again :crossed_fingers:)

hlissner:
> (sorry, after two blunders, I now understand that PRs must go first to draft state; won't happen again 🤞)

Sorry for the confusion; I put PRs into draft mode when there's more to do or if I plan to tackle it differently myself or elsewhere (what the "moved" tag means), and not because you did anything wrong (you didn't). To be specific, the deprecation affects both `magit-stage-buffer-file` and `magit-unstage-buffer-file`, but this PR only fixed the first, so I've done them both in 26b9776.

Thanks for bringing it to my attention in any case!

panchoh:
Thanks for the comprehensive explanation, @hlissner, I appreciate it.

May the storms of Mars stay away from your path.

PR #8345, opened 2025-04-07
fix(org): call `keywordp', not `keyword'

simendsjo:


hlissner (approved):

hlissner:
Whoops! Good catch. Thanks again for your help!

PR #8361, opened 2025-04-15
Fix(python): python-ts-mode behavior with emacs 30.1

Eising:
Recent emacs versions (at least 30.1) invokes python-ts-mode directly without going through python-mode.

This meant that the LSP initialization and other similar hook calls were not executed.

This PR fixes this by conditionally running `(add-hook!)` with the `python-ts-mode-local-vars-hook` if python is enabled with the `+tree-sitter` flag.

I've chosen to keep the `python-mode-local-vars-hook` in place to ensure
backwards compatibility.

-----

### DIFF
--- a/modules/lang/python/config.el
+++ b/modules/lang/python/config.el
@@ -21,6 +21,10 @@
         python-indent-guess-indent-offset-verbose nil)
 
   (when (modulep! +lsp)
+    (when (modulep! +tree-sitter)
+      ;; Append the lsp! entrypoint to to python-ts-mode
+      (add-hook 'python-ts-mode-local-vars-hook #'lsp! 'append))
+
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but I cannot accept it because there is a larger, and ongoing, effort to rewrite the `:tools tree-sitter` module to swap out tree-sitter with treesit.el. The `*-ts-mode` variants of major modes are for treesit integration, so implementing this now will only complicate the other effort, but rest assured hooks *like* these will be part of the treesit.el update, when it does come out. Thanks for bringing it to my attention, in any case!

Eising:
Hi @hlissner,

I appreciate the feedback. I would have hoped for a merge to at least deal with the present bug until the rewrite has been completed, but I understand. I think I'll revert to emacs 29 for the time being.

PR #8385, opened 2025-05-11
fix(terraform): fix incorrect mapping

debashisbiswas:
Both `fmt` and `destroy` are defined on `f`. In my case, this mean that which-key hinted `fmt`, but actually ran `destroy`. This mapping looks like a typo, and I have updated it to use `d`, matching the convention of using the first letter of the command.


hlissner (approved):

hlissner:
Yeah, that does look like a typo. Thanks for your help!

PR #8395, opened 2025-05-17
tweak(lib): interactively set opacity for current and future frames

lattarov:
## Context
I use Emacs in daemon mode and wanted a way to set opacity to all new Emacs frames as soon as they're spawned. 

## Analysis
doom/set-frame-opacity only sets the current frame's opacity, but all new frames are not affected by it, and I couldn't find any other readily available way to do so.

I found two workarounds:

1) adding a hook server-after-make-frame-hook in config.el:

`
(add-hook! 'server-after-make-frame-hook
    (doom/set-frame-opacity opacity))
`

Would be a nice addition to the documentation somewhere in the UI readme's FAQ, but from what I gathered [docs are DNP until the doom modules refactor is finished](https://github.com/orgs/doomemacs/projects/17?pane=issue&itemId=76207420).

2) Change doom/set-frame-opacity so that it affects all frames
Proposed by this PR.

## Previous work on opacity handling

- #4007 (not merged) intended to implement a module to handle interactive current frame opacity handling, but it didn't solve my problem for it doesn't work on other frames. 
- #7721 fixed a PGTK related regression, a priori this PR does not touch this part of the code.

## Testing
Tested on PopOS 22.04, Wayland session.

-----
- [ x] My changes are not on [the do-not-PR list](https://doomemacs.org/donotpr) for this project.
- [ x] My commits conform to [Doom's git conventions](https://doomemacs.org/d/git-conventions).
- [x ] Any relevant issues or PRs have been linked to.
- [x ] This a draft PR; I need more time to finish it.

### DIFF
--- a/lisp/lib/ui.el
+++ b/lisp/lib/ui.el
@@ -176,22 +176,37 @@ Use `winner-undo' to undo this. Alternatively, use
     (while (ignore-errors (windmove-down)) (delete-window))))
 
 ;;;###autoload
-(defun doom/set-frame-opacity (opacity)
-  "Interactively change the current frame's opacity.
+(defun doom/set-frame-opacity (opacity &optional all-frames?)
+  "Interactively change the current and all new frames' opacity.
 
-OPACITY is an integer between 0 to 100, inclusive."
+OPACITY is an integer between 0 to 100, inclusive. If nil, prompt user
+for value and if it should be applied to future frames.
+
+ALL-FRAMES? is an optional that if non-nil defines sets OPACITY to all
+future frames."
   (interactive '(interactive))
+  (setq prompt-user? (eq opacity 'interactive))
   (let* ((parameter
           (if (eq window-system 'pgtk)
               'alpha-background
             'alpha))
          (opacity
-          (if (eq opacity 'interactive)
+          (if prompt-user?
               (read-number "Opacity (0-100): "
                            (or (frame-parameter nil parameter)
                                100))
             opacity)))
-    (set-frame-parameter nil parameter opacity)))
+
+    ;; current frame
+    (set-frame-parameter nil parameter opacity)
+
+    (if prompt-user?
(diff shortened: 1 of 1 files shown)


hlissner:
`doom/set-frame-opacity` exists for changing the opacity of a frame interactively. I'll consider extending it to operate on all current (and future) frames if passed the prefix argument:

```elisp
(defun doom/set-frame-opacity (opacity &optional all-frames?)
  "Interactively change the current frame's opacity.

OPACITY is an integer between 0 to 100, inclusive. If ALL-FRAMES? is non-nil,
apply this to all existing frames, as well as future frames."
  (interactive (list 'interactive current-prefix-arg))
  (let* ((parameter
          (if (eq window-system 'pgtk)
              'alpha-background
            'alpha))
         (opacity
          (if (eq opacity 'interactive)
              (read-number "Opacity (0-100): "
                           (or (frame-parameter nil parameter)
                               100))
            opacity)))
    (if (not all-frames?)
        (set-frame-parameter nil parameter opacity)
      (modify-all-frames-parameters (cons parameter opacity))
      (setf (alist-get parameter default-frame-alist) opacity))))
```

However, from the way you describe it, `default-frame-alist` is better suited to your purpose:

```elisp
;;; add to $DOOMDIR/config.el
(add-to-list 'default-frame-alist '(alpha-background . 95))
```

hlissner:
Sorry for the delay, I've pushed a fixed `doom/set-frame-opacity` in 2c18b61 which:

- Now takes a `FRAMES` argument (a list of frames or `t` for all open and future frames).
- If called interactively:
  - With the prefix arg, `FRAMES` defaults to `t`
  - Without the prefix arg, `FRAMES` defaults to `(list (selected-frame))` (the old behavior)

Which should satisfy the requirements of these PR, while also allowing `doom/set-frame-opacity` to be used programmatically, if desired. Thanks for bringing it to my attention, in any case!

PR #8399, opened 2025-05-24
fix(workspaces): avoid duplicate workspace creation when no frame exists

kilesduli:
When Emacs is started as a daemon and no frames exist, `emacsclient -c` should simply switch to the main workspace. However, before `+workspaces-associate-frame-fn` is called, the new frame is already added to the frame list, which causes `persp-frame-list-without-daemon` to always return a non-empty list. A new workspace will be created instead of switching to the main one.

Before:
``` bash
emacs --daemon
emacsclient --eval "(+workspace-list-names)"
# => ("main")
emacsclient -c --eval "(+workspace-list-names)"
# => ("main" "#1")
```

Affter:
``` bash
emacsclient -c --eval "(+workspace-list-names)"
# => ("main")
```

-----

### DIFF
--- a/modules/ui/workspaces/autoload/workspaces.el
+++ b/modules/ui/workspaces/autoload/workspaces.el
@@ -532,16 +532,16 @@ created."
 (defun +workspaces-associate-frame-fn (frame &optional _new-frame-p)
   "Create a blank, new perspective and associate it with FRAME."
   (when persp-mode
-    (if (not (persp-frame-list-without-daemon))
-        (+workspace-switch +workspaces-main t)
-      (with-selected-frame frame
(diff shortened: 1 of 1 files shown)


kilesduli:
Supplement: When running Emacs as a daemon, if no frames have been created yet, after invoking `emacsclient -c` and before the GUI frame actually appears, the selected-frame is `teminal-frame`. As a result, even if we switch to the "main" workspace, the newly created GUI frame is not bound to "main", but instead ends up in the "none" workspace.

This causes an error when we call `+workspace/kill` after startup. To fix this, I moved the if statement below `with-selected-frame`.

hlissner (approved):

Issue #8425, opened 2025-06-24
commit-msg hook runs forever on special URL


real-or-random:
### I confirm that...

- [x] I have searched the [issue tracker](https://github.com/doomemacs/doomemacs/issues), [documentation](https://docs.doomemacs.org), [FAQ](https://docs.doomemacs.org/-/faq), [Discourse](https://discourse.doomemacs.org), and [Google](https://google.com), in case this issue has already been reported/resolved.

- [x] I have read "[How to Debug Issues](https://doomemacs.org/d/how2debug)", and will use it to provide as much information about this issue as possible.

- [x] The issue can be reproduced on the **latest** available commit of Doom.

- [x] The issue can be reproduced on a stable release of Emacs, such as 27, 28, 29, or 30. *(Unstable versions end in .50, .60, or .9x)*

### Expected behavior

The hook should terminate.

### Current behavior

The hook runs forever, keeping the CPU busy.

### Steps to reproduce

1. Save this file to `/tmp/msg.txt`:

```
tweak(latex): foo bar foooooo

 https://this-is-a-line-starting-with-two-spaces-https-and-it-is-longer-than-72-chars
```

2. Run `~/.config/emacs/bin/doom --no-color ci hook commit-msg /tmp/msg.txt`

### System Information

https://gist.github.com/real-or-random/5ec2d6f3d5a7059f9394cf7e7f103683

hlissner:
As of ed9190e this should be resolved. Let me know if that isn't the case and I'll reopen it. Thanks for bringing it to my attention!

PR #8426, opened 2025-06-24
tweak(latex): don't insert braces after sub/superscript

real-or-random:
This avoids inserting braces when you type "a_", so you end up typing "a_i" instead of "a_{i}".

The official LaTeX manual says that braces are correct, but it doesn't matter for many cases, e.g., when the sub/superscript is just a single token like "i". It's simply a matter of taste in these cases.
 - Frank Mittelbach, a LaTeX maintainer, says that he types "a_i": https://tex.stackexchange.com/questions/82329/how-bad-for-tex-is-omitting-braces-even-if-the-result-is-the-same#comment176261_82331
 - David Carlisle, another LaTeX maintainer, says that a_i "arguably improves the look of the source code": https://tex.stackexchange.com/a/82337
 - Joseph Wright, yet another LaTeX maintainer, says that he "would always use braces here, so favour a_{i}": https://tex.stackexchange.com/a/1929

And I think if it is a matter of taste, then the sane default is to avoid any magic that gets in the way of users. Anyone who prefers braces can still enable this argument, but people who do not like them won't get upset every time they type "_" or "^".

-----

----

Note that I skipped the linter due to #8425.

### DIFF
--- a/modules/lang/latex/config.el
+++ b/modules/lang/latex/config.el
@@ -59,8 +59,6 @@ okular and pdf-tools.")
       TeX-source-correlate-method 'synctex
       ;; Don't start the Emacs server when correlating sources.
       TeX-source-correlate-start-server nil
-      ;; Automatically insert braces after sub/superscript in `LaTeX-math-mode'.
-      TeX-electric-sub-and-superscript t
       ;; Just save, don't ask before each compilation.
       TeX-save-query nil)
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
LGTM! Thanks again!

PR #8428, opened 2025-06-26
fix: *-local-vars-hooks not being run in tree-sitter modes

JasZhe:
I have some bespoke code that I had in the local-vars hook that checked a variable set in a dir-locals file. Recently I've been trying to use more tree-sitter based modes, and I was perplexed as to why my hook only worked in non-tree-sitter modes. 

I don't know if there's a third way to go about this but I thought it was either some string manipulation using `symbol-name` or `derived-mode-all-parents`. I think the latter is less brittle so that's what I've done here.

Thanks! 

Fix: #0000
Ref: #0000
Close: #0000

-----
- [-] I searched the issue tracker and this hasn't been PRed before.
- [-] My changes are not on [the do-not-PR list](https://doomemacs.org/donotpr) for this project.
- [-] My commits conform to [Doom's git conventions](https://doomemacs.org/d/git-conventions).


hlissner:
Thanks for the PR, but I must turn it down. There is already ongoing work in the [feature/treesit](https://github.com/doomemacs/doomemacs/tree/feature/treesit) branch to bring treesit support to Doom, which addresses this issue, but not by triggering the hooks of the parent mode(s), but by simply adding functions to the children hooks (because it's not always desirable that a major mode and its derived treesit mode share the same hooks). Thanks for bringing it to my attention, in any case!

PR #8433, opened 2025-07-02
tweak(macos): use full path for open

neuhalje:
Use the full path for the MacOS "open" command.

Using plain "open" will conflict with nushell's "open" command. Use the full path for open to make it working.

### DIFF
--- a/modules/os/macos/autoload.el
+++ b/modules/os/macos/autoload.el
@@ -13,12 +13,11 @@
                               (dired-get-file-for-visit)
                             (buffer-file-name)))
                  nil t)))
-         (command (format "/usr/bin/open %s"
-                          (if app-name
-                              (format "-a %s '%s'" (shell-quote-argument app-name) path)
-                            (format "'%s'" path)))))
-    (message "Running: %s" command)
-    (shell-command command)))
+         (args (cons "open"
+                     (append (if app-name (list "-a" app-name))
+                             (list path)))))
+    (message "Running: %S" args)
+    (apply #'doom-call-process args)))
 
 (defmacro +macos--open-with (id &optional app dir)
   `(defun ,(intern (format "+macos/%s" id)) ()
--- a/modules/os/macos/autoload.el
+++ b/modules/os/macos/autoload.el
@@ -13,7 +13,7 @@
                               (dired-get-file-for-visit)
                             (buffer-file-name)))
                  nil t)))
-         (command (format "open %s"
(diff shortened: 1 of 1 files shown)


hlissner:
I'd rather avoid full paths where I can (so users can still shim or replace `open` at the system level). Might be better to swap out `shell-command` for `doom-call-process` instead. Could you drop this advice into `$DOOMDIR/config.el` and see if it solves this for you?

```elisp
(defadvice! fixed-+macos-open-with (&optional app-name path)
  :override #'+macos-open-with
  (let* ((path (expand-file-name
                (replace-regexp-in-string
                 "'" "\\'"
                 (or path (if (derived-mode-p 'dired-mode)
                              (dired-get-file-for-visit)
                            (buffer-file-name)))
                 nil t)))
         (args (cons "open"
                     (append (if app-name (list "-a" app-name))
                             (list path)))))
    (message "Running: %S" args)
    (apply #'doom-call-process args)))
```

neuhalje:
That works

hlissner (changes requested):
Then in that case, please make this change to `+macos-open-with`:

```diff
@@ -13,12 +13,11 @@ (defun +macos-open-with (&optional app-name path)
                               (dired-get-file-for-visit)
                             (buffer-file-name)))
                  nil t)))
-         (command (format "open %s"
-                          (if app-name
-                              (format "-a %s '%s'" (shell-quote-argument app-name) path)
-                            (format "'%s'" path)))))
-    (message "Running: %s" command)
-    (shell-command command)))
+         (args (cons "open"
+                     (append (if app-name (list "-a" app-name))
+                             (list path)))))
+    (message "Running: %S" args)
+    (apply #'doom-call-process args)))

```

jdelStrother:
@neuhalje presumably this is because `shell-file-name` points to `nu`? Maybe instead you might set it to a more 'normal' shell like bash - I remember I had to do this a while back when I had SHELL=fish: 

```elisp
(setq shell-file-name (executable-find "bash"))
(let ((fishpath (executable-find "fish")))
  (when fishpath
    (setq explicit-shell-file-name fishpath
          vterm-shell fishpath)))
```

 Otherwise I ran into a lot of problems with emacs modules that expected `shell-command` to work with regular bashisms.


hlissner's suggestion looks good though - I don't think there's any need to use a shell for macos-open-with.

hlissner (approved):

hlissner:
LGTM! Thanks for your help!

Issue #8410, opened 2025-06-04
Improve undo/redo in the inferior shell


egr95:
### I confirm that...

- [x] I have searched the [issue tracker](https://github.com/doomemacs/doomemacs/issues), [documentation](https://docs.doomemacs.org), [FAQ](https://docs.doomemacs.org/-/faq), [Discourse](https://discourse.doomemacs.org), and [Google](https://google.com), in case this issue has already been reported/resolved.

- [x] I have read "[How to Debug Issues](https://doomemacs.org/d/how2debug)", and will use it to provide as much information about this issue as possible.

- [x] The issue can be reproduced on the **latest** available commit of Doom.

- [x] The issue can be reproduced on a stable release of Emacs, such as 27, 28, 29, or 30. *(Unstable versions end in .50, .60, or .9x)*

### Expected behavior

As discussed with Henrik, the following could be a good addition to the `shell` module:
- When in the inferior shell, undo/redo should only apply to any edits made after the last prompt. One should not be allowed to undo the output of the command that was previously run, nor the previous command itself.

The same holds for other REPL buffers such as eshell and those from [emacs-jupyter](https://github.com/emacs-jupyter/jupyter), but I would expect the inferior shell is the most widely used and therefore the most relevant here.

Thanks!

### Current behavior

.

### Steps to reproduce

.

### System Information

.

hlissner:
I've made some progress with this. Please give it a try and let me know if it's what you were expecting:

- For comint/shell:
  ```elisp
  (add-hook 'comint-exec-hook #'buffer-disable-undo)
  (defadvice! doom--comint-enable-undo-a (process _string)
    :after #'comint-output-filter
    (let* ((start-marker comint-last-output-start)
           (end-marker (or (if process (process-mark process))
                           (point-max-marker))))
      (when (and (< start-marker end-marker)
                 (eq (char-before start-marker) ?\n)) ;; Account for some of the IELM’s wilderness.
        (let ((inhibit-read-only t))
          (buffer-enable-undo)
          (setq buffer-undo-list nil)))))
  ```

- For eshell, you'd need to comment out [these lines](https://github.com/doomemacs/doomemacs/blob/6a6ac8f315cdf7aba13480c7ede2724ae98d81e0/modules/term/eshell/config.el#L90-L96), and use:
  ```elisp
  (add-hook 'eshell-pre-command-hook #'buffer-disable-undo)
  (add-hook! 'eshell-post-command-hook
[...]

PR #8434, opened 2025-07-14
fix(:term): ensure process-buffer is current

DennieTeMolder:
Recent changes have add advice to `comint-output-filter` to protect previous terminal output from being modified. 
However, these advises assume the `comint-output-filter` is called from the buffer associated to the provided process. This turns out to not always be the case as for example `ess--flush-accumulated-output` calls the output filter from inside of a separate accumulation buffer, resulting in the following error: "error in process filter: Wrong type argument: number-or-marker-p, nil". 

To fix this issue and other potential side effects, this PR wraps the function body in a `with-current-buffer` form.

-----

### DIFF
--- a/lisp/doom-ui.el
+++ b/lisp/doom-ui.el
@@ -356,13 +356,14 @@ windows, switch to `doom-fallback-buffer'. Otherwise, delegate to original
   (add-hook 'comint-exec-hook #'buffer-disable-undo)
   (defadvice! doom--comint-enable-undo-a (process _string)
     :after #'comint-output-filter
-    (let ((start-marker comint-last-output-start))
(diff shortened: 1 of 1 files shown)


hlissner (approved):

rinconjc:
hey folks, I wonder if this PR would fixe the frequent `error in process filter: and: Wrong type argument: number-or-marker-p, nil` I'm seeing in the latest Doom emacs. Specially, when running shell commands.

DennieTeMolder:
I haven't seen the error occur for `shell-command`, but this PR does fix this issue for  `async-shell-command`. Have a look at my latest comment on #8410 if you want a more detailed explanation.

PR #8442, opened 2025-07-22
tweak: update dockerfile template

root-goblin:
Modernizes the template for new `Dockerfile`s

* modules/editor/file-templates/templates/dockerfile-mode/__:
  - preserves file indentation
  - replace deprecated `MAINTAINER` instruction with `LABEL`, as per: https://docs.docker.com/reference/dockerfile/

Side note: there's an argument to be made for replacing `phusion/baseimage` with something more bland as a default, since the advent of `docker run --init` solves the same problem.

-----
- [ ._. ] I am blindly checking these off.

### DIFF
--- a/modules/editor/file-templates/templates/dockerfile-mode/__
+++ b/modules/editor/file-templates/templates/dockerfile-mode/__
@@ -1,4 +1,6 @@
+# expand-env: ((yas-indent-line 'fixed) (yas-wrap-around-region 'nil))
+# --
 FROM ${1:phusion/baseimage:latest}
-MAINTAINER ${2:`user-full-name` <`user-mail-address`>}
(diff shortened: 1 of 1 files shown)


hlissner (approved):

root-goblin:
One thing I noticed is that the magit hook using the cli lint doesn't pass when I use `tweak(file-templates): ...`

hlissner:
LGTM! Thanks for your help!

> One thing I noticed is that the magit hook using the cli lint doesn't pass when I use tweak(file-templates): ...

Sorry about that! I'm in the middle of backporting bits and pieces of a rewrite for Doom's CLI, which includes moving linter rules out into `.doom` config files, but I haven't backported the parts responsible for reading `.doom` files yet! Until that's done, feel free to ignore the linter.

PR #8448, opened 2025-08-04
fix: Mode map error in PHP config

ribaricplusplus:
Very small fix. On the current Tree Sitter branch, opening a PHP file results in an error because the mode map is not passed correctly to the `map!` macro.


ribaricplusplus:
This is not a proper fix, I'll make a new PR.

ribaricplusplus:
@hlissner I'm guessing this might've been marked as draft because of eval. I removed eval now :)

Let me know if any other changes needed.

modules/lang/php/config.el
```diff
@@ -25,26 +25,26 @@
...
-    :str "string"
-    :bool "list"
-    ;; Flow
-    :not "!"
-    :and "&&" :and "and"
-    :or "||" :or "or"
-    :for "for"
-    :return "return"
-    :yield "use")
+                  ;; Functional
```
kaiwk, line 28: Both `set-ligatures!` and `set-company-backend!` are not indented well. 

See https://github.com/doomemacs/doomemacs/blob/2129fffc79189bb6d4fd3bb665a4a936247411c4/modules/ui/ligatures/autoload/ligatures.el#L35

The body of the `set-ligatures!` should be indented like a defun body not a plain list.

hlissner:
I've resolved the whitespace issues (and removed an unused lexical variable) in 31f5b60. Thanks for bringing it to my attention!

PR #8453, opened 2025-08-09
feat: Update gptel

ribaricplusplus:
Update gptel so it supports GPT 5 and other improvements.


hlissner (approved):

hlissner:
The greater module category (:tools) was overdue a bump, plus newer commits of gptel were available, so to capture those and conform this to our commit conventions, I've bumped it in 09e5bc9. Thanks for bringing it to my attention!

PR #8456, opened 2025-08-14
feat(literate): indent output config with +pretty

PrismaPhonic:
This adds auto indenting of the generated config file, from the user's literate config, when the +pretty flag is added to the literate module. It's equivalent to opening the generated config.el file, selecting the entire buffer, and using the `indent-region` command.

-----

### DIFF
--- a/modules/config/literate/autoload.el
+++ b/modules/config/literate/autoload.el
@@ -166,6 +166,18 @@
         (indent-region (point-min) (point-max))
         (save-buffer)))))
 
+;;;###autoload
+(defun +literate-indent-elisp-src-blocks ()
+  "Indent all emacs-lisp src blocks in the current Org buffer, then save."
+  (interactive)
+  (save-excursion
+    (org-babel-map-src-blocks nil
+      (when (string= lang "emacs-lisp")
+        (org-babel-do-in-edit-buffer
+          (emacs-lisp-mode)
+          (indent-region (point-min) (point-max)))))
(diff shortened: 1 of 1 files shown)


PrismaPhonic:
I've just added a new commit, of an additional helper function, which can be used to auto indent all emacs-lisp babel src blocks in the current file. It's intended to help users to auto format their emacs-lisp blocks in their literate config, improving the styling of each respective codeblock.

hlissner:
I'm not sure I understand the utility of this. Are your elisp blocks minified/unformatted or something? Does org un-format elisp blocks in the tangling process? 

Not to mention, my understanding of literate configs is that the org file(s) themselves are supposed to be the human-facing part of the project, not the files that they tangle out to, so prettifying the end result seems like an odd choice, except perhaps when debugging the tangling process.

PrismaPhonic:
To your first point: it's not uncommon for a user to reuse sections of someone else's literate config, where the babel blocks may have a superfluous level of indention to all of the contents in the babel block. Some users prefer the look of this, but then would not want the generated config to apply an extra level of indentation to everything. 

To your second point, that's why I added the helper method which can be used to clean up the indentation of each babel block in the literate config itself. If a user were to use this, then the automatic cleanup of the tangled file would not yield any changes. 

I could see an argument for only keeping the helper function in the tip commit. For me personally I like the piece of mind in knowing that my generated config file is always properly formatted

PR #8469, opened 2025-08-28
tweak(notmuch): include the user's name in From: header

brenns10:
The notmuch/compose function allows selecting between the email addresses that the user has configured, but the email address is used directly in the From header. Messages display much nicer in people's inbox when the From contains a name in addition to an emeail address, so construct the From header with the user's name as well as the chosen email address.

Before this change, running "notmuch/compose" creates a buffer with the From: field populated just with my email address. After, the from field is in the form `From: My Name <my_email@example.com>`.

Thanks!

-----

### DIFF
--- a/modules/email/notmuch/autoload.el
+++ b/modules/email/notmuch/autoload.el
@@ -116,7 +116,9 @@
   (notmuch-mua-mail
    nil
    nil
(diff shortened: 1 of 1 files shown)


hlissner:
LGTM! Thanks for your help!

Issue #8455, opened 2025-08-11
Treemacs closes when switching perspectives


jlandahl:
### I confirm that...

- [x] I searched the issue tracker and found no similar issues.

- [x] I searched the [documentation](https://docs.doomemacs.org), [FAQ](https://docs.doomemacs.org/-/faq), [discussion board](https://doomemacs.org/discuss), and [Google](https://google.com) for solutions to my issue.

- [x] I read [the debugging guide](https://doomemacs.org/d/how2debug) and researched the issue to the best of my ability.

- [x] My issue can be observed on the latest available commit of Doom.

- [x] My issue can be observed on a stable release of Emacs (i.e. doesn't end in `.50`, `.60`, or `.90`–`.99`)

### Describe your issue

When I have a perspective open with Treemacs visible (enabled via `SPC o p` if that matters), if I switch to another perspective and then back, Treemacs is no longer visible in the first perspective.

### Steps to reproduce

The workflow is something like this: I open a project with `SPC p p`, then enable Treemacs via `SPC o p`. Let's say this is now perspective 2. I switch to the original perspective via `M-1`, then back to the project via `M-2`. Treemacs is no longer visible, requiring me to reopen it with `SPC o p`. This happens consistently when switching back and forth between perspectives.

### System information

https://gist.github.com/jlandahl/c4b020c4f4a3563d1678764c6b944317

jlandahl:
I've tried the versions of persp-mode between the previous pinned version (40e9993) and the new pinned version (f146ddc) to find which change caused the problem. (The doom version bump was made [here](https://github.com/doomemacs/doomemacs/commit/a1121acc94dd7eee65c152df6f07495e72dc3561)).

The version that caused the breakage is this one:
https://github.com/Bad-ptr/persp-mode.el/commit/5adf105f7168cd88357702d07768844b4788507a

I'll look into the changes to see if I can figure out what went wrong.

jlandahl:
I've done some initial digging into the code in that change, but have to put it on hold for the moment.

For anyone else running into this problem, the workaround is to change your `~/.config/emacs/modules/ui/workspaces/packages.el` file to read as follows:

```elisp
(package! persp-mode :pin "124f4430008859a75b25521c474f37aa9f75afeb")
```

Then run `doom sync` and restart.

[...]

PR #8474, opened 2025-08-31
bump: :ui persp-mode.el

JasZhe:
Bad-ptr/persp-mode.el@f146ddccaf59 -> Bad-ptr/persp-mode.el@fab4bf769274

Fix: #8455
Ref: #0000
Close: #0000

Upstream issue: https://github.com/Bad-ptr/persp-mode.el/issues/154

-----

Side windows (doom popups for instance) weren't being restored/associated with workspaces prior to this change in persp-mode.


jlandahl:
As mentioned [https://github.com/doomemacs/doomemacs/issues/8455#issuecomment-3240468712](here), this causes more problems than it solves. Doom (or persp-mode?) gets really confused when opening new perspectives, requiring manual cleanup to get first and new perspectives into a normal-ish state. Not sure if Doom needs some additionally compatibility code or if there are new persp-mode settings needed.

JasZhe:
huh.. I also tested it out on an empty doom config on the latest doom master via `doom run --doomdir nodoomconfig/` and it seemed fine on my end. But maybe I'll try using/testing this a bit more during the work week and see if anything comes up

JasZhe:
https://github.com/user-attachments/assets/322cf7bf-393c-45dc-9b88-7634207d2a5c

Bottom side windows seem to be respected but not the dirvish-side window for some reason.. 

but treemacs seems to be fine as well 

https://github.com/user-attachments/assets/f3620c1b-f851-48b5-84ae-8fe0687238e3

jlandahl:
I recently updated to Emacs 30.2, so I don't know if some of the (very!) weird things I've seen are specific to 30.2 or if they also affect 30.1. I don't really have a simple way to test under 30.1 now, either. I ended up making my own fork of Doom for now so I could revert the breaking changes while still getting updates in other areas.

JasZhe:
I was on emacs30.1, but the change log between .1 and .2 was a bug fix so I don't think that would be the reason why

hlissner:
Until persp-mode sorts things out upstream (or I get around to replacing persp-mode), I've reverted it back to a commit before all of its recent breaking changes. Thanks for bringing this to my attention though!

PR #8477, opened 2025-09-01
fix(erlang): correct tree-sitter config

katafrakt:
Missing `:url` in the config caused error during the initialization, if Erlang module with `+tree-sitter` was turned on in `init.el`.

-----


hlissner:

PR #8479, opened 2025-09-01
Fix syntax for setting tree-sitter URL in Erlang config

jcf:
Quick heads up for anyone using Tree Sitter and Erlang — small syntax issue in the grammar specification causes Tree Sitter to fall over.

Just need to add a `:url` keyword to the specification and you're good to go!

-----


jcf:
Shared over in Discord. Closing as the fix is super trivial.

PR #8488, opened 2025-09-03
fix: Run hooks only after switch-to-buffer for server

real-or-random:
When a file is visited via `emacsclient`, server.el does the following in this order:
 - create a buffer `b` visiting the file
 - run `(set-buffer b)`
 - trigger `server-visit-hook`
 - run `(switch-buffer b)`
 - trigger `server-switch-hook`

Thus, the right hook for `doom-run-switch-buffer-hooks-h` is `server-switch-hook` because the "switch buffer" hooks may assume that the buffer has already been switched to.

This fixes an org error that occurs when running

    emacsclient --create-file --no-wait foo.txt

while there's a frame containing an org-roam file. Without this commit, the server will create a new frame and set the current buffer to foo.txt. But the new frame will still display the (duplicated) window for the org-roam file. Then `server-visit-hook` will be triggered and eventually run `+org-roam-manage-backlinks-buffer-h`, which will try to enable the org-roam backlinks buffer. But this will error because the current buffer is not an org(-roam) buffer.

Amend: 4a6de2419c81

-----

### DIFF
--- a/lisp/doom-ui.el
+++ b/lisp/doom-ui.el
@@ -737,7 +737,7 @@ triggering hooks during startup."
   (add-hook 'window-selection-change-functions #'doom-run-switch-window-hooks-h)
   (add-hook 'window-buffer-change-functions #'doom-run-switch-buffer-hooks-h)
   ;; `window-buffer-change-functions' doesn't trigger for files visited via the server.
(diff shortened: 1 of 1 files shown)


hlissner:

PR #8496, opened 2025-09-05
Fix (void-variable evil-windows-map) on startup of fresh install

Seanld:
It's very possible this is a me problem, because I'm shocked nobody else has reported it yet if not.

While trying to do a fresh install of Doom, I ran into `/modules/config/default/+evil-bindings.el` having a `(void-variable evil-windows-map)` error. It looks to be related to #8298, but between previous fresh installs of Doom and now, I don't think I've changed anything Evil-related in my config.

Simply binding `evil-window-map` after Evil loads seems to do the trick. 

If this is a me problem, I'd like to know how I can fix this in my config.

-----
- [x] My changes are not on [the do-not-PR list](https://doomemacs.org/donotpr) for this project. (The closest DNP only mentions no conditional wrapping.)

### DIFF
--- a/modules/config/default/+evil-bindings.el
+++ b/modules/config/default/+evil-bindings.el
@@ -341,7 +341,7 @@
       :desc "Org Capture"           "X"    #'org-capture
       ;; C-u is used by evil
       :desc "Universal argument"    "u"    #'universal-argument
-      :desc "window"                "w"    evil-window-map
+      (:after evil :desc "window"   "w"    evil-window-map)
       :desc "help"                  "h"    help-map
 
       (:when (modulep! :ui popup)


hlissner:
It's hard to tell _why_ it's happening to you without more info, but I can certify that it is local and not an actual bug. Evil is configured to load eagerly (i.e. immediately, upon loading the evil module), so unless `:config default` is ~~below~~ above `:editor evil` in your `doom!` block OR evil has failed to load for some unrelated reason (could be there's another error at startup short-circuiting Doom's startup process), this error should not occur.

In any case, I must turn down this PR, but feel free to follow up here.

Seanld:
Well I don't know what fixed it, but I had an old `init.el` I've been holding onto, and after you mentioned the order of the `doom!` definitions, I tried swapping around the order of the definitions. That didn't seem to fix it. So I decided it's probably time to update my `init.el` using the newest example file. It loads just fine now after doing that. Thanks for the pointer.

utsamir:
> Well I don't know what fixed it, but I had an old `init.el` I've been holding onto, and after you mentioned the order of the `doom!` definitions, I tried swapping around the order of the definitions. That didn't seem to fix it. So I decided it's probably time to update my `init.el` using the newest example file. It loads just fine now after doing that. Thanks for the pointer.

Can confirm this, it's not only a you problem @Seanld - it's a you *and* me problem at this point. Just ran into exactly the same problem after upgrading from a 5 mo old config. It definitely seems to be related to `init.el`.

I promise to post some more details if I get around to do some comparison between configs.

qxuken:
I ran into the same issue on macOS 26.3 (Emacs 30.2). I don't really know Elisp, but I copied what was done above and it worked perfectly:
`      :desc "window"                "w"    #'evil-window-map`

Issue #8497, opened 2025-09-05
HTML TS recipe is wrong


ispringle:
### I confirm that...

- [x] I searched the issue tracker and found no similar issues.

- [x] I searched the [documentation](https://docs.doomemacs.org), [FAQ](https://docs.doomemacs.org/-/faq), [discussion board](https://doomemacs.org/discuss), and [Google](https://google.com) for solutions to my issue.

- [x] I read [the debugging guide](https://doomemacs.org/d/how2debug) and researched the issue to the best of my ability.

- [x] My issue can be observed on the latest available commit of Doom.

- [x] My issue can be observed on a stable release of Emacs (i.e. doesn't end in `.50`, `.60`, or `.90`–`.99`)

### Describe your issue

When I try to install the HTML TS grammar it doesn't work

As you can see from the source, the pin is the same for both CSS and HTML. The chances of two hashes being the same is pretty much impossible.

```
((css "https://github.com/tree-sitter/tree-sitter-css" "v0.23.0" nil nil nil
      "6a442a3cf461b0ce275339e5afa178693484c927")
 (html "https://github.com/tree-sitter/tree-sitter-html" "v0.23.0" nil nil nil
       "6a442a3cf461b0ce275339e5afa178693484c927")
```

I believe the correct hash would be `14bdaf0da9e26e2de9b30178c2242539d2b0b285` as that's the on associated with 0.23.0, though it might be better to just upgrade to 0.23.2 at this point.

I lied, I'm on v31 but this is version agnostic. This is just a config issue.

### Steps to reproduce

Install the HTML TS parser.

### System information

https://pastebin.com/prLRv5eh

ispringle:
I've created #8498 to resolve this.

hlissner:
Resolved in https://github.com/doomemacs/doomemacs/commit/8f55404781edacf66fa330205533b002de3fb5ee.

PR #8498, opened 2025-09-05
Update +html.el

ispringle:
Updates tree-sitter-html grammar to 0.23.2, this also resolves the incorrect commit hash which is causing errors when trying to build.

Fix: #8497
Ref: #8497
Close: #8497

-----

### DIFF
--- a/modules/lang/web/+html.el
+++ b/modules/lang/web/+html.el
@@ -175,8 +175,8 @@
   :init
   (set-tree-sitter! 'html-mode 'html-ts-mode
     '((html :url "https://github.com/tree-sitter/tree-sitter-html"
-            :rev "v0.23.0"
-            :commit "6a442a3cf461b0ce275339e5afa178693484c927"))))
+            :rev "v0.23.2"
+            :commit "5a5ca8551a179998360b4a4ca2c0f366a35acc03"))))
 
 
 (use-package! mhtml-ts-mode  ; 31+ only


hlissner:
I chose 0.23.0 because 0.23.1 and newer failed to load on Emacs 29.x (ran fine on 30+ though). Can you confirm that 0.23.2 runs on 29.x for you? (And on 31+, `:commit` has precedence over `:rev` so the latter is ignored). The commit hash issue is legit though.

If we want Emacs 30.x users to benefit from `0.23.2`, use `(if (< (treesit-library-abi-version) 15) "v0.23.0" "v0.23.2")` instead.

ispringle:
I did not know that about the rev vs commit, however that tree-sitter-html repo didn't have a release for v0.23.0, so even on 31 (which I am using) it fails because there just wasn't a release cut for that version. https://github.com/tree-sitter/tree-sitter-html/releases

I'll just keep it on v0.23.0, looking at their commits there and honestly the only updates are deps and tooling, so there's no significant reason to pick 0.23.X over plain 0.23.0, this will keep the config simpler.

hlissner:
Turned out *both* :commit hashes were wrong, and since the latest commits are essentially 0.23.2, I've removed :commit altogether and made `:rev` conditional in 8f55404. Thanks for bringing it to my attention, in any case!

hlissner:
> I did not know that about the rev vs commit, however that tree-sitter-html repo didn't have a release for v0.23.0, so even on 31 (which I am using) it fails because there just wasn't a release cut for that version. https://github.com/tree-sitter/tree-sitter-html/releases

Ack, I was under the impression that it clones the repo if no release/archive is available for a tag. My assumptions need double checking...

PR #8505, opened 2025-09-12
fix(elixir): Consolidate package configurations

prashantvithani:
Move the `flycheck-credo` and `exunit` configurations into the `+elixir-common-config` function to group all common setup logic in one place to make it available for both `elixir-mode` and `elixir-ts-mode`.

### DIFF
--- a/modules/lang/elixir/config.el
+++ b/modules/lang/elixir/config.el
@@ -10,17 +10,17 @@
 
 (defun +elixir-common-config (mode)
   (set-ligatures! mode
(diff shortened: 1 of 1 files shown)


hlissner:
Moving the two use-package blocks into `+elixir-common-config` verbatim wasn't enough; its keymaps and hooks had to be updated as well (plus I wanted to avoid the whitespace issues). I've done this in 6fcdb49. Thanks for bringing it to my attention!

PR #8506, opened 2025-09-14
fix: rebar3 does not work with fmt -v arguments

apostaat:

Fix: #0000
Ref: #0000
Close: #0000

-----
- [v] I searched the issue tracker and this hasn't been PRed before.
- [v] My changes are not on [the do-not-PR list](https://doomemacs.org/donotpr) for this project.
- [v] My commits conform to [Doom's git conventions](https://doomemacs.org/d/git-conventions).
- [v] Any relevant issues or PRs have been linked to.

`doom doctor` utility raises a warning however the command it tries to invoke does not exist in `rebar3`
 i changed it to proper command call, and added check for `erlfmt -v` command, changed the text of warnings accordingly

### DIFF
--- a/modules/lang/erlang/doctor.el
+++ b/modules/lang/erlang/doctor.el
@@ -10,5 +10,10 @@
          "This module requires (:tools tree-sitter)")
 
 (when (modulep! :editor format)
-  (unless (and (executable-find "rebar3") (zerop (car (doom-call-process "rebar3" "fmt" "-v"))))
+  (unless (executable-find "erlfmt")
     (warn! "Couldn't find erlfmt. Formatting will be disabled.")))
+
+(when (modulep! :editor format)
+  (unless (and (executable-find "rebar3")
+               (zerop (car (doom-call-process "rebar3" "format"))))
+    (warn! "Couldn't find rebar3 with format plugin installed (needed in Erlang module). Formatting will be disabled.")))


hlissner:
Do erlang users need *both* `erlfmt` and `rebar3 format`, for some reason? The module is only configured to recognized [the latter](https://github.com/doomemacs/doomemacs/blob/master/modules/lang/erlang/config.el#L8-L9).

apostaat:
> erlang users need both erlfmt and rebar3 format, for some reason

@hlissner looks like there is no need in both. `rebar3` is more popular rather than `erlfmt`, however it needs to be configured to make it work (`rebar3 format` does not work from the box, `format` is a plugin that needs to be installed), while `erlfmt` installation does not need any configuration by default. 

In the [piece of code](https://github.com/doomemacs/doomemacs/blob/286f04c6ef5c1985ea7548d26ed09b1d76f8ee8b/modules/lang/erlang/config.el#L8-L9) you mentioned we need the same fix probably:  

 ```
 (set-formatter! 'erlfmt '("rebar3" "format")
    :modes '(erlang-mode erlang-ts-mode))
```

Should I open separate PR for it?

hlissner:
I've consolidated both changes (and conformed the commit to our git conventions) into dd8b706 (and 96498a9, but that was pushed a bit prematurely). Thanks for bringing it to my attention!

PR #8507, opened 2025-09-15
Fix typo: s/defaul/default/

tomekw:


hlissner:
Good catch! Thanks for your help.

PR #8514, opened 2025-09-19
fix(ada): use gpr-specific modes and indent properly

brownts:
GNAT Project modes (`gpr-mode` and `gpr-ts-mode`) exist to handle .gpr files. These modes are now used instead of the Ada major mode.

Lines may need to be re-indented when RET is pressed.  This is to handle cases of incomplete syntax and ambiguity in what may be entered when an empty line is initially indented.  Re-indenting after text has been entered corrects incorrectly guessed initial indentation.  To accommodate this scenario, RET is remapped to `reindent-then-newline-and-indent`.

Also updates documentation to reflect these changes.

-----

### DIFF
--- a/modules/lang/ada/README.org
+++ b/modules/lang/ada/README.org
@@ -20,8 +20,8 @@ This module adds support for the Ada and SPARK programming languages.
 ** Module flags
 # Flags should be in alphanumerical order.
 - +lsp ::
-  Enable LSP support for ~ada-mode~ / ~ada-ts-mode~. Requires [[doom-module::tools lsp]]
-  and a langserver (supports [[https://github.com/AdaCore/ada_language_server][ada_language_server]]).
+  Enable LSP support for ~ada-mode~ / ~ada-ts-mode~, ~gpr-mode~ / ~gpr-ts-mode~.
+  Requires [[doom-module::tools lsp]] and a langserver (supports [[https://github.com/AdaCore/ada_language_server][ada_language_server]]).
 - +tree-sitter ::
   Leverages tree-sitter for better syntax highlighting and structural text
   editing. Requires [[doom-module::tools tree-sitter]] and Emacs 29.1+.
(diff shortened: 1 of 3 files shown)


brownts:
Hi @hlissner:

I'm not a seasoned DoomEmacs user, so maybe I'm missing something, but I'm use to Vanilla Emacs default behavior of `RET` being bound to `newline`.  It appears in DoomEmacs, `RET` is bound to `newline-and-indent` which changes the behavior.  When `electric-indent-mode` is enabled and RET is bound to `newline`, pressing `RET` causes the current line to be reindented, then a newline is inserted and finally the empty line is indented.  However, because DoomEmacs appears to remap `newline` to `newline-and-indent` (I believe in `modules/config/default/config.el`), reindentation of the current line does not occur.  Note that the implementation of `newline-and-indent` explicitly disables `electric-indent-mode` around its call of `newline`, disabling the "electric behavior" (i.e., reindenting the original line) of `newline`.

For example, in the following scenario (assuming `ada-ts-mode`), with Vanilla Emacs, pressing `RET` (at the end of the `X : Integer;` line) places point immediately below `X`, (since you might have multiple variable declarations you're writing).  So far, that is the same behavior as DoomEmacs.

```ada
declare
   X : Integer;
   <-- point is here.
```

With Vanilla Emacs, if I then type `begin` and press `RET`, the `begin` keyword will be reindented to the same level as the `declare` keyword before inserting a newline and indenting the empty line, and I end up with the following:

```ada
declare
   X : Integer;
begin
   <-- point is here.
```

With DoomEmacs, if I type `begin` and press `RET`, the `begin` keyword is not reindented and I end up with the following:

```ada
declare
   X : Integer;
   begin
      <-- point is here.
```

I assume DoomEmacs uses `newline-and-indent` over `newline` for some reason, but I wasn't really sure why.  I noticed that in the SML mode configuration (i.e., `module/lang/sml/config.el`) it was also remapping `RET` to `reindent-then-newline-and-indent`, and I assumed it was for this same reason.

Is there some better way to get the expected indentation behavior in DoomEmacs?

hlissner:
I wanted electric.el to handle this, perhaps with `+electric-indent-words` (from the `:emacs electric` module, perhaps by adding keywords like `begin` to it using `set-electric!` like we have [in the :lang sh module](https://github.com/doomemacs/doomemacs/blob/feature/tabspaces/modules/lang/sh/config.el#L20)) or `electric-indent-chars`, but I might've jumped the gun with my initial review — this is something I haven't formally (or fully) begun standardizing yet (why it was allowed to happen in the sml module).

Once v3 is done I'll revisit the matter, but for now using `reindent-then-newline-and-indent` is fine. Thanks for bringing it to my attention (and for the PR)!

tomekw:
Thank you @brownts! :blue_heart:

PR #8518, opened 2025-09-20
feat: add +respect-visual flag to :editor evil

larrasket:
Adds a new `+respect-visual` flag to the :editor evil module.
When enabled, commands that operate on visual regions will prefer
the active visual selection instead of point. This makes operator
and motion behavior more intuitive for users who rely on visual
mode.

### DIFF
--- a/modules/editor/evil/README.org
+++ b/modules/editor/evil/README.org
@@ -15,7 +15,11 @@ This holy module brings the Vim editing model to Emacs.
 - +everywhere ::
   Enable evilified keybinds everywhere possible. Uses the [[https://github.com/emacs-evil/evil-collection][evil-collection]] plugin
   as a foundation.
-
+- +respect-visual ::
+  Whether movement commands respect ~visual-line-mode~.If enabled, motions such
+  as j and k navigate by visual lines (on the screen) rather than "physical"
+  lines (defined by newline characters).  If nil, the setting of
+  ~visual-line-mode~ is ignored.
 ** Packages
 - [[doom-package:evil]]
 - [[doom-package:evil-args]]
--- a/modules/editor/evil/config.el
+++ b/modules/editor/evil/config.el
@@ -22,7 +22,7 @@ directives. By default, this only recognizes C directives.")
 (defvar evil-want-C-w-delete t)
 (defvar evil-want-Y-yank-to-eol t)
 (defvar evil-want-abbrev-expand-on-insert-exit nil)
-(defvar evil-respect-visual-line-mode nil)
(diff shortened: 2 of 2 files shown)


hlissner:
Thanks for the PR but I must turn it down. Module flags are not intended to be a substitute for setting trivial variables (except in rare cases where the flag represents a broader feature that *isn't* implementation specific). Thanks for bringing it to my attention, in any case!

larrasket:
Henrik,

https://github.com/doomemacs/doomemacs/commit/c27621a777c11354a4913c7eb455db3766984709
Is not a good solution either IMHO. You do not map all what evil does
map:

   (when evil-respect-visual-line-mode
     (evil-define-minor-mode-key 'motion 'visual-line-mode
       "j" 'evil-next-visual-line
       "gj" 'evil-next-line
       "k" 'evil-previous-visual-line
       "gk" 'evil-previous-line
       "0" 'evil-beginning-of-visual-line
       "g0" 'evil-beginning-of-line
       "$" 'evil-end-of-visual-line
       "g$" 'evil-end-of-line
       "V" 'evil-visual-screen-line))

Moreover, even if we update the code from that commit to match current
evil behavior, this add maintaibility overhead in case evil's
implementation changes, instead of leaving the responsability of
handling this variable totally to evil. 

I suggest to consider having this as a flag again. Or, at least, let's
find a more reliable solution.

Henrik Lissner ***@***.***> writes:

> *
> hlissner left a comment (doomemacs/doomemacs#8518)
>
> Thanks for the PR but I must turn it down. Module flags are not intended to be a substitute for setting trivial variables
> (except in rare cases where the flag represents a broader feature that isn't implementation specific). Thanks for bringing it
> to my attention, in any case!
>
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you authored the thread.

-- 
Whatever
Salih

hlissner:
> You do not map all what evil does map:

That isn't necessary. Evil handles [the rest](https://github.com/emacs-evil/evil/blob/master/evil-integration.el#L435-L445).

> Moreover, even if we update the code from that commit to match current
evil behavior, this add maintaibility overhead in case evil's
implementation changes, instead of leaving the responsability of
handling this variable totally to evil.

Why would we need to? This is meant to extend how Evil handles `evil-respect-visual-line-mode`, not replace it.

larrasket:
The variable `evil-respect-visual-line-mode` will not be respected. It must be set before evil is loaded.

If someone puts `evil-respect-visual-line-mode` in their config.el, I believe it won't be loaded before evil since we are loading it by the init.el.

PR #8523, opened 2025-09-21
feat: add retry mechanism for straight operations

NightMachinery:
## Summary
- Add monkey-patch for straight to retry failed operations up to 10 times
- This helps handle transient network failures when fetching packages
- Configurable retry count via `elpa-straight-retry-count` variable

## Test plan
- [ ] Test package installation with network interruptions
- [ ] Verify retry behavior works as expected
- [ ] Confirm no breaking changes to existing straight functionality

🤖 Generated with [Claude Code](https://claude.ai/code)


NightMachinery:
I forked this patch from https://www.d12frosted.io/posts/2021-04-08-straight-el-retries .

You might want to rename the functions and put them somewhere better.

This patch is essential for me, as I have around 500 packages and the internet here is rather unstable.

hlissner:
A more robust solution was needed, but needs testing on an actually unreliable network. Could you try [this gist out](https://gist.github.com/hlissner/166691ff4744933add72de619df4bd33) on your local system, replacing your current solution, and let me know if it works as you expect?

hlissner:
Just a heads up: I fixed a bug in the gist just now, so if you're already using it, it needs to be updated.

NightMachinery:
I tested this (the latest version just now) by restarting my proxy server to simulate a network failure, but it didn't work:
```
❯ doom sync -u 
LANG=en_US@calendar=persian.UTF-8 cannot be used, using en_US.UTF-8 instead.
- Using Emacs 29.2 @ /opt/homebrew/Cellar/emacs-plus@29/29.2/Emacs.app/Contents/MacOS/Emacs
> Synchronizing "default" profile...
  > Regenerating envvars file
    ✓ Generated ~/.emacs.d/.local/env
  > Ensuring packages are installed and built...
    - No packages need attention
  > Updating all packages (this may take a while)...
    x The package manager threw an error
    x Last 16 lines of straight's error log:
      [Return code: 0]
      $ cd /Users/evar/.emacs.d/.local/straight/repos/shrink-path.el/
      $ git log --oneline --no-merges c14882c8599aec79a6e8ef2d06454254bb3e1e41 \^c14882c8599aec79a6e8ef2d06454254bb3e1e41
      [Return code: 0]
      $ cd /Users/evar/.emacs.d/.local/straight/repos/ucs-utils/
      $ git rev-parse HEAD
      91b9e0207fff5883383fd39c45ad5522e9b90e65
      [Return code: 0]
      $ cd /Users/evar/.emacs.d/.local/straight/repos/ucs-utils/
      $ git config --get remote.origin.url
      https://github.com/rolandwalker/ucs-utils.git
      [Return code: 0]
      $ cd /Users/evar/.emacs.d/.local/straight/repos/ucs-utils/
      $ git fetch origin
      fatal: unable to access 'https://github.com/rolandwalker/ucs-utils.git/': Failure when receiving data from the peer
      [Return code: 128]
    ! Wrote extended straight log to ~/.emacs.d/.local/state/logs/cli.doom.250922135303.85448.error
✓ Finished in 5.76625s
```

NightMachinery:
I also notice that the new code doesn't have a sleep before retry. I think having that will help, as network failures tend to be temporally close. Perhaps even an exponential backoff sleep could work.

NightMachinery:
Sth else I noticed; `print!` seems to eat the last line of output. E.g., in the log I pasted above, we cannot see this line anymore:
```
    > (5/483) Fetching pcache...
```

hlissner:
After some testing of my own (`systemd-run --user --property=PrivateNetwork=yes --same-dir --pty` is quite useful), I've zeroed in on the solution in 4de1621. Give that a try and let me if know if it gives you any issues. In any case, thank you for bringing this to my attention!

PR #8529, opened 2025-09-25
feat: upload add functions to register buffer-local ssh-deploy mapping

stfl:
Added functions to register a local file to a remote file and set this mapping as a buffer-local variable.
Keep track of buffer-local mappings via a global variable that can be cleared at once. This is handy if the remote is gone and not coming back and we want to remove several mappings at once.
The universal arg C-u is used to remove the mapping for a local file.

I would also like to add key maps. I currently have the following but since the module does not register mappings I was unsure where to put this.

```lisp
(map! :map ssh-deploy-menu-map
        :leader
        :prefix "r"
        "l" #'+upload-register-mapping
        "L" #'+upload-clear-mappings)
 ```

### DIFF
--- a/modules/tools/upload/config.el
+++ b/modules/tools/upload/config.el
@@ -50,4 +50,55 @@
         (unless ssh-deploy-root-local
           (setq ssh-deploy-root-local (doom-project-root)))
         (when ssh-deploy-automatically-detect-remote-changes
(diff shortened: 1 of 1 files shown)


stfl:
Thank you for adopting and improving on my implementation. <3 doom

PR #8547, opened 2025-10-18
fix: set-formatter! align with documentation

peterhoeg:
The documentation for `set-formatter!` states:

    MODES is a major mode, a list thereof, or a list of two-element sublists with
    the structure: (MAJOR-MODE FORM).
    
It however, assumed it would a list and would blow up when passed a single major mode.

I figured it was easier to fix the code than adjust the documentation...


hlissner:
Thanks for your help!

PR #8552, opened 2025-10-22
fix(org): reconcile ob-async with new org, lazy loading

aisamu:
#### fix(org): check :async on already loaded languages
Running a regular/`:sync` source block would prevent successive
invocations of blocks of that same language with `:async` from loading
`ob-async`.

#### fix(org): don't export noweb-cache-related var
Org added a global cache for noweb expansion that includes a buffer
object. Those aren’t readable by the Lisp reader across processes, so
when async.el serializes the parent environment and hits that cons
`'(#<buffer *new*> . 739 )`, the child later tries to read it and errors
with `Invalid read syntax: "#<"'`

Fix: https://github.com/astahlman/ob-async/issues/99

-----

### DIFF
--- a/modules/lang/org/config.el
+++ b/modules/lang/org/config.el
@@ -298,11 +298,11 @@ Also adds support for a `:sync' parameter to override `:async'."
   "Load babel libraries lazily when babel blocks are executed."
   (defun +org--babel-lazy-load (lang &optional async)
     (cl-check-type lang (or symbol null))
+    ;; ob-async has its own agenda for lazy loading packages (in the child
+    ;; process), so we only need to make sure it's loaded.
+    (when async
+      (require 'ob-async nil t))
     (unless (cdr (assq lang org-babel-load-languages))
-      (when async
-        ;; ob-async has its own agenda for lazy loading packages (in the child
-        ;; process), so we only need to make sure it's loaded.
-        (require 'ob-async nil t))
       (prog1 (or (run-hook-with-args-until-success '+org-babel-load-functions lang)
                  (require (intern (format "ob-%s" lang)) nil t)
                  (require lang nil t))


modules/lang/org/config.el
```diff
@@ -291,11 +293,12 @@ Also adds support for a `:sync' parameter to override `:async'."
   "Load babel libraries lazily when babel blocks are executed."
   (defun +org--babel-lazy-load (lang &optional async)
     (cl-check-type lang (or symbol null))
+    ;; ob-async has its own agenda for lazy loading packages (in the child
+    ;; process), so we only need to make sure it's loaded.
+    (when async
+      (unless (featurep 'ob-async)
```
hlissner: Remove `(unless (featurep 'ob-async) ...)` check. It's redundant with `require`, which won't load the package if it is already loaded.

modules/lang/org/config.el
```diff
@@ -215,7 +215,9 @@ Is relative to `org-directory', unless it is absolute. Is used in Doom's default
   ;; won't likely complete in time, and will instead output an ob-async hash
   ;; instead of the wanted evaluation results.
   (after! ob
-    (add-to-list 'org-babel-default-lob-header-args '(:sync)))
+    (add-to-list 'org-babel-default-lob-header-args '(:sync))
+    (after! async
+      (add-to-list 'async-inject-variables-exclude-regexps "\\`org-babel-expand-noweb-references--cache-buffer\\'")))
```
hlissner: Please replace this with:

```diff
   (after! ob
     (add-to-list 'org-babel-default-lob-header-args '(:sync)))
-    (add-to-list 'org-babel-default-lob-header-args '(:sync))
-    (after! async
-      (add-to-list 'async-inject-variables-exclude-regexps "\\`org-babel-expand-noweb-references--cache-buffer\\'")))
+    (add-to-list 'org-babel-default-lob-header-args '(:sync)))
+
+  (defadvice! +org--exclude-expand-noweb-references-a (fn &rest args)
+    :around #'ob-async-org-babel-execute-src-block
+    (let ((async-inject-variables-exclude-regexps
+           (cons "\\`org-babel-expand-noweb-references--cache-buffer\\'"
+                 async-inject-variables-exclude-regexps)))
+      (apply fn args)))
```

This isolates the effect to `ob-async`'s interactions with `ob`, avoiding any potential side-effects in unrelated uses down the road.

hlissner (changes requested):

modules/lang/org/config.el
aisamu: That was the first implementation, but a quick-and-dirty benchmark showed one or two orders of magnitude difference!

The absolute values were still small (something like 1s vs 0.1 for 10000 invocations) but I didn't want to be the guy known for slowing down Doom!

aisamu:
```emacs-lisp
(require 'benchmark)
(require 'ob-async nil 't) 

(let ((N 200000))
  (message "featurep xN: %S" (benchmark-run N (unless (featurep 'ob-async) t)))
  (message "require  xN: %S" (benchmark-run N (require  'ob-async nil 't))))
  
featurep xN: (0.030702 0 0.0)
require  xN: (2.2516190000000003 0 0.0)
```
The above also holds for `benchmark-run-compiled`

Surprisingly, if I run the above via `eval-buffer` instead of `eval-last-sexp`, I get no difference! 
Color me confused.

aisamu:
Done, thanks for the suggestions!
If both are OK, give me the 👌🏻 and I'll rebase/clean-up the branch

aisamu:
Done. Will only rebase from now on

hlissner:
Sorry for taking so long to get to this. Thanks for your help, in any case!

PR #8567, opened 2025-11-05
fix: off by one when switching workspaces

JasZhe:
The workspaces in the echo area (via `+workspaces/display` for instance) are 1-indexed and I believe the intention is that `5 gt` for example, takes me to the 5th workspace rather than the 6th.

But that actually takes me to the 6th workspace since the actual internal workspace list is 0-indexed (via `+workspace-list-names` used by `+workspace/switch-to`).

The `max` here is to just avoid funny behaviour if we set the prefix-arg to <1 (i.e. 0 or negative, which is easier to just go to the first workspace) If we wanted to swap X workspaces relative to the current one then +workspace/switch-left +workspace/switch-right would be more appropriate

-----

### DIFF
--- a/modules/ui/workspaces/autoload/evil.el
+++ b/modules/ui/workspaces/autoload/evil.el
@@ -30,10 +30,10 @@
 (evil-define-command +workspace:switch-next (&optional count)
   "Switch to next workspace. If COUNT, switch to COUNT-th workspace."
   (interactive "<c>")
-  (if count (+workspace/switch-to count) (+workspace/cycle +1)))
+  (if count (+workspace/switch-to (max (1- count) 0)) (+workspace/cycle +1)))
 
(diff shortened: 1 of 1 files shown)


wchmb:
Hi, I'm sorry, but I disagree.
Simple maths.
If you are in the 1st seat and move 5 times to the next seat, then you should end up in the 6th seat.

JasZhe:
Yeah that's the behaviour we get for `+workspace/switch-right` or `+workspace/switch-left` which does the "seat moving" as you say, but I think the intention of these tow commands `+workspace:switch-next` `+workspace:switch-previous` is to behave like vim where it jumps to Xth seat i.e. it doesn't matter what seat I'm in now, I just want to go to the 2nd seat or the 5th seat. Right now, I'm off by one, and jump to the 6th seat when I want to go to the 5th seat.

PR #8581, opened 2025-11-16
perf(unicode): drastically reduce initial startup time

kisaragi-hiu:
font-utils.el disable its cache on X and Windows so every check for the existence of a font recomputes the font list, which takes a non-negligible amount of time, at least on X. This default is reasonable as stated in [the docstring of font-utils-use-memory-cache](https://github.com/rolandwalker/font-utils/blob/master/font-utils.el#L175):

> Disabled on X11 and MS Windows by default, because font-family-list
> often gives truncated results before Emacs is fully initialized.

This is, however, unreasonable from unicode-fonts, where calls to check the existence of fonts happens in a loop; during the loop, Emacs is not going to suddenly become fully initialized, so each call is actually recomputing the same font list over and over again.

Enabling the font-utils memory cache, at least for the duration of unicode-fonts-setup, thus drastically reduces the initial startup time, or the startup time when the pcache store is reset for some reason.

On my device, this makes it so a fresh `unicode-fonts-setup` (delete ~/.config/emacs/.local/cache/pcache then restart Emacs) takes only about 3 seconds, instead of the 30+ seconds before.

-----

### DIFF
--- a/modules/ui/unicode/autoload.el
+++ b/modules/ui/unicode/autoload.el
@@ -8,6 +8,10 @@ necessary."
     (setq-default bidi-display-reordering t)
     (+unicode-setup-fonts-h (selected-frame))))
 
+;; From font-utils.el, required by unicode-fonts.el.
+;; This only marks the variable as special locally.
+(defvar font-utils-use-memory-cache)
+
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
LGTM! Thanks for your help!

PR #8582, opened 2025-11-19
fix(cc-mode): Wrap tree-sitter remap hack in conditional

mlaves:
C/C++ files open in tree-sitter modes (e.g., `c++-ts-mode`) even when the `:tools tree-sitter` module is **not** enabled in `init.el`. Interestingly, this only happens for the second file (first file opens in `c++-mode`, all other in `c++-ts-mode`).

Root cause: Doom unconditionally removes cc-mode's protective "null entries" from `major-mode-remap-defaults`.

Fix: Only remove the null entries when the `+tree-sitter` flag is enabled.

This is an issue because in `c++-ts-mode`:
- CC Mode settings are ignored: `c-basic-offset`, `c-offsets-alist`, `c-default-style`,
- Custom indentation styles don't apply, and
- Hooks like `c++-mode-hook` don't run.

-----

### DIFF
--- a/modules/lang/cc/config.el
+++ b/modules/lang/cc/config.el
@@ -45,8 +45,9 @@ This is ignored by ccls.")
   ;; HACK: cc-mode adds null entries to `major-mode-remap-defaults', which
   ;;   overrides our tree-sitter remappings, causing the first remap to succeed,
   ;;   but future remaps to fail.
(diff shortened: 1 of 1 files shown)


hlissner (approved):

hlissner:
Thanks for your help!

PR #8609, opened 2025-12-24
tweak: remove ligatures from Elixir

katafrakt:
This removes default settings for extra ligatures in Elixir module.

The choice of the ligatures here feels quite arbitrary and sometimes wrong. For example:
* `def` is covered, but `defp` is not
* `use` has a special ligature, but it does not have any special meaning in Elixir that would justify that
* `return` has a ligature, but there is no `return` in Elixir
* these ligatures also show up in comments and doc strings, making them less readable
Screenshot 2025-12-23 at 21 59 54

In general, I don't feel like these ligatures from "hard" functional programming are establish enough to warrant having them as a default config. Anyone interested can set them themselves.

-----

### DIFF
--- a/modules/lang/elixir/config.el
+++ b/modules/lang/elixir/config.el
@@ -9,19 +9,6 @@
 ;;; Packages
 
 (defun +elixir-common-config (mode)
-  (set-ligatures! mode
-    ;; Functional
-    :def "def"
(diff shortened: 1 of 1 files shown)


hlissner:
LGTM! Thanks for your help!

PR #8613, opened 2025-12-30
fix(macos): fullscreen correctly

verhovsky:
This setting makes Emacs behave [differently from every other app](https://github.com/d12frosted/homebrew-emacs-plus/issues/861#issue-3739702257) on macOS and it has been annoying me for 5 years.

If the issue is that you don't want emacs to start in fullscreen (which in my opinion is good behavior, when have you ever wanted to use a little 56x24 character box to edit text

Screenshot 2025-12-29 at 22 12 17

instead of your full screen? I have had this in my config.el

```emacs
;; start in fullscreen
(add-to-list 'initial-frame-alist '(fullscreen . fullboth))
```

since a few months after you added this line 5 years ago) then this is surely not the way to configure this.

### DIFF
--- a/modules/os/macos/config.el
+++ b/modules/os/macos/config.el
@@ -11,10 +11,6 @@
 ;;
 ;;; Compatibilty fixes
 
-;; Curse Lion and its sudden but inevitable fullscreen mode!
-;; This is meaningless to railwaycat's emacs-mac build though.
-(setq ns-use-native-fullscreen nil)
-
 ;; Visit files opened outside of Emacs in existing frame, not a new one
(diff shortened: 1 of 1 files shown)


hlissner:
That setting was a layaway from when I was a MacOS user, *long* ago; I *hated* native fullscreen's UI animation, isolating Emacs in its own workspace, and it monopolizing not only that display, but all other connected displays. This was back in the Lion days, so the UX must have improved since. In any case, I no longer use MacOS, so I leaned into the consistency-with-other-apps argument and merged this. Thanks for the help!

PR #8615, opened 2025-12-30
fix(org): bump org-roam

dustinfarris:
A recent breaking change in org-roam was reverted.  Doom users had reported impact on their workflows and latest main should resolve these issues.


hlissner:
The whole module was overdue a larger bump so I've included this bump in 60b20b4. Thanks for bringing it to my attention, in any case!

PR #8649, opened 2026-02-10
fix(clojure): Enable CIDER for clojurec and clojurescript

prashantvithani:
Extends CIDER support to `clojurec-mode` and `clojurescript-mode` by matching the existing behavior for `clojure-mode`.


hlissner (approved):

hlissner:
Thanks again for your help!

Issue #8601, opened 2025-12-09
org-capture script creates frames under X11 despite PGTK Emacs running on Wayland


nairadithya:
### Describe your issue

I have bound the org-capture shell script to a hotkey. When triggered, I receive this warning:
> You are trying to run Emacs configured with the "pure-GTK" interface under the X Window System. That configuration is unsupported and will lead to sporadic crashes during transfer of large selection data. It will also lead to various problems with keyboard input.

**Expected behavior:**  The org-capture frame should open on Wayland, consistent with normal Emacs frames which work correctly under Wayland.
**Actual behavior:** The org-capture frame attempts to open under XWayland, triggering the PGTK/X11 incompatibility warning.

### Steps to reproduce

- Install Emacs compiled with PGTK support for Wayland and setup Doom Emacs. Usually distributed as `emacs-wayland`(AUR) or `emacs-pgtk`(DNF)
- Run a Wayland compositor (e.g., Sway, Hyprland, GNOME Wayland)
- Ensure XWayland is running (this provides the DISPLAY variable that triggers the issue)
- Run the command: `exec ~/.config/emacs/bin/org-capture`
- The X11/PGTK compatibility issue displays on the newly created frame.

### System information

https://pastebin.com/hSgT070D

### Disclosures

- [ ] This issue was written by AI.

- [ ] I used AI extensively to investigate this issue.

- [x] My issue was observed on a stable and supported release of GNU Emacs (must not end in `.50`, `.60`, or `.90`–`.99`, e.g. `30.0.50`).

- [x] A [backtrace](https://discourse.doomemacs.org/t/85) is provided for each error I've reported (if any).

nairadithya:
My current fix is to unset the DISPLAY variable in the script. This makes the newly created frame operate under Wayland but I am unsure of whether it might break other things.

```sh
if ! emacsclient --suppress-output --eval nil 2>/dev/null; then
  unset DISPLAY
  export GDK_BACKEND=wayland
  emacs --daemon
  trap cleanup EXIT INT TERM
fi
```

This was obtained by adding an extra line to the original check, written below:

```sh
if ! emacsclient --suppress-output --eval nil 2>/dev/null; then
  export GDK_BACKEND=wayland
  emacs --daemon
  trap cleanup EXIT INT TERM
fi
```

real-or-random:
I'm pretty sure that this is not a doom issue.

Quoting myself from #8652:
[...]

PR #8652, opened 2026-02-13
fix(bin): org-capture uses Wayland when available

kovan:
## Summary
- Set `GDK_BACKEND=wayland` before starting the temporary Emacs daemon when `WAYLAND_DISPLAY` is set
- Without this, PGTK Emacs falls back to X11 via XWayland when `DISPLAY` is also present, triggering the "pure-GTK interface under X Window System" incompatibility warning

## Approach
- Only applies when the script starts its own temporary daemon (not when connecting to an existing one)
- Only activates on Wayland (`$WAYLAND_DISPLAY` check)
- Does **not** `unset DISPLAY`, which could break XWayland apps launched from Emacs

Fix: #8601

---

### DIFF
--- a/bin/org-capture
+++ b/bin/org-capture
@@ -16,6 +16,11 @@ cleanup() {
 # If emacs isn't running, we start a temporary daemon, solely for this window.
 if ! emacsclient --suppress-output --eval nil 2>/dev/null; then
   echo "No Emacs daemon/server is available! Starting one..."
+  # FIX(#8601): Ensure PGTK Emacs uses Wayland when available, rather than
+  #   falling back to X11 via XWayland (which triggers incompatibility warnings).
+  if [ -n "$WAYLAND_DISPLAY" ]; then
+    export GDK_BACKEND=wayland
+  fi
   emacs --daemon
   trap cleanup EXIT INT TERM
 fi


real-or-random:
I wonder if this should qualify as a doom issue at all. If a `emacs --daemon` with a pgtk installation runs an X11 instance, then there should be something wrong in the user's config. 

I don't get this warning when I run `emacs --daemon` with Emacs 30.2 on Wayland. I have WAYLAND_DISPLAY and DISPLAY set. GDK_BACKEND is unset.

kovan:
The issue is specific to Doom's `bin/org-capture` script, which starts its own temporary daemon. When both `WAYLAND_DISPLAY` and `DISPLAY` are set (the default on any Wayland session with XWayland — which is most setups), GTK picks X11 unless `GDK_BACKEND` is explicitly set. Users don't typically set `GDK_BACKEND` globally because most GTK apps handle this fine — it's PGTK Emacs that's uniquely sensitive to this.

So the user's config isn't wrong — having both `WAYLAND_DISPLAY` and `DISPLAY` set is normal. The fix just ensures the daemon that Doom's own script spawns uses the right backend.

real-or-random:
> When both `WAYLAND_DISPLAY` and `DISPLAY` are set (the default on any Wayland session with XWayland — which is most setups), GTK picks X11 unless `GDK_BACKEND` is explicitly set. Users don't typically set `GDK_BACKEND` globally because most GTK apps handle this fine — it's PGTK Emacs that's uniquely sensitive to this.

This exactly contradicts what I've reported above. I have `WAYLAND_DISPLAY` and `DISPLAY` set, and `GDK_BACKEND` unset. When I run `emacs --daemon -Q` (with PGTK Emacs 30.2), then I get a Wayland connection and no X11 connection. I verified that when I connect to it via `emacsclient`, I get a Wayland frame. 
 
> The issue is specific to Doom's `bin/org-capture` script, which starts its own temporary daemon.

How so? What's special about the way it starts a daemon?


I really would appreciate a reply that hasn't been generated by an LLM.

kovan:
You're right — I tested this on WSL2 with PGTK Emacs 30.2 (WAYLAND_DISPLAY=wayland-0, DISPLAY=:0, GDK_BACKEND unset) and the daemon correctly connects via Wayland without any intervention. Can't reproduce the X11 fallback.

@nairadithya could you share your Emacs version (`M-x emacs-version`), GTK version, and distro? This might be version-specific.

Closing for now since it's not reproducible on 30.2.

Issue #8650, opened 2026-02-11
+popup/toggle should restore the last closed popup


sgleizes:
### Describe your issue

After quitting a popup with `+popup/quit-window`, I would expect that using `+popup/toggle` would restore that popup, but it always shows the *Messages* buffer, which makes this command hard to rely on.

From the doc of `+popup/toggle`, I understand that this is expected behavior:
```
Toggle any visible popups.

If no popups are available, display the *Messages* buffer in a popup window.
```
But, I think this command is not very helpful if it doesn't handle the usecase of restoring the last closed popup. Since popup buffers seem to stay alive after quitting their window, it seems possible to restore them.

For now I have redefined `+popup/quit-window` with the following to quickly support this:
```
(defun +popup/quit-window (&optional arg)
  "The regular `quit-window' sometimes kills the popup buffer and switches to a
buffer that shouldn't be in a popup. We prevent that by remapping `quit-window'
to this commmand."
  (interactive "P")
  (let ((orig-buffer (current-buffer)))
    (when (and +popup--remember-last (+popup-buffer-p))
      (+popup--remember (list (selected-window))))
    (quit-window arg)
    (when (and (eq orig-buffer (current-buffer))
               (+popup-buffer-p))
      (+popup/close nil 'force))))
```
But this is likely not the cleanest / most compatible way.

What do you think?

### Steps to reproduce

1. With the `(popup +all +defaults)` module enabled, open a helpful buffer with e.g. `helpful-callable`.
2. Quit the popup window with `+popup/quit-window` (`q`).
3. Run +popup/toggle` using M-x or a binding.
The *Messages* buffer displays, whereas I would expect that the closed popup be restored.

### System information

https://pastebin.com/BpsbMXJP

### Disclosures

- [ ] This issue was written with/by AI.

hlissner:
As of 29bc91b this should be resolved. Let me know if that isn't the case and I'll reopen it. Thanks for bringing it to my attention!

sgleizes:
Thanks for the quick fix!

Unfortunately in my case it doesn't work for popups where I have changed the rules with `:quit 'current`.
I have this in my config:
```
  (set-popup-rules!
    ;; Do not quit buffers in `help-mode', `helpful-mode' from other windows (with ESC).
    '(("^\\*\\([Hh]elp\\|Apropos\\)"
       :slot 2 :vslot -8 :size 0.42 :select t :quit 'current)))
```

[...]

PR #8653, opened 2026-02-13
fix(popup): +popup/toggle restores last quit popup

kovan:
## Summary
- `+popup/quit-window` (bound to `q`) calls `quit-window`, which destroys the popup window before `+popup/close` can save its state via `+popup--remember`
- This means `+popup/toggle` can never restore the last closed popup and always falls back to `*Messages*`
- Fix: call `+popup--remember` **before** `quit-window` runs, so the window state is saved while the window is still live

## How it works
1. `+popup/quit-window` → `quit-window` deletes the window → `+popup/close` fallback condition is false → state never saved
2. With this fix: save state first → `quit-window` → `+popup/toggle` can now restore

Fix: #8650

---

### DIFF
--- a/modules/ui/popup/+hacks.el
+++ b/modules/ui/popup/+hacks.el
@@ -44,6 +44,10 @@ buffer that shouldn't be in a popup. We prevent that by remapping `quit-window'
 to this commmand."
   (interactive "P")
   (let ((orig-buffer (current-buffer)))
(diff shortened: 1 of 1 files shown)


hlissner:
Rather than wrap around `quit-window`, I simply remap to `+popup/close` directly in 29bc91b, which should fix this issue. Thanks for bringing it to my attention!

Issue #8572, opened 2025-11-07
$env:EMACS returns incorrectly formatted path and causes SPC H R R crash


Gael-Lopes-Da-Silva:
### I confirm that...

- [x] I searched the issue tracker and found no similar issues.

- [x] I searched the [documentation](https://docs.doomemacs.org), [FAQ](https://docs.doomemacs.org/-/faq), [discussion board](https://doomemacs.org/discuss), and [Google](https://google.com) for solutions to my issue.

- [x] I read [the debugging guide](https://doomemacs.org/d/how2debug) and researched the issue to the best of my ability.

- [x] My issue can be observed on the latest available commit of Doom.

- [x] My issue can be observed on a stable release of Emacs (i.e. doesn't end in `.50`, `.60`, or `.90`–`.99`)

### Describe your issue

On Windows 11, trying to run doom related command like SPC H R R return an error.

<img width="1920" height="345" alt="Image" src="https://github.com/user-attachments/assets/c660f6e3-d16f-45cc-a4cf-fc283165e429" />

Trying to debug, I found that the $env:EMACS variable return this:

<img width="1920" height="341" alt="Image" src="https://github.com/user-attachments/assets/652d0e9e-83ac-4760-918d-5a6fbe466288" />

Trying to only use Executable-Find "emacs.exe" work. So $env:EMACS path is incorrectly formatted.

### Steps to reproduce

1. Install doom emacs (On Windows 11)
2. Open emacs
3. Do SPC H R R to run doom sync

### System information

https://pastebin.com/CUMiGKM0

hlissner:
I need someone better versed in powershell to look at this. doom.ps1 worked fine in win10, so not sure if this is a win11 thing or a powershell issue and how to ensure backwards compatibility.

hlissner:
With 3ad9eeb I'm hoping this should be adequately resolved, at least until a powershell dev can suggest a better alternative. Let me know if the issue resurfaces and I'll reopen it. Thanks for bringing it to my attention!

PR #8654, opened 2026-02-13
fix(lib): don't shell-quote EMACS env var

kovan:
## Summary
- Remove `shell-quote-argument` from the `EMACS` environment variable in `doom-run-doom-command!`
- `shell-quote-argument` produces cmd.exe-style quoting on Windows, but env vars are raw strings — PowerShell can't parse the quoted path, breaking `doom/reload` (SPC h r r) and other commands that invoke `bin/doom`
- The other env vars in the same block (`EMACSDIR`, `DOOMDIR`, `DOOMLOCALDIR`) are already unquoted

Fix: #8572

---

### DIFF
--- a/lisp/lib/config.el
+++ b/lisp/lib/config.el
@@ -46,7 +46,7 @@
      (with-current-buffer
          (with-environment-variables
              (("PATH" (string-join exec-path path-separator))
-              ("EMACS" (shell-quote-argument emacs-bin))
+              ("EMACS" emacs-bin)
               ("EMACSDIR" doom-emacs-dir)
               ("DOOMDIR" doom-user-dir)
               ("DOOMLOCALDIR" doom-local-dir)


hlissner:
Shell-quoting is still necessary on Linux/MacOS, because bin/doom's uses `eval` [here](https://github.com/doomemacs/doomemacs/blob/master/bin/doom#L13) to allow multi-argument values for `$EMACS` (necessary for users who've installed Emacs via Snap or Flatpak). This is not a perfect solution. I'd rather someone more powershell-savvy look at [this line in bin/doom.ps1](https://github.com/doomemacs/doomemacs/blob/master/bin/doom.ps1#L32) and suggest an `eval` equivalent for interpolating `$emacs` more literally, but since Snap and Flatpak don't exist on Windows (and convoluted emacs.exe placements that don't involve WSL are more rare there), I'll settle for only escaping spaces in `$EMACS` on Windows, for now (as I've done in 3ad9eeb).

In any case, thanks for bringing this to my attention!

Issue #8543, opened 2025-10-12
dir-locals-set-class-variables  and indent-bars-mode


kvaneesh:
### I confirm that...

- [x] I searched the issue tracker and found no similar issues.

- [x] I searched the [documentation](https://docs.doomemacs.org), [FAQ](https://docs.doomemacs.org/-/faq), [discussion board](https://doomemacs.org/discuss), and [Google](https://google.com) for solutions to my issue.

- [x] I read [the debugging guide](https://doomemacs.org/d/how2debug) and researched the issue to the best of my ability.

- [x] My issue can be observed on the latest available commit of Doom.

- [x] My issue can be observed on a stable release of Emacs (i.e. doesn't end in `.50`, `.60`, or `.90`–`.99`)

### Describe your issue

This is related to https://github.com/doomemacs/doomemacs/issues/8499#issuecomment-3364432651 . I think the problem is we are enabling indent-bars-mode, before run-hooks got a chance to setup the dir-local variables. I am observing that  hack-dir-local-variables() get run after indent-bars-mode() is run and because of that indent-bars--guess-spacing() finds the value of `indent-bars-spacing` as 2 instead of what is set in dir-locals-set-class-variables

### Steps to reproduce

with the recreate mentioned in issue [8499](https://github.com/doomemacs/doomemacs/issues/8499)  enable indent-guides with a different indent-offset. 

### System information

https://paste.centos.org/view/309909b6

hlissner:
As of b468dae this should be resolved (so that indent-bars is activated on `after-change-major-mode-hook`, which fires after dir/file local variables are set). Let me know if that isn't the case and I'll reopen it. Thanks for bringing it to my attention!

PR #8656, opened 2026-02-13
fix(indent-guides): activate after dir-locals are applied

kovan:
## Summary
- `indent-bars-mode` activated on `prog-mode-hook`, but inside `run-mode-hooks` the hook ordering is: mode hooks → `hack-local-variables` → `after-change-major-mode-hook`
- This meant `indent-bars--guess-spacing` read default spacing instead of dir-local overrides (e.g. `c-basic-offset` set via `dir-locals-set-class-variables`)
- Move activation to `after-change-major-mode-hook` where dir-locals are guaranteed to be applied, with a `derived-mode-p` guard for prog/text/conf modes
- Runs exactly once, at the right time

Fix: #8543

---

### DIFF
--- a/modules/ui/indent-guides/config.el
+++ b/modules/ui/indent-guides/config.el
@@ -14,12 +14,18 @@ be enabled. If any function returns non-nil, the mode will not be activated."
 
 (use-package! indent-bars
   :unless noninteractive
-  :hook ((prog-mode text-mode conf-mode) . +indent-guides-init-maybe-h)
+  ;; FIX(#8543): Use `after-change-major-mode-hook' instead of mode-specific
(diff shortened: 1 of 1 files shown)


hlissner:
I opted for a slightly different approach in b468dae, so that:

- The package is deferred until the latest possible time during startup (i.e. waiting for the first buffer that is interactively opened by the user). `after-change-major-mode-hook` triggers *very* often during startup and would result in `indent-bars-setup` being called much too soon.
- `indent-bars-mode` is activated less indiscriminately (the ties to text-mode/prog-mode/etc weren't important, they simply casted the widest net for a non-global minor mode — which was imperfect because some modes or special buffers aren't derived from those three).

Issue #8521, opened 2025-09-21
undo-tree does not close buffers


captainepoch:
### I confirm that...

- [x] I searched the issue tracker and found no similar issues.

- [x] I searched the [documentation](https://docs.doomemacs.org), [FAQ](https://docs.doomemacs.org/-/faq), [discussion board](https://doomemacs.org/discuss), and [Google](https://google.com) for solutions to my issue.

- [x] I read [the debugging guide](https://doomemacs.org/d/how2debug) and researched the issue to the best of my ability.

- [x] My issue can be observed on the latest available commit of Doom.

- [x] My issue can be observed on a stable release of Emacs (i.e. doesn't end in `.50`, `.60`, or `.90`–`.99`)

### Describe your issue

When you try to undo changes with undo-tree, it creates a buffer than it cannot be closed. And, if you do another undo, it creates another, but it never closes the buffers.

Related to #8198.

Happens on GNU/Linux as well.

### Steps to reproduce

1. Setup DOOM Emacs to use undo-tree.
2. Open a file.
3. Make changes.
4. Try to undo those changes.

If you do the same steps, that diff buffer isn't reused, it creates another.

### System information

https://pastebin.com/PUsm46rT

Issue #8198, opened 2024-12-09
undo-tree does not close buffers


captainepoch:
### I confirm that...

- [x] I have searched the [issue tracker](https://github.com/doomemacs/doomemacs/issues), [documentation](https://docs.doomemacs.org), [FAQ](https://docs.doomemacs.org/-/faq), [Discourse](https://discourse.doomemacs.org), and [Google](https://google.com), in case this issue has already been reported/resolved.

- [x] I have read "[How to Debug Issues](https://doomemacs.org/d/how2debug)", and will use it to provide as much information about this issue as possible.

- [x] The issue can be reproduced on the **latest** available commit of Doom.

- [x] The issue can be reproduced on a stable release of Emacs, such as 27, 28, or 29. *(Unstable versions end in .50, .60, or .9x)*

### Expected behavior

When you undo (or redo) changes, the buffer created showing those changes should be closed along the undo-tree one.

### Current behavior

It creates two buffers: one showing the tree and another showing the diff, but it only destroys the one containing the tree.

### Steps to reproduce

When you try to undo changes:

 - It creates two buffers: one showing the diff and another showing a tree with one or different branches.
 - When you finish undoing those changes, press twice `q` and the tree buffer closes, but not the one showing the diff.

If you do the same steps, that diff buffer isn't reused, it creates another.

This is also for redoing changes

![Image](https://github.com/user-attachments/assets/9bed4fad-3a0b-4cda-a371-9a5e22796761)

![Image](https://github.com/user-attachments/assets/8f7bdb66-2149-4ae6-8d64-1cd3e9de9e6b)

### System Information

MacOS 13.7.1 Darwin 22.6.0 x86_64 ns (Emacs 29.4, latest doom-emacs available).

hlissner:
Did you find a solution to this?

captainepoch:
I did not, but since nobody said anything for a month, I consider it's not a relevant issue.

hlissner:
So you are still experiencing the issue? If so, could you provide:

1. More specific steps to reproduce your issue. I cannot reproduce it. `M-x undo-tree-visualize` followed by `q` cleans up both buffers as expected.
2. The output of `M-x doom/info` or `~/.config/emacs/bin/doom info`.

captainepoch:
Still happening, in both Linux (NixOS) and macOS (Ventura using Emacs 24.9). I updated Doom Emacs this morning just to see if it was an old issue, but still happens.

[...]

PR #8661, opened 2026-02-13
fix(undo): move undo-tree diff advice to correct use-package block

kovan:
## Summary
- undo-tree creates diff buffers that never close, piling up on repeated undo operations
- Root cause: the `defadvice!` overriding `undo-tree-visualizer-show-diff` (commit `fabce333e`, fixing #8198) was placed inside `(use-package! undo-fu ...)` which has `:unless (modulep! +tree)` — so it only loads when undo-tree is NOT active. The fix literally never runs.
- Move the advice to `(use-package! undo-tree ...)` where it belongs
- Also add an override for `undo-tree-visualizer-update-diff` that removes the `balance-windows` call which conflicts with Doom's popup system (the root cause identified in the HACK comment)

Fix: #8521

## Test plan
- [ ] Enable `(emacs undo +tree)` in init.el
- [ ] Open a file, make changes, invoke `undo-tree-visualize`
- [ ] Navigate the undo tree — diff window should appear and update correctly
- [ ] Quit the visualizer — diff buffer should be cleaned up
- [ ] Repeat multiple times — no orphaned diff buffers should accumulate

---

### DIFF
--- a/modules/emacs/undo/config.el
+++ b/modules/emacs/undo/config.el
@@ -21,24 +21,7 @@
               (define-key map (kbd "C-x r U") #'undo-fu-session-recover)
               map)
     :init-value nil
-    :global t)
-
-  ;; HACK: If undo-tree creates its diff window next to a popup/side window, the
-  ;;   `balance-window' calls in `undo-tree-visualizer-update-diff' can wreck
-  ;;   havoc on the window tree, making the diff window an unclosable "root"
-  ;;   window (which emacs will happily throw errors about when you call
-  ;;   `undo-tree-visualizer-quit'). Breakage ensues.
-  ;; REVIEW: Should be reported/addressed upstream, in undo-tree!
-  (defadvice! +undo-tree--show-visualizer-diff-safely-a (&optional node)
-    :override #'undo-tree-visualizer-show-diff
-    (setq undo-tree-visualizer-diff t)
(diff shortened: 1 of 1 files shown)


modules/emacs/undo/config.el
```diff
@@ -117,4 +100,29 @@
...
+      (set-window-buffer win buff)
+      (shrink-window-if-larger-than-buffer win)))
+
+  (defadvice! +undo-tree--update-visualizer-diff-safely-a (&optional node)
+    :override #'undo-tree-visualizer-update-diff
+    (with-current-buffer undo-tree-visualizer-parent-buffer
+      (undo-tree-diff node))
+    (let ((win (get-buffer-window undo-tree-diff-buffer-name)))
+      (when win
+        (shrink-window-if-larger-than-buffer win)))))
```
hlissner: Please simplify (and rename) this advice to:

```elisp
(defadvice! +undo-tree--suppress-balance-windows-a (fn &rest args)
  :around #'undo-tree-visualizer-update-diff
  (letf! ((#'balance-windows #'ignore))
    (apply fn args))
```

This way it's a little more resistant to unexpected changes upstream (and the new name better reflects its purpose).

hlissner:
LGTM! Thanks again!

Issue #8546, opened 2025-10-17
eglot: on-save checks, and fights between flymake and flycheck


fosskers:
### I confirm that...

- [x] I searched the issue tracker and found no similar issues.

- [x] I searched the [documentation](https://docs.doomemacs.org), [FAQ](https://docs.doomemacs.org/-/faq), [discussion board](https://doomemacs.org/discuss), and [Google](https://google.com) for solutions to my issue.

- [x] I read [the debugging guide](https://doomemacs.org/d/how2debug) and researched the issue to the best of my ability.

- [x] My issue can be observed on the latest available commit of Doom.

- [x] My issue can be observed on a stable release of Emacs (i.e. doesn't end in `.50`, `.60`, or `.90`–`.99`)

### Describe your issue

This is after a decent amount of experimentation and searching.

It doesn't seem possible to tell eglot to do diagnostic checks only on-save. I only switched to eglot yesterday to experiment with `basedpyright` (which wasn't working with `lsp-mode`), and these old flycheck settings were still laying around:

```emacs-lisp
(after! flycheck
  (setq flycheck-check-syntax-automatically '(save mode-enabled)))
```

Basically, this breaks eglot (through interaction with `flycheck-eglot`?), making it necessary to save the buffer multiple times for diagnostics to accurately reflect the current reality of the buffer. Manually turning flymake on-and-off, nor setting `(setq flymake-no-changes-timeout nil)` seems to work.

If I remove the above `flycheck` settings, sanity is restored and diagnostics reflect reality, but only automatically through the usual idle timer. I'd rather they be on-save only.

### Steps to reproduce

1. Set `(lsp +glot)`.
2. Add the above `flycheck` settings.
3. Open a language buffer that uses an LSP.
4. Edit some things and observe lagging diagnostics.

### System information

https://pastebin.com/pRmqKpjD

hlissner:
Does this fix the issue? (from #8662)

```lisp
;;; add to $DOOMDIR/config.el
(defadvice! +lsp--flycheck-eglot-prefer-diagnostics-a (fn diags &rest args)
  :around #'flycheck-eglot--report-eglot-diagnostics
  (let ((flycheck-check-syntax-automatically
         (append '(idle-change new-line) flycheck-check-syntax-automatically)))
    (apply fn diags args)))
```

PR #8662, opened 2026-02-13
fix(lsp): ensure eglot diagnostics bypass flycheck trigger gate

kovan:
## Summary
- With eglot + flycheck (no `+flymake`), setting `flycheck-check-syntax-automatically` to `(save mode-enabled)` breaks diagnostics — requires multiple saves for them to catch up
- Root cause: `flycheck-eglot` hard-codes `(idle-change new-line)` as trigger conditions when pushing LSP diagnostics to flycheck. When the user restricts the allowed conditions, those triggers are blocked and diagnostics are stored but never displayed
- Fix: add `:around` advice on `flycheck-eglot--report-eglot-diagnostics` that temporarily appends `(idle-change new-line)` to `flycheck-check-syntax-automatically`. Push-based LSP diagnostics always get through; the user's setting is preserved for all other checkers

Fix: #8546

## Test plan
- [ ] Enable `(tools lsp +eglot)` and `:checkers syntax` (without `+flymake`)
- [ ] Set `flycheck-check-syntax-automatically` to `(save mode-enabled)` in user config
- [ ] Open a file with a working LSP server
- [ ] Introduce an error and save — diagnostics should appear after a single save
- [ ] Fix the error and save — diagnostics should clear after a single save
- [ ] Non-LSP checkers should still respect the save-only setting

---

### DIFF
--- a/modules/tools/lsp/+eglot.el
+++ b/modules/tools/lsp/+eglot.el
@@ -85,4 +85,19 @@ server an expensive restart when its buffer is reverted."
 
 (use-package! flycheck-eglot
   :when (modulep! :checkers syntax -flymake)
-  :hook (eglot-managed-mode . flycheck-eglot-mode))
+  :hook (eglot-managed-mode . flycheck-eglot-mode)
+  :config
+  ;; HACK: `flycheck-eglot' hard-codes `(idle-change new-line)' as trigger
+  ;;   conditions when pushing LSP diagnostics to flycheck. If the user
+  ;;   restricts `flycheck-check-syntax-automatically' to e.g. `(save
+  ;;   mode-enabled)', those conditions are blocked and diagnostics are never
+  ;;   displayed. Temporarily expand the allowed conditions during eglot's
+  ;;   push-based diagnostic reporting so LSP results always get through.
+  ;; TODO: Remove after flycheck-eglot is bumped to include
+  ;;   flycheck/flycheck-eglot#17, which removes the condition argument
+  ;;   entirely.
+  (defadvice! +lsp--flycheck-eglot-prefer-diagnostics-a (fn &rest args)
+    :around #'flycheck-eglot--report-eglot-diagnostics
+    (let ((flycheck-check-syntax-automatically
+           (append '(idle-change new-line) flycheck-check-syntax-automatically)))
+      (apply fn args))))


modules/tools/lsp/+eglot.el
```diff
@@ -85,4 +85,16 @@ server an expensive restart when its buffer is reverted."
...
   :when (modulep! :checkers syntax -flymake)
-  :hook (eglot-managed-mode . flycheck-eglot-mode))
+  :hook (eglot-managed-mode . flycheck-eglot-mode)
+  :config
+  ;; HACK: `flycheck-eglot' hard-codes `(idle-change new-line)' as trigger
+  ;;   conditions when pushing LSP diagnostics to flycheck. If the user
+  ;;   restricts `flycheck-check-syntax-automatically' to e.g. `(save
+  ;;   mode-enabled)', those conditions are blocked and diagnostics are never
+  ;;   displayed. Temporarily expand the allowed conditions during eglot's
+  ;;   push-based diagnostic reporting so LSP results always get through.
```
hlissner: This seems like an issue that should be addressed upstream. Please add this annotation:

```elisp
  ;; REVIEW: PR this upstream!
```

hlissner (changes requested):

kovan:
Upstream PR submitted: https://github.com/flycheck/flycheck-eglot/pull/17

hlissner (approved):

kovan:
**Update:** The upstream fix was merged at flycheck/flycheck-eglot#17 — it removes the condition argument from `flycheck-buffer-automatically` entirely, so push-based eglot diagnostics are unconditional.

This advice remains useful as a stopgap until Doom bumps `flycheck-eglot` to include that fix. Once bumped, the advice becomes a no-op and can be removed.

Worth noting (from the upstream maintainer): the user's original goal of save-only diagnostics can't be achieved through `flycheck-check-syntax-automatically` — eglot diagnostics are push-based by design. To suppress real-time diagnostics, users would need to configure it at the LSP server level or via `eglot-report-progress` / server-specific settings. We may want to document this in the checkers/syntax or lsp module.

hlissner:
That got merged quicker than expected. In which case I've gone ahead and bumped eglot-flymake in 8c5c6b5. Thanks for bringing it to my attention (and for the PR upstream)!

Issue #8391, opened 2025-05-14
Can't interrupt running code with C-c C-c in Python shell


yonghuier:
### I confirm that...

- [x] I have searched the [issue tracker](https://github.com/doomemacs/doomemacs/issues), [documentation](https://docs.doomemacs.org), [FAQ](https://docs.doomemacs.org/-/faq), [Discourse](https://discourse.doomemacs.org), and [Google](https://google.com), in case this issue has already been reported/resolved.

- [x] I have read "[How to Debug Issues](https://doomemacs.org/d/how2debug)", and will use it to provide as much information about this issue as possible.

- [x] The issue can be reproduced on the **latest** available commit of Doom.

- [x] The issue can be reproduced on a stable release of Emacs, such as 27, 28, 29, or 30. *(Unstable versions end in .50, .60, or .9x)*

### Expected behavior

The `C-c C-c` command should send an interrupt (SIGINT) to the Python process and stop the execution.

### Current behavior

Nothing happens; the Python shell becomes unresponsive.

I have to manually kill the buffer or restart Emacs.

### Steps to reproduce

1. run program in python shell
2. push C-c C-c 
3. doesn't work

### System Information

https://pastebin.com/XikgRXJh

hlissner:
I can't reproduce this. Some things to try:

- Interrupt Emacs itself by executing `pkill -USR2 emacs` in the terminal. This should (hopefully) force Emacs to stop what it's doing and produce a backtrace. That should tell us where it's hanging.
- Use `C-h v` to inspect the value of the following variables and report them to me:
  - `python-interpreter`
  - `python-interpreter-args`
  - `python-shell-interpreter`
  - `python-shell-interpreter-args`

yonghuier:
I used command `pkill -USR2 /opt/homebrew/Cellar/emacs-plus@30/30.1/Emacs.app/Contents/MacOS/Emacs`, but it didn't work. My variables setup as follow:

- `python-interpreter: "python"`
- `python-interpreter-args: ""`
- `python-shell-interpreter: "python3"`
- `python-shell-interpreter-args: "-i"`

hlissner:
> I used command pkill -USR2 /opt/homebrew/Cellar/emacs-plus@30/30.1/Emacs.app/Contents/MacOS/Emacs, but it didn't work.

[...]

PR #8665, opened 2026-02-13
fix(python): disable pyrepl to fix C-c C-c interrupt in REPL

kovan:
## Summary
- `C-c C-c` (`comint-interrupt-subjob`) fails to interrupt running Python code in the inferior REPL on Python 3.13+, particularly on macOS
- Root cause: Python 3.13's pyrepl mishandles SIGINT under `TERM=dumb` (which comint sets). The `^C` character is treated as literal input rather than triggering an interrupt, causing `SyntaxError: invalid non-printable character U+0003`
- Fix: set `PYTHON_BASIC_REPL=1` in `python-shell-process-environment` to disable pyrepl and use the classic readline-based REPL. This is the same approach VS Code uses ([vscode-python#25164](https://github.com/microsoft/vscode-python/issues/25164))
- Safe to apply universally — on systems where pyrepl works fine, the classic REPL is functionally equivalent under `TERM=dumb`

Fix: #8391

## Test plan
- [ ] Open a Python file, run `M-x run-python`
- [ ] Execute an infinite loop: `while True: print(1)`
- [ ] Press `C-c C-c` — the loop should be interrupted
- [ ] Verify normal REPL usage works (eval region, send buffer, etc.)

---

### DIFF
--- a/modules/lang/python/config.el
+++ b/modules/lang/python/config.el
@@ -70,6 +70,13 @@
              (executable-find "python3"))
     (setq python-shell-interpreter "python3"))
 
+  ;; HACK: Python 3.13's pyrepl mishandles SIGINT under Emacs's comint
+  ;;   (TERM=dumb), particularly on macOS. The ^C character is treated as
+  ;;   literal input rather than triggering an interrupt signal. Disabling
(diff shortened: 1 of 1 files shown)


hlissner:
LGTM! Thanks again!

Issue #1908, opened 2019-10-16
[REQUEST] Make next-error sane


rgrinberg:
I've discussed this on discord with @hlissner so I'm just repeating things here so that we don't forget about this issue:

The current behavior of `next-error` (`] e`), is to look for a buffer with an error (in practice this ends up being either a flychecked or compilation-mode buffer) and then jump to the next error in this buffer. There's a couple of problems with this behavior. Both are related to the fact that the location of your cursor after the jump is unpredictable. The issues are:

* Sometimes one expects a local jump that should occur only within the current buffer. What we really want here is `next-error-in-buffer`.

* If one is looking for an error in any buffer, the error should still be opened in the same window we called `next-error` from. Currently, `next-error` may find the error in a different window, if the buffer in that window contains an error.

Proposed fixes:

* Have a command to cycle through errors only in the current buffer

* Fix `next-error` to never jump to another window.

hlissner:
As of c174b25 this should be resolved. Let me know if that isn't the case and I'll reopen it. Thanks for bringing it to my attention!

PR #8671, opened 2026-02-13
fix(evil): make ]e/[e cycle errors in current buffer first

kovan:
## Summary

- Add `+evil/next-error` and `+evil/previous-error` wrappers that cycle errors within the current buffer when `next-error-function` is set (flycheck, flymake, compilation), and fall back to cross-buffer `next-error` wrapped in `save-selected-window` to prevent cursor from jumping to another window
- Rebind `]e` / `[e` to use these new wrappers instead of raw `next-error` / `previous-error`

Fix: #1908

## Context

Currently `]e` / `[e` map directly to `next-error` / `previous-error`, which has two problems:
1. `next-error` may jump the cursor to a different window if that window's buffer has errors
2. There's no way to cycle errors only in the current buffer

The new wrappers check `next-error-function` — when set (by flycheck, flymake, or compilation-mode), errors cycle within the current buffer. When unset, the default cross-buffer behavior is preserved but wrapped in `save-selected-window` to prevent window-jumping.

## Test plan

- [ ] Open a file with flycheck/flymake errors — `]e` / `[e` cycles through errors in that buffer
- [ ] Open a compilation buffer in a split — `]e` from a non-error buffer shows the error but does NOT move cursor to the compilation window
- [ ] Verify count prefix works (e.g. `3]e` jumps 3 errors ahead)
- [ ] Verify `[e` navigates in reverse direction

---

### DIFF
--- a/modules/editor/evil/autoload/unimpaired.el
+++ b/modules/editor/evil/autoload/unimpaired.el
@@ -75,6 +75,24 @@ See `+evil/next-preproc-directive' for details."
   (interactive "p")
   (+evil/next-comment (- count)))
 
+;;; ]e / [e
+;;;###autoload
+(defun +evil/next-error (count)
+  "Jump to the next error in the current buffer.
+Falls back to `next-error' if the current buffer has no error source."
+  (interactive "p")
+  (if next-error-function
+      (next-error count)
+    (save-selected-window
+      (next-error count))))
+
+;;;###autoload
+(defun +evil/previous-error (count)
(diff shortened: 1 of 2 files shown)


hlissner:
Between https://github.com/doomemacs/doomemacs/commit/c174b25d5749317535c5bc7bd729f692274e141b and https://github.com/doomemacs/doomemacs/commit/dc0fc9d4f46dca2ad5af8870ade9ee3d7da69d67 this should be resolved. Thanks for bringing it to my attention!

Issue #6027, opened 2022-01-22
Support xdg-desktop-portal dark mode


A6GibKm:
### Describe your request

Recently a new [setting](https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.Settings) was added to xdg-desktop-portal to declare the user preference for dark mode. This setting is DE-agnostic¹ and it is sandbox friendly.

How to implement: Emacs has to read via dbus the value of the setting and subscribe to changes, then set the value of the GtkSetting property [gtk-application-prefer-dark-theme](https://docs.gtk.org/gtk3/property.Settings.gtk-application-prefer-dark-theme.html) so that the headerbar of emacs follows the preference, and optionally set the `doom-theme` to something that makes sense.

1. it will be officially suported in GNOME 42, elementary OS already supports it and there is a merge request for KDE that will most likely included in the next release.

### Briefly explain its use-case

Avoid this:

![image](https://user-images.githubusercontent.com/11528996/150654444-bb09fc29-e193-4a49-bc28-a97a66367b8b.png)

A6GibKm:
The following snippet can be used to read and listen the dbus interface.

```lisp
(defun handler (a)
  (message "current value %s" (car (car a))))

(defun signal-handler (namespace key value)
  (if (and
       (string-equal namespace "org.freedesktop.appearance")
       (string-equal key "color-scheme"))
      (message "got new value: %s" (car value)) nil))

(dbus-call-method-asynchronously
 :session
 "org.freedesktop.portal.Desktop"
 "/org/freedesktop/portal/desktop"
 "org.freedesktop.portal.Settings"
 "Read"
 #'handler
 "org.freedesktop.appearance"
 "color-scheme")

(dbus-register-signal
 :session
 "org.freedesktop.portal.Desktop"
 "/org/freedesktop/portal/desktop"
 "org.freedesktop.portal.Settings"
 "SettingChanged"
 #'signal-handler)
```

real-or-random:
See also https://github.com/LionyxML/auto-dark-emacs

Issue #8119, opened 2024-10-21
Incompatible with packages using `enable-theme` (e.g., auto-dark-mode) in daemon mode


real-or-random:
### I confirm that...

- [x] I have searched the [issue tracker](https://github.com/doomemacs/doomemacs/issues), [documentation](https://docs.doomemacs.org), [FAQ](https://docs.doomemacs.org/-/faq), [Discourse](https://discourse.doomemacs.org), and [Google](https://google.com), in case this issue has already been reported/resolved.

- [x] I have read "[How to Debug Issues](https://doomemacs.org/d/how2debug)", and will use it to provide as much information about this issue as possible.

- [x] The issue can be reproduced on the **latest** available commit of Doom.

- [ ] The issue can be reproduced on a stable release of Emacs, such as 27, 28, or 29. *(Unstable versions end in .50, .60, or .9x)* (I'm guilty, I haven't tested but really, it's sufficiently clear that this is not the issue...)

### Expected behavior

Doom should reload the theme stored in the outdated `doom-theme` available, after a theme has been enabled via `enable-theme`.

### Current behavior

In, doom-ui.el, we advice for `load-theme`, which sets `doom-theme`. In daemon mode, whenever a frame is created (`'server-after-make-frame-hook`), doom reloads the theme `doom-theme`  So far, so good, and this works with any packages that use `load-theme` to change the theme, e.g., auto-dark-mode.

However, this won't work if some package uses `enable-theme`, which is a recent optimization in auto-dark-mode:
 https://github.com/LionyxML/auto-dark-emacs/blob/1b6c1841c5b15388ee78cb03a4139fd14f42783b/auto-dark.el#L250-L253

### Steps to reproduce

1. Start emacs in daemon mode.
2. Create a frame in (say) in dark mode. Auto-dark-mode will `load-theme` the dark theme, `doom-theme` is set to the dark theme.
3. Switch system to light mode. Auto-dark-mode will `load-theme` the light theme, `doom-theme` is set to the light theme.
4. Switch system to dark mode. Auto-dark-mode will `enable-theme` the dark theme, `doom-theme` is left untouched.
5. Create a frame. Doom will `load-theme` the `doom-theme`, which is the light theme. This is not what we want.

### Possible fixes

Fixing this seems tricky. I'm not sure what the right/best fix is and who should fix it (will report to auto-dark-mode in a minute!) Here are some ideas:

[...]

PR #8673, opened 2026-02-13
module(auto-dark): add :ui auto-dark module

kovan:
## Summary

- Add new `:ui auto-dark` module that integrates the [`auto-dark`](https://github.com/LionyxML/auto-dark-emacs) package with Doom's theme system
- Enables automatic light/dark theme switching based on OS appearance settings
- Supports Linux (D-Bus/xdg-desktop-portal), macOS (AppleScript), Windows (registry), and Android (Termux)

## Context

When the OS dark mode preference changes (e.g. GNOME/KDE scheduled dark mode, macOS auto-appearance), Emacs doesn't react — `doom-theme` is set once and never changes. The `auto-dark` package handles cross-platform detection with real-time D-Bus signal listeners on Linux, but needs integration glue to work properly with Doom's theme infrastructure.

The module overrides `auto-dark--enable-themes` to use `enable-theme` (faster, no flash) instead of `load-theme` (full reload) for already-loaded themes, and syncs `doom-theme` on switches so `doom/reload-theme` stays correct.

### Usage

Enable in `init.el`:
```elisp
(doom! ...
       :ui
       auto-dark
       ...)
```

Configure in `config.el`:
```elisp
(setq auto-dark-themes '((doom-one)          ;; dark theme(s)
                          (doom-one-light)))   ;; light theme(s)
```

Fix: #6027

## Test plan

- [ ] Enable `:ui auto-dark` module, set `auto-dark-themes` with a dark/light pair
- [ ] On Linux: toggle system dark mode via `gsettings set org.gnome.desktop.interface color-scheme prefer-dark` / `prefer-light` — Emacs should switch in real-time
- [ ] `doom/reload-theme` reloads the correct (auto-selected) theme
- [ ] `doom-theme` variable reflects the active theme after each switch
- [ ] `solaire-mode` and other `doom-load-theme-hook` subscribers fire on each switch
- [ ] Daemon mode: new frames get the correct theme

---

### DIFF
--- a/modules/ui/auto-dark/config.el
+++ b/modules/ui/auto-dark/config.el
@@ -1,0 +1,32 @@
+;;; ui/auto-dark/config.el -*- lexical-binding: t; -*-
+
+(use-package! auto-dark
+  :hook (doom-load-theme . auto-dark-mode)
+  :config
+  ;; HACK: `auto-dark--enable-themes' disables all current themes then loads
+  ;;   the target themes from scratch via `load-theme'. Doom's `load-theme'
+  ;;   advice already handles enabling themes and running hooks, so this causes
+  ;;   redundant work and can flash. Override to only disable non-target themes
+  ;;   and enable (not reload) themes that are already loaded.
+  (defadvice! +auto-dark--use-doom-theme-loading-a (&optional themes)
+    "Use Doom's theme infrastructure when switching themes."
+    :override #'auto-dark--enable-themes
+    (let ((target-themes (remq 'user (delete-dups (or themes custom-enabled-themes)))))
+      ;; Disable themes not in the target set
+      (dolist (theme custom-enabled-themes)
+        (unless (memq theme target-themes)
+          (disable-theme theme)))
+      ;; Enable target themes, loading if necessary
+      (dolist (theme (reverse target-themes))
+        (if (custom-theme-p theme)
+            (enable-theme theme)
+          (load-theme theme t)))))
+
+  ;; Sync `doom-theme' when auto-dark switches themes, so that
+  ;; `doom/reload-theme' reloads the correct (current) theme.
+  (add-hook! '(auto-dark-dark-mode-hook auto-dark-light-mode-hook)
+    (defun +auto-dark-sync-doom-theme-h ()
+      "Update `doom-theme' to reflect auto-dark's current theme set."
+      (when-let* ((themes (auto-dark--themes-for-mode
+                           auto-dark--last-dark-mode-state)))
+        (setq doom-theme (if (cdr themes) themes (car themes)))))))
--- a/modules/ui/auto-dark/packages.el
+++ b/modules/ui/auto-dark/packages.el
@@ -1,0 +1,4 @@
+;; -*- no-byte-compile: t; -*-
+;;; ui/auto-dark/packages.el
+
+(package! auto-dark :pin "a71e791e47d09c5bf4bcbc2bbd7300b71ff72f1a")


real-or-random:
I've been using this package for a while, 

It's nice to see a PR for this, but I believe the way it's integrated in Doom is a bit too simplistic.

### Configuration

I think it's more natural to have built-in support in Doom for a light and a dark theme (i.e., without a module), e.g., by setting `doom-theme` to a list of two themes. There could still be an optional module that performs the switch automatically. On the other hand, this design would couple some Doom core functionality to an optional module. I'd be curious to hear what @hlissner thinks.

### `doom-theme` and `enable-theme`/`load-theme`

> The module overrides `auto-dark--enable-themes` to use `enable-theme` (faster, no flash) instead of `load-theme` (full reload) for already-loaded themes, and syncs `doom-theme` on switches so `doom/reload-theme` stays correct.

This has a history. You should look at https://github.com/doomemacs/doomemacs/issues/8119 and see if what this PR does matches the conclusion there. (I don't think so, but I'm not sure... I currently don't have the time to load all of this into my brain.) 

### `custom-safe-themes`

There's an issue with `auto-dark` and `doom` where one needs to set `custom-safe-themes` to `t`, see https://github.com/LionyxML/auto-dark-emacs/issues/64. I believe everyone should set this to `t` because the attacker model is misguided. If you have malicious code in your load path, then it's game over anyway. But this needs to be thought through carefully... And maybe there's a less intrusive fix, e.g., trust only those themes the user set explicitly in their config.

### When to activate `(auto-dark-mode)`

See https://github.com/LionyxML/auto-dark-emacs/issues/64#issuecomment-3157315461 for my current config and here for [an explanation](https://github.com/LionyxML/auto-dark-emacs/issues/64#issuecomment-3159840636). This does some logic to make sure that `(auto-dark-mode)` is postponed as much as possible. I needed this on Linux in daemon mode to make sure that DBus is available (to query the system dark/light status) when the mode is enabled.

kovan:
Closing this PR. As @real-or-random pointed out, hlissner's 50b9afb already addressed #8119 by moving to `enable-theme-functions`, which means both the `auto-dark--enable-themes` override and the `doom-theme` sync hook in this PR are redundant — upstream auto-dark already uses `enable-theme` for loaded themes, and Doom's `doom-enable-theme-h` already syncs `doom-theme` on any theme switch.

The remaining `custom-safe-themes` and DBus timing concerns also need more thought than a simple module wrapper can address. Better to let users configure auto-dark directly until there's a clearer design direction from upstream.

real-or-random:
> until there's a clearer design direction from upstream.

The package works, and the README suggests a working way to configure it in Doom Emacs. I don't think guidance from upstream is needed to integrate it here. It's just that someone needs to do the work. Or, if you think otherwise and you think that upstream can provide "design direction", then I suggest opening an issue in upstream asking a specific question.

kovan:
Fair point — by "upstream" I meant Doom's core, not auto-dark itself. To clarify: the specific reason for closing is that after reviewing hlissner's 50b9afb and the current `doom-enable-theme-h` hook, both overrides in this PR turned out to be unnecessary — Doom already handles `enable-theme` sync and auto-dark already uses `enable-theme` for loaded themes. So the PR was adding complexity for no benefit.

The remaining open questions (`custom-safe-themes`, DBus timing in daemon mode) are real but orthogonal — they'd need to be addressed in any auto-dark integration, and I didn't want to leave a half-baked module up. If someone wants to pick this up with those issues handled properly, the minimal module is really just `(use-package! auto-dark :hook (doom-load-theme . auto-dark-mode))` plus the package pin — no overrides needed.

PR #8683, opened 2026-02-15
fix(vertico): stop using deprecated consult sources

gagbo:
This allows a smoother bump of consult later

Ref: <https://github.com/minad/consult/commit/3a8ac46916c0699f0905ac4fe65d72768add3efb>


hlissner:
I went ahead and incorporated this into a bump of the vertico module (and the rest of `:completion`) in 5a9c459. Thanks for bringing it to my attention, in any case!

gagbo:
All good, I knew the PR was a bit lazy (since it did not bump the module), but at least that had the commit link there just in case

PR #8684, opened 2026-02-17
fix: silence straight package.el warning

Frestein:
Now it's definitely fixed.

Fix: https://github.com/doomemacs/doomemacs/issues/8568

-----

### DIFF
--- a/lisp/doom-straight.el
+++ b/lisp/doom-straight.el
@@ -55,10 +55,12 @@
       ;; no affect on packages that are pinned, however (run 'doom sync --gc' to
       ;; compact those after-the-fact). Some packages break when shallow cloned
       ;; (like magit and org), but we'll deal with that elsewhere.
-      straight-vc-git-default-clone-depth '(1 single-branch)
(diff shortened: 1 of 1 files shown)


real-or-random:
I know I asked you to PR this, but now that I see the diff here, I don't see how making the `setq` dependent on `(fboundp 'straight-use-recipes)` is supposed to fix the issue. (If anything, adding a condition should make the setq less likely to make an effect, no?)

hlissner:
This is indeed a post hoc error and will not fix the issue any more than the top-level `setq` already does. This is a load-order issue: `straight` is loaded before the variable is set (presumably because `doom-straight.el` is configured to load *after* `straight.el` loads, by which point it's too late). The most effective solution would be to set `straight-package--warning-displayed` earlier, like in `doom.el`, or [to suppress it](https://github.com/doomemacs/doomemacs/issues/8568#issuecomment-3917015796). I'd prefer the latter. I'll experiment with this further.

In any case, thank you for the PR, but I must turn it down.

Frestein:
This solves the problem for me. :/

PR #8367, opened 2025-04-19
feat(racket): allow toggling smart bracket mode

Jdogzz:
Adds a custom variable to allow disabling smart bracket mode.

Fix: #1452

-----
- [x] I searched the issue tracker and this hasn't been PRed before.
- [x] My changes are not on [the do-not-PR list](https://doomemacs.org/donotpr) for this project.
- [x] My commits conform to [Doom's git conventions](https://doomemacs.org/d/git-conventions).
- [ ] My changes are visual; I've included before and after screenshots.
- [ ] I am blindly checking these off.
- [x] Any relevant issues or PRs have been linked to.
- [ ] This a draft PR; I need more time to finish it.

PR #8694, opened 2026-02-26
feat(racket): add smart bracket mode toggle

### LINKED ISSUE #1452: In racket-mode, square brackets are automatically changed to parenthesis
In racket-mode, whenever I type a `[`, a `(` is automatically inserted. This is quite irritating, because actually I do want a `[` to be inserted. And currently the only way would be pasting one in.

I'm sure there probably is a configuration option (probably in smartparens) which would fix this. However, I have been unable to find such an option.

kovan:
Fixes #1452 — `racket-smart-open-bracket-mode` converts `[` to `(` with no way to disable it.

Adds `+racket-smart-open-bracket-mode` defvar (default `t`). Set to `nil` in `$DOOMDIR/config.el` to disable. Also fixes README which incorrectly documented the default. Acknowledges draft PR #8367.

**Verified** on Emacs 30.2: reproduced bracket conversion, confirmed `[` preserved when mode disabled.

---

### DIFF
--- a/modules/lang/racket/README.org
+++ b/modules/lang/racket/README.org
@@ -62,13 +62,19 @@ Formatting is handled using the [[doom-module::editor format]] module via [[http
 #+end_quote
 
 ** racket-smart-open-bracket-mode
-~racket-smart-open-bracket-mode~ gets turned off automatically if you use
-~parinfer~, ~lispy~. If you wish to enable it:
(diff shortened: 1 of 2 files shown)


hlissner:
Rather than add a new option altogether, I opted to simply remove the hook as a default for the module in 951d972, since it is a mote too magical for my liking and easier to opt into than out of. Thanks for bringing it to my attention, in any case!

PR #8699, opened 2026-03-01
fix(vertico): consult-register can load evil registers

liaowang11:
Fixes doomemacs/doomemacs#8698

The `:completion vertico` module advises Consult to list Evil registers via `evil-register-list`, which includes virtual registers like `"`, `1`-`9`, `*`, and `+`.

Only advising `consult-register--alist` makes these registers appear in `consult-register`, but selecting them fails during insertion/loading because `consult-register-load` eventually calls `insert-register`, which consults Emacs' real `register-alist`.

This PR applies the same advice to `consult-register-load` so selecting these registers inserts successfully.

### DIFF
--- a/modules/completion/vertico/config.el
+++ b/modules/completion/vertico/config.el
@@ -127,14 +127,26 @@ orderless."
     :before (list #'consult-recent-file #'consult-buffer)
     (recentf-mode +1))
 
-  (defadvice! +vertico--use-evil-registers-a (fn &rest args)
-    "Use `evil-register-list' if `evil-mode' is active."
-    :around #'consult-register--alist
-    (let ((register-alist
-           (if (bound-and-true-p evil-local-mode)
-               (evil-register-list)
-             register-alist)))
-      (apply fn args)))
+  (when (modulep! :editor evil)
+    (defadvice! +vertico--use-evil-registers-in-get-register-a (fn register &rest args)
+      "Use `evil-get-register' for Evil's special registers."
+      :around #'get-register
+      (if (and (bound-and-true-p evil-local-mode)
+               (fboundp 'evil-get-register)
+               (or (memq register '(?\" ?* ?+ ?% ?# ?/ ?: ?. ?- ?= ?_))
+                   (<= ?1 register ?9)))
+          (evil-get-register register t)
+        (apply fn register args)))
+
+    (defadvice! +vertico--use-evil-registers-a (fn &rest args)
+      "Use `evil-register-list' if `evil-local-mode' is active."
+      :around #'consult-register--alist
+      (let ((register-alist
+             (if (and (bound-and-true-p evil-local-mode)
+                      (fboundp 'evil-register-list))
+                 (evil-register-list)
+               register-alist)))
+        (apply fn args))))
 
   (setq consult-project-function #'doom-project-root
         consult-narrow-key "<"


hlissner:
How about simply advising `get-register`? Seems to be the root of all these issues, and the original goal *was* to make Evil's registers accessible in more places. Plus, `evil-register-list` is (enough of) a superset of `register-alist` to be its substitute.

liaowang11:
There are some changes but I'm no sure if you like this approach.
1. If we are trying to make "Evil's registers accessible in more places", maybe the advice should live in another place? `modules/editor/evil/config.el`?
2. If we advice `get-register`, there maybe recursion as `evil-get-register` relies on `get-register` so the advice have to check for the `special registers`.

hlissner:
After some experimentation, I zeroed in on the implementation in 88e2a0d, which ensures decent coverage for `register.el`, dodges a type error `evil-get-register` emits if `register` isn't a character, handles `=` (which prompts the user when retrieved), and handles cases where Evil isn't enabled in the current buffer (but enabled globally). (EDIT: And it gates all this behind `:editor (evil +everywhere)`, so it can be opted out of)

Give it a try and let me know if it gives you any trouble. Thanks for bringing it to my attention!

liaowang11:
> After some experimentation, I zeroed in on the implementation in [88e2a0d](https://github.com/doomemacs/doomemacs/commit/88e2a0d5dea807487d880b36e8afab30067deb5b), which ensures decent coverage for `register.el`, dodges a type error `evil-get-register` emits if `register` isn't a character, handles `=` (which prompts the user when retrieved), and handles cases where Evil isn't enabled in the current buffer (but enabled globally). (EDIT: And it gates all this behind `:editor (evil +everywhere)`, so it can be opted out of)
> 
> Give it a try and let me know if it gives you any trouble. Thanks for bringing it to my attention!

I have tested for a while and it works nicely. Thanks for your work.

sadrux:
> After some experimentation, I zeroed in on the implementation in [88e2a0d](https://github.com/doomemacs/doomemacs/commit/88e2a0d5dea807487d880b36e8afab30067deb5b), which ensures decent coverage for `register.el`, dodges a type error `evil-get-register` emits if `register` isn't a character, handles `=` (which prompts the user when retrieved), and handles cases where Evil isn't enabled in the current buffer (but enabled globally). (EDIT: And it gates all this behind `:editor (evil +everywhere)`, so it can be opted out of)
> 
> Give it a try and let me know if it gives you any trouble. Thanks for bringing it to my attention!

After this change, `jump-to-register` no longer works correctly. Specifically, after commit [88e2a0d](https://github.com/doomemacs/doomemacs/commit/88e2a0d5dea807487d880b36e8afab30067deb5b)￼, invoking `jump-to-register  ` fails to behave like the vanilla implementation from register.el.

hlissner:
@sadrux See if [d23bbe8](https://github.com/doomemacs/doomemacs/commit/d23bbe87721c61f4d5a605f2914b32780bb89949) fixes that.

sadrux:
> @sadrux See if [d23bbe8](https://github.com/doomemacs/doomemacs/commit/d23bbe87721c61f4d5a605f2914b32780bb89949) fixes that.

Yes, Now it works. Thank you for your work @hlissner

PR #8704, opened 2026-03-05
tweak(tree-sitter): use maintained systemverilog grammar

andrewpeck:
[tree-sitter-verilog](https://github.com/gmlarumbe/tree-sitter-verilog) has unfortunately not been maintained for years.

Gonzalo maintains [tree-sitter-systemverilog](https://github.com/gmlarumbe/tree-sitter-systemverilog) which boasts

- Full implementation of the latest SystemVerilog standard (IEEE 1800-2023)
- Robust and reliable: sv-tests results
- Actively maintained
- Implements node fields
- Supports parsing of code snippets (e.g., always block outside of a module)
- Basic preprocessing capabilities
- Thoroughly tested (~3000 tests) including:

verilog is strictly a subset of systemverilog so the systemverilog grammar is backwards compatible with verilog features. The same emacs major mode is used for both verilog and systemverilog.

tree-sitter-verilog has significant issues and fails parsing many basic verilog files.

I use tree-sitter-systemverilog daily and have had no issues since switching.

I'm not 100% sure if tweak is the right category for this commit; please let me know if that should change.

Thank you!

### DIFF
--- a/modules/tools/tree-sitter/config.el
+++ b/modules/tools/tree-sitter/config.el
@@ -129,7 +129,7 @@
                  (surface "https://github.com/connorlay/tree-sitter-surface" nil nil nil nil)
                  (toml "https://github.com/tree-sitter/tree-sitter-toml" nil nil nil nil)
                  (typst "https://github.com/uben0/tree-sitter-typst" "master" "src" nil nil)
-                 (verilog "https://github.com/gmlarumbe/tree-sitter-verilog" nil nil nil nil)
+                 (verilog "https://github.com/gmlarumbe/tree-sitter-systemverilog" nil nil nil nil)
                  (vhdl "https://github.com/alemuller/tree-sitter-vhdl" nil nil nil nil)
                  (vue "https://github.com/tree-sitter-grammars/tree-sitter-vue" nil nil nil nil)
                  (wast "https://github.com/wasm-lsp/tree-sitter-wasm" nil "wast/src" nil nil)


hlissner (approved):

hlissner:
LGTM! Thanks for your help!

modules/tools/tree-sitter/config.el
```diff
@@ -129,7 +129,7 @@
                  (surface "https://github.com/connorlay/tree-sitter-surface" nil nil nil nil)
                  (toml "https://github.com/tree-sitter/tree-sitter-toml" nil nil nil nil)
                  (typst "https://github.com/uben0/tree-sitter-typst" "master" "src" nil nil)
-                 (verilog "https://github.com/gmlarumbe/tree-sitter-verilog" nil nil nil nil)
+                 (verilog "https://github.com/gmlarumbe/tree-sitter-systemverilog" nil nil nil nil)
```
real-or-random, line 132: Shouldn't this be like this?
```suggestion
                 (systemverilog "https://github.com/gmlarumbe/tree-sitter-systemverilog" nil nil nil nil)
```

With just `verilog`, `treesit-install-language-grammar` fails for me. (See also https://github.com/gmlarumbe/tree-sitter-systemverilog/pull/29)

hlissner: verilog-ts-mode [expects it to be `verilog`](https://github.com/gmlarumbe/verilog-ts-mode/blob/master/verilog-ts-mode.el#L2826). The key to fixing this is likely [this `treesit-load-name-override-list` entry](https://github.com/gmlarumbe/verilog-ts-mode/blob/master/verilog-ts-mode.el#L2824), but it isn't set until the mode is activated, which isn't early enough for `set-tree-sitter!` (which tries to auto-detect and install the grammar at the major-mode-remap level, before *-ts-modes are activated). They really should be setting it at top-level...

Anyhow, see if this fixes it for you:

```elisp
;;; add to $DOOMDIR/config.el
(with-eval-after-load 'treesit
  (add-to-list 'treesit-load-name-override-list '(verilog "libtree-sitter-systemverilog" "tree_sitter_systemverilog")))
```

If it does, I'll add it to the tree-sitter module.

real-or-random: Ah, I don't even use verilog. I noticed this when trying to write a function that (re-)installs grammars for defined languages. But then I realized that's not super useful and settled on a function that updates/re-installs the installed grammars. I think that would be a nice addition to doom. Would you be interested in a PR?

---

But I looked into this verilog issue more and played around a bit. I'm pretty confident that what I suggested above is the right thing to do. In fact, verilog-ts-mode [does the same](https://github.com/gmlarumbe/verilog-ts-mode/blob/a804a3e8b744edeaf879705c85aae196f761cba8/verilog-ts-mode.el#L2766-L2774). This ensures that `treesit-install-language-grammar` will be able to install the grammar. Yes, verilog-ts-mode then depends on the `'verilog` grammar, but when the mode is enabled and wants to enable tree-sitter, the override will make sure that the right shared library is found.  And then it's okay that the override is only added when enabling the mode. 

Your suggested fix doesn't help. Then `(treesit-install-language-grammar 'verilog)` installs the file to `libtree-sitter-verilog.so`, and then loading it won't work. (Not with the override but also not without it: the file is named `libtree-sitter-verilog.so` but the exported symbol in it is `tree_sitter_systemverilog`.)    

> They really should be setting it at top-level...

My conclusion is that they shouldn't set it at all. The code comment cites an Emacs naming convention:

https://github.com/gmlarumbe/verilog-ts-mode/blob/a804a3e8b744edeaf879705c85aae196f761cba8/verilog-ts-mode.el#L2822-L2824

But as far as I understand the docs, that convention is for loading grammars only, i.e., it only says that the grammar for some language `'systemverilog` should be in a file `libtree-sitter-systemverilog.ext` with some specific exported symbol in it. But in no way, it says that a mode named `verilog-ts-mode` shouldn't depend on `'systemverilog`. I think the simplest thing would be to let `verilog-ts-mode` depend use  `'systemverilog` directly without any override/alias.

cc @gmlarumbe

gmlarumbe: Hi @real-or-random ,

Thanks for taking the time to do some research on this issue.

>  I think the simplest thing would be to let verilog-ts-mode depend use 'systemverilog directly without any override/alias.

You are right. It seems that `verilog-ts-mode` was depending on `'verilog` instead of on `'systemverilog` because during development it started using a fork of `tree-sitter-verilog`.

I have updated it in this latest [commit](https://github.com/gmlarumbe/verilog-ts-mode/commit/32d00f783711eb2f03615aa8d5739d5c9a50a7f9) and removed the override.

PR #8725, opened 2026-03-25
docs(config/use-package): add README

LiamMcCausland:
Adds a README.org for the :config use-package module.

Documents its purpose and Doom-specific behavior while following existing
module documentation style.

### DIFF
--- a/modules/config/use-package/README.org
+++ b/modules/config/use-package/README.org
@@ -1,0 +1,81 @@
+:PROPERTIES:
+:ID:       20010bc4-a661-4289-bb8b-8635329f83fb
+:END:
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but I must turn it down because the `:config use-package` module is temporary and [it's enabled state is hard-coded](https://github.com/doomemacs/doomemacs/blob/master/lisp/lib/modules.el#L74-L76) for backwards compatibility. Once v3 is published, the module (and Doom's modifications to `use-package` will be, at most, removed entirely, or moved to a compatibility module for folks who want more time to adapt their configs to v3's changes.

Thanks for the help, in any case!

PR #8728, opened 2026-03-25
revert: fix(lispy): evil-escape keybind collision on j

real-or-random:
evil-escape had been disabled in lispy because of a key binding conflict involving the evil-escape-key-sequence "jk". But the latter is nil since 68ec9e9c8600, so we can re-enable evil-escape.

Ref: 68ec9e9c8600
Revert: ccd20847c7ce

-----

### DIFF
--- a/modules/editor/lispy/config.el
+++ b/modules/editor/lispy/config.el
@@ -47,7 +47,4 @@
           additional
           additional-insert))
   :config
-  (lispyville-set-key-theme)
-  (add-hook! 'evil-escape-inhibit-functions
-    (defun +lispy-inhibit-evil-escape-fn ()
-      (and lispy-mode (evil-insert-state-p)))))
+  (lispyville-set-key-theme))


hlissner:
Shouldn't we keep this in case the user re-enables `evil-escape-key-sequence`?

real-or-random:
> Shouldn't we keep this in case the user re-enables `evil-escape-key-sequence`?

Hm, I had this thought, too, but I rejected it in the end:
 - I suspect users who deliberately set an `evil-escape-key-sequence` will actively use it and will rather want to change the lispyville binding to resolve the conflict. If your muscle memory makes you press `jk` a couple of times per minute, you will be very annoyed that it doesn't work in one particular mode. It would make me very annoyed for sure, though I don't have the conflict because I use `fd`.
 - There's also something to having the code base self-consistent: I assume that in a world where `evil-escape-key-sequence` was `nil` in Doom from the beginning, we wouldn't have added 68ec9e9c86009623ced5849343534bced2e6fcd3 in the anticipation that lispyville would conflict with someone setting the sequence to `jk`.
 - If someone sets `jk` and they want to disable it in lisp mode, they can still add the hook.

hlissner:
Fair enough. Thanks again for your help!

PR #8729, opened 2026-03-26
eglot support zig lsp

unship:
aglot already support zig lsp


hlissner:
I've rearranged the section in b1ce26f instead, which removes the line you removed. Thanks for bringing it to my attention!

Issue #8362, opened 2025-04-15
doomscript broken after "missing lexical binding" silencing change


marienz:
### I confirm that...

- [x] I have searched the [issue tracker](https://github.com/doomemacs/doomemacs/issues), [documentation](https://docs.doomemacs.org), [FAQ](https://docs.doomemacs.org/-/faq), [Discourse](https://discourse.doomemacs.org), and [Google](https://google.com), in case this issue has already been reported/resolved.

- [x] I have read "[How to Debug Issues](https://doomemacs.org/d/how2debug)", and will use it to provide as much information about this issue as possible.

- [x] The issue can be reproduced on the **latest** available commit of Doom.

- [x] The issue can be reproduced on a stable release of Emacs, such as 27, 28, 29, or 30. *(Unstable versions end in .50, .60, or .9x)*

### Expected behavior

Doomscripts (like the examples provided in docs/example.org) work.

### Current behavior

```
Error: end-of-file nil
  command-line-1(("--eval" "\"(setq" "warning-inhibit-types" "'((files" "missing-lexbind-cookie)))\"" "--load" "/home/marienz/src/doomemacs/early-init" "--load" "/tmp/doomscript" "--"))
  command-line()
  normal-top-level()
End of file during parsing
```

strace confirms doomscript invokes Emacs like:

```
execve("/etc/profiles/per-user/marienz/bin/emacs", ["emacs", "-q", "--no-site-file", "--batch", "--eval", "\"(setq", "warning-inhibit-types", "'((files", "missing-lexbind-cookie)))\"", "--load", "/home/marienz/src/doomemacs/early-init", "--load", "/tmp/doomscript", "--"], 0x2d315950 /* 80 vars */) = 0
```

(The double quotes in the value of `$emacs` after commit def4579a9f45bac90c0874b567b5225d03cf4135 do not have the intended effect: they are not stripped, and what is intended as a single argument to `--eval` is split.)

Note: this only affects `bin/doomscript`. `bin/doom` is fine.

If `bin/doomscript` was a bash script, I would have tried to fix this by converting the `emacs` variable from a string to an array. But my portable-shell-fu is not strong enough (or I'm just blanking on an obvious solution) to fix this cleanly for any `/bin/sh` without pulling the `--eval` out of `$emacs` and repeating it everywhere that variable is used.

[...]

PR #8743, opened 2026-04-04
fix(cli): doomscript: end-of-file error

marienz:
534b6403508136e6667a26b43c12c63f053f9b35 introduced the same regression previously introduced in bac90c0 (see #8362): it adds a quoted argument including whitespace to `$emacs`, but the contents of that variable are not evaluated by the shell. This means the quotes do not prevent word splitting and are not removed, resulting in Emacs getting invoked with separate arguments like `"(setq` and `warning-inhibit-types`:

```
$ bin/doomscript /tmp/doomscript

Error: end-of-file nil
  command-line-1(("--eval" "\"(setq" "warning-inhibit-types" "'((files" "missing-lexbind-cookie)))\"" "--load" "/home/marienz/src/doomemacs/early-init" "--load"
 "/tmp/doomscript" "--"))
  command-line()
  normal-top-level()
End of file during parsing
```

If we were using Bash we could fix this by using an array, but in portable shell our options are more limited.

Fortunately, `$emacs` is only used twice, and one of them does not need most of the arguments we are adding because it adds `-Q` and does not load Doom. So fix it by replacing that one with `$EMACS --batch` and inlining the rest of `$emacs` into its only remaining user.

Amend: 534b6403508136e6667a26b43c12c63f053f9b35

-----

### DIFF
--- a/bin/doomscript
+++ b/bin/doomscript
@@ -35,18 +35,6 @@ case "$EMACS" in
        ;;
 esac
 
-# Careful not to use -Q! It implies --no-site-lisp; it omits the site-lisp
-# directory from `load-path' which prevent Doom from manually loading the site
-# files later. Site files are important on some systems or deployment methods
-# (like Snap or NixOS).
-emacs="$EMACS -q --no-site-file --batch"
-
-# Emacs complains about missing lexical-binding settings in elisp files, but
-# this warning is unhelpful to end users, polluting the output of any shell
-# script derived from this script, so silence it (it's not enough to set it *in*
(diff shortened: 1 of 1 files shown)


hlissner:
I wanted to avoid quoting $EMACS, because I want to support commands with spaces (e.g. for users who install Emacs via Snap/Flatpak/Docker). Besides that (and moving the comments into `bin/doom`), inlining switches into the call sights made good sense, so I've addressed both ideas into 4a5046c. Thanks for your help and for bringing it to my attention!

PR #8753, opened 2026-04-16
fix(rss): off-by-one error in elfeed-search-selected

wchmb:
Hello, 
  
This is a bug I'm experiencing as evil user when I visual-select some entries. I didn't reproduce it in regular emacs using mark regions, sorry.
  
1. In evil visual line mode, when you select lines 5-7, emacs sets `region-end` to the beginning of line 8
  (the character after the newline of line 7).
2. elfeed-search-selected does [this](https://github.com/skeeto/elfeed/blob/master/elfeed-search.el#L756C8-L756C30)
3. Since `region-end` is at the beginning of line 8, `line-number-at-pos end` returns 8, so the loop iterates over lines 5, 6, 7 **and 8**

If I operate on selected lines, i.e marking them as read (u), I'd expect to operate on entries at lines 5-7 but it actually operates on entries at lines 5-**8**

-----
- [X] This PR contains AI-generated work.

### DIFF
--- a/modules/app/rss/config.el
+++ b/modules/app/rss/config.el
@@ -52,6 +52,23 @@ easier to scroll through.")
       shr-put-image-function #'+rss-put-sliced-image-fn
       shr-external-rendering-functions '((img . +rss-render-image-tag-without-underline-fn))))
 
+  (defadvice! +rss--elfeed-search-selected-fix-a (fn &optional ignore-region-p)
+    "Prevent off-by-one when region ends at the beginning of the next line"
+    :around #'elfeed-search-selected
+    (if (and (not ignore-region-p)
+             (use-region-p)
+             (let ((end (region-end)))
+               (and (> end (region-beginning))
+                    (= end (save-excursion
+                             (goto-char end)
+                             (line-beginning-position))))))
+        (let ((orig-region-end (symbol-function 'region-end))
+              (fixed-end (1- (region-end))))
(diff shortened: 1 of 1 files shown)


hlissner:
Does this fix still work if you substitute your advice with this?

```elisp
(defadvice! +rss--elfeed-search-selected-fix-a (fn &optional ignore-region-p)
  "Prevent off-by-one when region ends at the beginning of the next line"
  :around #'elfeed-search-selected
  (if (and (not ignore-region-p)
           (bound-and-true-p evil-local-mode)
           (doom-region-active-p)
           (let ((end (region-end)))
             (and (> end (region-beginning))
                  (save-excursion (goto-char end) (bolp)))))
      (letf! ((fixed-end (1- (region-end)))
              (defun region-end () fixed-end))
        (funcall fn nil))
    (funcall fn ignore-region-p)))
```

wchmb:
yes! your fix still working :)

only this warning `Malformed function ‘(fixed-end (1- (region-end)))’ ` from your macro letf!
that I cleaned refactoring it a bit

``` elisp
  (let ((fixed-end (1- (region-end))))
    (letf! (defun region-end () fixed-end)
      (funcall fn nil)))
```

hlissner:
After inspecting `elfeed-search-selected` I was able to shrink it a little more into what we have in b6bffa2. Thanks for bringing it to my attention!

PR #8758, opened 2026-04-24
feat(vterm): add eshell-like keybindings

lattarov:
Currently in vterm module:
- "d d" clears the tty input
- "V" highlights the whole line, including tty data
- "0" goes to the beginning of the line, past the tty data

Eshell seems to have a more intuitive behaviour, which is to avoid highlighting/going backwards past the tty/lambda data and "d d" clears the input, and this PR tries to port that to the vterm module:

I haven't found a way to do it with evil-collection, `(evil-collection-vterm-setup)` does not create this behaviour. Also, [evil-mode isn't yet supported in vterm upstream](https://github.com/akermu/emacs-libvterm/issues/313). #4373 was never merged but evil-collection does include most of those bindings.  

edit: add evil-collection and other PR linking paragraph

-----

### DIFF
--- a/modules/term/vterm/config.el
+++ b/modules/term/vterm/config.el
@@ -16,7 +16,21 @@
   :config
   (set-popup-rule! "^\\*vterm" :size 0.25 :vslot -4 :select t :quit nil :ttl 0)
 
-  (map! :map vterm-mode-map "C-q" #'vterm-send-next-key)
+  (map! :map vterm-mode-map "C-q" #'vterm-send-next-key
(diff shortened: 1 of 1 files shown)


hlissner:
The `V` keybinding was too rigid (not everyone's prompt has a `$` in it), so I omitted it in c44363f, but I've incorporated the other two commands into the module. I do see evil-collection has commands that *seem* like they should had this behavior, but I couldn't get them to work either. I'll look into it further when I have time.

PR #8760, opened 2026-04-27
fix(cli) properly escape ' in zsh completion help strings

kitt-cat:
help strings in zsh completions are not properly escaped, leading to errors. 

for example typing `doom -` and pressing <kbd>tab</kbd> prints:

```
_doom:351: unmatched `
```

this change escapes `'` as well as backslashing `[]:\` following
recommendations in zshcompsys(1) and empirical observations.

...i didn't actually notice any of these latter characters in doom's current
help strings, but i did add some to confirm the this change causes zsh to
render them as expected.

**future**: newlines in these strings are also not desirable, but this change
makes no attempt to address that  (zsh renders them as a literal `\n`).

-----

### DIFF
--- a/lisp/cli/make/completions.el
+++ b/lisp/cli/make/completions.el
@@ -93,9 +93,24 @@
                                 (if (and args (string-prefix-p "--" (caar switches)))
                                     "=" "")))
                       (format "'[%s]%s'"
-                              (replace-regexp-in-string "'" "''" (cdr option))
+                              (escape-help-string (cdr option))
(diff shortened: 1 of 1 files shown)


hlissner:
Thanks for the PR, but I must turn it down due to its timing. I'm in the middle of a hefty rewrite of Doom's CLI and these changes wouldn't survive the eventual merge. Sorry for the trouble, but thanks for the help, in any case (if a completion script generator makes it into v3, I'll factor the fix into it though)!

PR #8775, opened 2026-05-18
tweak: disable line numbers in minimal-UI buffers

wchmb:
According to [doom-ui.el#L598](https://github.com/doomemacs/doomemacs/blob/master/lisp/doom-ui.el#L598) Doom only enables `display-line-numbers-mode` for `prog-mode`, `text-mode`, and `conf-mode`. So in theory the gutter is never enabled in them yet it is displayed when entering from a line-numbered buffer, and it persists until the next input event triggers a redisplay.

I think this is a redisplay artifact: when a window transitions to a buffer whose `display-line-numbers` is nil, Emacs doesn't always recompute the window's left margin in time, and the previous buffer's gutter pixels stays. 

To be honest, I'm not sure this is the cleanest fix but I started by adding something like this to my config.el and it worked :)
```elisp
  ;; config.el
  (add-hook! '(vterm-mode-hook eshell-mode-hook magit-mode-hook term-mode-hook)
    #'doom-disable-line-numbers-h)
```

So I think the hooks in this commit are meaningful even for buffers that derive from `special-mode` and would never have enabled line numbers in the first place. I've followed the following criterion: if modeline is hidden here [bead7a](https://github.com/doomemacs/doomemacs/commit/bead7a77483d4b11404f408c4b218f6dc7c220ee) then line numbers are also useless. I hope I haven't left anyone out

-----

### DIFF
--- a/lisp/doom-ui.el
+++ b/lisp/doom-ui.el
@@ -562,6 +562,8 @@ windows, switch to `doom-fallback-buffer'. Otherwise, delegate to original
 (add-hook 'completion-list-mode-hook #'mode-line-invisible-mode)
 (add-hook 'Man-mode-hook #'mode-line-invisible-mode)
 
+(add-hook 'completion-list-mode-hook #'doom-disable-line-numbers-h)
+(add-hook 'Man-mode-hook #'doom-disable-line-numbers-h)
 
 ;;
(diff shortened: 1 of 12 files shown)


hlissner:
LGTM! Thanks again!

PR #8787, opened 2026-05-30
fix(default): prioritize neotree/treemacs sidebar bindings over dirvish

LemonBreezes:
I think most people prefer to use Treemacs or Neotree over Dirvish Side if they explicitly have it enabled in their config since these keybindings are the primary entry-points for those packages.


hlissner:
I've merged this into doomemacs/modules@8f07f59, since module development has moved to a new repo. Thanks for the PR and for bringing this to my attention!

PR #8852, opened 2026-07-15
refactor(cli): doom env: read allow/deny cookies with doom-file-cookie

fdeitylink:
I realized there is a simpler way of preventing the cookies from being evaluated as function calls; rather than quoting, just use the cookie function that doesn't eval (`doom-file-cookie` instead of `doom-file-cookie-p`).

This does cause a small breaking change for anyone who pulled in the previous fix (a305b91f44d4), because, without evaluating, the cookie `'nil` (for instance) is a list with two symbols (`(quote nil)`) rather than an empty list so `quote` and `nil` will be treated as regexes to match against. This results in `doom env` reporting `Wrong type argument: (stringp quote)` after reloading the envvars file. However, the fix is easy: edit the envvars file to remove the quoting from the file cookies.

Sorry for all the back and forth on `doom env`, and thanks again for all your help!

-----

### DIFF
--- a/bin/doom-env
+++ b/bin/doom-env
@@ -95,16 +95,16 @@ Why this over exec-path-from-shell?
     (setq-local coding-system-for-write 'utf-8-unix)
     (unless (equal output-file "-")
       (print! (start "%s envvars file...") (if reload? "Reloading" "Generating")))
-    (let* ((deny  (if reload? (doom-file-cookie-p output-file "deny")))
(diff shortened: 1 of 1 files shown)


hlissner (changes requested):
Please wrap those `doom-file-cookie` calls in `(doom-unquote ...)`, that will resolve any breakage.

hlissner (approved):

hlissner:
LGTM! Thanks for your help!

